PR #2, opened 2014-06-12
Split API classes and interfaces

jmikola:
Split previous API files into many more classes and interfaces. Manager and Server APIs were left as-is.

It may be helpful to create a class hierarchy/diagram.


jmikola:
Oops, didn't mean to link to this from #1943.

PR #26, opened 2015-03-12
Store WriteConcern in WriteResult to allow reporting of acknowledgement

jmikola:
* https://jira.mongodb.org/browse/PHPC-168
 * https://jira.mongodb.org/browse/PHPC-170
 * https://jira.mongodb.org/browse/PHPC-177

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -486,9 +486,11 @@ bool phongo_execute_write(mongoc_client_t *client, char *namespace, mongoc_bulk_
 	mongoc_bulk_operation_set_client(bulk, client);
 
 	/* If a write concern was not specified, libmongoc will use the client's
-	 * write concern. */
+	 * write concern; however, we should still fetch it for the write result. */
 	if (write_concern) {
 		mongoc_bulk_operation_set_write_concern(bulk, write_concern);
+	} else {
+		write_concern = mongoc_client_get_write_concern(client);
 	}
 
 	efree(dbname);
@@ -511,6 +513,7 @@ bool phongo_execute_write(mongoc_client_t *client, char *namespace, mongoc_bulk_
 	}
 
 	writeresult = phongo_writeresult_init(return_value, &bulk->result, server_hint TSRMLS_CC);
+	writeresult->write_concern = mongoc_write_concern_copy(write_concern);
 
 	/* The Write failed */
 	if (!hint) {
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1051,21 +1051,17 @@ void php_phongo_read_preference_to_zval(zval *retval, mongoc_read_prefs_t *read_
 
 void php_phongo_write_concern_to_zval(zval *retval, mongoc_write_concern_t *write_concern) /* {{{ */
 {
-	char                   *wtag;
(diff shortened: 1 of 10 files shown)


php_phongo.c
```diff
@@ -1051,21 +1060,18 @@ void php_phongo_read_preference_to_zval(zval *retval, mongoc_read_prefs_t *read_
...
 		add_assoc_string_ex(retval, ZEND_STRS("w"), wtag, 1);
+	} else if (mongoc_write_concern_get_wmajority(write_concern)) {
+		add_assoc_string_ex(retval, ZEND_STRS("w"), "majority", 1);
 	} else {
-		if (!mongoc_write_concern_get_wmajority(write_concern)) {
-			add_assoc_long_ex(retval, ZEND_STRS("w"), mongoc_write_concern_get_w(write_concern));
-		}
+		const int32_t w = mongoc_write_concern_get_w(write_concern);
+
+		add_assoc_long_ex(retval, ZEND_STRS("w"), w == MONGOC_WRITE_CONCERN_W_DEFAULT ? 1 : w);
```
bjori: Thats not true.
It may feel like w=1, but it isn't. It means server defaults, which can be 5, or majority, or even 0.

bjori: See http://docs.mongodb.org/manual/reference/command/replSetGetConfig/#replSetGetConfig.settings.getLastErrorDefaults

php_phongo.c
```diff
@@ -1051,21 +1060,18 @@ void php_phongo_read_preference_to_zval(zval *retval, mongoc_read_prefs_t *read_
 
 void php_phongo_write_concern_to_zval(zval *retval, mongoc_write_concern_t *write_concern) /* {{{ */
 {
-	char                   *wtag;
+	const char *wtag = mongoc_write_concern_get_wtag(write_concern);
 
 	array_init_size(retval, 5);
 
-	if (mongoc_write_concern_get_w(write_concern) == MONGOC_WRITE_CONCERN_W_DEFAULT) {
-		return;
```
bjori: This was relatively intentional.
If w=MONGOC_WRITE_CONCERN_W_DEFAULT, and fsync and j aren't specifcally provided, then there is no writeConcern.
In pre-2.6 it means that getLastError will be called, with no arguments, and the server defaults applied.
In 2.6+ it means no writeConcern document provided with the write, and the server defaults applied

I'm not sure how mongoc deals with fsync and j defaults.. I guess it doesn't care if they are explicitly provided or not (e.g. overwrite server defaults if explicitly provided).......

bjori:
Other then the comments above, lgtm

php_phongo.c
jmikola: I changed this to omit `w` if it's equal to `MONGOC_WRITE_CONCERN_W_DEFAULT`. We'll still print `wtag`, `"majority"`, or the numeric value for all other cases.

I'd like to do the same for `journal` and `fsync`, but that depends on mongodb/mongo-c-driver#194 being merged. As for `wtimeout`, 0 is considered the empty value for that. Ultimately, I think we need support for `null` arguments to WriteConcern or to simply take an `$options` array and be done with it. One argument is that we're locking ourselves into having an argument for `fsync`, which we already know is deprecated.

PR #33, opened 2015-04-03
PHPC-224: Consolidate Result and Cursor classes into one

jmikola:
* https://jira.mongodb.org/browse/PHPC-224
 * https://jira.mongodb.org/browse/PHPC-225
 * https://jira.mongodb.org/browse/PHPC-215


tests/functional/cursorid-001.phpt
```diff
@@ -16,16 +16,14 @@ $query = new MongoDB\Driver\Query(array(), array(
...
-$cursor = $qr->getIterator();
+$cursor = $manager->executeQuery(NS, $query);
+
 $cursorid = $cursor->getId();
 $s1 = (string)$cursorid;
 var_dump(
     $cursorid,
     $s1
 );
-$cursor->kill();
```
jmikola: This method was removed. Killing of cursors when they go out of scope will be implemented in [PHPC-225](https://jira.mongodb.org/browse/PHPC-225).

tests/functional/cursor-002.phpt
```diff
@@ -1,236 +0,0 @@
...
-var_dump(
-    $cursor->getBatchSize(),
-    $cursor->setBatchSize(15),
-    $cursor->getBatchSize(),
-    $cursor->isDead()
-);
-
-
-
-for($j=0, $cursor->rewind(); $cursor->valid(); $cursor->next(), $j++) {
```
jmikola: This test wasn't relevant anymore, since we're no longer implementing Iterator.

That said, I don't understand why the cursor was dumped on the 10th iteration and we expected `firstBatch` to be:

```
array(1) {
  ["username"]=>
  string(16) "abernathy.audrey"
}
```

tests/functional/cursor-001.phpt
```diff
@@ -14,35 +14,16 @@ $query = new MongoDB\Driver\Query(array(), array(
...
 
 foreach ($cursor as $document) {
     echo $document['username'] . "\n";
 }
 
-var_dump(
-    $cursor->getBatchSize(),
-    $cursor->setBatchSize(15),
-    $cursor->getBatchSize(),
-    $cursor->isDead()
```
jmikola: The batchSize methods were removed, as we no longer allow this to be changed during iteration (we simply initialize it based on the Query's option). `isDead()` can be tested elsewhere.

tests/functional/cursorid-001.phpt
jmikola: Tests for kill on destruct are now added.

php_phongo.c
```diff
@@ -1638,34 +1627,32 @@ static void phongo_result_iterator_move_forward(zend_object_iterator *iter TSRML
...
 	}
-	if (mongoc_cursor_next(result->cursor, &doc)) {
-		MAKE_STD_ZVAL(result->visitor_data.zchild);
-		bson_to_zval(bson_get_data(doc), doc->len, &result->visitor_data);
+	if (mongoc_cursor_next(cursor->cursor, &doc)) {
+		MAKE_STD_ZVAL(cursor->visitor_data.zchild);
+		bson_to_zval(bson_get_data(doc), doc->len, &cursor->visitor_data);
 	} else {
-		iter->funcs->invalidate_current(iter TSRMLS_CC);
+		/* TODO: is this really necessary? */
```
bjori: Because it'll be called during the dtor?
Maybe its not strictly necessary but I'd prefer to keep it in case someone can find a quirky way of triggering move_forward outside of normal routines.
.. remove the comment?

jmikola: The reason I didn't think it was necessary was that we already call `php_phongo_cursor_free_current()` at the top of this function. Thereafter, the `cursor->visitor_data` is only populated in two possible places:

 * `bson_iter_next()` on `first_batch_iter` succeeded, and the element was a document. `bson_to_zval()` is invoked to populate `cursor->visitor_data` and we return immediately.
 * `mongoc_cursor_next()` returned successfully and sets the `doc` pointer. We then call `bson_to_zval()` to populate `cursor->visitor_data`.

In the latter case (cursor iteration), this line of code is the `else` condition (i.e. cursor iteration didn't have a next document). I don't see any point where `cursor->visitor_data` could have been populated.

bjori: If coverity agrees with you, then just remove it?
I don't see the point in a TODO that will not ever change :]

bjori:
391 additions and 994 deletions

PR #34, opened 2015-04-03
Docs small fix

serebro:


bjori:
Whops. Thanks for the fix!

FYI; the toArray() behaviour is change in the 0.4.0 release, see: https://jira.mongodb.org/browse/PHPC-203

I've merged your fix for now -- we'll have to update the example again once we release 0.4.0 (hopefully early next week).

PR #35, opened 2015-04-14
PHPC-239: Cursor refcount issues

bjori:


php_phongo.c
```diff
@@ -1561,13 +1561,13 @@ static void php_phongo_cursor_iterator_dtor(zend_object_iterator *iter TSRMLS_DC
 {
 	php_phongo_cursor_iterator *cursor_it = (php_phongo_cursor_iterator *)iter;
 
+	php_phongo_cursor_free_current(cursor_it->cursor);
+
```
jmikola: Since this only frees `cursor->visitor_data.zchild` if it's not null, was there a specific reason to move it up? I've no objection if it was just for style points, but I'm curious if there was another reason.

php_phongo.c
```diff
@@ -1634,7 +1634,6 @@ static void php_phongo_cursor_iterator_rewind(zend_object_iterator *iter TSRMLS_
 	php_phongo_cursor_iterator *cursor_it = (php_phongo_cursor_iterator *)iter;
 	php_phongo_cursor_t        *cursor = cursor_it->cursor;
 
-	php_phongo_cursor_free_current(cursor);
```
jmikola: Since we can rewind a command cursor multiple times to iterate through the first batch (before we trigger a getmore), wouldn't this invite a memory leak?

php_phongo.c
bjori: cursor_it->cursor is the php_phongo_cursor_t* of the cursor_it->intern.data zval*.

If you destroy the zval first, the php_phongo_cursor_t* will be invalid.

php_phongo.c
bjori: How do you rewind it?
You can only get here from foreach() -- which will allocate its own get_iterator(), so each foreach() is isolated case.

bjori: Turns out you were somewhat right.

These free_current() are all harmless, so I left them -- but I managed to reproduce the issue of iterating over a cursor twice -- where the second time we would only access the cursor->firstBatch document.

tests/standalone/bug0239.phpt
```diff
@@ -0,0 +1,133 @@
+--TEST--
+TEST-0239: Iterator segfault
```
jmikola: This should be PHPC and omit the leading zero.

jmikola:
LGTM.

PR #37, opened 2015-04-14
PHPC-240: Cannot iterate beyond command cursor's first batch

jmikola:
https://jira.mongodb.org/browse/PHPC-240

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1568,8 +1568,6 @@ static void php_phongo_cursor_iterator_dtor(zend_object_iterator *iter TSRMLS_DC
 {
 	php_phongo_cursor_iterator *cursor_it = (php_phongo_cursor_iterator *)iter;
 
-	php_phongo_cursor_free_current(cursor_it->cursor);
-
 	if (cursor_it->intern.data) {
 		zval_ptr_dtor((zval**)&cursor_it->intern.data);
 		cursor_it->intern.data = NULL;
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -235,7 +235,7 @@ static void php_phongo_log(mongoc_log_level_t log_level, const char *log_domain,
 /* }}} */
 
 /* {{{ Init objects */
-void phongo_cursor_init(zval *return_value, mongoc_cursor_t *cursor, const bson_t *bson, mongoc_client_t *client, zend_bool is_command_cursor TSRMLS_DC) /* {{{ */
+void phongo_cursor_init(zval *return_value, mongoc_cursor_t *cursor, mongoc_client_t *client TSRMLS_DC) /* {{{ */
 {
 	php_phongo_cursor_t *intern;
 
@@ -245,8 +245,6 @@ void phongo_cursor_init(zval *return_value, mongoc_cursor_t *cursor, const bson_
 	intern->cursor = cursor;
 	intern->server_id = mongoc_cursor_get_hint(cursor);
 	intern->client = client;
-	intern->is_command_cursor = is_command_cursor;
-	intern->firstBatch = bson ? bson_copy(bson) : NULL;
 } /* }}} */
 
 void phongo_server_init(zval *return_value, mongoc_client_t *client, int server_id TSRMLS_DC) /* {{{ */
@@ -589,7 +587,7 @@ int phongo_execute_query(mongoc_client_t *client, const char *namespace, const p
 		return true;
 	}
 
-	phongo_cursor_init(return_value, cursor, doc, client, 0 TSRMLS_CC);
+	phongo_cursor_init(return_value, cursor, client TSRMLS_CC);
 	return true;
 } /* }}} */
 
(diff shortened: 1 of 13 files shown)


php_phongo.c
```diff
@@ -245,8 +245,6 @@ void phongo_cursor_init(zval *return_value, mongoc_cursor_t *cursor, const bson_
 	intern->cursor = cursor;
 	intern->server_id = mongoc_cursor_get_hint(cursor);
 	intern->client = client;
-	intern->is_command_cursor = is_command_cursor;
```
jmikola: I suggest removing this because, after simplifying our iteration handlers, it would only be used in the Cursor's debug handler.

php_phongo.c
```diff
@@ -1561,11 +1553,6 @@ static void php_phongo_cursor_iterator_dtor(zend_object_iterator *iter TSRMLS_DC
 {
 	php_phongo_cursor_iterator *cursor_it = (php_phongo_cursor_iterator *)iter;
 
-	if (cursor_it->cursor->firstBatch) {
-		bson_clear(&cursor_it->cursor->firstBatch);
-		cursor_it->cursor->firstBatch = NULL;
-	}
-
```
jmikola: Is the `php_phongo_cursor_free_current()` below actually necessary? `php_phongo_cursor_free()` already calls `php_phongo_cursor_free_current()`, and that will be invoked by `php_phongo_cursor_free_object()` when the zval is destroyed.

php_phongo.c
```diff
@@ -1561,11 +1553,6 @@ static void php_phongo_cursor_iterator_dtor(zend_object_iterator *iter TSRMLS_DC
...
 	php_phongo_cursor_free_current(cursor_it->cursor);
```
jmikola: Is this `php_phongo_cursor_free_current()` call actually necessary? `php_phongo_cursor_free()` already calls `php_phongo_cursor_free_current()`, and that will be invoked by `php_phongo_cursor_free_object()` when the zval is destroyed.

php_phongo.c
```diff
@@ -1638,30 +1613,16 @@ static void php_phongo_cursor_iterator_rewind(zend_object_iterator *iter TSRMLS_
 {
 	php_phongo_cursor_iterator *cursor_it = (php_phongo_cursor_iterator *)iter;
 	php_phongo_cursor_t        *cursor = cursor_it->cursor;
+	const bson_t               *doc;
 
 	php_phongo_cursor_free_current(cursor);
 	cursor_it->current = 0;
```
jmikola: Since the cursor is initialized to point to its first result, the rewind handler really need not do anything other than load the current element. `cursor_it->current` is already initialized to zero when allocating the structure. 

I actually think we should not reset `cursor_it->current` to zero here, as rewinding should be a NOP. Alternatively, we could be more aggressive and throw an exception if `cursor_it->current` is greater than zero (to notify users that they shouldn't be rewinding after they've started iteration).

Whatever we decide to do, we can easily test by wrapping our Cursor in an IteratorIterator, which allows us to trigger the rewind handler by calling `rewind()` in userland.

php_phongo.c
```diff
@@ -604,36 +602,43 @@ int phongo_execute_command(mongoc_client_t *client, const char *db, const bson_t
...
-						cursor->is_command = false;
-						phongo_cursor_init(return_value, cursor, &first_batch, client, 1 TSRMLS_CC);
-						return true;
-					}
+				ns = bson_iter_utf8(&child, &cursor->nslen);
+				bson_strncpy(cursor->ns, ns, sizeof cursor->ns);
+			} else if (BSON_ITER_IS_KEY(&child, "firstBatch")) {
+				if (BSON_ITER_HOLDS_ARRAY(&child) && bson_iter_recurse(&child, &cid->first_batch_iter)) {
+					cid->in_first_batch = true;
 				}
```
bjori: Does this resolve PHPC-230 ?

php_phongo.c
```diff
@@ -1276,17 +1281,7 @@ void php_phongo_cursor_to_zval(zval *retval, php_phongo_cursor_t *cursor) /* {{{
...
-		php_phongo_bson_state  state = PHONGO_BSON_STATE_INITIALIZER;
-
-		MAKE_STD_ZVAL(state.zchild);
-		bson_to_zval(bson_get_data(cursor->firstBatch), cursor->firstBatch->len, &state);
-		add_assoc_zval_ex(retval, ZEND_STRS("firstBatch"), state.zchild);
-	} else {
-		add_assoc_null_ex(retval, ZEND_STRS("firstBatch"));
-	}
 	add_assoc_long_ex(retval, ZEND_STRS("server_id"), cursor->server_id);
-	add_assoc_bool_ex(retval, ZEND_STRS("is_command_cursor"), cursor->is_command_cursor);
```
jmikola: @derickr: Take note that I'm removing "firstBatch" and "is_command_cursor" from the debug output. If this is merged, I will create HHVM tickets for you.

php_phongo.c
jmikola: Ah, I forgot about [PHPC-230](https://jira.mongodb.org/browse/PHPC-230). This should certainly fix it, though, since we are now relying on the final return statement after the common `phongo_cursor_init()` below.

php_phongo.c
bjori: cursor_it->cursor is the `intern` of the cursor_it->intern.data zval (which is destroyed one line below), isn't it?
So yes, it must be freed first.

jmikola: If the `zval_ptr_dtor((zval**)&cursor_it->intern.data);` below destroys the last reference to the zval, wouldn't its "free_object" handler get invoked?

bjori: Precisely -- if ts the last reference.
And it doesn't know about us assigning random pointers -- so the data becomes corrupt.

php_phongo.c
bjori: Then retrieving the key would be off-by-one between calling rewind() and next(), wouldn't it?

bjori:
lgtm

php_phongo.c
jmikola: But the "free_object" handler already calls `php_phongo_cursor_free_current()`. So it would either implicitly be invoked by us dtor-ing the zval here (if it's the last reference) or later on (if there is some other reference). And if there was another reference, there's probably no need to free the current element. If a second iterator is created for the Cursor (i.e. `php_phongo_cursor_t` struct), then its own rewind would free the current element anyway, but that's a side effect of two PHP iterators sharing the same Cursor.

php_phongo.c
jmikola: I don't think so. `cursor_it->current` is zero from "get_iterator" allocating the `php_phongo_cursor_iterator` struct. The "move_forward" handler isn't called until we advance to the element beyond the first, since only "rewind" would be called (obviously after "get_iterator") when iterating through the first element in a `foreach`.

jmikola: I have two more tests I can add to the PR tomorrow. Basically, I want to iterate through the Cursor with a `foreach` and call `rewind()` at each iteration and assert that it's a NOP (both for altering the current element *and* the index). Since we can't very well rewind earlier into the BSON buffer (especially after a getmore overwrites a previous batch), I'd just like to test that rewinding doesn't change the value or key progression.

bjori:
After reading the `cursor-iterator_handlers-001.phpt` testcase, I think we have to throw an exception in `rewind()` if the cursor has been iterated over, or in the middle of iteration.
Repeated calls to `rewind()` without `next()` in between is fine (although doesn't make a lot of sense), but after iteration has started a call to `rewind()` really has to throw `OutOfBoundsException`.

It is really confusing if calling `rewind()` is somewhat of a `NOOP` and you can continue stepping through some parts of it.

```
rewind($cursor); $first = current($cursor); next($cursor);
rewind($cursor); $firstAgain = current($cursor);
```

```
$first must be == $firstAgain
```

And since cannot do that without re-executing the query, which we clearly don't want to do -- I think we must throw an exception.

tests/standalone/cursor-getmore-001.phpt
```diff
@@ -0,0 +1,36 @@
...
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+Inserted: 5
+Found ID: 0
+Found ID: 1
+Found ID: 2
+Found ID: 3
+Found ID: 4
+===DONE===
```
bjori: Since you are adding tests for this -- maybe add one test where the results fit nicely within the batchSize?

derickr:
I think calling rewind more than once (even when no next() is called), should throw that exception.

bjori:
lgtm -- see also https://jira.mongodb.org/browse/PHPC-255

PR #38, opened 2015-04-22
add --with-libbson option

remicollet:
Per Fedora Packaging Guidelines, use of bundled libraries is not allowed.
 See https://fedoraproject.org/wiki/Packaging:No_Bundled_Libraries

This change add an --with-libbson option with allow to buld with system libbson when available (relying only on pkgconfig output)


bjori:
Thanks!
Merged in https://github.com/10gen-labs/mongo-php-driver-prototype/commit/a2770118b9b94d3d94a4de80e08e37bb7324ad76

I amended the commit message to add the ticket# since we autogenerate the changelogs

PR #40, opened 2015-04-25
PHPC-274: Fix zval_to_bson() encoding of BSON\Serializable objects

jmikola:
* https://jira.mongodb.org/browse/PHPC-274
 * https://jira.mongodb.org/browse/PHPC-275

----

Previously, only BSON\Persistable objects were handled.

### DIFF
--- a/src/bson.c
+++ b/src/bson.c
@@ -607,34 +607,43 @@ void object_to_bson(zval *object, php_phongo_bson_flags_t flags, const char *key
 
 	if (instanceof_function(Z_OBJCE_P(object), php_phongo_type_ce TSRMLS_CC)) {
 		if (instanceof_function(Z_OBJCE_P(object), php_phongo_serializable_ce TSRMLS_CC)) {
-			zval *retval = NULL;
+			zval *obj_data = NULL;
+			bson_t child;
+			HashTable *tmp_ht;
 
-			zend_call_method_with_0_params(&object, NULL, NULL, BSON_SERIALIZE_FUNC_NAME, &retval);
-			if (retval) {
-				bson_t      child;
-				HashTable  *tmp_ht;
+			zend_call_method_with_0_params(&object, NULL, NULL, BSON_SERIALIZE_FUNC_NAME, &obj_data);
 
-				convert_to_array_ex(&retval);
-				tmp_ht = HASH_OF(retval);
-
-				if (tmp_ht) {
-					tmp_ht->nApplyCount++;
-				}
-
-				bson_append_document_begin(bson, key, key_len, &child);
-				if (instanceof_function(Z_OBJCE_P(object), php_phongo_persistable_ce TSRMLS_CC)) {
-					if (flags & PHONGO_BSON_ADD_CHILD_ODS) {
-						bson_append_binary(&child, PHONGO_ODM_FIELD_NAME, -1, 0x80, (const uint8_t *)Z_OBJCE_P(object)->name, strlen(Z_OBJCE_P(object)->name));
-					}
-				}
-				zval_to_bson(retval, flags, &child, NULL TSRMLS_CC);
-				bson_append_document_end(bson, &child);
-
-				if (tmp_ht) {
-					tmp_ht->nApplyCount--;
-				}
-				zval_ptr_dtor(&retval);
+			if (!obj_data) {
+				/* zend_call_method() failed */
+				return;
 			}
+
+			if (Z_TYPE_P(obj_data) != IS_ARRAY) {
+				phongo_throw_exception(PHONGO_ERROR_RUNTIME TSRMLS_CC, "Expected %s() to return an array, %s given", BSON_SERIALIZE_FUNC_NAME, zend_get_type_by_const(Z_TYPE_P(obj_data)));
+				zval_ptr_dtor(&obj_data);
+
+				return;
+			}
+
+			tmp_ht = HASH_OF(obj_data);
+
+			if (tmp_ht) {
+				tmp_ht->nApplyCount++;
+			}
+
+			bson_append_document_begin(bson, key, key_len, &child);
+			if (instanceof_function(Z_OBJCE_P(object), php_phongo_persistable_ce TSRMLS_CC)) {
+				if (flags & PHONGO_BSON_ADD_CHILD_ODS) {
+					bson_append_binary(&child, PHONGO_ODM_FIELD_NAME, -1, 0x80, (const uint8_t *)Z_OBJCE_P(object)->name, strlen(Z_OBJCE_P(object)->name));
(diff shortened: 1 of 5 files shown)


bjori:
It is really difficult to understand the effect of this change in the modified test case.
Can you create a test case that currently fails, and this patch fixes, for clarity?

src/bson.c
```diff
@@ -607,34 +607,43 @@ void object_to_bson(zval *object, php_phongo_bson_flags_t flags, const char *key
...
 
-				bson_append_document_begin(bson, key, key_len, &child);
-				if (instanceof_function(Z_OBJCE_P(object), php_phongo_persistable_ce TSRMLS_CC)) {
-					if (flags & PHONGO_BSON_ADD_CHILD_ODS) {
-						bson_append_binary(&child, PHONGO_ODM_FIELD_NAME, -1, 0x80, (const uint8_t *)Z_OBJCE_P(object)->name, strlen(Z_OBJCE_P(object)->name));
-					}
-				}
-				zval_to_bson(retval, flags, &child, NULL TSRMLS_CC);
-				bson_append_document_end(bson, &child);
+				return;
```
jmikola: For PHPC-275: this conditional and the preceding `(!obj_data)` conditional (when `zend_call_method()` fails) above are the relevant changes for the `bson-encode_error-002.phpt` test case.

src/bson.c
```diff
@@ -763,23 +772,30 @@ PHONGO_API void zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *
 
 	switch(Z_TYPE_P(data)) {
 		case IS_OBJECT:
-			if (instanceof_function(Z_OBJCE_P(data), php_phongo_persistable_ce TSRMLS_CC)) {
+			if (instanceof_function(Z_OBJCE_P(data), php_phongo_serializable_ce TSRMLS_CC)) {
```
jmikola: For PHPC-274: this swap of `php_phongo_persistable_ce` to `php_phongo_serializable_ce` is the relevant change for satisfying the non-embedded test cases in `bson-encode-002.phpt` (the embedded test cases already passed before this).

src/bson.c
```diff
@@ -763,23 +772,30 @@ PHONGO_API void zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *
...
+					phongo_throw_exception(PHONGO_ERROR_RUNTIME TSRMLS_CC, "Expected %s() to return an array, %s given", BSON_SERIALIZE_FUNC_NAME, zend_get_type_by_const(Z_TYPE_P(obj_data)));
+					zval_ptr_dtor(&obj_data);
 
-					break;
+					return;
 				}
+
+				ht_data = HASH_OF(obj_data);
+
+				if (instanceof_function(Z_OBJCE_P(data), php_phongo_persistable_ce TSRMLS_CC)) {
```
jmikola: For PHPC-274: since we swapped the `instanceof` check above, we still want to respect the `PHONGO_BSON_ADD_ODS` flag, so this conditional has been added.

jmikola:
The attached `bug0274.phpt` test is an isolated regression test (extracted from `bson-encode-002.phpt`).

src/bson.c
```diff
@@ -607,34 +607,43 @@ void object_to_bson(zval *object, php_phongo_bson_flags_t flags, const char *key
...
+			if (!obj_data) {
+				/* zend_call_method() failed */
+				return;
+			}
 
-				if (tmp_ht) {
-					tmp_ht->nApplyCount++;
-				}
+			if (Z_TYPE_P(obj_data) != IS_ARRAY) {
+				phongo_throw_exception(PHONGO_ERROR_RUNTIME TSRMLS_CC, "Expected %s() to return an array, %s given", BSON_SERIALIZE_FUNC_NAME, zend_get_type_by_const(Z_TYPE_P(obj_data)));
```
jmikola: This should be an UnexpectedValueException.

bjori: Change this to UnexpectedValueException please

src/bson.c
```diff
@@ -763,23 +772,30 @@ PHONGO_API void zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *
...
-				zend_call_method_with_0_params(&data, NULL, NULL, BSON_SERIALIZE_FUNC_NAME, &obj_data);
-				if(obj_data) {
-					if (Z_TYPE_P(obj_data) == IS_ARRAY) {
-						ht_data = HASH_OF(obj_data);
-					} else {
-						phongo_throw_exception(PHONGO_ERROR_RUNTIME TSRMLS_CC, "%s", "Return value expected to be array");
-						zval_ptr_dtor(&obj_data);
-					}
+				if (Z_TYPE_P(obj_data) != IS_ARRAY) {
+					phongo_throw_exception(PHONGO_ERROR_RUNTIME TSRMLS_CC, "Expected %s() to return an array, %s given", BSON_SERIALIZE_FUNC_NAME, zend_get_type_by_const(Z_TYPE_P(obj_data)));
```
jmikola: This should also be an UnexpectedValueException.

src/bson.c
```diff
@@ -763,23 +772,30 @@ PHONGO_API void zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *
...
 		case IS_OBJECT:
-			if (instanceof_function(Z_OBJCE_P(data), php_phongo_persistable_ce TSRMLS_CC)) {
+			if (instanceof_function(Z_OBJCE_P(data), php_phongo_serializable_ce TSRMLS_CC)) {
+				zend_call_method_with_0_params(&data, NULL, NULL, BSON_SERIALIZE_FUNC_NAME, &obj_data);
 
-				if (flags & PHONGO_BSON_ADD_ODS) {
-					bson_append_binary(bson, PHONGO_ODM_FIELD_NAME, -1, 0x80, (const uint8_t *)Z_OBJCE_P(data)->name, strlen(Z_OBJCE_P(data)->name));
+				if (!obj_data) {
+					/* zend_call_method() failed */
+					return;
```
jmikola: Instead of returning here and after an exception below, we should break and then change the `if (ht_data && ht_data->nApplyCount > 1)` conditional later to also return if `ht_data` is `NULL`. Additionally, ensure we initialize `ht_data` to `NULL` at the top of this function.

bjori:
lgtm

src/bson.c
jmikola: https://github.com/10gen-labs/mongo-php-driver-prototype/pull/43

src/bson.c
jmikola: https://github.com/10gen-labs/mongo-php-driver-prototype/pull/43

PR #41, opened 2015-04-30
PHPC-278: nModified may be null for legacy writes

jmikola:
https://jira.mongodb.org/browse/PHPC-278

----

A test isn't included in the PR because we don't presently have 2.4 in the build matrix (see: [PHPC-279](https://jira.mongodb.org/browse/PHPC-279)).


bjori:
lgtm

PR #46, opened 2015-05-20
PHPC-305: add --with-libmongo option

remicollet:
Notice: check use --atleast 1.1.5 when should be 1.2.0 but .pc in generated file (from git snapshot, 1.2.0-dev still report as 1.1.5-dev...)

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -315,18 +315,42 @@ dnl libmongoc stuff {{{
   AC_MSG_CHECKING(configuring libmongoc)
   AC_MSG_RESULT(...)
 
+PHP_ARG_WITH(libmongoc, Use system libmongoc,
+[  --with-libmongoc           Use system libmongoc], no, no)
+
   AC_DEFINE(HAVE_MONGOC, 1, [Kinda useless extension without it..])
 
-  CPPFLAGS="$CPPFLAGS -DBSON_COMPILATION -DMONGOC_COMPILATION -DMONGOC_TRACE"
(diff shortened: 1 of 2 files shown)


bjori:
I've noticed the wrong mongoc version reporting to -- haven't bothered fixing it but I guess we'll have to now :)

Thanks for this patch!

I'm a littlebit worried that this branch will become stale as its not something I would actively test -- but I guess you would spot things quickly if I accidentally broke it ;)

Do you have an environment where you test these builds and I can  proactively verify everything is OK?

remicollet:
> Do you have an environment where you test these builds and I can proactively verify everything is OK?

I'm used to build everything as RPM (in clean chrooted env). Updating the RPM is mostly just to bump the commit date and reference, then I can quickly launch various build (Fedora 18-22, RHEL 5-7, 32/64bits, NTS/ZTS, PHP 5.4/5.5/5.6/7.0, ...).

A solution could be to use a Copr repository, designed for such need.

bjori:
If I understand Copr correctly I have to
- create src.rpm
- put it online somewhere
- Update the src.rpm URL in Copr

And then Copr will build the binary RPM for the various target platforms then put the package into a RPM repository users can use, similar to launchpad PPA.

It seems targeting rolling out releases rather then CI environment.

Wondering if it makes sense to maintain the rpm spec file, and the debian packaging, in this repo. hmmm.

PR #47, opened 2015-06-17
[WIP] PHPC-316: config.w32 should define MONGOC_ENABLE_SASL

jmikola:
https://jira.mongodb.org/browse/PHPC-316

### DIFF
--- a/config.w32
+++ b/config.w32
@@ -5,7 +5,7 @@ ARG_WITH("mongodb-sasl", "MongoDB: Build against Cyrus-SASL", "yes");
 ADD_EXTENSION_DEP("mongodb", "openssl", false);
 
 if (PHP_MONGODB != "no") {
-  ADD_FLAG("CFLAGS_MONGODB", " /I" + configure_module_dirname + " /I" + configure_module_dirname + "/src/libmongoc/src/mongoc" + " /I" + configure_module_dirname + "/src/libbson/src/bson/ /I" + configure_module_dirname + "/src/libbson/src /I" + configure_module_dirname + "/src/contrib  /D HAVE_MONGODB_SASL=1 /D BSON_COMPILATION=1 /D MONGOC_COMPILATION=1 /D MONGOC_TRACE=1 /D MONGOC_NO_AUTOMATIC_GLOBALS=1 /D BSON_EXTRA_ALIGN=0");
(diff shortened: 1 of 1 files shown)


config.m4
```diff
@@ -406,6 +406,12 @@ if test "$PHP_MONGODB_SASL" != "no"; then
   ], [
     -L$MONGODB_SASL_DIR/$PHP_LIBDIR
   ])
+
+  PHP_CHECK_LIBRARY(sasl2, sasl_client_done,
+  [
+    MONGOC_HAVE_SASL_CLIENT_DONE=1
+    AC_SUBST(MONGOC_HAVE_SASL_CLIENT_DONE)
+  ])
```
jmikola: I noticed elsewhere in `config.m4` we split the constant definition and `AC_SUBST` into separate statements. Is there any reason we don't simplify as `AC_SUBST(MONGOC_HAVE_SASL_CLIENT_DONE, 1)`?

bjori: No idea. I imagine version dependencies? No prior art in php-src makes the use of it "risky"

config.w32
```diff
@@ -20,9 +20,14 @@ if (PHP_MONGODB != "no") {
 
   AC_DEFINE("HAVE_MONGOC", 1);
 
-  if (!CHECK_LIB("libsasl.lib", "mongodb", PHP_MONGODB)) {
-	  WARNING("mongodb libsasl support not enabled, libs not found");
+  if (PHP_MONGODB_SASL != "no") {
```
jmikola: This is based on what we were doing for 1.x [here](https://github.com/mongodb/mongo-php-driver/blob/239a4324911094ff6c755afaf78aadc3981194b1/config.w32).

config.w32
```diff
@@ -20,9 +20,14 @@ if (PHP_MONGODB != "no") {
...
+    if (CHECK_LIB("libsasl.lib", "mongodb", PHP_MONGODB)) {
+      ADD_FLAG("CFLAGS_MONGODB", " /D MONGOC_ENABLE_SASL=1");
```
jmikola: In this case, I think we still need to check for `sasl_client_done()` (as we do in `config.m4` when calling `PHP_CHECK_LIBRARY(sasl2, sasl_client_done, ...)`). I'm currently looking into whether `CHECK_FUNC_IN_HEADER()` can be used.

bjori:
If it works on windows, lgtm

jmikola:
I'll revisit this once I can test the make build/test targets with your Windows VM.

bjori:
Since the Windows build instructions have very specific dependencies I don't believe config.w32 generally contains as complex checks as config.m4, especially when it comes to version/function existence.

I'm not sure about the PHPC-316 commit, but the PHPC-306 is good and should maybe be merged standalone?

jmikola:
I split off the `config.m4` change into #55 and left this PR for PHPC-316 (despite its branch name) since we had relevant comments above. We can revisit this when the Windows build environment is ready.

jmikola:
This PR should be obsolete. A proper check for `MONGOC_ENABLE_SASL` was implemented in https://github.com/mongodb/mongo-php-driver/pull/276 for [PHPC-629](https://jira.mongodb.org/browse/PHPC-629).

PR #48, opened 2015-06-19
PHPC-309: Move BSON types and functions to MongoDB\BSON namespace

derickr:


bjori:
Since the pecl/bson ext will be released from the same actual source code it helps if we don't just replace the actual namespace name like this, as it would make it impossible to use the same sources.

I've made the necessary changes and pushed https://github.com/10gen-labs/mongo-php-driver-prototype/commit/61c44329217eb028ab4c60ba535418f15f5ab671

PR #49, opened 2015-06-21
LDAP VM does not require python-pip (or epel)

jmikola:
When building the VMs on Friday, I encountered some errors with selecting a mirror for the epel repository. Ultimately, I noticed that the LDAP VM doesn't actually depend on `python-pip`, which was the only reason we were installing epel. This removes them both.


jmikola:
I'll wait until we have the test suite running (especially connection and LDAP tests) before merging this.

PR #51, opened 2015-06-29
PHPC-317: Ensure "sort" query option serializes as BSON object

jmikola:
https://jira.mongodb.org/browse/PHPC-317

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -356,7 +356,7 @@ bool phongo_query_init(php_phongo_query_t *query, zval *filter, zval *options TS
 				return false;
 			}
 
-			convert_to_array_ex(&sort);
+			convert_to_object_ex(&sort);
 			Z_ADDREF_P(sort);
 			add_assoc_zval_ex(zquery, ZEND_STRS("$orderby"), sort);
 		}
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -329,7 +329,6 @@ bool phongo_query_init(php_phongo_query_t *query, zval *filter, zval *options TS
 				return false;
 			}
 
-			convert_to_array_ex(&modifiers);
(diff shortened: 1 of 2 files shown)


php_phongo.c
```diff
@@ -342,7 +341,6 @@ bool phongo_query_init(php_phongo_query_t *query, zval *filter, zval *options TS
 				return false;
 			}
 
-			convert_to_array_ex(&projection);
 			query->selector = bson_new();
```
bjori: How is this related to $sort ?
Is it similar issue, in which case, should it not be convert_to_object_ex() then?

bjori:
I think you need to kick travis to see if it likes it

php_phongo.c
jmikola: The projection and query document (i.e. what we merge modifiers into above) didn't need this treatment since they are top-level documents and you encode them as such later on. Sort was the unique case where it gets stuffed in an `$orderby` field and could be encoded as an array inadvertently.

I do mean to split these changes up into separate commits, but the `convert_to_array_ex()` deletions here and for modifiers are basically just removing dead code.

jmikola:
Travis looks good. This will need a test for sorting on `{ "0": 1 }` and `{ "0": 1, "1": 1 }` to test for proper serialization, though.

php_phongo.c
bjori: oki lgtm

jmikola:
Updated commit to test for `{ "0" : 1 }`, which was all that's really needed. We can verify that the serialization is correct through Query's debug handler.

Additionally, 3.0.x seems to ignore the `$orderby` value if it's not an object. Testing that the documents come back in an order other than their natural (inserted) order helps to test that the server is actually taking our `$orderby` value into account.

bjori:
lgtm

PR #52, opened 2015-06-29
Update install docs now that we have a PECL pcakage

jmikola:


jmikola:
Forgot that I already opened #50 last week.

PR #55, opened 2015-06-30
PHPC-306: Check for sasl_client_done in config.m4

jmikola:
https://jira.mongodb.org/browse/PHPC-306

This is split off from #47.


jmikola:
@bjori: I'll wait for Travis to report green, but I believe this has a LGTM from https://github.com/10gen-labs/mongo-php-driver-prototype/pull/47#issuecomment-117277615.

bjori:
This doesn't actually do anything until we bump libmongoc.
Do that maybe at the same time?

lgtm either way

bjori:
I take that back.

The AC_SUBST() needs to run no matter what, and the MONGOC_HAVE_SASL_CLIENT_DONE "initialized" to 0.

Otherwise we wind up with pretty weird looking mongoc-config.h

config.m4
```diff
@@ -406,7 +406,18 @@ if test "$PHP_MONGODB_SASL" != "no"; then
...
   ])
+
+  PHP_CHECK_LIBRARY(sasl2, sasl_client_done,
+  [
+    AC_SUBST(MONGOC_HAVE_SASL_CLIENT_DONE, 1)
+  ], [
+    AC_SUBST(MONGOC_HAVE_SASL_CLIENT_DONE, 0)
+  ])
+else
+  AC_SUBST(MONGOC_ENABLE_SASL, 0)
```
jmikola: I realized that `MONGOC_ENABLE_SASL` was also missing when we configured without SASL (same issue that you brought up with `MONGOC_HAVE_SASL_CLIENT_DONE`). Relevant libmongoc changes are https://github.com/mongodb/mongo-c-driver/commit/c3c14f431864c82f670581c0e380cef9047cfb96.

jmikola:
This is currently waiting on [CDRIVER-731](https://jira.mongodb.org/browse/CDRIVER-731) and possibly https://github.com/mongodb/mongo-c-driver/pull/246. As-is, the SASL mutex functions generate extra trace output that we can't capture, which causes spurious test failures.

As a side note, bumping libmongoc did introduce some other issues with server selection timing out. It's completely unrelated to this config change, but is something I'll also want to look into before merging this into master.

PR #56, opened 2015-07-06
->[to|from]Array() -> [to|from]PHP(). "root" typemapping. Default "root" to stdclass, not array

jmikola:
* https://jira.mongodb.org/browse/PHPC-311

Rename BSON\to|from]Array() to BSON\[to|from]PHP()

 * https://jira.mongodb.org/browse/PHPC-315

Support setting the "root" (container) value to explicit type.
->setTypemap(["root" => "array"]);

 * https://jira.mongodb.org/browse/PHPC-319

Use stdClass, not array, for top-level containers by default

### DIFF
--- a/src/bson.c
+++ b/src/bson.c
@@ -905,10 +905,10 @@ int bson_to_zval(const unsigned char *data, int data_len, php_phongo_bson_state 
 	/* If php_phongo_bson_visit_binary() finds an ODM class, it supersedes our
 	 * document type. */
 	if (state->odm) {
-		state->map.document_type = PHONGO_TYPEMAP_CLASS;
+		state->map.root_type = PHONGO_TYPEMAP_CLASS;
 	}
 
-	switch (state->map.document_type) {
+	switch (state->map.root_type) {
 		case PHONGO_TYPEMAP_NATIVE_ARRAY:
 			/* Nothing to do here */
 			break;
@@ -916,11 +916,11 @@ int bson_to_zval(const unsigned char *data, int data_len, php_phongo_bson_state 
 		case PHONGO_TYPEMAP_CLASS:
 			/* If the class implements Unserializable, initialize the object
 			 * from our array data; otherwise, fall through to native object. */
-			if (instanceof_function(state->odm ? state->odm : state->map.document, php_phongo_unserializable_ce TSRMLS_CC)) {
+			if (instanceof_function(state->odm ? state->odm : state->map.root, php_phongo_unserializable_ce TSRMLS_CC)) {
 				zval *obj = NULL;
 
 				MAKE_STD_ZVAL(obj);
-				object_init_ex(obj, state->odm ? state->odm : state->map.document);
+				object_init_ex(obj, state->odm ? state->odm : state->map.root);
 				zend_call_method_with_1_params(&obj, NULL, NULL, BSON_UNSERIALIZE_FUNC_NAME, NULL, state->zchild);
 				zval_ptr_dtor(&state->zchild);
 				state->zchild = obj;
@@ -970,7 +970,6 @@ void php_phongo_bson_typemap_to_state(zval *typemap, php_phongo_bson_typemap *ma
 		char                  *classname;
 		int                    classname_len;
 		zend_bool              classname_free = 0;
-		zend_class_entry      *array_ce = NULL, *document_ce = NULL;
 
 		classname = php_array_fetchl_string(typemap, "array", sizeof("array")-1, &classname_len, &classname_free);
 		if (classname_len) {
@@ -979,8 +978,8 @@ void php_phongo_bson_typemap_to_state(zval *typemap, php_phongo_bson_typemap *ma
 			} else if (!strcasecmp(classname, "stdclass") || !strcasecmp(classname, "object")) {
 				map->array_type = PHONGO_TYPEMAP_NATIVE_OBJECT;
 			} else {
+				zend_class_entry *array_ce = zend_fetch_class(classname, classname_len, ZEND_FETCH_CLASS_AUTO TSRMLS_CC);
 				map->array_type = PHONGO_TYPEMAP_CLASS;
-				array_ce = zend_fetch_class(classname, classname_len, ZEND_FETCH_CLASS_AUTO TSRMLS_CC);
 
 				if (instanceof_function(array_ce, php_phongo_unserializable_ce TSRMLS_CC)) {
 					map->array = array_ce;
@@ -998,11 +997,30 @@ void php_phongo_bson_typemap_to_state(zval *typemap, php_phongo_bson_typemap *ma
 			} else if (!strcasecmp(classname, "stdclass") || !strcasecmp(classname, "object")) {
 				map->document_type = PHONGO_TYPEMAP_NATIVE_OBJECT;
 			} else {
+				zend_class_entry *document_ce = zend_fetch_class(classname, classname_len, ZEND_FETCH_CLASS_AUTO TSRMLS_CC);
 				map->document_type = PHONGO_TYPEMAP_CLASS;
-				document_ce = zend_fetch_class(classname, classname_len, ZEND_FETCH_CLASS_AUTO TSRMLS_CC);
 
 				if (instanceof_function(document_ce, php_phongo_unserializable_ce TSRMLS_CC)) {
 					map->document = document_ce;
+				}
+			}
+			if (classname_free) {
+				efree(classname);
+			}
+		}
+
+		classname = php_array_fetchl_string(typemap, "root", sizeof("root")-1, &classname_len, &classname_free);
+		if (classname_len) {
+			if (!strcasecmp(classname, "array")) {
+				map->root_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
+			} else if (!strcasecmp(classname, "stdclass") || !strcasecmp(classname, "object")) {
+				map->root_type = PHONGO_TYPEMAP_NATIVE_OBJECT;
+			} else {
+				zend_class_entry *root_ce = zend_fetch_class(classname, classname_len, ZEND_FETCH_CLASS_AUTO TSRMLS_CC);
+				map->root_type = PHONGO_TYPEMAP_CLASS;
+
+				if (instanceof_function(root_ce, php_phongo_unserializable_ce TSRMLS_CC)) {
+					map->root = root_ce;
 				}
 			}
 			if (classname_free) {
--- a/src/bson.c
+++ b/src/bson.c
@@ -896,17 +896,40 @@ int bson_to_zval(const unsigned char *data, int data_len, php_phongo_bson_state 
 		return 0;
 	}
 
+	/* We initialize an array because it will either be returned as-is (native
+	 * array in type map), passed to bsonUnserialize() (ODM class), or used to
+	 * initialize a stdClass object (native object in type map). */
 	array_init(state->zchild);
 	bson_iter_visit_all(&iter, &php_bson_visitors, state);
-	if ((state->map.array || state->odm) && instanceof_function(state->odm ? state->odm : state->map.array, php_phongo_unserializable_ce TSRMLS_CC)) {
-		zval *obj = NULL;
 
-		MAKE_STD_ZVAL(obj);
-		object_init_ex(obj, state->odm ? state->odm : state->map.array);
-		zend_call_method_with_1_params(&obj, NULL, NULL, BSON_UNSERIALIZE_FUNC_NAME, NULL, state->zchild);
-		SEPARATE_ZVAL(&state->zchild);
-		zval_dtor(state->zchild);
-		ZVAL_ZVAL(state->zchild, obj, 1, 1);
+	/* If php_phongo_bson_visit_binary() finds an ODM class, it supersedes our
+	 * document type. */
+	if (state->odm) {
+		state->map.document_type = PHONGO_TYPEMAP_CLASS;
+	}
+
+	switch (state->map.document_type) {
+		case PHONGO_TYPEMAP_NATIVE_ARRAY:
+			/* Nothing to do here */
+			break;
+
+		case PHONGO_TYPEMAP_CLASS:
+			/* If the class implements Unserializable, initialize the object
+			 * from our array data; otherwise, fall through to native object. */
+			if (instanceof_function(state->odm ? state->odm : state->map.document, php_phongo_unserializable_ce TSRMLS_CC)) {
+				zval *obj = NULL;
+
+				MAKE_STD_ZVAL(obj);
+				object_init_ex(obj, state->odm ? state->odm : state->map.document);
+				zend_call_method_with_1_params(&obj, NULL, NULL, BSON_UNSERIALIZE_FUNC_NAME, NULL, state->zchild);
+				zval_ptr_dtor(&state->zchild);
+				state->zchild = obj;
+				break;
+			}
+
+		case PHONGO_TYPEMAP_NATIVE_OBJECT:
+		default:
+			object_and_properties_init(state->zchild, zend_standard_class_def, Z_ARRVAL_P(state->zchild));
 	}
 
 	if (bson_reader_read(reader, &eof) || !eof) {
--- a/src/bson.c
+++ b/src/bson.c
@@ -943,9 +943,9 @@ int bson_to_zval(const unsigned char *data, int data_len, php_phongo_bson_state 
 	return 1;
 }
 
-/* {{{ proto string BSON\fromArray(array|object data)
-   Returns the BSON representation of a value */
-PHP_FUNCTION(fromArray)
+/* {{{ proto string BSON\fromPHP(array|object $value)
+   Returns the BSON representation of a PHP value */
+PHP_FUNCTION(fromPHP)
 {
 	zval   *data;
 	bson_t *bson;
@@ -1029,9 +1029,9 @@ void php_phongo_bson_typemap_to_state(zval *typemap, php_phongo_bson_typemap *ma
 		}
 	}
 }
-/* {{{ proto string BSON\toArray(string data [, array $typemap = array()])
-   Returns the PHP representation of a BSON value, optionally converting them into custom types/classes */
-PHP_FUNCTION(toArray)
+/* {{{ proto array|object BSON\toPHP(string $bson [, array $typemap = array()])
+   Returns the PHP representation of a BSON value, optionally converting it into a custom class */
+PHP_FUNCTION(toPHP)
 {
 	char                  *data;
 	int                    data_len;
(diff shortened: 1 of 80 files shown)


tests/manager/manager-executeQuery-001.phpt
```diff
@@ -73,7 +73,7 @@ object(MongoDB\Driver\Cursor)#%d (%d) {
       ["mode"]=>
       int(1)
       ["tags"]=>
-      array(0) {
+      object(stdClass)#%d (0) {
```
jmikola: @bjori: When `mongoc_read_prefs_t` is initialized, its tags are also initialized to an empty BSON document. In turn, `php_phongo_read_preference_to_zval()` will do:

```
if (read_prefs->tags.len) {
    php_phongo_bson_state  state = PHONGO_BSON_STATE_INITIALIZER;

    MAKE_STD_ZVAL(state.zchild);
    bson_to_zval(bson_get_data(&read_prefs->tags), read_prefs->tags.len, &state);
    add_assoc_zval_ex(retval, ZEND_STRS("tags"), state.zchild);
} else {
    add_assoc_null_ex(retval, ZEND_STRS("tags"));
}
```

That `else` condition will really never be reached. Would you rather I change the condition to `read_prefs->tags.len > 5` and add a comment that we only dump it if it's a non-empty document? Otherwise, I can leave this as-is.

tests/manager/manager-var-dump-001.phpt
```diff
@@ -45,10 +45,10 @@ object(MongoDB\Driver\Manager)#%d (%d) {
       ["is_passive"]=>
       bool(false)
       ["tags"]=>
-      array(0) {
+      object(stdClass)#%d (0) {
```
jmikola: @bjori: This and the `last_is_master` field below are stdClass instances because `php_phongo_server_to_zval()` invokes `bson_to_zval()` on some of its BSON documents. I assume this is a debug handler where we definitely want to dump things as arrays, correct?

If so, perhaps this happens often enough that I should just create an alternative constant to `PHONGO_BSON_STATE_INITIALIZER` intended for this purpose.

jmikola: Also, I noticed that `php_phongo_server_to_zval()` uses fresh stack-allocated `php_phongo_bson_state` variables for each `bson_to_zval()` call, while `php_phongo_writeresult_get_debug_info()` re-uses them. I don't see any harm in re-using them, as it would only be modified if the `__pclass` field is picked up and ODM behavior is activated (and I wouldn't expect that for this BSON data).

jmikola: @bjori: One more question. If `Server::getTags()` and `Server::getInfo()` return the tags and isMaster response, respectively, shouldn't those return stdClass instances by default? Or should they be consistent with our debug handler and decode BSON objects as arrays? This would primarily affect the method signature for `Server::getTags()`, since it's currently documented as returning an array.

I'm leaning towards using arrays for both of this, since they're both largely used for diagnostics and shouldn't have any array/document confusion.

tests/manager/manager-executeQuery-001.phpt
bjori: I think all the else append null checks were all added at the same time (dozens of them) after I realized not everything was initialized everytime, so var_dump()ing had the tendency to crash.

I don't think I ever checked individual cases, but instead always added a fallback path.

I'd say in this particular case, tags = null would look weird. tags = array() looks best.

Why is this stdClass now? that looks weird?

tests/manager/manager-var-dump-001.phpt
bjori: new ReadPreference($readPreference, array( array("dc" => "nyc"), array("disk" => "ssd")));

We set these things as arrays. I think this should all be arrays in the debug output too.

tests/manager/manager-executeQuery-001.phpt
jmikola: > Why is this stdClass now? that looks weird?

I'm changing `php_phongo_read_preference_to_zval()`, `Server::getTags()`, `Server::getInfo()`, and `php_phongo_server_to_zval()` to use native array decoding, as we discussed earlier.

tests/manager/manager-var-dump-001.phpt
jmikola: :+1:

src/bson.c
```diff
@@ -896,17 +896,44 @@ int bson_to_zval(const unsigned char *data, int data_len, php_phongo_bson_state
...
-		SEPARATE_ZVAL(&state->zchild);
-		zval_dtor(state->zchild);
-		ZVAL_ZVAL(state->zchild, obj, 1, 1);
+
+	/* If php_phongo_bson_visit_binary() finds an ODM class, it supersedes our
+	 * document type.
+	 *
+	 * TODO: decide if native array/object should be allowed to override this,
+	 * in which case we only supersede PHONGO_TYPEMAP_NONE.
+	 */
```
bjori: no code todo please. They are destined to rot.
If its realistic, add a ticket. If not. swallow it :]

tests/bson/bson-encode-004.phpt
```diff
@@ -22,7 +22,7 @@ $hannes->addFriend($mikola);
 
 var_dump($hannes);
 
-$s = fromArray(array($hannes));
+$s = fromArray($hannes);
```
bjori: Not a fan. it causes bigass changes in the diff...
Is there a compelling reason to do this?
If you want to add the second variant as a new test case, thats cool - but I'm not a fan of manipulating the existing testcase in this way.

tests/bson/bson-generate-document-id.phpt
```diff
@@ -48,23 +48,23 @@ array(3) {
     string(8) "New York"
   }
   ["_id"]=>
-  object(%s\ObjectID)#%d (%d) {
+  object(%s\ObjectID)#%d (1) {
```
bjori: no need. If anything -- these should be instead be replaced with %d everywhere :)

tests/bson/bson-generate-document-id.phpt
```diff
@@ -48,23 +48,23 @@ array(3) {
...
   }
 }
 
 Dumping fetched user document:
-array(3) {
+object(stdClass)#%d (3) {
   ["_id"]=>
-  object(%s\ObjectID)#%d (%d) {
+  object(%s\ObjectID)#%d (1) {
     ["oid"]=>
```
bjori: keep as %d please.
Who knows. maybe we'll add extra data to the ObjectID (like seperating out the timestamp in a var_dump()), and then changing this one is annoying (especially since it creates a weird looking failed diff until you realize whats wrong)

tests/bson/bson-generate-document-id.phpt
```diff
@@ -48,23 +48,23 @@ array(3) {
...
     string(24) "%s"
   }
   ["username"]=>
   string(3) "bob"
   ["address"]=>
-  object(stdClass)#%d (%d) {
+  object(stdClass)#%d (2) {
```
bjori: same as above

tests/functional/query-sort-004.phpt
```diff
@@ -26,6 +26,11 @@ var_dump($query);
...
 $cursor = $manager->executeQuery(NS, $query);
 
+/* Numeric keys of stdClass instances cannot be directly accessed, so ensure the
+ * document is decoded as a PHP array.
+ */
+$cursor->setTypeMap(array('document' => 'array'));
+
 foreach ($cursor as $document) {
     echo $document['0'] . "\n";
 }
```
bjori: Does "document" apply to the root? That looks very wrong.
It didn't before, and I don't think it was part of this ticket either.

tests/replicaset/manager-getservers-001.phpt
```diff
@@ -76,7 +76,7 @@ array(3) {
       ["primary"]=>
       string(19) "192.168.112.10:3000"
       ["tags"]=>
-      object(stdClass)#%d (%d) {
+      array(2) {
```
bjori: How did this happen?

tests/replicaset/manager-getservers-001.phpt
```diff
@@ -158,7 +158,7 @@ array(3) {
...
```
bjori: This also looks suspicious. Why are things changing from stdclass->array?
The ticket is about the top-level array become a stdclass.....

tests/replicaset/manager-getservers-001.phpt
jmikola: This is the result of `php_phongo_server_to_zval()` using `state.map.document_type = PHONGO_TYPEMAP_NATIVE_ARRAY;` when decoding `last_is_master`, and that mapping applying both to top-level and embedded documents alike.

Same concern that you raised in https://github.com/10gen-labs/mongo-php-driver-prototype/pull/56/files#r34193151. In order to fix this, I believe [PHPC-319](https://jira.mongodb.org/browse/PHPC-319) should depend on [PHPC-315](https://jira.mongodb.org/browse/PHPC-315) and I need to create the "root" mapping first.

tests/replicaset/manager-getservers-001.phpt
jmikola: Same as https://github.com/10gen-labs/mongo-php-driver-prototype/pull/56/files#r34196036.

src/bson.c
jmikola: Created [PHPC-329](https://jira.mongodb.org/browse/PHPC-329), so I'll remove this.

tests/replicaset/manager-getservers-001.phpt
bjori: I don't see why it has to depend on PHPC-315, this ticket should be able to be resolved alone (by not applying 'document' mapping to the top-level).
Its not going to do that anyway even after PHPC-315.

I don't mind postponing this until PHPC-315 is finished if you feel that is easier then to fix the current 'document' behavior.

jmikola: Well, without PHPC-315 and introducing a new option to control the top-level document, we have no way to request native array decoding with `bson_to_zval()` when we want it (e.g. debug output). Otherwise, we'd always get a stdClass back unless the ODM behavior picks up a class name in our binary field.

I think PHPC-315 should be integrated into this PR, and then I can replace the `state.map.document_type` assignments with `state.map.root_type`.

tests/bson/bson-generate-document-id.phpt
jmikola: I didn't see the harm in hard-coding these because the elements themselves are rigidly defined below. If more elements were added, we'd have to update both anyway.

I can make a commit changing this across the board, though. In the long run, it would reduce the size of diffs when we add elements down the line and take away one more thing that might get out of sync.

bjori:
I've fixed the failing test on travis and am kicking the builds now

jmikola:
@bjori: I cleaned up the cases where I replaced `(%d)` expectations with hard-coded numbers, but I wasn't about to do a global search-and-replace in this PR. Feel free to create a ticket to do this later. I'm not sure if you just wanted stdClass and array dumps covered, or also things like string lengths.

jmikola:
A few issues still remaining:

 * [PHPC-329](https://jira.mongodb.org/browse/PHPC-329): Only allow ODM behavior to supersede unspecified type map (not array/stdClass/custom mode)
 * [PHPC-330](https://jira.mongodb.org/browse/PHPC-330): Include `__pclass` in `bsonUnserialize()` when possible
 * [PHPC-331](https://jira.mongodb.org/browse/PHPC-331): Throw UnexpectedValueExceptions for unsupported (de)serialization cases

src/bson.c
```diff
@@ -957,7 +979,7 @@ void php_phongo_bson_typemap_to_state(zval *typemap, php_phongo_bson_typemap *ma
 				map->array_type = PHONGO_TYPEMAP_NATIVE_OBJECT;
 			} else {
 				map->array_type = PHONGO_TYPEMAP_CLASS;
-				array_ce = zend_fetch_class(classname, classname_len, ZEND_FETCH_CLASS_AUTO TSRMLS_CC);
+				zend_class_entry *array_ce = zend_fetch_class(classname, classname_len, ZEND_FETCH_CLASS_AUTO TSRMLS_CC);
```
bjori: Variable declarations need to be done first.
This won't compile on VS.. And I thought this was part of our --enable-developer-flags (-Wdeclaration-after-statement).

src/bson.c
```diff
@@ -976,7 +998,7 @@ void php_phongo_bson_typemap_to_state(zval *typemap, php_phongo_bson_typemap *ma
 				map->document_type = PHONGO_TYPEMAP_NATIVE_OBJECT;
 			} else {
 				map->document_type = PHONGO_TYPEMAP_CLASS;
-				document_ce = zend_fetch_class(classname, classname_len, ZEND_FETCH_CLASS_AUTO TSRMLS_CC);
+				zend_class_entry *document_ce = zend_fetch_class(classname, classname_len, ZEND_FETCH_CLASS_AUTO TSRMLS_CC);
```
bjori: Same as above, declaration-after-statement

src/bson.c
bjori: https://travis-ci.org/10gen-labs/mongo-php-driver-prototype/jobs/70487040#L878

src/bson.c
```diff
@@ -986,11 +1008,30 @@ void php_phongo_bson_typemap_to_state(zval *typemap, php_phongo_bson_typemap *ma
...
+
+		classname = php_array_fetchl_string(typemap, "root", sizeof("root")-1, &classname_len, &classname_free);
+		if (classname_len) {
+			if (!strcasecmp(classname, "array")) {
+				map->root_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
+			} else if (!strcasecmp(classname, "stdclass") || !strcasecmp(classname, "object")) {
+				map->root_type = PHONGO_TYPEMAP_NATIVE_OBJECT;
+			} else {
+				map->root_type = PHONGO_TYPEMAP_CLASS;
+				zend_class_entry *root_ce = zend_fetch_class(classname, classname_len, ZEND_FETCH_CLASS_AUTO TSRMLS_CC);
```
bjori: same as above

tests/bson/bson-encode-002.phpt
```diff
@@ -36,38 +36,38 @@ class NumericArray implements BSON\Serializable, BSON\Unserializable
...
 
 echo "Testing top-level AssociativeArray:\n";
-$bson = BSON\fromArray(new AssociativeArray);
-echo BSON\toJSON($bson), "\n";
+$bson = fromPHP(new AssociativeArray);
+echo toJSON($bson), "\n";
 echo "Encoded BSON:\n";
 hex_dump($bson);
-$value = BSON\toArray($bson, array("document" => 'AssociativeArray'));
+$value = toPHP($bson, array("root" => 'AssociativeArray'));
```
bjori: Why was the document -> root change here needed?
AFAICT this test has now been repurposed to cover both "document" and "root" ?

I'd prefer to leave it covering "document" and adding another one for "root"

tests/bson/bson-encode-002.phpt
```diff
@@ -36,38 +36,38 @@ class NumericArray implements BSON\Serializable, BSON\Unserializable
...
 
 echo "\nTesting top-level NumericArray:\n";
-$bson = BSON\fromArray(new NumericArray);
-echo BSON\toJSON($bson), "\n";
+$bson = fromPHP(new NumericArray);
+echo toJSON($bson), "\n";
 echo "Encoded BSON:\n";
 hex_dump($bson);
-$value = BSON\toArray($bson, array("document" => 'NumericArray'));
+$value = toPHP($bson, array("root" => 'NumericArray'));
```
bjori: same as above, this looks like a bad change

tests/bson/bson-encode-004.phpt
```diff
@@ -74,33 +74,34 @@ object(Person)#%d (5) {
...
+    A0 : 69 70 00 c8 00 00 00 02 63 6f 75 6e 74 72 79 00  [ip......country.]
+    B0 : 08 00 00 00 49 63 65 6c 61 6e 64 00 00 00 04 66  [....Iceland....f]
+    C0 : 72 69 65 6e 64 73 00 5a 00 00 00 03 30 00 52 00  [riends.Z....0.R.]
+    D0 : 00 00 05 5f 5f 70 63 6c 61 73 73 00 06 00 00 00  [...__pclass.....]
+    E0 : 80 50 65 72 73 6f 6e 02 6e 61 6d 65 00 07 00 00  [.Person.name....]
+    F0 : 00 4a 65 72 65 6d 79 00 10 61 67 65 00 15 00 00  [.Jeremy..age....]
+   100 : 00 04 61 64 64 72 65 73 73 65 73 00 05 00 00 00  [..addresses.....]
+   110 : 00 04 66 72 69 65 6e 64 73 00 05 00 00 00 00 00  [..friends.......]
+   120 : 00 00 00                                         [...]
+object(stdClass)#%d (1) {
```
bjori: And thats why changes "just because" aren't cool!
but ok. fine.

bjori:
LGTM
- [ ] Fix the build
- [ ] Confirm no new build warnings
- [ ] Confirm tests/bson/bson-encode-002.phpt

src/bson.c
jmikola: Good catch. I surely must have missed this stderr output when building locally.

tests/bson/bson-encode-004.phpt
jmikola: AFAIK, this diff wasn't for any of the referenced tickets.

I specifically made it in the separate "Use plural property name to store list of Address objects" commit.

tests/bson/bson-encode-002.phpt
jmikola: The test cases, of which there are four, are the same:

 * Testing top-level AssociativeArray
 * Testing embedded AssociativeArray
 * Testing top-level NumericArray
 * Testing embedded NumericArray

For the first case here, "document" would have previously not applied to anything and the decoded BSON would have been an array (since changed to stdClass in this PR). Given the name of the test, I feel it should actually be testing "root", which didn't exist at the time it was originally created.

Otherwise, I would have simply changed the expected output from array to stdClass. That assertion is effectively covered by the very next "Testing embedded AssociativeArray" case.

tests/bson/bson-encode-002.phpt
jmikola: Intentional: https://github.com/10gen-labs/mongo-php-driver-prototype/pull/56/files#r34633822

jmikola:
Rebased the PHPC-315 commit to fix the build warnings, and also adjusted my build scripts to I should catch stderr output more easily in the future.

I left an explanation of why I changed the `bson-encode-002.phpt` test, too. Hopefully that's acceptable.

bjori:
oklgtm

PR #57, opened 2015-07-06
PHPC-325: bson_to_zval() leaks when buffer contains multiple documents

jmikola:
https://jira.mongodb.org/browse/PHPC-325

### DIFF
--- a/src/bson.c
+++ b/src/bson.c
@@ -885,37 +885,37 @@ int bson_to_zval(const unsigned char *data, int data_len, php_phongo_bson_state 
 
 	reader = bson_reader_new_from_data(data, data_len);
 
-	if (!(b = bson_reader_read(reader, &eof))) {
+	if (!(b = bson_reader_read(reader, NULL))) {
 		php_error_docref(NULL TSRMLS_CC, E_WARNING, "Could not read document from reader");
 		return 0;
 	}
 
-	do {
-		if (!bson_iter_init(&iter, b)) {
-			bson_reader_destroy(reader);
-			return 0;
-		}
-
-		array_init(state->zchild);
-		bson_iter_visit_all(&iter, &php_bson_visitors, state);
-		if ((state->map.array || state->odm) && instanceof_function(state->odm ? state->odm : state->map.array, php_phongo_unserializable_ce TSRMLS_CC)) {
-			zval *obj = NULL;
-
-			MAKE_STD_ZVAL(obj);
-			object_init_ex(obj, state->odm ? state->odm : state->map.array);
-			zend_call_method_with_1_params(&obj, NULL, NULL, BSON_UNSERIALIZE_FUNC_NAME, NULL, state->zchild);
-			SEPARATE_ZVAL(&state->zchild);
-			zval_dtor(state->zchild);
-			ZVAL_ZVAL(state->zchild, obj, 1, 1);
-		}
-	} while ((b = bson_reader_read(reader, &eof)));
-
-	bson_reader_destroy(reader);
-
-	if (!eof) {
-		php_error_docref(NULL TSRMLS_CC, E_WARNING, "Invalid BSON detected");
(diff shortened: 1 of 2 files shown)


jmikola:
@bjori: review plz

src/bson.c
```diff
@@ -885,38 +885,38 @@ int bson_to_zval(const unsigned char *data, int data_len, php_phongo_bson_state
 
 	reader = bson_reader_new_from_data(data, data_len);
 
-	if (!(b = bson_reader_read(reader, &eof))) {
+	if (!(b = bson_reader_read(reader, NULL))) {
```
jmikola: I'm passing `NULL` here because we don't actually check `eof` until the second call, which your original do/while loop was accomplishing.

src/bson.c
```diff
@@ -885,38 +885,38 @@ int bson_to_zval(const unsigned char *data, int data_len, php_phongo_bson_state
...
+		SEPARATE_ZVAL(&state->zchild);
+		zval_dtor(state->zchild);
+		ZVAL_ZVAL(state->zchild, obj, 1, 1);
+	}
 
-	if (!eof) {
-		php_error_docref(NULL TSRMLS_CC, E_WARNING, "Invalid BSON detected");
+	if (bson_reader_read(reader, &eof) || !eof) {
+		php_error_docref(NULL TSRMLS_CC, E_WARNING, "Reading document did not exhaust input buffer");
+		bson_reader_destroy(reader);
```
bjori: This needs to dtor the state->zchild.
There is no way for the caller to know on failure if it has to clean it up or not depending on the failure.

src/bson.c
```diff
@@ -1007,6 +1007,7 @@ PHP_FUNCTION(toArray)
 
 	MAKE_STD_ZVAL(state.zchild);
 	if (!bson_to_zval((const unsigned char *)data, data_len, &state)) {
+		zval_ptr_dtor(&state.zchild);
```
bjori: This shouldn't be needed.
What if it failed with "Could not initialize BSON iterator"? Then you are dtoring uninitialized value.

tests/bson/bug0325.phpt
```diff
@@ -0,0 +1,21 @@
...
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$bson1 = MongoDB\BSON\fromJSON('{"x": "y"}');
+$bson2 = MongoDB\BSON\fromJSON('{"a": "b"}');
+$value = MongoDB\BSON\toArray($bson1 . $bson2);
+
```
bjori: Please use the global fromJSON() and toArray() functions.

src/bson.c
jmikola: `MAKE_STD_ZVAL(state.zchild)` is always called before we venture into `bson_to_zval()`, which is why we free. This shouldn't depend on whether we called `array_init()` within `bson_to_zval()`.

tests/bson/bug0325.phpt
jmikola: Noted. I pulled this right from the ticket, but I did see you made those global functions for a reason :)

src/bson.c
jmikola: I avoided placing the `zval_ptr_dtor()` within `bson_to_zval()` because every calling context (excluding the `toArray()` function, which I address in this PR) was already freeing the zval on its own. Since this function expects the zval to already be allocated, the calling contexts are able to make that decision.

src/bson.c
bjori: oh. you are right. I was mixing things

src/bson.c
bjori: No you are right.
I was tricked by the array_init().

bjori:
lgtm

PR #60, opened 2015-07-20
PHPC-330: Always include __pclass property when unserializing BSON

jmikola:
https://jira.mongodb.org/browse/PHPC-330


derickr:
LGTM

PR #64, opened 2015-07-27
BSON serialization spec compliance

jmikola:
* https://jira.mongodb.org/browse/PHPC-336
 * https://jira.mongodb.org/browse/PHPC-329
 * https://jira.mongodb.org/browse/PHPC-331
 * https://jira.mongodb.org/browse/PHPC-343
 * https://jira.mongodb.org/browse/PHPC-344


src/bson.c
```diff
@@ -617,10 +621,22 @@ void object_to_bson(zval *object, php_phongo_bson_flags_t flags, const char *key
...
 
-			if (Z_TYPE_P(obj_data) != IS_ARRAY) {
-				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Expected %s() to return an array, %s given", BSON_SERIALIZE_FUNC_NAME, zend_get_type_by_const(Z_TYPE_P(obj_data)));
+			if (Z_TYPE_P(obj_data) != IS_ARRAY && !(Z_TYPE_P(obj_data) == IS_OBJECT && instanceof_function(Z_OBJCE_P(obj_data), zend_standard_class_def TSRMLS_CC))) {
+				const char *classname = NULL;
+				zend_uint classname_len = 0;
+				int classname_free = 0;
+
+				if (Z_TYPE_P(obj_data) == IS_OBJECT) {
+					classname_free = !zend_get_object_classname(obj_data, &classname, &classname_len TSRMLS_CC);
```
jmikola: @bjori: I noticed you were `Z_OBJCE_P(object)->name` directly when creating the `__pclass` value. I found `zend_get_object_classname()` used in core in several places where a class name needed to be reported, so I assume it's fine to use for this exception message.

I can't imagine why folks would want to report a different class name, but I expect the ODM behavior is fine sticking to the `name` value from the struct.

jmikola:
@derickr: Feel free to take a look at this and my test changes. I was able to get the 5.3 build working (although I still seem some local failures about missing OpenSSL types), so I'll plan to merge this in tomorrow.

config.m4
```diff
@@ -111,6 +111,12 @@ if test "$MONGODB" != "no"; then
     PHP_CHECK_GCC_ARG(-Wdeclaration-after-statement,    _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wdeclaration-after-statement")
     dnl PHP_CHECK_GCC_ARG(-Werror,                          _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Werror")
 
+    PHP_MONGODB_FOUND_VERSION=`${PHP_CONFIG} --version`
+    PHP_MONGODB_FOUND_VERNUM=`echo "${PHP_MONGODB_FOUND_VERSION}" | $AWK 'BEGIN { FS = "."; } { printf "%d", ([$]1 * 100 + [$]2) * 100 + [$]3;}'`
+
+    if test "$PHP_MONGODB_FOUND_VERNUM" -ge "50400"; then
+      PHP_CHECK_GCC_ARG(-Wno-cast-qual,                 _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wno-cast-qual")
+    fi
```
derickr: You shouldn't do this. Instead, fix the consts.

src/bson.c
```diff
@@ -62,6 +66,9 @@
 
 #define PHONGO_ODM_FIELD_NAME "__pclass"
 
+#define IS_CLASS_INSTANTIATABLE(ce) \
```
derickr: It's common to prefix macros, or, if you don't, #undefine them when you're done with them

tests/bson/bson-encode-002.phpt
```diff
@@ -67,7 +67,7 @@ $bson = fromPHP(array('embed' => new NumericArray));
 echo toJSON($bson), "\n";
 echo "Encoded BSON:\n";
 hex_dump($bson);
-$value = toPHP($bson, array("document" => 'NumericArray'));
+$value = toPHP($bson, array("array" => 'NumericArray'));
```
derickr: Doesn't this change the meaning of the test?

tests/utils/tools.php
```diff
@@ -147,16 +147,16 @@ function START($id, array $options = array()) {
     }
 
     $options += array("name" => "mongod", "id" => $id);
-    $opts = [
-        "http" => [
+    $opts = array(
+        "http" => array(
```
derickr: Oh, I missed this. I had created a specific PR for this already at https://github.com/10gen-labs/mongo-php-driver-prototype/pull/65

tests/bson/bson-toPHP-003.phpt
```diff
@@ -0,0 +1,470 @@
+--TEST--
+BSON\toPHP(): Tests from serialization specification
+--SKIPIF--
```
derickr: Is this just a copy of my test?

tests/bson/bson-encode-002.phpt
jmikola: The original test was written before allowing Serializables to encode as a BSON array. The original intention of this test is just to round-trip the data from NumericArray to BSON and back to NumericArray. Since the value does get stored as a BSON array, the type map needed an update.

tests/bson/bson-toPHP-003.phpt
jmikola: It should have the same test cases (from the spec itself), but it wasn't using traits or short array syntax. Additionally, I'm using the dynamic BSON namespace and encode/decode functions that @bjori created in `tools.php`.

config.m4
jmikola: Just for 5.3?

5.4+ rightfully lists the args as `const char *`, so there's no need there.

src/bson.c
jmikola: I can toss a `PHONGO` prefix on this, unless you think it's better to undefine at the bottom of the file. I currently have no use for this outside of `bson.c`.

config.m4
bjori: I'm fine with this.
Just for reference, which exact case prompted this change?
Isn't it also valid for the arginfos?

src/bson.c
bjori: `zend_get_object_classname` has been removed in PHP-7.
Please continue to use `Z_OBJCE_P(object)->name`

src/bson.c
```diff
@@ -617,10 +624,22 @@ void object_to_bson(zval *object, php_phongo_bson_flags_t flags, const char *key
...
+			if (Z_TYPE_P(obj_data) != IS_ARRAY && !(Z_TYPE_P(obj_data) == IS_OBJECT && instanceof_function(Z_OBJCE_P(obj_data), zend_standard_class_def TSRMLS_CC))) {
+				const char *classname = NULL;
+				zend_uint classname_len = 0;
+				int classname_free = 0;
+
+				if (Z_TYPE_P(obj_data) == IS_OBJECT) {
+					classname_free = !zend_get_object_classname(obj_data, &classname, &classname_len TSRMLS_CC);
+				}
+
+				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Expected %s() to return an array or stdClass, %s given", BSON_SERIALIZE_FUNC_NAME, (classname_len ? classname : zend_get_type_by_const(Z_TYPE_P(obj_data))));
```
bjori: You should probably include the classname too.
There are plenty matches for the function name so it doesn't help mutch having that in the error message without the classname

src/bson.c
```diff
@@ -964,71 +1000,63 @@ PHP_FUNCTION(fromPHP)
...
-				if (instanceof_function(array_ce, php_phongo_unserializable_ce TSRMLS_CC)) {
-					map->array = array_ce;
-				}
-			}
-			if (classname_free) {
-				efree(classname);
-			}
+			apply_classname_to_state(classname, classname_len, &map->array_type, &map->array TSRMLS_CC);
+		}
+		if (classname_free) {
```
bjori: How could classname_free be set if classname_len is 0?

src/bson.c
jmikola: :+1:

src/bson.c
jmikola: :+1:

src/bson.c
jmikola: Based on `char *php_array_zval_to_string(zval *z, int *plen, zend_bool *pfree)` in `php_array_api.h`, `pfree` is only set to zero if `zval *z` is `NULL` or a string, in which case its direct value is returned. I believe I encountered a memory leak in the test where I passed in `null` to each type map key. This uses `STR_EMPTY_ALLOC()`, which might return an interned or allocated, empty string (which is why I started using `str_efree()` on it, since `efree()` isn't aware of interned strings).

config.m4
jmikola: arginfos appear to be declared as `static const` in PHP 5.3. I only saw these warnings with `zend_parse_parameters()` and possibly one other macro (not arginfo).

src/bson.c
bjori: If classname_len is 0, then there isn't anything to free, so classname_free can hardly be set?
Shouldn't this then be moved inside the if (classname_len) ?

jmikola: AFAIK, an empty string could still be allocated as a single null byte and `classname_len` would be zero. I think it's quite independent from `classname_len`.

config.m4
jmikola: For reference, here are a few of these errors from 5.3:

```
/home/jmikola/workspace/mongodb/phpc/src/bson.c: In function ‘zif_fromPHP’:
/home/jmikola/workspace/mongodb/phpc/src/bson.c:967:55: warning: passing argument 3 of ‘zend_parse_parameters’ discards ‘const’ qualifier from pointer target type
  if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "A", &data) == FAILURE) {
                                                       ^
In file included from /home/jmikola/bin/php-bin/php-5.3/include/php/main/php.h:38:0,
                 from /home/jmikola/workspace/mongodb/phpc/src/bson.c:29:
/home/jmikola/bin/php-bin/php-5.3/include/php/Zend/zend_API.h:245:14: note: expected ‘char *’ but argument is of type ‘const char *’
 ZEND_API int zend_parse_parameters(int num_args TSRMLS_DC, char *type_spec, ...);
              ^
/home/jmikola/workspace/mongodb/phpc/src/bson.c:974:1: warning: passing argument 3 of ‘_estrndup’ discards ‘const’ qualifier from pointer target type
  RETVAL_STRINGL((const char *) bson_get_data(bson), bson->len, 1);
 ^
```

A full log of the stderr output from a 5.3 build is here: https://gist.github.com/jmikola/e3a3f7da387333187864

src/bson.c
```diff
@@ -31,6 +31,10 @@
 #include <Zend/zend_hash.h>
 #include <Zend/zend_interfaces.h>
 
+#if PHP_VERSION_ID >= 50400
+# include <Zend/zend_string.h>
+#endif
```
bjori: I feel this should really be in the compat header as a #else condition for the str_efree() ?
Or is that pushing it a bit far?

src/bson.c
bjori: ok

bjori:
I'm not sure why so many unrelated things are in the same PR!

I haven't read the latest and greatest "BSON serialization spec" so I don't know if this does indeed confirm to it, or makes actual sense.

The code itself looks good to me though.

config.m4
derickr: We don't care any more, as we can drop 5.3 :-D

src/bson.c
derickr: I think that'd be neater too.

tests/bson/bson-toPHP-003.phpt
derickr: With 5.3 gone, you can use traits and short array syntax again though!

src/bson.c
jmikola: Will do.

tests/bson/bson-toPHP-003.phpt
jmikola: Soon.

jmikola:
> I'm not sure why so many unrelated things are in the same PR!

I created half of the tickets after starting PHPC-329 and most of the code was inter-dependent.

PR #65, opened 2015-07-28
Fixed syntax for PHP 5.3

derickr:


jmikola:
As you mentioned, this is already in #64, so I'll merge it with that PR.

PR #66, opened 2015-07-28
Fixes due to some inconsistencies in HHVM vs PHP

derickr:


tests/bson/bson-decode-001.phpt
```diff
@@ -62,7 +62,7 @@ Test#2 { "0" : null }
 OK
 Test#3 { "0" : 123 }
 OK
-Test#4 { "0" : 4.23 }
+Test#4 { "0" : 4.23%d }
```
jmikola: AFAIK, this shouldn't be necessary if you bump libbson. Older versions printed more precision than was necessary.

Also, `%d` doesn't have a "zero or more" variant (like `%s` and `%S`), so this could fail for newer libbson versions that only print `4.23`, no?

tests/bson/bson-encode-001.phpt
```diff
@@ -77,7 +77,7 @@ Test#7 { "0" : null }
      0 : 08 00 00 00 0a 30 00 00                          [.....0..]
 Test#8 { "0" : 123 }
      0 : 0c 00 00 00 10 30 00 7b 00 00 00 00              [.....0.{....]
-Test#9 { "0" : 4.23 }
+Test#9 { "0" : 4.23%d }
```
jmikola: Same: https://github.com/10gen-labs/mongo-php-driver-prototype/pull/66/files#r35662922

tests/bson/bug0325.phpt
```diff
@@ -16,6 +16,6 @@ var_dump($value);
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
-Warning: %s\toPHP(): Reading document did not exhaust input buffer in %s on line %d
+Warning:%sReading document did not exhaust input buffer in %s on line %d%a
```
jmikola: We seldom use `%a`. What is HHVM printing that requires greedy matching of multiple lines?

tests/bson/bson-decode-001.phpt
derickr: Oh, perhaps. Hadn't realize this was a libbson version. Will try that.

tests/bson/bug0325.phpt
derickr: A stack trace (as my commit message says).

jmikola: :+1:

tests/bson/bug0334.phpt
```diff
@@ -19,13 +19,11 @@ class MyClass implements BSON\Persistable {
...
+var_dump($php['__pclass']->getData());
 
 ?>
-===DONE===
-<?php exit(0); ?>
 --EXPECT--
-     0 : 28 00 00 00 05 5f 5f 70 63 6c 61 73 73 00 07 00  [(....__pclass...]
-    10 : 00 00 80 4d 79 43 6c 61 73 73 02 66 6f 6f 00 04  [...MyClass.foo..]
-    20 : 00 00 00 62 61 72 00 00                          [...bar..]
-===DONE===
```
jmikola: As discussed in IRC, we'll take this change but I'll create a `bug0334-002.phpt` file, which you can skip, that actually tests that we don't encode duplicate keys in the BSON (using `hex_dump()`).

tests/bson/bug0325.phpt
jmikola: Will fix with [PHPC-345](https://jira.mongodb.org/browse/PHPC-345).

bjori: @derickr Can you paste the full error message here for future reference?
Also, is it all messages or just thig particular one?

jmikola: He said it was argument-passing errors as well, which he was ignoring. I believe HHVM is just throwing stacktraces on all error/warning messages.

jmikola:
Two failures for PHP 5.5 and 5.6:

> Fatal error: Cannot use object of type MyClass as array in /home/travis/build/10gen-labs/mongo-php-driver-prototype/tests/bson/bug0334.php on line 20

That should be resolved with #64.

tests/bson/bug0334.phpt
```diff
@@ -19,13 +19,11 @@ class MyClass implements BSON\Persistable {
...
 
-hex_dump(fromPHP(new MyClass));
+$bson = fromPHP(new MyClass);
+$php = toPHP($bson, array('root' => 'array'));
+
+var_dump($php['__pclass']->getData());
 
 ?>
-===DONE===
-<?php exit(0); ?>
```
jmikola: IIRC, we discussed `===DONE===` previously and it should stick around.

bjori:
This change seems to have broken travis:

========DIFF========

001+ Fatal error: Cannot use object of type MyClass as array in /home/travis/build/10gen-labs/mongo-php-driver-prototype/tests/bson/bug0334.php on line 20

001- string(7) "MyClass"

========DONE========

FAIL PHPC-334: Injected __pclass should override a __pclass key in bsonSerialize() return value [tests/bson/bug0334.phpt]

tests/bson/bug0325.phpt
derickr: That's correct - well, almost all. It looks like:

```
Warning: Reading document did not exhaust input buffer in /home/derick/dev/php/derickr-mongo-php-driver-prototype/tests/utils/tools.php on line 379
    #0 MongoDB\BSON\toPHP(), called at [/home/derick/dev/php/derickr-mongo-php-driver-prototype/tests/utils/tools.php:379]
    #1 toPHP(), called at [/home/derick/dev/php/derickr-mongo-php-driver-prototype/tests/bson/bug0325.phpt:11]
```

jmikola:
@bjori: Correct. I mentioned that in https://github.com/10gen-labs/mongo-php-driver-prototype/pull/66#issuecomment-125681623.

PR #70, opened 2015-07-30
Restore -Werror compile flag for developer options

jmikola:
See: 3051d3b18469751db1256559714ae99172613b01


jmikola:
Still broken. See: mongodb/mongo-c-driver#256

PR #71, opened 2015-07-31
Renamed subtype to subType as per naming guidelines

derickr:
It also matches the method name getSubType(), and fixes an inconsistency with
Hippo.


bjori:
Why are we calling it a subType anyway?
Why isn't it called 'type'?

The only reason why the bsonspec calls it explicitly "subtype" is to avoid confusion with other bson main types.

When you have an instance of Binary, there is no confusion what $binary->getType() means.

derickr:
Superseeded by https://github.com/10gen-labs/mongo-php-driver-prototype/pull/71

PR #76, opened 2015-08-05
Added the error message to the failure cases.

derickr:
This makes it much easier to debug what goes wrong, en-masse. In my case, I already found a "FIXME ADD EXCEPTION" instead of having to debug what the f goes on :-)

### DIFF
--- a/tests/utils/tools.php
+++ b/tests/utils/tools.php
@@ -243,13 +243,13 @@ function throws($function, $exceptionname, $infunction = null) {
                 if (strcasecmp($function, $infunction) == 0) {
                     printf("OK: Got %s thrown from %s\n", $exceptionname, $infunction);
                 } else {
-                    printf("ALMOST: Got %s - but was thrown in %s, not %s\n", $exceptionname, $function, $infunction);
+                    printf("ALMOST: Got %s - but was thrown in %s, not %s (%s)\n", $exceptionname, $function, $infunction, $e->getMessage());
                 }
                 return $e->getMessage();
             }
             printf("OK: Got %s\n", $exceptionname);
         } else {
-            printf("ALMOST: Got %s - expected %s\n", get_class($e), $exceptionname);
+            printf("ALMOST: Got %s - expected %s: %s\n", get_class($e), $exceptionname, $e->getMessage());
         }
         return $e->getMessage();
     }


bjori:
Could you have it trim newlines?
There are unfortunately messages generated by external components that have newlines in them (a lot of them even) :(

bjori:
lgtm after that

tests/utils/tools.php
```diff
@@ -243,13 +243,13 @@ function throws($function, $exceptionname, $infunction = null) {
...
                 } else {
-                    printf("ALMOST: Got %s - but was thrown in %s, not %s\n", $exceptionname, $function, $infunction);
+                    printf("ALMOST: Got %s - but was thrown in %s, not %s (%s)\n", $exceptionname, $function, $infunction, $e->getMessage());
                 }
                 return $e->getMessage();
             }
             printf("OK: Got %s\n", $exceptionname);
         } else {
-            printf("ALMOST: Got %s - expected %s\n", get_class($e), $exceptionname);
+            printf("ALMOST: Got %s - expected %s: %s\n", get_class($e), $exceptionname, $e->getMessage());
```
jmikola: @bjori: Are you asking for an [`rtrim()`](http://php.net/rtrim) (or `rtrim("\r\n")`) call on the exception message parameters to `printf()`? If so, I can make that change tomorrow morning and merge this.

bjori: No ... That will only trim trailing new lines, not in the middle of the message

jmikola: What components are creating messages with newlines (not sure I've seen those before).

And a second question: is this just to allow `%s` in EXPECTF blocks, instead of requiring multi-line catch-alls?

bjori: SSL, SASL, and I think mongoc compounds some error message somewhere, although its been a while so maybe not.

Its about being all in one line yes, so you can differentiate what is coming from what in the output.

jmikola:
Manually merged in https://github.com/10gen-labs/mongo-php-driver-prototype/commit/2aa53295397ca5c464af57b539c0b2db2ac4c8e6.

Newline stripping added in https://github.com/10gen-labs/mongo-php-driver-prototype/commit/e333fc035726a8da6590294d5b086d044d3630c1.

PR #77, opened 2015-08-05
Manager getters and options for read prefs and write concerns

jmikola:
* https://jira.mongodb.org/browse/PHPC-196
 * https://jira.mongodb.org/browse/PHPC-353


src/MongoDB/Manager.c
```diff
@@ -81,17 +81,55 @@ PHP_METHOD(Manager, __construct)
...
+	intern->client = php_phongo_make_mongo_client(uri, driverOptions TSRMLS_CC);
+	mongoc_uri_destroy(uri);
+
 	if (!intern->client) {
-		phongo_throw_exception(PHONGO_ERROR_RUNTIME TSRMLS_CC, "%s", "Failed to create Manager from URI");
+		char *str;
+		size_t str_len;
+
+		str = bson_as_json(&bson_options, &str_len);
+		phongo_throw_exception(PHONGO_ERROR_RUNTIME TSRMLS_CC, "Failed to create Manager from URI '%s' and options: %s", uri_string, str);
```
bjori: I don't think its a good idea to serialize the entire options array as json and include it in the exception...

tests/manager/manager-getwriteconcern-001.phpt
```diff
@@ -0,0 +1,161 @@
...
+    array(STANDALONE . '/?w=majority', array('wtimeoutms' => 1000)),
+    array(STANDALONE . '/?w=customTagSet', array('wtimeoutms' => 1000)),
+);
+
+foreach ($tests as $i => $test) {
+    list($uri, $options) = $test;
+
+    $manager = new MongoDB\Driver\Manager($uri, $options);
+    var_dump($manager->getWriteConcern());
+}
```
bjori: since this method uses return_value_used, this test needs to include a test for calling the method without assigning the return value.

jmikola: Added.

src/MongoDB/Manager.c
jmikola: I reverted to just printing the URI for the two original exceptions (URI parsing fails and client initialization fails). I did change the URI parsing exception to use InvalidArgumentException, which I think makes more sense.

For the later two exceptions, I've changed them to InvalidArgumentException and just report that we couldn't apply the RP or WC options to the Manager.

jmikola:
@bjori: see my replies on https://github.com/jmikola/mongo-php-driver-prototype/commit/2b20d34dabcf852608feb96b0df7be7c8fbf3be6 and https://github.com/jmikola/mongo-php-driver-prototype/commit/80ef0e9390a7803d3404ee69617c1ee0f733a030 and let me know if anything remains.

php_phongo.c
```diff
@@ -1591,6 +1607,178 @@ void php_phongo_populate_default_ssl_ctx(php_stream_context *ctx, zval *driverOp
...
 } /* }}} */
 
+bool php_phongo_apply_rp_options_to_client(mongoc_client_t *client, bson_t *options TSRMLS_DC) /* {{{ */
+{
+	bson_iter_t iter;
+	mongoc_read_prefs_t *new_rp;
+	const mongoc_read_prefs_t *old_rp;
+
+	if (!(old_rp = mongoc_client_get_read_prefs(client))) {
+		MONGOC_WARNING("Client does not have a read preference.");
```
bjori: I'm really not a fan of using MONGOC_WARNING(). I've used MONGOC_DEBUG() before for.. debugging. But this feels backwards..

It turns out php_phongo_log() doesn't actually throw exception or anything -- does even raise a php-error by default -- for _WARNING.
So something doesn't work and the user has no idea why?

php_phongo.c
```diff
@@ -1591,6 +1607,178 @@ void php_phongo_populate_default_ssl_ctx(php_stream_context *ctx, zval *driverOp
...
+	}
+
+	/* Warn on conflict, since read preference will be validated later */
+	if (mongoc_read_prefs_get_mode(new_rp) == MONGOC_READ_PRIMARY &&
+	    !bson_empty(mongoc_read_prefs_get_tags(new_rp))) {
+		MONGOC_WARNING("Primary read preference mode conflicts with tags.");
+	}
+
+	if (!mongoc_read_prefs_is_valid(new_rp)) {
+		mongoc_read_prefs_destroy(new_rp);
```
bjori: Shouldn't this throw exception?

bjori: Ah IC. the caller is reponsible for throwing exception when this function returns false.
But at that point, we've lost the reason -- how should the user know that the ReadPreferences were invalid?

jmikola: The exception thrown by the caller will report that we couldn't apply read preferences (since we know that `php_phongo_apply_rp_options_to_client()` is what failed).

php_phongo.c
jmikola: This is no difference than the warnings raised by libmongoc during URI parsing, though. They end up logged (if logging is enabled) or ignored. I can certainly remove these if you think it has no value, though.

This particular line (checking `old_rp`) is more of an assertion, because the client should always have a valid RP and WC struct.

php_phongo.c
bjori: But why weren't they applied?
I feel like we'll get some support questions about that right off the bat

php_phongo.c
bjori: The mongoc use of MONGOC_WARNING() for error messaging is problematic.
All we can do is log it and fail. And the failure contains no information as to why it failed. It just failed.

For example, monogc will issue MONGOC_WARNING() as to why a mongoc_bulk_operation_replace_one() was not added to the bulk. The function declaration is void, so its really difficult for a user to determine if an bulk operation has been added to the bulk or not, and doing so programmatically is a nightmare.

From phongo perspective, these log messages are just that.. logs. That are sent to /dev/null unless you turn on 'mongodb.debug' php.ini

I think, when something fails, or we don't do something, or interrupt early, or whatever, we should throw. throw an exception with why that happened.
If it isn't an exception worthy situation, then we should make it a debug or info level log message.
Why? Simply because WARNING runs the risk of the next person after you seeing this code assuming it means something that it doesn't. It doesn't trigger exception. It doesn't abort. It doesn't indicate any sorts of failure, and it doesn't tell the user anything by default.

bjori:
lgtm

What happened to the coveralls reporting? Would be nice to see the test coverage for this

PR #79, opened 2015-08-13
Trim #includes, since src/contrib is already an include path

jmikola:


bjori:
And coveralls is back!
lgtm

PR #81, opened 2015-08-18
Updates for 1.0.0alpha1 release

jmikola:


tests/functional/phpinfo-1.phpt
```diff
@@ -16,7 +16,7 @@ phpinfo();
 mongodb
 
 mongodb support => enabled
-mongodb version => 0.%d.%d
+mongodb version => %d.%d.%d
```
bjori: I'd just bump it to 1.. We do want to verify the version is atleast realistic..
It'd be embarrassing if the first stable release was 6.6.6

php_phongo.h
```diff
@@ -30,8 +30,8 @@
 extern zend_module_entry mongodb_module_entry;
 
 /* FIXME: Its annoying to bump version. Move into phongo_version.h.in */
-#define MONGODB_VERSION_S "0.6.4"
-#define MONGODB_STABILITY_S "devel"
+#define MONGODB_VERSION_S "1.0.0"
+#define MONGODB_STABILITY_S "alpha"
```
bjori: Remember to call it alpha*1* when you package it

tests/manager/manager-debug-001.phpt
```diff
@@ -29,7 +29,7 @@ foreach($content as $line) {
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
-[%s]     PHONGO: DEBUG   > Creating Manager, phongo-0.%d.%d[%s] - mongoc-1.%s, libbson-1.%s
+[%s]     PHONGO: DEBUG   > Creating Manager, phongo-%d.%d.%d[%s] - mongoc-1.%s, libbson-1.%s
```
jmikola: @bjori: Should I also rename "phongo", or do we expect this to differ between HHVM and PHP? AFAIK, we're only using "phongo" in the source filenames. This may be one of the few places it's still presented publicly.

bjori: I'd imagine hhvm would need to skip debug messaging like this, considering the trace will likely be subtly different.
I don't have a preference for replacing 'phongo' with %s or not

php_phongo.h
jmikola: :+1:

PR #82, opened 2015-08-19
Allow for PHONGO-SERVERS.json in ~ (and fix a test case name)

derickr:


tests/utils/basic.inc
```diff
@@ -1,8 +1,11 @@
 <?php
 require __DIR__ . "/" . "tools.php";
 
-
-$FILENAME = sys_get_temp_dir() . "/PHONGO-SERVERS.json";
+if (file_exists("{$_SERVER['HOME']}/PHONGO-SERVERS.json")) {
+	$FILENAME = "{$_SERVER['HOME']}/PHONGO-SERVERS.json";
```
bjori: That feels a littlebit dodge. Like the idea though
How about environment variable? PHONGO_SERVERS=/path/to/PHONGO-SERVERS.json

derickr: Pushed a new commit for that.

PR #85, opened 2015-08-21
Merge RELEASE notes

bjori:
Archive 0.y.z release notes into RELEASE-0 and start a RELEASE-1.0 file.
Also updated the release script to update (preprend) into RELEASE-x.y.

...And pumped version back to -dev after the 1.0.0-alpha2 release


Makefile.frag
```diff
@@ -137,11 +141,13 @@ docs:
...
 
-package.xml:
+package.xml: RELEASE
 	php bin/prep-release.php $(MONGODB_VERSION)-$(MONGODB_STABILITY)
 
 RELEASE:
-	@git log --pretty=format:"%ad  %an  <%ae>%n%x09* %s%n" --date short --since="$$(git show -s --format=%ad `git rev-list --tags --max-count=1`)" > RELEASE-$(MONGODB_VERSION)
+	echo "RELEASE $(MONGODB_VERSION)" >> RELEASE-$(MONGODB_VERSION)
+	echo "-------------" >> RELEASE-$(MONGODB_VERSION)
+	@git log --pretty=format:"%ad  %an  <%ae>%n%x09* %s%n" --date short --since="$$(git show -s --format=%ad `git rev-list --tags --max-count=1`)" >> RELEASE-$(MONGODB_VERSION)
```
jmikola: This currently generates a new release file with the full version. Is that intentional, as you only join it during packaging?

jmikola:
LGTM.

Makefile.frag
bjori: Thats the only way we can prepend stuff to a file, by creating a temporary file we append things to in order, and then rename it.

I just release i didn't suppress the command execution with @.. Let me amend that

PR #88, opened 2015-08-26
PHPC-335: Create debug handlers for BSON type classes

jmikola:
https://jira.mongodb.org/browse/PHPC-335

### DIFF
--- a/src/BSON/Javascript.c
+++ b/src/BSON/Javascript.c
@@ -44,6 +44,7 @@
 
 PHONGO_API zend_class_entry *php_phongo_javascript_ce;
 
+zend_object_handlers php_phongo_handler_javascript;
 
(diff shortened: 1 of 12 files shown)


derickr:
LGTM

bjori:
I'm allergic to skipifs.
Any reason bson-utcdatetime-002 can't do 
```
%rint(|string(13) "|%r1416445411987%r"|)%r
```
?

jmikola:
@bjori: It's totally possible to use that for the EXPECTF, but the main difference was in constructing the UTCDateTime objects. Those are really two different tests.

bjori:
Why ? The string format must work in both.
And this is testing a debug handler, the integer literal doesn't provide any value here

jmikola:
Good point. Will do: https://jira.mongodb.org/browse/PHPC-414

PR #89, opened 2015-08-27
Split tests for invalid arguments out into their own files, and ignore them for HHVM

derickr:
And we also add a cast because of HHVM not allowing current() on an object

### DIFF
--- a/tests/bson/bson-utcdatetime_error-001.phpt
+++ b/tests/bson/bson-utcdatetime_error-001.phpt
@@ -1,0 +1,24 @@
+--TEST--
+BSON BSON\UTCDateTime #001 error
+--INI--
+date.timezone=America/Los_Angeles
+--SKIPIF--
+<?php if (defined("HHVM_VERSION_ID")) exit("skip HHVM handles parameter parsing differently"); ?>
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; CLEANUP(STANDALONE) ?>
+--FILE--
+<?php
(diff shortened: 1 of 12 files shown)


jmikola:
LGTM.

tests/bson/bson-utcdatetime_error-001.phpt
```diff
@@ -0,0 +1,24 @@
...
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(STANDALONE);
+
+$classname = BSON_NAMESPACE . "\\UTCDateTime";
+
+throws(function() use($classname) {
+    $d = new $classname;
```
bjori: This use looks a bit weird? $classname isn't used outside of it

derickr: I don't quite understand what you are trying to say, but this is how all the tests were... I just split them up.

bjori: $classname is defined outside of the closure for no apparent reason, which then requires use($classname).

Defining $classname inside the closure seems like would be simpler.

It might have made sense in the original test cases if the variable was used in couple of places, but in this test where it is only used once from within the closure it looks funny to declare it outside and then use() it.

Did that clarify the comment?

jmikola: #92 should resolve this.

PR #90, opened 2015-08-27
PHP7: Change create_object functions to support new PHP7 function prototype

pboyd04:
This pull change the create_object prototypes for the object classes so that they should work with the new PHP7 create_object prototype as well as a few other miscellaneous changes to the files to compile against PHP7.

### DIFF
--- a/src/BSON/Binary.c
+++ b/src/BSON/Binary.c
@@ -84,7 +84,11 @@ PHP_METHOD(Binary, getData)
 		return;
 	}
 
+#if PHP_VERSION_ID >= 70000
+        RETURN_STRINGL(intern->data, intern->data_len);
+#else
 	RETURN_STRINGL(intern->data, intern->data_len, 1);
+#endif
 }
 /* }}} */
 /* {{{ proto integer Binary::getType()
@@ -130,18 +134,34 @@ static zend_function_entry php_phongo_binary_me[] = {
 
 
 /* {{{ php_phongo_binary_t object handlers */
-static void php_phongo_binary_free_object(void *object TSRMLS_DC) /* {{{ */
+static void php_phongo_binary_free_object(phongo_free_object_arg* object TSRMLS_DC) /* {{{ */
 {
-	php_phongo_binary_t *intern = (php_phongo_binary_t*)object;
+	php_phongo_binary_t *intern = Z_BINARY_OBJ(object);
 
-	zend_object_std_dtor(&intern->std TSRMLS_CC);
+        zend_object_std_dtor(&intern->std TSRMLS_CC);
 
-	if (intern->data) {
-		efree(intern->data);
-	}
-	efree(intern);
+        if (intern->data) {
+                efree(intern->data);
+        }
+#if PHP_VERSION_ID < 70000
+        efree(intern);
+#endif
 } /* }}} */
+#if PHP_VERSION_ID >= 70000
+zend_object* php_phongo_binary_create_object(zend_class_entry *class_type TSRMLS_DC) /* {{{ */
+{
+        php_phongo_binary_t *intern;
 
+        intern = (php_phongo_binary_t *)ecalloc(1, sizeof(php_phongo_binary_t)+zend_object_properties_size(class_type));
+
+        zend_object_std_init(&intern->std, class_type TSRMLS_CC);
+        object_properties_init(&intern->std, class_type);
+
+        intern->std.handlers = &php_phongo_handler_binary;
+
+        return &intern->std;
+} /* }}} */
+#else
 zend_object_value php_phongo_binary_create_object(zend_class_entry *class_type TSRMLS_DC) /* {{{ */
 {
 	zend_object_value retval;
@@ -158,6 +178,7 @@ zend_object_value php_phongo_binary_create_object(zend_class_entry *class_type T
 
 	return retval;
 } /* }}} */
+#endif
 
 HashTable *php_phongo_binary_get_debug_info(zval *object, int *is_temp TSRMLS_DC) /* {{{ */
 {
@@ -168,7 +189,11 @@ HashTable *php_phongo_binary_get_debug_info(zval *object, int *is_temp TSRMLS_DC
 	*is_temp = 1;
 	array_init_size(&retval, 2);
 
+#if PHP_VERSION_ID >= 70000
+        add_assoc_stringl_ex(&retval, ZEND_STRS("data"), intern->data, intern->data_len);
+#else
 	add_assoc_stringl_ex(&retval, ZEND_STRS("data"), intern->data, intern->data_len, 1);
+#endif
 	add_assoc_long_ex(&retval, ZEND_STRS("type"), intern->type);
 
 	return Z_ARRVAL(retval);
@@ -189,6 +214,10 @@ PHP_MINIT_FUNCTION(Binary)
 
 	memcpy(&php_phongo_handler_binary, phongo_get_std_object_handlers(), sizeof(zend_object_handlers));
 	php_phongo_handler_binary.get_debug_info = php_phongo_binary_get_debug_info;
+#if PHP_VERSION_ID >= 70000
+	php_phongo_handler_binary.free_obj = php_phongo_binary_free_object;
+	php_phongo_handler_binary.offset = XtOffsetOf(php_phongo_binary_t, std);
+#endif
 
 	zend_declare_class_constant_long(php_phongo_binary_ce, ZEND_STRL("TYPE_GENERIC"), BSON_SUBTYPE_BINARY TSRMLS_CC);
 	zend_declare_class_constant_long(php_phongo_binary_ce, ZEND_STRL("TYPE_FUNCTION"), BSON_SUBTYPE_FUNCTION TSRMLS_CC);
--- a/src/BSON/Binary.c
+++ b/src/BSON/Binary.c
@@ -189,11 +189,7 @@ HashTable *php_phongo_binary_get_debug_info(zval *object, int *is_temp TSRMLS_DC
 	*is_temp = 1;
 	array_init_size(&retval, 2);
 
-#if PHP_VERSION_ID >= 70000
-        add_assoc_stringl_ex(&retval, ZEND_STRS("data"), intern->data, intern->data_len);
-#else
-	add_assoc_stringl_ex(&retval, ZEND_STRS("data"), intern->data, intern->data_len, 1);
-#endif
+        ADD_ASSOC_STRING(&retval, "data", intern->data);
 	add_assoc_long_ex(&retval, ZEND_STRS("type"), intern->type);
 
 	return Z_ARRVAL(retval);
--- a/src/BSON/Regex.c
+++ b/src/BSON/Regex.c
@@ -169,20 +169,22 @@ static zend_function_entry php_phongo_regex_me[] = {
 
 
 /* {{{ php_phongo_regex_t object handlers */
-static void php_phongo_regex_free_object(void *object TSRMLS_DC) /* {{{ */
+static void php_phongo_regex_free_object(phongo_free_object_arg *object TSRMLS_DC) /* {{{ */
 {
-	php_phongo_regex_t *intern = (php_phongo_regex_t*)object;
+	php_phongo_regex_t *intern = Z_REGEX_OBJ(object);
 
 	zend_object_std_dtor(&intern->std TSRMLS_CC);
 
-    if (intern->pattern) {
-        efree(intern->pattern);
-    }
+	if (intern->pattern) {
+		efree(intern->pattern);
+	}
 
-    if (intern->flags) {
-        efree(intern->flags);
-    }
+	if (intern->flags) {
+		efree(intern->flags);
+	}
+#if PHP_VERSION_ID < 70000
 	efree(intern);
+#endif
 } /* }}} */
 
 #if PHP_VERSION_ID >= 70000
@@ -232,7 +234,8 @@ PHP_MINIT_FUNCTION(Regex)
 	zend_class_implements(php_phongo_regex_ce TSRMLS_CC, 1, php_phongo_type_ce);
         memcpy(&php_phongo_handler_regex, phongo_get_std_object_handlers(), sizeof(zend_object_handlers));
 #if PHP_VERSION_ID >= 70000
-        php_phongo_handler_regex.free_obj = php_phongo_regex_free_object;
+	php_phongo_handler_regex.free_obj = php_phongo_regex_free_object;
+	php_phongo_handler_regex.offset = XtOffsetOf(php_phongo_regex_t, std);
 #endif
 
 	return SUCCESS;
--- a/src/BSON/Regex.c
+++ b/src/BSON/Regex.c
@@ -44,6 +44,8 @@
 
 PHONGO_API zend_class_entry *php_phongo_regex_ce;
 
+zend_object_handlers php_phongo_handler_regex;
+
 /* {{{ proto BSON\Regex Regex::__construct(string $pattern, string $flags)
    Constructs a new regular expression. */
 PHP_METHOD(Regex, __construct)
@@ -85,8 +87,11 @@ PHP_METHOD(Regex, getPattern)
 		return;
 	}
 
-
+#if PHP_VERSION_ID >= 70000
+        RETURN_STRINGL(intern->pattern, intern->pattern_len);
+#else
 	RETURN_STRINGL(intern->pattern, intern->pattern_len, 1);
+#endif
 }
 /* }}} */
 /* {{{ proto void Regex::getFlags()
@@ -102,8 +107,11 @@ PHP_METHOD(Regex, getFlags)
 		return;
 	}
 
-
+#if PHP_VERSION_ID >= 70000
+        RETURN_STRINGL(intern->flags, intern->flags_len);
+#else
 	RETURN_STRINGL(intern->flags, intern->flags_len, 1);
+#endif
 }
 /* }}} */
 /* {{{ proto void Regex::__toString()
@@ -123,7 +131,11 @@ PHP_METHOD(Regex, __toString)
 
 
 	regex_len = spprintf(&regex, 0, "/%s/%s", intern->pattern, intern->flags);
+#if PHP_VERSION_ID >= 70000
+        RETVAL_STRINGL(regex, regex_len);
+#else
 	RETVAL_STRINGL(regex, regex_len, 0);
+#endif
 }
 /* }}} */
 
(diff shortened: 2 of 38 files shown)


bjori:
That looks awesome, thanks!

I was worried we might have to split the src/MongoDB/*.c files between PHP5 and PHP7 files to keep things clean, but this looks like it would work great.

I've got a chunk for php_phongo.c that I haven't pushed yet, I'll review this later today

bjori:
Looks like we have to update PHONGO_CE_INIT() too, which uses ZEND_ACC_FINAL_CLASS (appears it only called ZEND_ACC_FINAL now)

pboyd04:
Yeah I couldn't find the final class flag in the PHP 7 header files. So I wasn't quiet sure how to fix that part. I also debated creating a new macro in phongo_compat.h for the RETURN_STRINGL macro so that the entire code base didn't need the #if's around that macro. Otherwise I'm fairly happy with this patch.

src/BSON/Binary.c
```diff
@@ -142,6 +146,21 @@ static void php_phongo_binary_free_object(void *object TSRMLS_DC) /* {{{ */
...
+
+        intern = (php_phongo_binary_t *)ecalloc(1, sizeof(php_phongo_binary_t)+zend_object_properties_size(class_type));
+
+        zend_object_std_init(&intern->std, class_type TSRMLS_CC);
+        object_properties_init(&intern->std, class_type);
+
+        intern->std.handlers = &php_phongo_handler_binary;
+
+        return &intern->std;
+}
```
bjori: The closing folding markers, /* }}} */, are missing (looks like they are missing

src/MongoDB/Cursor.c
```diff
@@ -72,17 +72,32 @@ PHP_METHOD(Cursor, setTypeMap)
 
 static int php_phongo_cursor_to_array_apply(zend_object_iterator *iter, void *puser TSRMLS_DC) /* {{{ */
 {
-	zval **data, *return_value = (zval*)puser;
+#if PHP_VERSION_ID >= 70000
+	zval *data, *return_value = (zval*)puser;
+
+        data = iter->funcs->get_current_data(iter);
```
bjori: There is some odd alignment here. is this spaces vs tabs?

src/MongoDB/Cursor.c
```diff
@@ -72,17 +72,32 @@ PHP_METHOD(Cursor, setTypeMap)
...
 	}
+#if PHP_VERSION_ID >= 70000
+        if (data == NULL) {
+#else
 	if (data == NULL || *data == NULL) {
+#endif
 		return ZEND_HASH_APPLY_STOP;
 	}
+#if PHP_VERSION_ID >= 70000
+        Z_ADDREF_P(data);
```
bjori: Shouldn't this be Z_TRY_ADDREF_P() ? Are we sure its guaranteed to need a addref ?

src/MongoDB/Cursor.c
```diff
@@ -72,17 +72,32 @@ PHP_METHOD(Cursor, setTypeMap)
 
 static int php_phongo_cursor_to_array_apply(zend_object_iterator *iter, void *puser TSRMLS_DC) /* {{{ */
 {
-	zval **data, *return_value = (zval*)puser;
+#if PHP_VERSION_ID >= 70000
+	zval *data, *return_value = (zval*)puser;
```
bjori: Since the return_value is zval* in both cases, I think its better to split this line into two.
The original is also inconsistent with how we do this everywhere else in the code so I think that will only be a good thing.

I realize it will need an extra if version check for the data assignment below, but I think that will have to be ok

src/MongoDB/Exception/WriteException.c
```diff
@@ -57,8 +57,11 @@ PHP_METHOD(WriteException, getWriteResult)
 		return;
 	}
 
-
+#if PHP_VERSION_ID >= 70000
+        zend_read_property(php_phongo_writeexception_ce, getThis(), ZEND_STRL("writeResult"), 0 TSRMLS_CC, writeresult);
```
bjori: The TSRMLS_CC can be removed here, but more importantly is this right?
What does zend_read_property() return now?

src/MongoDB/Manager.c
```diff
@@ -58,8 +58,12 @@ PHP_METHOD(Manager, __construct)
 {
 	php_phongo_manager_t     *intern;
 	zend_error_handling       error_handling;
+#if PHP_VERSION_ID >= 70000
+        zend_string              *uri_string;
+#else
```
bjori: I don't think this was needed? the zend_parse_parameter() with 's' looks like would be fine as we are still passing char* to phongo_manager_init(), which this change does anyway too..

src/MongoDB/Manager.c
```diff
@@ -477,19 +504,29 @@ zend_object_value php_phongo_manager_create_object(zend_class_entry *class_type
 
 	return retval;
 } /* }}} */
+#endif
 
 bool phongo_add_server_debug(void *item, void *ctx)
 {
+#if PHP_VERSION_ID >= 70000
+        mongoc_server_description_t *server = item;
+        zval                        *retval = ctx;
```
bjori: Leave the server and retval outside of the version check, they are the same anyway so its nice to make that obvious

bjori:
Looks great!

I made a few comments, most of them excessively pedantic ;)
Let me know if you want to work on it, otherwise I can take over if you are done.

bjori:
As for a new macro wrapping the RETURN_STRINGL & add_assoc_stringl_ex() and friends, yeah, these inline version checks are annoying so a macro would have been nice for it (see phongo_compat.[c|h]), but its fine as it is now.

src/MongoDB/Manager.c
```diff
@@ -462,6 +474,21 @@ static void php_phongo_manager_free_object(void *object TSRMLS_DC) /* {{{ */
...
+{
+        php_phongo_manager_t *intern;
+
+        intern = (php_phongo_manager_t *)ecalloc(1, sizeof(php_phongo_manager_t)+zend_object_properties_size(class_type));
+
+        zend_object_std_init(&intern->std, class_type TSRMLS_CC);
+        object_properties_init(&intern->std, class_type);
+
+        intern->std.handlers = &php_phongo_handler_manager;
+
```
bjori: Wait. Something is off here. Where is the free_object_storage defined?
We need to call php_phongo_manager_free_object() to cleanup ourself, mongoc things in particular

src/MongoDB/Manager.c
pboyd04: I had a crash at this line when I used 's' and according to the core file the uri_string (char*) version was garbage. This on the other hand worked.

src/MongoDB/Cursor.c
pboyd04: The original implementation had Z_ADDREF_PP(data); I just kept as close to the original implementation as possible.

src/BSON/Binary.c
pboyd04: Updated. See new commit.

src/MongoDB/Exception/WriteException.c
pboyd04: Good catch. zend_read_property() still return the property. The added zval field is to indicate success or failure of the property read.

src/MongoDB/Manager.c
pboyd04: Updated.

src/MongoDB/Manager.c
bjori: Thats weird, but ok. I guess we can dig deeper into it later

src/MongoDB/Cursor.c
bjori: ADDREF on a scalar value in PHP7 and will cause an assert() or crash.
See the 5th bulletpoint on https://wiki.php.net/phpng-upgrading#zval

src/MongoDB/Manager.c
pboyd04: Yeah, the PHP7 porting documentation isn't that great yet. Apparently you need to add the free to the handlers. I've done this with BSON\Binary. Could you take a quick look and see if that looks right to you before I do the rest of the objects?

src/BSON/Binary.c
```diff
@@ -130,9 +134,15 @@ static zend_function_entry php_phongo_binary_me[] = {
 
 
 /* {{{ php_phongo_binary_t object handlers */
-static void php_phongo_binary_free_object(void *object TSRMLS_DC) /* {{{ */
+#if PHP_VERSION_ID >= 70000
+static inline php_phongo_binary_t * php_phongo_binary_t_fetch_object(zend_object *obj)
+{
+      return (php_phongo_binary_t *)((char *)obj - XtOffsetOf(php_phongo_binary_t, std));
+}
+
```
bjori: I did this already, see ./php_phongo_structs-7.h

You can now fetch the object using Z_BINARY_OBJ_P() (both in PHP5 and PHP7)

src/BSON/Binary.c
```diff
@@ -142,6 +152,32 @@ static void php_phongo_binary_free_object(void *object TSRMLS_DC) /* {{{ */
...
+        zend_object_std_init(&intern->std, class_type TSRMLS_CC);
+        object_properties_init(&intern->std, class_type);
+
+        intern->std.handlers = &php_phongo_handler_binary;
+
+        return &intern->std;
+} /* }}} */
+#else
+static void php_phongo_binary_free_object(void *object TSRMLS_DC) /* {{{ */
+{
```
bjori: oboy. If the only thing in these two free handlers is void* vs zend_object* which we then pass to Z_BINARY_OBJ_P() to get the php_phongo_binary_t*....

How about adding in phongo_compat.h a `phongo_free_object_arg` which is defined as void or zend_object?
Then we don't have to duplicate any of the free_object implementations at all.
Its a bit fugly, but I think its better then duplicating loads of code that risks diverging in the future

src/BSON/Binary.c
```diff
@@ -142,6 +152,32 @@ static void php_phongo_binary_free_object(void *object TSRMLS_DC) /* {{{ */
...
+        php_phongo_binary_t *intern = (php_phongo_binary_t*)object;
+
+        zend_object_std_dtor(&intern->std TSRMLS_CC);
+
+        if (intern->data) {
+                efree(intern->data);
+        }
+        efree(intern);
```
bjori: According to https://wiki.php.net/phpng-upgrading#zend_object_handlers this efree() should be removed

src/BSON/Binary.c
```diff
@@ -189,6 +230,9 @@ PHP_MINIT_FUNCTION(Binary)
 
 	memcpy(&php_phongo_handler_binary, phongo_get_std_object_handlers(), sizeof(zend_object_handlers));
 	php_phongo_handler_binary.get_debug_info = php_phongo_binary_get_debug_info;
+#if PHP_VERSION_ID >= 70000
+        php_phongo_handler_binary.free_obj = php_phongo_binary_free_object;
```
bjori: We also need to add
```
php_phongo_handler_binary.offset = XtOffsetOf(php_phongo_binary_t, std);
```

bjori:
btw. if you are rebasing this pr ontop of current PHP7, beware.. I changed config.m4 to only build php_phongo.c for now :)

src/BSON/Binary.c
pboyd04: That macro only works if the pointer is a zval containing an zend_object that is a php_phongo_binary_t* in this case we have the raw zend_object... So I think we need a new macro.

Compile errors with that macro:
In file included from /usr/include/php/Zend/zend.h:31:0,
                 from /usr/include/php/main/php.h:35,
                 from /root/mongo-php-driver-prototype/src/BSON/Binary.c:32:
/root/mongo-php-driver-prototype/src/BSON/Binary.c: In function 'php_phongo_binary_free_object':
/usr/include/php/Zend/zend_types.h:529:31: error: 'zend_object' has no member named 'value'
 #define Z_OBJ(zval)     (zval).value.obj
                               ^
/usr/include/php/Zend/zend_types.h:530:28: note: in expansion of macro 'Z_OBJ'
 #define Z_OBJ_P(zval_p)    Z_OBJ(*(zval_p))
                            ^
./php_phongo_classes.h:36:64: note: in expansion of macro 'Z_OBJ_P'
 # define Z_BINARY_OBJ_P(zv)            php_binary_fetch_object(Z_OBJ_P(zv));
                                                                ^
/root/mongo-php-driver-prototype/src/BSON/Binary.c:139:32: note: in expansion of macro 'Z_BINARY_OBJ_P'
  php_phongo_binary_t *intern = Z_BINARY_OBJ_P(object);
                                ^
make: *** [src/BSON/Binary.lo] Error 1

bjori: Ahhhhh. You are right, sorry.

Maybe add a new one then, one that does the cast (for php5) or calls _fetch_object(zo) for php7.
```
#define Z_OBJ_BINARY(zo) php_binary_fetch_object(zo)
else
#define Z_OBJ_BINARY(zo) (php_phongo_binary_t *)zo
```
should work?

src/MongoDB/Exception/WriteException.c
```diff
@@ -50,15 +50,20 @@ PHONGO_API zend_class_entry *php_phongo_writeexception_ce;
...
+#endif
 
 
 	if (zend_parse_parameters_none() == FAILURE) {
 		return;
 	}
 
-
+#if PHP_VERSION_ID >= 70000
+        writeresult = zend_read_property(php_phongo_writeexception_ce, getThis(), ZEND_STRL("writeResult"), 0 TSRMLS_CC, &rv);
```
bjori: I'd just drop the TSRMLS_CC here. It should (generally) be last, but in 7 it doesn't do anything anymore anyway.

src/BSON/Binary.c
```diff
@@ -168,7 +205,11 @@ HashTable *php_phongo_binary_get_debug_info(zval *object, int *is_temp TSRMLS_DC
 	*is_temp = 1;
 	array_init_size(&retval, 2);
 
+#if PHP_VERSION_ID >= 70000
+        add_assoc_stringl_ex(&retval, ZEND_STRS("data"), intern->data, intern->data_len);
+#else
 	add_assoc_stringl_ex(&retval, ZEND_STRS("data"), intern->data, intern->data_len, 1);
```
bjori: I change my mind on this one when I realized we already had similar macro for dealing with int64 (ADD_INDEX_INT64()/ADD_ASSOC_INT64()), so I've added ADD_ASSOC_STRING().
See https://jira.mongodb.org/browse/PHPC-385

bjori:
Let me know if I'm working you to hard ;)
I can take over and finish this on Monday

pboyd04:
Nope I just pushed updated free_obj prototype with the suggestions.

bjori:
I'm not sure what happened, but something happened to this pr containing existing changes in the PHP7 branch making it weird to review.

Could you try rebasing it ontop of PHP7?

pboyd04:
Fixed, looks like I didn't have the latest change to phongo_compat.h

bjori:
Travis claims PHP5 builds are segfaulting now

src/BSON/Regex.c
```diff
@@ -202,7 +230,10 @@ PHP_MINIT_FUNCTION(Regex)
 	php_phongo_regex_ce = zend_register_internal_class(&ce TSRMLS_CC);
 
 	zend_class_implements(php_phongo_regex_ce TSRMLS_CC, 1, php_phongo_type_ce);
-
+        memcpy(&php_phongo_handler_regex, phongo_get_std_object_handlers(), sizeof(zend_object_handlers));
+#if PHP_VERSION_ID >= 70000
+        php_phongo_handler_regex.free_obj = php_phongo_regex_free_object;
```
bjori: Missing .offset assignment

pboyd04:
Sorry fixing merge conflicts made me push some stuff before I was ready.

pboyd04:
Ok I think that's it. Once we get PHPC-373 and the removal of zval_used_for_init then I think the build will work and the library will have all the symbols. Who knows if the tests will pass, but hey it's a start.

bjori:
I haven't done another detailed pass yet, skimming over it - looks great.
I've got two comments;
- This contains far more then just object creation/destructing -- could the commits be organized into specifics changes, and squashing related ones?
- The `Z_COMMAND_OBJ()` (and similar) should be named `Z_OBJ_COMMAND(zo)`
This is to differentiate between `zval`, and `zend_object`, (`Z_OBJ_`=`zend_object`=`zo`).
the `Z_*_P()` macros mean `zval*`, and `Z_*()` generally mean `zval`

I probably won't have time to look at this again until Monday :/
Thanks for all your work!

src/bson.c
```diff
@@ -215,7 +221,13 @@ bool php_phongo_bson_visit_binary(const bson_iter_t *iter ARG_UNUSED, const char
 	TSRMLS_FETCH();
 
 	if (v_subtype == 0x80 && strcmp(key, PHONGO_ODM_FIELD_NAME) == 0) {
+#if PHP_VERSION_ID >= 70000
+                zend_string *zs_classname = zend_string_init((char *)v_binary, v_binary_len, 0);
```
bjori: zend_string_init() expects a const char*

src/bson.c
```diff
@@ -800,8 +876,15 @@ PHONGO_API void zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *
 
 				if (instanceof_function(Z_OBJCE_P(data), php_phongo_persistable_ce TSRMLS_CC)) {
 					if (flags & PHONGO_BSON_ADD_ODS) {
+#if PHP_VERSION_ID >= 70000
+						zend_string* field_name = zend_string_init(PHONGO_ODM_FIELD_NAME, sizeof(PHONGO_ODM_FIELD_NAME)-1, 0);
+						bson_append_binary(bson, PHONGO_ODM_FIELD_NAME, -1, 0x80, (const uint8_t *)ZSTR_VAL(Z_OBJCE_P(data)->name), ZSTR_LEN(Z_OBJCE_P(data)->name));
+                                                zend_hash_del(ht_data, field_name);
```
bjori: Use the zend_hash_str_del() rather then allocating new zend_string all the time

pboyd04:
Ok things have been squashed and Jira issues opened. Let me know what else you see.

bjori:
The individual commits are still not very concrete self-standing changes that can be merged.
And there is a lot of back-and-forward going in, along with unrelated fixes.

There are still several outstanding comments, such as closing folding markers, object offset registration and object free handlers.

Much of this is now directly in the path of things I hoped to work on today I'll see if I can rework these commits a bit and try to merge as much as I can without destabilizing the upcoming PHP5-stable release.

Thanks for all your work, it helps a lot !

pboyd04:
Glad it helps. Good luck with your work today!

PR #91, opened 2015-08-28
Skip phpinfo() and hex-dump tests for HHVM, with replacements for the latter

derickr:


jmikola:
LGTM.

Any plans to revise the `__pclass` order in HHVM long-term?

derickr:
Not really... any plays to revise the ```__pclass``` order in Phongo long-term? :-) As I mentioned before, it's tricky for me to change without making it a lot slower.

jmikola:
Agree to disagree :)

I forgot about the performance issue. I'll try to stop bringing this up in conversation going forward.

PR #93, opened 2015-09-01
Object property names missing due to lack of key length

pboyd04:
The key names aren't showing up because phongo_bson_append() is being passed a 0 length value.


pboyd04:
And yes that means I got it partly running.

bjori:
I've got handful of more fixes too.. Even starting to pass some tests now!

bjori:
Thanks!

PR #94, opened 2015-09-01
Fix phongo_stream_destroy crash on PHP7

pboyd04:
The debug print inside of phongo_stream_destroy() is causing a crash on PHP7. This patch turns that debug print off in PHP7.


bjori:
Hmm. Why is it crashing?
I thought the macro retrieved an identifying integer in PHP7 too?

bjori:
It is actually very important ;)
Due to the way we handle stream we need to confirm we aren't creating and destroying connections that should be left alone.

If you can repro the crash, can you in gdb:
(gdb) p stream->res->handle
(gdb) ptype stream->res->handle
(gdb) p *stream->res

pboyd04:
Neither stream_wrap or base_stream (the variables in the function have a res member). And the variable stream is null.

(gdb) p stream
$1 = 0
(gdb) p base_stream
$2 = (php_phongo_stream_socket *) 0x7f404ae67820
(gdb) p base_stream->res->handle
There is no member named res.
(gdb) p stream_wrap
$3 = (mongoc_stream_t *) 0x7f404ae67820
(gdb) p stream_wrap->res
There is no member named res.

pboyd04:
And res in the base_stream->stream is also NULL:
(gdb) p base_stream->stream->res
$5 = (zend_resource *) 0x0

pboyd04:
Here is the whole base_stream->stream variable contents:
(gdb) p *(base_stream->stream)
$8 = {ops = 0x7f404c09d940, abstract = 0x561587e7a660, readfilters = {head = 0x0, tail = 0x0, stream = 0x561587e6d1a0}, writefilters = {head = 0x0,
    tail = 0x0, stream = 0x561587e6d1a0}, wrapper = 0x0, wrapperthis = 0x0, wrapperdata = {value = {lval = 0, dval = 0, counted = 0x0, str = 0x0,
      arr = 0x0, obj = 0x0, res = 0x0, ref = 0x0, ast = 0x0, zv = 0x0, ptr = 0x0, ce = 0x0, func = 0x0, ww = {w1 = 0, w2 = 0}}, u1 = {v = {type = 0 '\000',
        type_flags = 0 '\000', const_flags = 0 '\000', reserved = 0 '\000'}, type_info = 0}, u2 = {var_flags = 0, next = 0, cache_slot = 0, lineno = 0,
      num_args = 0, fe_pos = 0, fe_iter_idx = 0}}, fgetss_state = 0, is_persistent = 1, mode = "r+", '\000' <repeats 13 times>, res = 0x0, in_free = 0,
  fclose_stdiocast = 0, stdiocast = 0x0, __exposed = 1, orig_path = 0x0, ctx = 0x0, flags = -2147483648, eof = 0, position = 406,
  readbuf = 0x561587e80070 "Y", readbuflen = 8192, readpos = 89, writepos = 89, chunk_size = 8192, enclosing_stream = 0x0}

bjori:
Ok. Thats weird.

Until we figure this out, can you instead update the PHONGO_STREAM_ID() for PHP7 to return "42" ?
And then file a ticket to look into how to get the uniq ID from a stream.

phongo_compat.h
```diff
@@ -156,7 +156,10 @@
 # define SUPPRESS_UNUSED_WARNING(x)
 # define DECLARE_RETURN_VALUE_USED int return_value_used = 1;
 # define EXCEPTION_P(_ex, _zp) ZVAL_OBJ(_zp, _ex)
-# define PHONGO_STREAM_ID(stream) stream->res->handle
+//TODO - PHPC-401 - Revert this change when this issue is closed
+//# define PHONGO_STREAM_ID(stream) stream->res->handle
+# define PHONGO_STREAM_ID(stream) 42
+//End Workaround
```
bjori: No need for the inline TODO, we have the ticket.
Besides, // is not a valid comment in c89 :)

PR #96, opened 2015-09-02
Fixed getServers test to include milliseconds in UTCDate dumps

derickr:


jmikola:
LGTM. Did I miss adding this to #88?

derickr:
Yeah, but I missed it in my original patch too as I wasn't testing with replica sets yet.

PR #97, opened 2015-09-02
PHP 7 Support for phongo!

bjori:


.travis.yml
```diff
@@ -1,9 +1,10 @@
 language: php
 php:
+  - '7.0'
+  - '5.6'
```
derickr: Why is 5.6 in here twice? :-)

bjori: Heh. Good catch, no idea. 
Travis only seems to be running 5.6 once though, so thats cool

phongo_compat.h
```diff
@@ -112,3 +145,76 @@
 # endif
 #endif
 
+#if PHP_VERSION_ID >= 70000
+# define phongo_char zend_string
+# define phongo_char_pdup(str) pestrdup(filename->val, 1)
```
derickr: Shouldn't this be "str->val" instead of "filename->val"?

phongo_compat.h
```diff
@@ -112,3 +145,76 @@
...
+# define ADD_ASSOC_BOOL_EX(_zv, _key, _value) add_assoc_bool_ex(_zv, ZEND_STRL(_key), _value);
+# define phongo_free_object_arg zend_object
+# define phongo_zpp_char_len size_t
+# define ZEND_HASH_APPLY_COUNT(ht) (ht)->u.v.nApplyCount
+# define PHONGO_RETVAL_STRINGL(s, slen) RETVAL_STRINGL(s, slen)
+# define PHONGO_RETURN_STRINGL(s, slen) RETURN_STRINGL(s, slen)
+# define PHONGO_RETURN_STRING(s) RETURN_STRING(s)
+#else
+# define phongo_char char
+# define phongo_char_pdup(str) pestrdup(filename, 1)
```
jmikola: Should `filename` be `str` here?

phongo_compat.h
bjori: I bet its only used in one place, and "filename" is the variable that is passed to it

php_phongo.c
```diff
@@ -45,13 +45,15 @@
...
-#include <ext/date/php_date.h>
 /* Stream wrapper */
 #include <main/php_streams.h>
 #include <main/php_network.h>
 /* Debug log writing */
 #include <main/php_open_temporary_file.h>
+/* For formating timestamp in the log */
+#include <ext/date/php_date.h>
+/* String manipulation */
+#include <Zend/zend_string.h>
```
jmikola: Should this be conditional for [PHP 7](http://lxr.php.net/xref/PHP_MASTER/Zend/zend_string.h) or [PHP 5.x](http://lxr.php.net/xref/PHP_5_6/Zend/zend_string.h)? The implementations are quite different.

bjori: I don't quite see the benefit of making it conditional?
Its probably even implicitly added already, we use str_efree() from that header file, and phongo_compat.h seems to include it.... So maybe just remove it if you want to reduce the includes.

jmikola:
Superseded by #136

PR #98, opened 2015-09-02
PHPC-381: Server::executeQuery/Command() should take a read preference

jmikola:
https://jira.mongodb.org/browse/PHPC-381

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -105,7 +105,7 @@ zend_class_entry* phongo_exception_from_phongo_domain(php_phongo_error_domain_t 
 			return php_phongo_connectionexception_ce;
 	}
 
-	MONGOC_ERROR("Resolving unknown exception domain!!!");
+	MONGOC_ERROR("Resolving unknown phongo error domain: %d", domain);
 	return php_phongo_runtimeexception_ce;
 }
 zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_domain_t */ domain, uint32_t /* mongoc_error_code_t */ code)
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -106,7 +106,7 @@ zend_class_entry* phongo_exception_from_phongo_domain(php_phongo_error_domain_t 
 	}
 
 	MONGOC_ERROR("Resolving unknown exception domain!!!");
-	return spl_ce_RuntimeException;
+	return php_phongo_runtimeexception_ce;
 }
 zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_domain_t */ domain, uint32_t /* mongoc_error_code_t */ code)
 {
@@ -166,7 +166,7 @@ zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_d
 			return phongo_ce_mongo_connection_exception;
 #endif
 		default:
-			return spl_ce_RuntimeException;
+			return php_phongo_runtimeexception_ce;
 	}
 }
 PHONGO_API zval* phongo_throw_exception(php_phongo_error_domain_t domain TSRMLS_DC, const char *format, ...)
--- a/src/MongoDB/Server.c
+++ b/src/MongoDB/Server.c
@@ -58,7 +58,7 @@ PHP_METHOD(Server, __construct)
 	phongo_throw_exception(PHONGO_ERROR_RUNTIME TSRMLS_CC, "Accessing private constructor");
 }
 /* }}} */
-/* {{{ proto MongoDB\Driver\Cursor Server::executeCommand(string $db, MongoDB\Driver\Command $command)
+/* {{{ proto MongoDB\Driver\Cursor Server::executeCommand(string $db, MongoDB\Driver\Command $command[, MongoDB\Driver\ReadPreference $readPreference = null]))
    Executes a command on this server */
 PHP_METHOD(Server, executeCommand)
 {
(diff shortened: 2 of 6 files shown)


tests/server/server-executeCommand-002.phpt
```diff
@@ -0,0 +1,78 @@
...
+
+$rp = new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_SECONDARY);
+$secondary = $manager->selectServer($rp);
+
+$command = new MongoDB\Driver\Command(array('profile' => 2));
+$cursor = $secondary->executeCommand(DATABASE_NAME, $command);
+$result = current($cursor->toArray());
+
+if (empty($result->ok)) {
+    exit("Could not set profile level\n");
```
bjori: exit?

jmikola: I just wanted to abort the test at this point. Throwing an exception would also work, but this seemed simpler.

php_phongo.c
```diff
@@ -105,7 +105,7 @@ zend_class_entry* phongo_exception_from_phongo_domain(php_phongo_error_domain_t
 			return php_phongo_connectionexception_ce;
 	}
 
-	MONGOC_ERROR("Resolving unknown exception domain!!!");
+	MONGOC_ERROR("Resolving unknown phongo error domain!");
```
bjori: Since you are updatingthe line, maybe include the domain in the message?

tests/server/server-executeQuery_error-001.phpt
```diff
@@ -0,0 +1,24 @@
...
+echo throws(function() use ($secondary) {
+    $secondary->executeQuery(NS, new MongoDB\Driver\Query(array("x" => 1)));
+}, "MongoDB\Driver\Exception\RuntimeException"), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+OK: Got MongoDB\Driver\Exception\RuntimeException
+not master and slaveOk=false
```
bjori: Thats surprising. I thought SS dictated all queries on secondaries had to be slaveOkay

bjori:
The tests don't seem to be testing what happens when you pass in a $rp that doesn't match the readPreference of the server.
The $rp used seems to be the one passed to ->selectServer() that got us this rp to begin with.

Other then that, the change itself is lgtm

tests/server/server-executeQuery_error-001.phpt
jmikola: https://jira.mongodb.org/browse/CDRIVER-829

jmikola: I since changed this test to use a conflicting read preference as you requested in an earlier comment. And per the SS spec, the primary read preference should still trigger a "not master and slaveOk=false" error on the secondary, which means this test should still work once [CDRIVER-829](https://jira.mongodb.org/browse/CDRIVER-829) is implemented.

jmikola:
> The tests don't seem to be testing what happens when you pass in a $rp that doesn't match the readPreference of the server.

Revised `server-executeCommand_error-001.phpt` and `server-executeQuery_error-001.phpt` to take care of that :+1:

src/MongoDB/Server.c
```diff
@@ -391,11 +393,13 @@ PHP_METHOD(Server, isPassive)
 ZEND_BEGIN_ARG_INFO_EX(ai_Server_executeCommand, 0, 0, 2)
 	ZEND_ARG_INFO(0, db)
 	ZEND_ARG_OBJ_INFO(0, command, MongoDB\\Driver\\Command, 0)
+	ZEND_ARG_OBJ_INFO(0, readPreference, MongoDB\\Driver\\ReadPreference, 1)
```
derickr: Is there any specific reason why this is ,1 at the end? (And below as well)

jmikola: I had to double-check, but according to [this](http://lxr.php.net/xref/PHP_5_6/Zend/zend_API.h#103) the 1 is for allowing null. And the ReadPreference args here allow null, just like in Manager.

bjori: Optional parameters should allow `null` for most of our arguments, especially `objects` where it must.
We don't want to be in the situation again where users have to
```
if ($arg2) {
 $obj->method($arg1, $arg2);
} else {
 $obj->method($arg1);
}
```

tests/server/server-executeCommand-002.phpt
```diff
@@ -0,0 +1,78 @@
+--TEST--
+MongoDB\Driver\Server::executeCommand() takes a read preference
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; NEEDS("REPLICASET"); ?>
```
bjori: I'm not sure how it happened, but I somehow stopped using `__DIR__ . "/" ."../utils/basic-skipif.inc";` and regret it a lot now.

Having the leading `/` in quotes of its own allows the second component to be fully qualified filename relative to the cwd.
That means all editors will be able to jump to that file trivially. As it is now, the leading `/` makes it into nonexisting absolute path :(

tests/server/server-executeCommand-002.phpt
```diff
@@ -0,0 +1,78 @@
...
+
+var_dump($profileEntry->command);
+
+$command = new MongoDB\Driver\Command(array('profile' => 0));
+$cursor = $secondary->executeCommand(DATABASE_NAME, $command);
+$result = current($cursor->toArray());
+
+if (empty($result->ok)) {
+    exit("Could not set profile level\n");
+}
```
bjori: This should say something on success to verify it succeeded.
I'm not 100% how run-tests really deals with exit messages.
Also, what if it segfaults while checking `$result->ok`? now this test would pass

tests/server/server-executeQuery_error-001.phpt
bjori: I'm pretty sure this is fixed in mongoc-1.2-dev now, meaning it won't raise "not master" but set the slaveOk.

A query on a standalone node is readPreference=primary, so I think that is correct behaviour

tests/server/server-executeQuery-005.phpt
```diff
@@ -0,0 +1,66 @@
...
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(REPLICASET);
+
+$rp = new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_SECONDARY);
+$secondary = $manager->selectServer($rp);
+
+$command = new MongoDB\Driver\Command(array('profile' => 2));
```
bjori: Does this clear out the profile collection?
If it doesn't, then as long as this test has passed once, it will pretty much always pass again even though it should break, as the record is always there

bjori:
The code change itself lgtm.
The tests I'm a little skeptical about. But if they work great.

src/MongoDB/Server.c
jmikola: :+1: That's the idea. So really nothing to fix here.

tests/server/server-executeCommand-002.phpt
jmikola: Is this just a matter of supporting certain IDEs and editor plugins? I personally never had such functionality, so I didn't notice this.

bjori: I think most IDEs allow you to hover over a filename and "click it" to goto the file.
In vim it is `gf` ("goto file"), or `gF` (if the string is in the format of `filename:linenumber`, then goto file on line `linenumber`)

tests/server/server-executeCommand-002.phpt
jmikola: There is still a `===DONE===` line below, so I don't see the concern with this inadvertently passing on a segfault.

bjori: Oh. right. Thats why we have it! :dancers:

tests/server/server-executeQuery-005.phpt
jmikola: https://jira.mongodb.org/browse/PHPC-411

tests/server/server-executeCommand-002.phpt
derickr: I did not know about gF!

PR #99, opened 2015-09-03
Bump libmongoc to 1.2-beta1

bjori:


jmikola:
LGTM. Nothing for libbson, correct?

bjori:
correct. thats as up2date as it can be

PR #102, opened 2015-09-09
PHPC-408: Ensure object_to_bson() always creates a BSON document

jmikola:
https://jira.mongodb.org/browse/PHPC-408

### DIFF
--- a/src/bson.c
+++ b/src/bson.c
@@ -717,6 +717,11 @@ void object_to_bson(zval *object, php_phongo_bson_flags_t flags, const char *key
 			return;
 		}
 	}
+
+	/* Even if we don't know how to encode the object, ensure that we at least
+	 * create an empty BSON document. */
+	bson_append_document_begin(bson, key, key_len, &child);
+	bson_append_document_end(bson, &child);
 }
 void phongo_bson_append(bson_t *bson, php_phongo_bson_flags_t flags, const char *key, long key_len, int entry_type, zval *entry TSRMLS_DC)
 {
--- a/tests/bson/bson-utcdatetime-001.phpt
+++ b/tests/bson/bson-utcdatetime-001.phpt
@@ -51,8 +51,8 @@ Test#1 { "0" : { "$date" : 1416445411987 } }
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 bool(true)
-Test#2 { }
-string(3) "{ }"
-string(3) "{ }"
+Test#2 { "0" : {  } }
+string(14) "{ "0" : {  } }"
+string(14) "{ "0" : {  } }"
 bool(false)
 ===DONE===
--- a/tests/bson/bson-utcdatetime-001.phpt
+++ b/tests/bson/bson-utcdatetime-001.phpt
@@ -12,21 +12,21 @@ $manager = new MongoDB\Driver\Manager(STANDALONE);
 
 $classname = BSON_NAMESPACE . "\\UTCDateTime";
 $utcdatetime = new $classname("1416445411987");
-$result = $manager->executeInsert(NS, array('_id' => 1, 'x' => $utcdatetime));
+
+$manager->executeInsert(NS, array('_id' => 1, 'x' => $utcdatetime));
+
 $query = new MongoDB\Driver\Query(array('_id' => 1));
 $cursor = $manager->executeQuery(NS, $query);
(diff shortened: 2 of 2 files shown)


tests/bson/bson-utcdatetime-001.phpt
```diff
@@ -12,21 +12,21 @@ $manager = new MongoDB\Driver\Manager(STANDALONE);
...
+$datetime = $utcdatetime->toDateTime();
+var_dump($datetime->format(DATE_RSS));
+var_dump((string) $utcdatetime);
 
 $tests = array(
     array($utcdatetime),
-    array($array[0]->x),
-    array($date),
+    array($results[0]->x),
+    array($datetime),
```
bjori: This really wasn't intentional.

I would remove this check from the test, and make a totally new testcase for the exact ticket.

bjori:
I don't think this is a good idea. Silently serializing objects as empty?
Its a clear data loss at least without any indication of the error.

I would rather see an exception thrown.

tests/bson/bson-utcdatetime-001.phpt
```diff
@@ -51,8 +51,8 @@ Test#1 { "0" : { "$date" : 1416445411987 } }
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 bool(true)
-Test#2 { }
-string(3) "{ }"
-string(3) "{ }"
+Test#2 { "0" : {  } }
+string(14) "{ "0" : {  } }"
+string(14) "{ "0" : {  } }"
```
bjori: This doesn't make a whole lot of sense to me.
```
php -r '$d = new DateTime("now"); var_dump($d);'                                                                                                                                                                     <<<
object(DateTime)#1 (3) {
  ["date"]=>
  string(26) "2015-09-09 16:13:49.000000"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(19) "America/Los_Angeles"
}
```

Why wouldn't at least the public properties be serialized?

derickr: No... DateTime only creates those when you make a var_dump() or use the debug handler.... (yes, that is odd).

tests/bson/bson-utcdatetime-001.phpt
```diff
@@ -12,21 +12,21 @@ $manager = new MongoDB\Driver\Manager(STANDALONE);
 
 $classname = BSON_NAMESPACE . "\\UTCDateTime";
 $utcdatetime = new $classname("1416445411987");
-$result = $manager->executeInsert(NS, array('_id' => 1, 'x' => $utcdatetime));
-$query = new MongoDB\Driver\Query(array('_id' => 1));
-$cursor = $manager->executeQuery(NS, $query);
-$array = iterator_to_array($cursor);
```
derickr: Please don't change test cases when changing code that affects the test. That makes it really difficult to see whether you broke the test case, or the code.

jmikola: I isolated this change in https://github.com/jmikola/mongo-php-driver-prototype/commit/870ceb6a5f38453a22f2050dd8c4e5cfec4548d1. The only functional change was removing the `$result` assignment after the insert (since it was never used) and changing the DateTime `echo` to a `var_dump()` with a string cast.

jmikola:
> I don't think this is a good idea. Silently serializing objects as empty? Its a clear data loss at least without any indication of the error. I would rather see an exception thrown.

As @derickr mentioned, https://github.com/mongodb-labs/mongo-php-driver-prototype/pull/103 is my proposed change for that. I just broke these into separate PRs to address the issue Derick raised in PHPC-408 independently.

src/bson.c
```diff
@@ -717,6 +717,11 @@ void object_to_bson(zval *object, php_phongo_bson_flags_t flags, const char *key
 			return;
 		}
 	}
+
+	/* Even if we don't know how to encode the object, ensure that we at least
+	 * create an empty BSON document. */
+	bson_append_document_begin(bson, key, key_len, &child);
+	bson_append_document_end(bson, &child);
```
derickr: I thought you were going to throw an exception in this case?

jmikola: https://github.com/mongodb-labs/mongo-php-driver-prototype/pull/103/files#diff-f7d29a680148f52d6601f59ed787f577R712 throws an exception for an unsupported BSON Type, but any other object will ideally get its public properties serialized.

PR #103, opened 2015-09-09
PHPC-410: object_to_bson() behavior for non-stdClass and unsupported types

jmikola:
https://jira.mongodb.org/browse/PHPC-410

Based on #102

### DIFF
--- a/tests/bson/bson-fromPHP-002.phpt
+++ b/tests/bson/bson-fromPHP-002.phpt
@@ -1,0 +1,30 @@
+--TEST--
+BSON\fromPHP(): Encoding non-Persistable objects as a document
+--INI--
+date.timezone=America/Los_Angeles
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"?>
+--FILE--
+<?php
+use MongoDB\BSON as BSON;
+
+require_once __DIR__ . "/../utils/basic.inc";
+
+class MyDocument {
+    private $foo = 1;
+    protected $bar = 2;
+    public $baz = 3;
+}
+
+$s = fromPHP(new MyDocument);
+echo "Test ", toJSON($s), "\n";
+hex_dump($s);
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+Test { "foo" : 1, "bar" : 2, "baz" : 3 }
+     0 : 20 00 00 00 10 66 6f 6f 00 01 00 00 00 10 62 61  [ ....foo......ba]
+    10 : 72 00 02 00 00 00 10 62 61 7a 00 03 00 00 00 00  [r......baz......]
+===DONE===
--- a/tests/bson/bson-fromPHP_error-002.phpt
+++ b/tests/bson/bson-fromPHP_error-002.phpt
@@ -1,0 +1,32 @@
+--TEST--o
+BSON\fromPHP(): Encoding unknown Type objects as a document field value
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"?>
+--FILE--
+<?php
+use MongoDB\BSON as BSON;
+
+require_once __DIR__ . "/../utils/basic.inc";
+
+class UnknownType implements BSON\Type {}
+
+$tests = array(
+    array(new UnknownType()),
+    array('x' => new UnknownType()),
+);
+
+foreach ($tests as $document) {
+    echo throws(function() use ($document) {
+        fromPHP($document);
+    }, 'MongoDB\Driver\Exception\UnexpectedValueException'), "\n";
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+OK: Got MongoDB\Driver\Exception\UnexpectedValueException
+Unexpected %SBSON\Type instance: UnknownType
+OK: Got MongoDB\Driver\Exception\UnexpectedValueException
+Unexpected %SBSON\Type instance: UnknownType
+===DONE===
--- a/tests/bson/bson-utcdatetime-001.phpt
+++ b/tests/bson/bson-utcdatetime-001.phpt
@@ -51,8 +51,8 @@ Test#1 { "0" : { "$date" : 1416445411987 } }
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 bool(true)
-Test#2 { "0" : {  } }
-string(14) "{ "0" : {  } }"
-string(14) "{ "0" : {  } }"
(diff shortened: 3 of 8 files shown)


tests/bson/bson-fromPHP-002.phpt
```diff
@@ -0,0 +1,55 @@
...
+Test { }
+     0 : 05 00 00 00 00                                   [.....]
+Test { "date" : "2014-11-20 01:03:31.987000000", "timezone_type" : 1, "timezone" : "+00:00" }
+     0 : 55 00 00 00 02 64 61 74 65 00 1e 00 00 00 32 30  [U....date.....20]
+    10 : 31 34 2d 31 31 2d 32 30 20 30 31 3a 30 33 3a 33  [14-11-20 01:03:3]
+    20 : 31 2e 39 38 37 30 30 30 30 30 30 00 10 74 69 6d  [1.987000000..tim]
+    30 : 65 7a 6f 6e 65 5f 74 79 70 65 00 01 00 00 00 02  [ezone_type......]
+    40 : 74 69 6d 65 7a 6f 6e 65 00 07 00 00 00 2b 30 30  [timezone.....+00]
+    50 : 3a 30 30 00 00                                   [:00..]
+Test { "foo" : 1, "bar" : 2, "baz" : 3 }
```
jmikola: This will most certainly change once we address [PHPC-393](https://jira.mongodb.org/browse/PHPC-393) and stop serializing private/protected properties.

derickr: That's a change that has to make it into 1.0 though! And is there a ticket for me? :-)

tests/bson/bson-utcdatetime-001.phpt
```diff
@@ -51,8 +51,8 @@ Test#1 { "0" : { "$date" : 1416445411987 } }
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 bool(true)
-Test#2 { }
-string(3) "{ }"
-string(3) "{ }"
+Test#2 { "0" : { "date" : "2014-11-20 01:03:31.987000000", "timezone_type" : 1, "timezone" : "+00:00" } }
```
derickr: Please use %d for the trailing bits of 0.

jmikola: I assume for the "31.987000000" and not "+00:00"?

derickr: Yes :-)

tests/bson/bson-fromPHP-002.phpt
jmikola: As discussed earlier, [HHVM-99](https://jira.mongodb.org/browse/HHVM-99), which already works according to spec.

tests/bson/bson-utcdatetime-001.phpt
```diff
@@ -51,8 +51,8 @@ Test#1 { "0" : { "$date" : 1416445411987 } }
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 bool(true)
-Test#2 { }
-string(3) "{ }"
-string(3) "{ }"
+Test#2 { "0" : { "date" : "2014-11-20 01:03:31.987%r(0+)%r", "timezone_type" : 1, "timezone" : "+00:00" } }
+string(98) "{ "0" : { "date" : "2014-11-20 01:03:31.987%r(0+)%r", "timezone_type" : 1, "timezone" : "+00:00" } }"
+string(98) "{ "0" : { "date" : "2014-11-20 01:03:31.987%r(0+)%r", "timezone_type" : 1, "timezone" : "+00:00" } }"
```
jmikola: @derickr: `%r(0+)%r` seemed better than `%d`. I would expect all trailing numbers to be zero since we construct the date with a specific degree of precision.

derickr:
LGTM

jmikola:
@bjori suggested removing DateTime from the new tests, so I'll make that change.

tests/bson/bson-fromPHP_error-002.phpt
```diff
@@ -0,0 +1,29 @@
...
+<?php require __DIR__ . "/../utils/basic-skipif.inc"?>
+--FILE--
+<?php
+use MongoDB\BSON as BSON;
+
+require_once __DIR__ . "/../utils/basic.inc";
+
+class UnknownType implements BSON\Type {}
+
+echo throws(function() {
```
derickr: This misses a test case for just "fromPHP(new UnknownType)".

tests/bson/bson-utcdatetime-001.phpt
```diff
@@ -51,8 +51,8 @@ Test#1 { "0" : { "$date" : 1416445411987 } }
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 string(37) "{ "0" : { "$date" : 1416445411987 } }"
 bool(true)
-Test#2 { "0" : {  } }
-string(14) "{ "0" : {  } }"
-string(14) "{ "0" : {  } }"
+Test#2 { "0" : { "date" : "2014-11-20 01:03:31.987%r(0+)%r", "timezone_type" : 1, "timezone" : "+00:00" } }
```
derickr: This is not quite right, as DateTime doesn't have any real properties - I guess you should remove the test as per @bjori 's comments from your email.

jmikola: I removed DateTime from the *new* tests, but planned to leave it as-is here. AFAICT, it does have undocumented public properties. `var_dump(new DateTime);` produces:

```
class DateTime#1 (3) {
  public $date =>
  string(26) "2015-09-15 12:50:49.000000"
  public $timezone_type =>
  int(3)
  public $timezone =>
  string(16) "America/New_York"
}
```

tests/bson/bson-fromPHP_error-002.phpt
jmikola: Correct. I wasn't sure if we wanted to do that (before your email response and spec change). I'll add a second test for it, though.

jmikola:
@derickr: I revised the PR to conform to your [serialization spec change](https://github.com/mongodb-labs/mongo-hhvm-driver-prototype/commit/eb71a1c3653e0f2e1196058628cad2f3386f5b45).

There may be one point where we differ. I'm throwing if the user attempts to serialize a Type instance (other than Serializable) as a root document, while the spec only says to throw "if an object (root or embedded) of an unknown class implementing the `MongoDB\BSON\Type` interface." I think this is reasonable behavior, since an ObjectId or UTCDateTime is really unfit for a top-level document -- and the result would be empty anyway, since our types don't have public properties.

PR #101, opened 2015-09-09
mongoc 1.2.0-rc0 fixes

bjori:

PR #104, opened 2015-09-11
report libraries runtime versions

remicollet:
See CDRIVER-832: export library verson in the API
See PHPC 305 Allow to build with system libmongoc

In first implementation of --with-libmongoc I have set version check to 1.1.5, as 1.2.0 doesn't exists yet, so I use this PR to raise it to correct 1.2.0 value.

I have add a symbol check to avoid breakage with 1.2.0beta1 which doesn't have the symbol yet (this check could perhaps be removed later...)

Of course https://github.com/mongodb/mongo-c-driver/pull/265 should be merged before this one.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -2375,8 +2375,22 @@ PHP_MINFO_FUNCTION(mongodb)
 	php_info_print_table_header(2, "mongodb support", "enabled");
 	php_info_print_table_row(2, "mongodb version", MONGODB_VERSION_S);
 	php_info_print_table_row(2, "mongodb stability", MONGODB_STABILITY_S);
+#ifdef HAVE_MONGOC_GET_VERSION
+	{
+	char *tmp;
+
+	spprintf(&tmp, 0, "%d.%d.%d", bson_get_major_version(), bson_get_minor_version(), bson_get_micro_version());
+	php_info_print_table_row(2, "libmongoc headers version", MONGOC_VERSION_S);
+	php_info_print_table_row(2, "libmongoc library version", mongoc_get_version());
+	php_info_print_table_row(2, "libbson headers version", BSON_VERSION_S);
+	php_info_print_table_row(2, "libbson library version", tmp);
+	efree(tmp);
+	}
+#else
+	/* Bundled libraries, buildtime = runtime */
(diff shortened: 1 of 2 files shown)


bjori:
I don't think we need the symbol check.
When we bump the dependency to mongoc-beta1 we'll immediately depend on it as minimum version due to pr #101 and pr #105 among others.

remicollet:
Il will drop this check when 1.2.0beta2 (or rc) will be released (with this symbol)

php_phongo.c
```diff
@@ -2331,8 +2331,22 @@ PHP_MINFO_FUNCTION(mongodb)
...
 	php_info_print_table_row(2, "mongodb stability", MONGODB_STABILITY_S);
+#ifdef HAVE_MONGOC_GET_VERSION
+	{
+	char *tmp;
+
+	spprintf(&tmp, 0, "%d.%d.%d", bson_get_major_version(), bson_get_minor_version(), bson_get_micro_version());
+	php_info_print_table_row(2, "libmongoc headers version", MONGOC_VERSION_S);
+	php_info_print_table_row(2, "libmongoc library version", mongoc_get_version());
+	php_info_print_table_row(2, "libbson headers version", BSON_VERSION_S);
+	php_info_print_table_row(2, "libbson library version", tmp);
```
jmikola: Now that 1.2.0-rc0 of libbson and libmongoc have `bson_get_version()` and `mongoc_get_version()` functions, respectively, should this be changed? The `*_get_version()` functions are both wrappers for returning the `*_VERSION_S` constants, so I'm not sure why the headers and library rows can't use the same.

The `ssprintf()` definitely appears problematic, as it'd miss the "rc0" suffix, which is present in the constant (and returned by the `*_get_version()` function).

remicollet: previous version use this because bson_get_version doesn"t exists yet.

remicollet:
In latest commit:
* drop symbol check
* add HAVE_LIBBSON and HAVE_LIBMONGOC constants

Notice I can't really test for now, build seems broken
(error: 'mongoc_write_result_t' has no member named 'writeConcernError')

remicollet:
Rebased see PR #120

PR #108, opened 2015-09-15
PHPC-418: Getters for ReadPreference and WriteConcern properties

jmikola:
https://jira.mongodb.org/browse/PHPC-418


src/MongoDB/WriteConcern.c
```diff
@@ -100,6 +99,93 @@ PHP_METHOD(WriteConcern, __construct)
...
+		RETURN_STRING(PHONGO_WRITE_CONCERN_W_MAJORITY, 1);
+	}
+
+	RETURN_LONG(intern->write_concern->w);
+}
+/* }}} */
+
+/* {{{ proto string|integer WriteConcern::getWtimeout()
+   Returns the WriteConcern "wtimeout" option */
+PHP_METHOD(WriteConcern, getWtimeout)
```
derickr: shouldn't that be getWTimeout ?

jmikola: http://docs.mongodb.org/master/core/write-concern/ referred to it simply as `wtimeout`, and our constructor argument is also just `$wtimeout`, so I wasn't inclined to capitalize it.

For what it's worth, I also think `getW()` looks really weird.

derickr: ok

PR #118, opened 2015-09-24
PHPC-393: Ignore non-public properties when encoding BSON

jmikola:
https://jira.mongodb.org/browse/PHPC-393


src/bson.c
```diff
@@ -774,12 +774,28 @@ void phongo_bson_append(bson_t *bson, php_phongo_bson_flags_t flags, const char
...
+
 PHONGO_API void zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *bson, bson_t **bson_out TSRMLS_DC) /* {{{ */
 {
 	HashPosition pos;
 	HashTable   *ht_data = NULL;
 	zval        *obj_data = NULL;
 
+	/* If we will be encoding a class that may contain protected and private
+	 * properties, we'll need to filter them out later. */
+	bool         ht_data_from_properties = false;
```
derickr: I doubt MS VC supports bool/false. Might be better to use int and = 0/=1.

derickr:
Check whether bool/false/true is used in other places. If not, you will have to change it I think. Otherwise: LGTM

bjori:
It works fine. We use it in several places already

PR #105, opened 2015-09-11
PHPC-415: Try authenticating once and include failure message

bjori:

PR #119, opened 2015-10-02
PHPC-409: Compatibility updates for libmongoc 1.2.0-rc0

jmikola:
* https://jira.mongodb.org/browse/PHPC-409
 * https://jira.mongodb.org/browse/PHPC-415

This also incorporates commits from #101 and #105.

Note that the following tests will still fail until [CDRIVER-903](https://jira.mongodb.org/browse/CDRIVER-903) is resolved (tracking in [PHPC-441](https://jira.mongodb.org/browse/PHPC-441)):

 * tests/server/server-executeCommand-002.phpt
 * tests/server/server-executeQuery-005.phpt


config.m4
```diff
@@ -217,7 +217,6 @@ if test "$MONGODB" != "no"; then
     bson-timegm.c \
     bson-utf8.c \
     bson-value.c \
-    bson-version.c \
```
bjori: This is not correct and will make https://github.com/mongodb-labs/mongo-php-driver-prototype/pull/104/files broken.

Please include all files. If the file was renamed, please use the new name

php_phongo.c
```diff
@@ -2007,32 +2020,37 @@ bool php_phongo_writeresult_get_write_errors(php_phongo_writeresult_t *writeresu
...
-	if (!bson_empty0(&writeresult->write_result.writeConcernError)) {
-		bson_iter_t iter;
-
-		if (bson_iter_init_find(&iter, &writeresult->write_result.writeConcernError, "code") && BSON_ITER_HOLDS_INT32(&iter)) {
-			code = bson_iter_int32(&iter);
-		}
-		if (bson_iter_init_find(&iter, &writeresult->write_result.writeConcernError, "errmsg") && BSON_ITER_HOLDS_UTF8(&iter)) {
-			err = bson_iter_utf8(&iter, NULL);
+	bson_iter_t iter;
+	bson_iter_t citer;
```
bjori: Remove the empty newline above them, and move it below them

php_phongo.c
```diff
@@ -2007,32 +2020,37 @@ bool php_phongo_writeresult_get_write_errors(php_phongo_writeresult_t *writeresu
...
+			bson_iter_init (&iter, &writeresult->write_result.writeConcernErrors) &&
+			bson_iter_next (&iter) &&
+			BSON_ITER_HOLDS_DOCUMENT (&iter) &&
+			bson_iter_recurse (&iter, &citer)) {
+		while (bson_iter_next (&citer)) {
+			if (BSON_ITER_IS_KEY (&citer, "errmsg")) {
+				err = bson_iter_utf8 (&citer, NULL);
+			} else if (BSON_ITER_IS_KEY (&citer, "code")) {
+				code = bson_iter_int32 (&citer);
+			}
```
bjori: Does it matter that this will be the last writeConcernErrors (not the first as said in the crud spec)?

php_phongo.c
```diff
@@ -2007,32 +2020,37 @@ bool php_phongo_writeresult_get_write_errors(php_phongo_writeresult_t *writeresu
...
 		}
 
 		bson_set_error(error, PHONGO_ERROR_WRITECONCERN_FAILED, code, "%s", err);
 		return true;
 	}
-
 	return false;
 } /* }}} */
-zval* php_phongo_throw_write_errors(php_phongo_writeresult_t *wr TSRMLS_DC) /* {{{ */
+zval* php_phongo_throw_write_error(php_phongo_writeresult_t *wr TSRMLS_DC) /* {{{ */
```
bjori: I'm confused why this function is being renamed. the field in our struct, and the bson from the server, is writeErrors, so.......

bjori:
> Note that the following tests will still fail until CDRIVER-903 is resolved:

Remember to include a PHPC tracking ticket for it, otherwise it might get fixed, but the test still failing for other reason, but will be ignored forever as "assumed to be CDRIVER-903" issue.

config.m4
jmikola: In this case, we also need the `mongoc-version-functions.c` added to `MONGOC_SOURCES` (it was never included). I'll fix both.

bjori: Good catch. mongoc-version-functions.c is also a new file in the rc0 and needs to be included

jmikola:
> Remember to include a PHPC tracking ticket for it, otherwise it might get fixed, but the test still failing for other reason, but will be ignored forever as "assumed to be CDRIVER-903" issue.

Added [PHPC-441](https://jira.mongodb.org/browse/PHPC-441) to the OP.

php_phongo.c
jmikola: Both of the fields in `mongoc_write_result_t` are plural (for write and write concern errors), but this function and the `php_phongo_writeresult_get_*` helper deal with throwing a single exception and fetching a single `bson_error_t`, respectively.

I can revert this and pluralize the write concern function if you feel strongly about it, though.

php_phongo.c
jmikola: Why would this be the last write concern error? This, and the code in `php_phongo_writeresult_get_write_error` upon which it was based, appear to inspect whether the first element of the BSON array is an object and then recurse into it and pull out the "errmsg" and "code" fields. The only loop here is via `while` for extracting the error document fields.

bjori: Ah, so the **5 line** `if ()` statement above seeks to the first `writeConcernError` and the `while ()` loop iterates over that specific document in search for `errmsg` and `code`.
Got it

php_phongo.c
bjori: I think the function is named after the field name in the struct+protocol. I'd keep it like that.
Not that we have particularly consistent and pretty internal API. I always meant to address that in one swooping change but never had the time.

php_phongo.c
jmikola: Yup. I believe you originally wrote this for extracting the first write error. I just adapted it for the write concern. The `while` is just there to be flexible about the error document's field ordering AFAICT.

@derickr pointed out that we don't have tests for this stuff, so I'm going to add some.

php_phongo.c
jmikola: Will do.

bjori:
lgtm

PR #120, opened 2015-10-04
report libraries runtime versions

remicollet:
Rebase against current master of pr #104

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -313,6 +313,7 @@ PHP_ARG_WITH(libbson, Use system libbson,
     fi
     PHP_EVAL_INCLINE($LIBBSON_INC)
     PHP_EVAL_LIBLINE($LIBBSON_LIB, MONGODB_SHARED_LIBADD)
+    AC_DEFINE(HAVE_LIBBSON, 1, [Use system libbson])
   else
     PHP_ADD_SOURCES_X(PHP_EXT_DIR(mongodb)[src/libbson/src/yajl], $YAJL_SOURCES,            [$STD_CFLAGS $MAINTAINER_CFLAGS], shared_objects_mongodb, yes)
     PHP_ADD_SOURCES_X(PHP_EXT_DIR(mongodb)[src/libbson/src/bson], $BSON_SOURCES,            [$STD_CFLAGS $MAINTAINER_CFLAGS], shared_objects_mongodb, yes)
@@ -331,7 +332,7 @@ PHP_ARG_WITH(libmongoc, Use system libmongoc,
     AC_PATH_PROG(PKG_CONFIG, pkg-config, no)
     AC_MSG_CHECKING(for libmongoc)
     if test -x "$PKG_CONFIG" && $PKG_CONFIG --exists libmongoc-1.0 && $PKG_CONFIG --exists libmongoc-priv; then
-      if $PKG_CONFIG libmongoc-1.0 --atleast-version 1.1.5; then
+      if $PKG_CONFIG libmongoc-1.0 --atleast-version 1.2.0; then
         LIBMONGOC_INC=`$PKG_CONFIG libmongoc-priv --cflags`
         LIBMONGOC_LIB=`$PKG_CONFIG libmongoc-priv --libs`
         LIBMONGOC_VER=`$PKG_CONFIG libmongoc-priv --modversion`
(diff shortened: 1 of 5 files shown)


config.m4
```diff
@@ -313,6 +313,7 @@ PHP_ARG_WITH(libbson, Use system libbson,
     fi
     PHP_EVAL_INCLINE($LIBBSON_INC)
     PHP_EVAL_LIBLINE($LIBBSON_LIB, MONGODB_SHARED_LIBADD)
+    AC_DEFINE(HAVE_LIBBSON, 1, [Use system libbson])
```
derickr: As to not make this confusing, perhaps we should use HAVE_SYSTEM_LIBBSON and HAVE_SYSTEM_LIBMONGOC? Because, we're always going to have a LIBBSON and LIBMONGOC.

remicollet:
@derickr macros renamed

php_phongo.c
```diff
@@ -2375,8 +2375,19 @@ PHP_MINFO_FUNCTION(mongodb)
 	php_info_print_table_header(2, "mongodb support", "enabled");
 	php_info_print_table_row(2, "mongodb version", MONGODB_VERSION_S);
 	php_info_print_table_row(2, "mongodb stability", MONGODB_STABILITY_S);
+#ifdef HAVE_SYSTEM_LIBMONGOC
+	php_info_print_table_row(2, "libmongoc headers version", MONGOC_VERSION_S);
+	php_info_print_table_row(2, "libmongoc library version", mongoc_get_version());
+#else
```
bjori: I would prefer to keep this consistent.
Otherwise we have to ask `what does phpinfo() report as "libmongoc headers version and "libmongoc library version" -- or -- libmongoc version?`.

Lets just always print out both compile time and lib

php_phongo.c
```diff
@@ -2375,8 +2375,19 @@ PHP_MINFO_FUNCTION(mongodb)
...
+	/* Bundled libraries, buildtime = runtime */
 	php_info_print_table_row(2, "libmongoc version", MONGOC_VERSION_S);
+#endif
+#ifdef HAVE_SYSTEM_LIBBSON
+	php_info_print_table_row(2, "libbson headers version", BSON_VERSION_S);
+	php_info_print_table_row(2, "libbson library version", bson_get_version());
+#else
```
bjori: same as above, lets always use this

php_phongo.c
remicollet: Or "libbson bundled version" ?
(I'd like to make explicit id system or bundled library is used)

bjori: Ah ok. Makes sense. Shall we make that the third line? mongoc: system/bundled ?
Then its easy `ctrl+f`able, along with being straight forward to ask/answer the question

remicollet: I think having 1 line for bundled library and 2 for system library is quite common and consistent with other ext : see GD, msgpack, libsodium ...)...  ok.... all added by me ;)

remicollet: Notice; I can change now, as bson_get_version is not yer defined in bundled lib...

remicollet: Notice: I CANNOT change now, as bson_get_version is not yer defined in bundled lib... (need to wait for PR #119)

bjori: hehe. ok. This is such a minor detail that as long as the output isn't vastly different then it isn't worth spending to much time on this.

One thing that just hit me though, maybe we should update the trace line that reports these versions: 
https://github.com/mongodb-labs/mongo-php-driver-prototype/blob/master/php_phongo.c#L1873

I am hoping the only question we will ever need to ask users for support cases is:

    Can you change your php.ini to include mongodb.debug=/tmp/ and rerun the script as send us the content?

Or

    Can you pass array("debug" => "/tmp/") as 3rd argument to the ctor

So it would be nice to know there if it was bundled/system.
See also: https://jira.mongodb.org/browse/PHPC-442

remicollet:
$ php -n -d extension=$PWD/modules/mongodb.so -r '$manager = new MongoDB\Driver\Manager("mongodb://localhost:27017",[],["debug" => "/tmp/"]);'
     $ $ cat /tmp/PHONGO*
     [2015-10-05T17:56:56+00:00]     PHONGO: DEBUG   > Creating Manager, phongo-1.0.0beta2[devel] - mongoc-1.2.0-beta1(bundled), libbson-1.2.0-beta(bundled), php-5.6.14

bjori:
Thanks!

jmikola:
Manually merged in https://github.com/mongodb-labs/mongo-php-driver-prototype/commit/71b3cb343cdcd8fdeeede44345d76e12e14c58d5. Thanks, @remicollet!

php_phongo.c
```diff
@@ -1870,7 +1870,23 @@ mongoc_client_t *php_phongo_make_mongo_client(const mongoc_uri_t *uri, zval *dri
 		php_phongo_populate_default_ssl_ctx(ctx, driverOptions);
 	}
 
-	MONGOC_DEBUG("Creating Manager, phongo-%s[%s] - mongoc-%s, libbson-%s", MONGODB_VERSION_S, MONGODB_STABILITY_S, MONGOC_VERSION_S, BSON_VERSION_S);
+	MONGOC_DEBUG("Creating Manager, phongo-%s[%s] - mongoc-%s(%s), libbson-%s(%s), php-%s",
+		MONGODB_VERSION_S,
+		MONGODB_STABILITY_S,
+		MONGOC_VERSION_S,
+#ifdef HAVE_SYSTEM_LIBMONGOC
```
jmikola: This appears to trigger [Compiler Error C2121](https://msdn.microsoft.com/en-us/library/1s467t6a.aspx) in VC11, because we're using a macro within the `MONGOC_DEBUG()` macro. I believe a quick solution would be to populate `const char *` variables beforehand and use those as-is in the `MONGOC_DEBUG()` macro.

remicollet: Sorry for this... :(

jmikola: No worries! I've missed more than my fair share of Windows errors :)

https://github.com/mongodb-labs/mongo-php-driver-prototype/commit/59177872a87324342d336052008c0fd3ac777747 should fix it.

PR #124, opened 2015-10-14
PHPC-407: Rename X509 error test and fix cert path

jmikola:
https://jira.mongodb.org/browse/PHPC-407

### DIFF
--- a/tests/connect/standalone-x509-0001.phpt
+++ b/tests/connect/standalone-x509-0001.phpt
@@ -1,5 +1,5 @@
 --TEST--
-Connect to MongoDB with using SSL and X.509 auth
+Connect to MongoDB with using SSL and X.509 auth (stream context)
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; NEEDS("STANDALONE_X509"); ?>
 --FILE--
--- a/tests/connect/standalone-x509-error-0001.phpt
+++ b/tests/connect/standalone-x509-error-0001.phpt
@@ -1,5 +1,5 @@
 --TEST--
-Connect to MongoDB with using X509 retrieving username from certificate #002
+X509 connection should not reuse previous stream after an auth failure
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; NEEDS("STANDALONE_X509"); ?>
 --FILE--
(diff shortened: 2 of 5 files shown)


tests/connect/standalone-x509-0001.phpt
```diff
@@ -1,5 +1,5 @@
 --TEST--
-Connect to MongoDB with using SSL and X.509 auth
+Connect to MongoDB with using SSL and X.509 auth (stream context)
```
jmikola: @bjori: please confirm that these title changes make sense. Best I could tell, tests 0002 and 0004 are analogous to 0001 and 0003, respectively, except they SSL context options passed directly to `$driverOptions` and processed by `php_phongo_populate_default_ssl_ctx()`, which only deals with a subset of the full context options found in 0001 and 0003.

tests/connect/standalone-x509-error-0001.phpt
```diff
@@ -29,12 +28,13 @@ $opts = array(
         "allow_self_signed" => false,
         "cafile" => $SSL_DIR . "/ca.pem", /* Defaults to openssl.cafile */
         "capath" => $SSL_DIR, /* Defaults to openssl.capath */
-        "local_cert" => $SSL_DIR . "/src/libmongoc/tests/certificates/client.pem",
+        "local_cert" => $SSL_DIR . "/client.pem",
```
jmikola: I'm not sure why this path was incorrect before. It doesn't affect the test, since we're going to provide an invalid username anyway.

tests/connect/standalone-x509-error-0001.phpt
```diff
@@ -29,12 +28,13 @@ $opts = array(
...
 );
+
+/* Wrong username */
 $parsed = parse_url(STANDALONE_X509);
 $dsn = sprintf("mongodb://username@%s:%d/%s?ssl=true&authMechanism=MONGODB-X509", $parsed["host"], $parsed["port"], DATABASE_NAME);
 
-
 $m1 = connect($dsn, $opts);
 $m2 = connect($dsn, $opts);
```
jmikola: This test appeared to be functionality equivalent to the one you attached to [PHPC-407](https://jira.mongodb.org/browse/PHPC-407), so I don't think there's any need to add that new test in.

tests/connect/standalone-x509-error-0001.phpt
bjori: Its not not correct, its just not bundled or generated anymore :)
Actually, it is incorrect indeed after we added the $SSL_DIR var.

certificates/client.pem is generated by the mongoc test suite so was probably there when the test was originally created :)

tests/connect/standalone-x509-error-0001.phpt
bjori: Thats likely. I added a testcase to the report so I wouldn't loose the magic combination of things, and then kept on going with it trying to decipher the issue and/or find others.

PR #125, opened 2015-10-15
PHPC-444: Support options array in BulkWrite and executeInsert()

jmikola:
https://jira.mongodb.org/browse/PHPC-444

### DIFF
--- a/src/MongoDB/Manager.c
+++ b/src/MongoDB/Manager.c
@@ -180,7 +180,7 @@ PHP_METHOD(Manager, executeBulkWrite)
 	phongo_execute_write(intern->client, namespace, bulk->bulk, phongo_write_concern_from_zval(zwrite_concern TSRMLS_CC), -1, return_value, return_value_used TSRMLS_CC);
 }
 /* }}} */
-/* {{{ proto MongoDB\Driver\WriteResult Manager::executeInsert(string $namespace, array|object $document[, MongoDB\Driver\WriteConcern $writeConcern = null])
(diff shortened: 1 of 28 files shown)


src/MongoDB/Manager.c
```diff
@@ -180,7 +180,7 @@ PHP_METHOD(Manager, executeBulkWrite)
 	phongo_execute_write(intern->client, namespace, bulk->bulk, phongo_write_concern_from_zval(zwrite_concern TSRMLS_CC), -1, return_value, return_value_used TSRMLS_CC);
 }
 /* }}} */
-/* {{{ proto MongoDB\Driver\WriteResult Manager::executeInsert(string $namespace, array|object $document[, MongoDB\Driver\WriteConcern $writeConcern = null])
+/* {{{ proto MongoDB\Driver\WriteResult Manager::executeInsert(string $namespace, array|object $document[, array $insertOptions = array()[, MongoDB\Driver\WriteConcern $writeConcern = null]])
```
jmikola: I'm pretty sure we already discussed this, but just to be sure: we still want to leave WriteConcern as its own type-hinted argument, instead of incorporating it into the `$options` array, correct?

It's clear that we wanted that for `executeBulkWrite()`, since you specify it at execution time (vs. other options going into BulkWrite's constructor), but I'm wondering if users would rather do:

```
$manager->executeInsert('db.collection', ['foo' => 'bar'], [], $writeConcern);
```

or

```
$manager->executeInsert('db.collection', ['foo' => 'bar'], ['writeConcern' => $writeConcern]);
```

Note: I wouldn't ask the same question of incorporating ReadPreference into the Query or Command, since we talked about that at length and the server selection makes more sense in the execute methods for those.

jmikola:
@bjori left a LGTM to this in IRC after general objections to the insert/update/delete convenience methods existing in the first place.

PR #126, opened 2015-10-19
PHPC-310: Exceptions aren't final; BSON types are and can't be serialized

jmikola:
https://jira.mongodb.org/browse/PHPC-310


bjori:
lgtm.

There are no tests to confirm the __wakeUp() is correct. Thats only a problem if in the future we forget and remove it without giving it a second thought.

jmikola:
Added `unserialize()` tests, although none of those actually trigger `Manager::__wakeUp()` and the "MongoDB\Driver objects cannot be serialized" message.

derickr:
The preventing serialization was not in the Jira ticket. That caused me a surprise when I saw the PR. PR LGTM though.

PR #127, opened 2015-10-20
PHPC-462: Remove unused DuplicateKeyException

jmikola:
https://jira.mongodb.org/browse/PHPC-462

This class was not being used, since executeBulkWrite() uses BulkWriteException and the single write methods use WriteException and WriteConcernException exclusively.


derickr:
LGTM

gmsantos:
@jmikola is this exception planned to be included back? 

We are migrating from legacy driver and catch `BulkWriteException` isn't so clear at all on our classes.

jmikola:
A DuplicateKeyException has no logical application in the new driver, as all writes are executed through [`MongoDB\Driver\Manager::executeBulkWrite()`](http://php.net/manual/en/mongodb-driver-manager.executebulkwrite.php).  The thrown [BulkWriteException](http://php.net/manual/en/class.mongodb-driver-exception-bulkwriteexception.php) might include any number of errors, which you'll need to infer from the [result object](http://php.net/manual/en/mongodb-driver-writeexception.getwriteresult.php). For the benefit of logging, we do concatenate all error messages into the BulkWriteException's message.

jmikola:
I've opened [PHPLIB-203](https://jira.mongodb.org/browse/PHPLIB-203) to discuss adding such a class to [mongodb/mongo-php-library](https://github.com/mongodb/mongo-php-library), where it could be thrown by its single-write operations.

PR #130, opened 2015-10-29
PHPC-475: toJSON and fromJSON should throw on error

jmikola:
https://jira.mongodb.org/browse/PHPC-475

### DIFF
--- a/src/bson.c
+++ b/src/bson.c
@@ -1096,6 +1096,7 @@ PHP_FUNCTION(toJSON)
 	      char          *data;
 	      int            data_len;
 	const bson_t        *b;
+	      bool           eof = false;
 	      bson_reader_t *reader;
 
 	(void)return_value_ptr; (void)this_ptr; (void)return_value_used; /* We don't use these */
@@ -1114,8 +1115,13 @@ PHP_FUNCTION(toJSON)
 		RETVAL_STRINGL(str, str_len, 1);
 		bson_free(str);
 	} else {
-		RETURN_NULL();
+		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Could not read document from BSON reader");
 	}
+
+	if (bson_reader_read(reader, &eof) || !eof) {
+		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Reading document did not exhaust input buffer");
+	}
+
 	bson_reader_destroy(reader);
 }
 /* }}} */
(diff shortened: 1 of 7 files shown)


tests/bson/bson-fromJSON_error-001.phpt
```diff
@@ -0,0 +1,23 @@
...
+}, 'MongoDB\Driver\Exception\UnexpectedValueException'), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+OK: Got MongoDB\Driver\Exception\UnexpectedValueException
+lexical error: invalid string in json text.
+                                       foo
+                     (right here) ------^
```
derickr: Does it really add the "right here" stuff? :-)

src/bson.c
```diff
@@ -1139,7 +1145,7 @@ PHP_FUNCTION(fromJSON)
 		RETVAL_STRINGL((const char *) bson_get_data(&b), b.len, 1);
 		bson_destroy(&b);
 	} else {
-		RETURN_NULL();
+		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "%s", error.message);
```
derickr: Are you sure error.message is always populated here, and not a NULL ptr?

derickr:
LGTM

tests/bson/bson-fromJSON_error-001.phpt
jmikola: Unfortunately, yes. It comes directly from yajl. We can attempt to shorten this to a one-liner at a later point.

src/bson.c
jmikola: Reasonably sure based on `bson_json_reader_read()`; however, it won't hurt to add a fall-back.

src/bson.c
```diff
@@ -1114,8 +1115,13 @@ PHP_FUNCTION(toJSON)
 		RETVAL_STRINGL(str, str_len, 1);
 		bson_free(str);
 	} else {
-		RETURN_NULL();
+		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Could not read document from BSON reader");
 	}
+
+	if (bson_reader_read(reader, &eof) || !eof) {
+		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Reading document did not exhaust input buffer");
```
bjori: This will yield in two exceptions being thrown because the former has no "return"

PR #136, opened 2015-11-07
PHPC-285: PHP 7 support

jmikola:
https://jira.mongodb.org/browse/PHPC-285

This PR supersedes #97

### DIFF
--- a/src/MongoDB/WriteConcern.c
+++ b/src/MongoDB/WriteConcern.c
@@ -215,40 +215,35 @@ static void php_phongo_writeconcern_free_object(void *object TSRMLS_DC) /* {{{ *
 	if (intern->write_concern) {
 		mongoc_write_concern_destroy(intern->write_concern);
 	}
+
+#if PHP_VERSION_ID < 70000
 	efree(intern);
+#endif
 } /* }}} */
 
-#if PHP_VERSION_ID >= 70000
-zend_object* php_phongo_writeconcern_create_object(zend_class_entry *class_type TSRMLS_DC) /* {{{ */
+phongo_create_object_retval php_phongo_writeconcern_create_object(zend_class_entry *class_type TSRMLS_DC) /* {{{ */
 {
-        php_phongo_writeconcern_t *intern;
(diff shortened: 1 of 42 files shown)


src/MongoDB/WriteConcern.c
```diff
@@ -91,7 +90,11 @@ PHP_METHOD(WriteConcern, __construct)
 	switch(ZEND_NUM_ARGS()) {
 		case 3:
 			if (Z_TYPE_P(journal) != IS_NULL) {
+#ifdef ZEND_ENGINE_3
+				mongoc_write_concern_set_journal(intern->write_concern, Z_TYPE_P(journal) == IS_TRUE);
```
bjori: This is not correct.

zpp accepts a 'z' here, so the type can be whatever. A truthful value, such as 1, will not match == IS_TRUE, but should still be considered as enabling journalling here.

    $wc = new MongoDB\Driver\WriteConcern(1, 0, 1);
    var_dump($wc->getJournal());

This should return true, but after this change -- it returns false

fifths:
make: 
In file included from /usr/local/php7/include/php/Zend/zend.h:39:0,
                 from /usr/local/php7/include/php/main/php.h:35,
                 from /home/lee/Code/mongo-php-driver-legacy/php_mongo.c:16:
/usr/local/php7/include/php/Zend/zend_variables.h:112:15: note: expected ‘struct zval *’ but argument is of type ‘struct zval **’
 ZEND_API void _zval_ptr_dtor(zval *zval_ptr ZEND_FILE_LINE_DC);
               ^
make: *** [php_mongo.lo] error 1

jmikola:
See #175.

src/MongoDB/WriteConcern.c
jmikola: Addressed in https://github.com/mongodb/mongo-php-driver/pull/175/files#diff-2255238768a806cc38ff567bdb9bc2b5R94, where @derickr uses `zend_is_true()`.

PR #148, opened 2015-12-03
Master branch is now 1.1.0-dev

jmikola:


php_phongo.h
```diff
@@ -30,7 +30,7 @@
 extern zend_module_entry mongodb_module_entry;
 
 /* FIXME: Its annoying to bump version. Move into phongo_version.h.in */
-#define MONGODB_VERSION_S "1.0.1"
+#define MONGODB_VERSION_S "1.1.0"
```
derickr: It should really be 1.1.0-dev though!

PR #151, opened 2015-12-07
3.2 changes

derickr:


jmikola:
LGTM. I have some local test failures on 3.2, but I don't think that has anything to do with your PR.

derickr:
Merged

bjori:
I recommend using the 3.2 GA, not old RC.
The RC has issues with for example bypassDocumentValidation and authentication

derickr:
It's ok - Jeremy is pushing that for the next PR.

PR #154, opened 2015-12-08
libmongoc 1.3 is not compatible

remicollet:
Just a sanity check.


derickr:
Thanks again!

jmikola:
@remicollet: Do you need a 1.0.2 release, or are you also handling this upstream?

remicollet:
no need for a release (thanks for the proposal)
pecl/mongodb is not yet in official repo, and this is more a sanity check for user buildling from sources

PR #156, opened 2015-12-09
PHPC-450: Fix test failures for MongoDB 3.2

jmikola:
https://jira.mongodb.org/browse/PHPC-450

### DIFF
--- a/tests/utils/tools.php
+++ b/tests/utils/tools.php
@@ -391,10 +391,8 @@ function fromJSON($var) {
     return $func($var);
 }
 
-
-
-/* NOTE: Using this function will take down mongod ! */
+/* Note: this fail point may terminate the mongod process, so you may want to
+ * use this in conjunction with a throwaway server. */
 function failGetMore(MongoDB\Driver\Manager $manager) {
     return configureFailPoint($manager, "failReceivedGetmore", "alwaysOn");
 }
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env:
 before_install:
   - sudo pip install cpp-coveralls
   - sudo pip install ${MONGODB_ORCHESTRATION_REPO}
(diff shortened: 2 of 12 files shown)


tests/server/server-executeQuery-005.phpt
```diff
@@ -50,8 +50,10 @@ printf("Set profile level to 0 successfully: %s\n", (empty($result->ok) ? 'no' :
 --EXPECTF--
 Set profile level to 2 successfully: yes
 object(stdClass)#%d (%d) {
-  ["$query"]=>
-  object(stdClass)#%d (%d) {
+  ["find"]=>
+  string(%d) "%s"
+  ["filter"]=>
+  object(stdClass)#%d (1) {
```
bjori: This looks like something that needs to be split into two tests.
One on pre 3.2, and one 3.2+

jmikola: Good point. I'll split these and the fail point test to use the 3.0 test server.

tests/utils/tools.php
```diff
@@ -391,10 +391,6 @@ function fromJSON($var) {
     return $func($var);
 }
 
-
-
-/* NOTE: Using this function will take down mongod ! */
```
bjori: Is this no longer true?
It used to quite literally (and intentionally) result in an abort() to make sure the getmore failed completely

jmikola: Oh, it's possible that getMore itself aborts (I don't recall offhand). I thought this was referring to the PHP function itself and sending the `configureFailPoint` command, which certainly wasn't true.

bjori: No no, the comment is referring to `mongod`

jmikola: Got it. I revised the comment below. :+1:

derickr:
LGTM

Issue #159, opened 2015-12-11
License?


oerdnj:
Folks, the package.xml (and PECL) says "New BSD License", but the LICENSE file is Apache 2.0. Could you please fix the discrepancy?

bjori:
I've updated the PECL page to say Apache, and sent PR for the package.xml generation script.

It was just the package file that was incorrect. All the files are Apache licensed.

jmikola:
Resolving, as #161 has been merged.

PR #161, opened 2015-12-11
Fix package license to be consistent with the source files

bjori:
Fixes https://github.com/mongodb/mongo-php-driver/issues/159


jmikola:
Merged in https://github.com/mongodb/mongo-php-driver/commit/f131403dc355493378067c9d21201b302fc90322. Thanks!

PR #165, opened 2015-12-14
PHPC-508: Construct valid stream URI for IPv6 literals

jmikola:
https://jira.mongodb.org/browse/PHPC-508


derickr:
LGTM

bjori:
You would want to update the `uniqid` too.
This also seems like something we should test for. It shouldn't be a problem to spin up a mongd on ipv6 in the vm.

Issue #133, opened 2015-11-04
No suitable servers found


prolic:
on travis, see: https://travis-ci.org/prooph/event-store-mongodb-adapter/jobs/89299198

prolic:
eventually related to: https://github.com/mongodb/mongo-php-driver/issues/112

bjori:
Are you sure mongod is actually running?
Can you add something like the following line after the "mongod" launch, then we can confirm mongod is running properly:

    mongo localhost:27017 --eval 'printjson(db.runCommand({"isMaster": 1}))'

prolic:
I am pretty sure mongod is running, as it always was with the legacy driver. Let's see.

prolic:
ups, command not found?? https://travis-ci.org/prooph/event-store-mongodb-adapter/jobs/89329883

bjori:
${PWD}/mongo localhost:.....

prolic:
is running:

    connecting to: localhost:27017/test
    
    {
    
        "ismaster" : true,
    
        "maxBsonObjectSize" : 16777216,
    
        "maxMessageSizeBytes" : 48000000,
    
        "maxWriteBatchSize" : 1000,
    
        "localTime" : ISODate("2015-11-04T23:08:01.148Z"),
    
        "maxWireVersion" : 3,
    
        "minWireVersion" : 0,
    
        "ok" : 1
    
    }

however:

MongoDB\Driver\Exception\RuntimeException: No suitable servers found (`serverselectiontryonce` set): [connection timeout calling ismaster on 'localhost:27017']

see: https://travis-ci.org/prooph/event-store-mongodb-adapter/jobs/89335661

gabrielalan:
I'm getting this error too, any solution? =(

olekhy:
same mistake, I have to wait for a speedy problem-solving?

jmikola:
Is this issue only happening on Travis, or is the error easily reproducible on a development environment, too?

jmikola:
Reading prooph/event-store-mongodb-adapter's [.travis.yml](https://github.com/prooph/event-store-mongodb-adapter/blob/master/.travis.yml):

 * You're fetching, unpacking, and starting the MongoDB binaries in `before_install`
 * `mongod` is launched with `&` in that block

Comparing this with the [.travis.yml](https://github.com/mongodb/mongo-php-library/blob/master/.travis.yml) file from [mongo-php-library](https://github.com/mongodb/mongo-php-library):

[...]

Issue #112, opened 2015-09-20
No suitable servers found - Windows, PHP 5.6 NTS - beta 1


adisoftbn:
Hello

I'm using new MongoDB\Driver\Manager("mongodb://localhost:27017");
I tested with the beta1 on windows with php 5.6 NTS(official dll build), but also with PHP7 in linux custom build(built by me from php7 branch).

I have the server locally for testing. It works with the old mongo extension on php 5.6.
I tried to enable debug with mongodb.debug but also mongodb.debug_log as in doc, but i get no debug info in logs

I use monogdb 3.0.6.

Any ideas what might be the problem ?

bjori:
`mongodb.debug` accepts either a directory or `stderr`/`stdout` as argument.
When running from the command line I usually do:

```
$ php -dmongodb.debug=stderr myscript.php
```

Not sure how well that will work on Windows, so can you try:
```
 php.exe -dmongodb.debug=. myscript.php
```

`.` means the current working directory, so it'll create a log file called `PHONGO-<timestamp><random>`
Can you then post the contents of that file?

adisoftbn:
The output is pretty big :) Cannot upload it here as file.

[2015-09-20T18:51:35+00:00]     PHONGO: DEBUG   > Connection string: 'mongodb://localhost:27017'
[2015-09-20T18:51:35+00:00]     PHONGO: TRACE   > ENTRY: php_phongo_make_mongo_client():1810
[2015-09-20T18:51:35+00:00]     PHONGO: DEBUG   > Creating Manager, phongo-1.0.0beta1[beta] - mongoc-1.1.2, libbson-1.1.3
[2015-09-20T18:51:35+00:00]     mongoc: TRACE   > ENTRY: mongoc_topology_description_init():51
[2015-09-20T18:51:35+00:00]     mongoc: TRACE   >  EXIT: mongoc_topology_description_init():67
[2015-09-20T18:51:35+00:00]     mongoc: TRACE   > ENTRY: mongoc_server_description_init():97
[2015-09-20T18:51:35+00:00]     mongoc: TRACE   >  EXIT: mongoc_server_description_init():132
[2015-09-20T18:51:35+00:00]    cluster: TRACE   > ENTRY: mongoc_cluster_init():1367
[2015-09-20T18:51:35+00:00]    cluster: TRACE   >  EXIT: mongoc_cluster_init():1390
[...]

PR #166, opened 2015-12-14
PHPC-512: Disable async stream connections

bjori:
Fixes: #133, #112. See: https://jira.mongodb.org/browse/PHPC-512

When doing async connections we can't properly determine if we should connect over IPv6 or IPv4. PHP will unfortunately default to connecting over IPv6 (when compiled in support, with minor runtime check), while mongod does not support IPv6 by default.

When then creating the IPv6 connection asynchronously, PHP hits EINPROGRESS and will happily return the socket as valid.

There currently is no way to configure these functions to prefer IPv4. The best we can do for now is to not connect asynchronously as then we'll properly discover that the IPv6 socket failed, and then continue through the list of addrinfo and next pickup the IPv4 address which works.


jmikola:
Merged in https://github.com/mongodb/mongo-php-driver/commit/7dad72fe7d87a51d18668b3e5eb6ab2abfcf1cd8.

PR #167, opened 2015-12-14
PHPC-448: Fix arginfo for ReadConcern constructor

jmikola:
https://jira.mongodb.org/browse/PHPC-448

The first argument is not required.


derickr:
Would this need a (simple) test case? Perhaps at least to make sure that HHVM does the same?

jmikola:
`readconcern-ctor-001.phpt` already tests that that constructor requires no arguments. This is just fixing some arginfo copypasta that I missed in the original PR.

derickr:
OK, LGTM

PR #168, opened 2015-12-14
PHPC-509: Default Manager URI to "mongodb://localhost:27017/"

jmikola:
https://jira.mongodb.org/browse/PHPC-509

### DIFF
--- a/src/MongoDB/Manager.c
+++ b/src/MongoDB/Manager.c
@@ -48,20 +48,21 @@
 #include "php_phongo.h"
 #include "php_bson.h"
 
+#define PHONGO_MANAGER_URI_DEFAULT "mongodb://localhost:27017/"
 
 PHONGO_API zend_class_entry *php_phongo_manager_ce;
 
 zend_object_handlers php_phongo_handler_manager;
 
-/* {{{ proto MongoDB\Driver\Manager Manager::__construct(string $uri[, array $options = array()[, array $driverOptions = array()]])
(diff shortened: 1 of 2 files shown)


tests/manager/manager-ctor-001.phpt
```diff
@@ -1,12 +1,12 @@
...
-MongoDB\Driver\Manager::__construct()
+MongoDB\Driver\Manager::__construct() with default URI
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; CLEANUP(STANDALONE) ?>
 --FILE--
 <?php
 require_once __DIR__ . "/../utils/basic.inc";
 
-$manager = new MongoDB\Driver\Manager(STANDALONE);
+$manager = new MongoDB\Driver\Manager();
```
derickr: I don't think this is going to work. The STANDALONE still points to a PHONGO SERVERS.json link - and "localhost" isn't always available is it? In any case, you shouldn't modify an existing case that doesn't change behaviour with the patch.

jmikola: `manager-ctor-001.phpt` was renamed to `manager-ctor-002.phpt`, so this is a new test.

The connection doesn't even happen until we do something that requires it, so this is basically just testing parsing of the default URI and the constructor allowing all params to be optional.

derickr:
LGTM

PR #179, opened 2015-12-22
PHPC-285: Windows compilation fixes for PHP 7

jmikola:
https://jira.mongodb.org/browse/PHPC-285


php_bson.h
```diff
@@ -71,7 +71,7 @@ typedef struct {
 } php_phongo_bson_state;
 
 #if PHP_VERSION_ID >= 70000
-#define PHONGO_BSON_STATE_INITIALIZER  { {}, { PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL}, NULL}
+#define PHONGO_BSON_STATE_INITIALIZER  { {{ 0 }}, { PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL}, NULL}
```
jmikola: @derickr: See http://stackoverflow.com/q/11551986 and http://stackoverflow.com/a/20405554. Windows required `{ 0 }` (as is already done in libmongoc), but the extra braces here are a concession for gcc's missing-braces warning. Windows seems to be OK with it, too, so I gathered this was preferable to disabling the warning altogether.

src/contrib/php-ssl.c
```diff
@@ -30,6 +30,10 @@
 /* LCOV_EXCL_START */
 #include "php-ssl.h"
 
+#if defined(NETWARE) || (defined(PHP_WIN32) && defined(_MSC_VER) && _MSC_VER >= 1900)
+#define timezone _timezone  /* timezone is called _timezone in LibC */
+#endif
```
jmikola: See: http://lxr.php.net/xref/PHP_7_0/ext/openssl/openssl.c#62

php_bson.h
jmikola: Also, kudos to @ksuarz to the quick insight into this!

derickr:
Lgtm

PR #185, opened 2015-12-26
Fix PHP 7 / 32bits build

remicollet:
/builddir/build/BUILD/php-pecl-mongodb-1.1.1/NTS/src/bson.c: In function 'php_phongo_bson_visit_int64':
     /builddir/build/BUILD/php-pecl-mongodb-1.1.1/NTS/src/bson.c:508:38: error: macro "add_assoc_stringl" passed 5 arguments, but takes just 4
       ADD_ASSOC_INT64(retval, key, v_int64);

As ADD_ASSOC_STRINGL imply string duplicate, also need  to free the temp (can perhaps be optimized for PHP 5... but more conditional for a race condition...)


jmikola:
Merged in https://github.com/mongodb/mongo-php-driver/commit/1f0f79da3bbf868f1223bdce7240b622da37251a. Thank you!

PR #193, opened 2015-12-31
PHPC-534: Enable configure without SASL

kaplanlior:
And fix some minor issues.

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -390,7 +390,7 @@ if test "$PHP_PCRE_DIR" != "no"; then
   fi
 fi
 
-PHP_ARG_WITH(mongodb-sasl, Build with Cyrus SASL support,
+PHP_ARG_WITH(mongodb-sasl, for Cyrus SASL support,
 [  --with-mongodb-sasl[=DIR]     mongodb: Include Cyrus SASL support], auto, yes)
 
 if test "$PHP_MONGODB_SASL" != "no"; then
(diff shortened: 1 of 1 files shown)


config.m4
```diff
@@ -390,8 +390,8 @@ if test "$PHP_PCRE_DIR" != "no"; then
   fi
 fi
 
-PHP_ARG_WITH(mongodb-sasl, Build with Cyrus SASL support,
-[  --with-mongodb-sasl[=DIR]     mongodb: Include Cyrus SASL support], auto, yes)
+PHP_ARG_WITH(mongodb-sasl, for Cyrus SASL support,
+[  --with-mongodb-sasl[=DIR]     mongodb: Include Cyrus SASL support], auto, no)
```
derickr: Any reason why you are changing the default from "yes" to "no" here?

kaplanlior:
When the default is yes, it doesn't respect --with-mongodb-sasl=no . The macros keep changing it to yes.
That also makes auto to become yes (and making the other check no relevant).

jmikola:
I can reproduce the behavior @kaplanlior mentioned above. On that note, I think we should make a similar change to `--with-pcre-dir[=DIR]`.

FWIW, this doesn't appear to be discussed directly in [the build docs](http://php.net/manual/en/internals2.buildsys.configunix.php). Their `--with-example[=FILE]` example would omit the "auto, (yes|no)" bit entirely. For our needs of having `--with-mongodb-sasl` default to auto if unspecified, but still respect "=no", Lior's solution seems to work nicely.

jmikola:
I rebased this with additional changes (and a link to a corresponding JIRA ticket) in https://github.com/mongodb/mongo-php-driver/pull/197.

PR #197, opened 2016-01-06
PHPC-534: Enable configure without SASL

jmikola:
https://jira.mongodb.org/browse/PHPC-534

Supersedes #193 /cc @kaplanlior

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -390,7 +390,7 @@ if test "$PHP_PCRE_DIR" != "no"; then
   fi
 fi
 
-PHP_ARG_WITH(mongodb-sasl, Build with Cyrus SASL support,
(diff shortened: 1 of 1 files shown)


derickr:
LGTM

kaplanlior:
I didn't change pcre default on #193 as its code doesn't fail configure even if it doesn't have pcre (it uses AC_MSG_RESULT instead of AC_MSG_ERROR). 

Also, "auto" has no use in pcre (it only checks for yes/no values). So having yes as default for pcre is actually what we have auto for in sasl.

jmikola:
@kaplanlior: Noted. I've removed the last commit in this PR, which was changing the PCRE default. :+1:

PR #225, opened 2016-02-09
PHPC-519: Add missing modifiers to final class methods

jmikola:
https://jira.mongodb.org/browse/PHPC-519

This is just being done for consistency. The classes are all already final (specified in their MINIT function). We also add some regression tests for extending these classes.


derickr:
LGTM, however I'm wondering whether these tests are going to break on HHVM...

jmikola:
> I'm wondering whether these tests are going to break on HHVM

Shouldn't be a problem, as we already had a few of these tests for various BSON types. This PR only adds a few more that were missing for other final classes.

PR #226, opened 2016-02-10
PHPC-550: Always encode ODS field when serializing Persistable documents

jmikola:
https://jira.mongodb.org/browse/PHPC-550

### DIFF
--- a/tests/bson/bson-ods-002.phpt
+++ b/tests/bson/bson-ods-002.phpt
@@ -1,5 +1,7 @@
 --TEST--
 BSON encoding: Encoding data into BSON representation, and BSON into Extended JSON
+--XFAIL--
+Persistable::bsonSerialize() should not return atomic modifiers (see: PHPC-550)
 --INI--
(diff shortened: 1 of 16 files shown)


tests/bson/bson-ods-002.phpt
```diff
@@ -1,131 +1,145 @@
...
 <?php
-use MongoDB\BSON as BSON;
-
 require_once __DIR__ . "/../utils/basic.inc";
 
-class Address implements BSON\Persistable {
-    protected $streetAddress;
-    protected $city;
-    protected $postalCode;
+use MongoDB\BSON as BSON;
```
derickr: Please don't change test cases! I realize it no longer works, and hence the XFAIL is fine, but I don't see why the test itself should be so massively changed.

jmikola: Removed the refactoring commit.

derickr:
LGTM

PR #227, opened 2016-02-10
PHPC-523: Parse readconcernlevel in URI options array

jmikola:
https://jira.mongodb.org/browse/PHPC-523

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1634,6 +1634,41 @@ void php_phongo_populate_default_ssl_ctx(php_stream_context *ctx, zval *driverOp
 #undef SET_STRING_CTX
 } /* }}} */
 
+static bool php_phongo_apply_rc_options_to_client(mongoc_client_t *client, bson_t *options TSRMLS_DC) /* {{{ */
+{
+	bson_iter_t iter;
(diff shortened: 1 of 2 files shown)


php_phongo.c
```diff
@@ -1644,6 +1644,41 @@ void php_phongo_populate_default_ssl_ctx(php_stream_context *ctx, zval *driverOp
...
 } /* }}} */
 
+static bool php_phongo_apply_rc_options_to_client(mongoc_client_t *client, bson_t *options TSRMLS_DC) /* {{{ */
+{
+	bson_iter_t iter;
+	mongoc_read_concern_t *new_rc;
+	const mongoc_read_concern_t *old_rc;
+
+	if (!(old_rc = mongoc_client_get_read_concern(client))) {
+		phongo_throw_exception(PHONGO_ERROR_MONGOC_FAILED TSRMLS_CC, "Client does not have a read concern");
```
derickr: I don't quite get what this is for. Is there a test for this? How would you trigger it? It it just a precaution?

jmikola: Just defensive programming. I haven't found a case where the pointer would actually be null without someone screwing around with the internal struct.

PR #230, opened 2016-02-12
Skip this test on HHVM, as its DateTime class is ancient and behaves differently.

derickr:


jmikola:
LGTM.

tests/bson/bson-utcdatetime-int-size.phpt
```diff
@@ -1,5 +1,7 @@
 --TEST--
 Test for UTCDateTime and integer parsing
+--SKIPIF--
+<?php if (defined("HHVM_VERSION_ID")) exit("skip HHVM implements only an older version of DateTime"); ?>
```
bjori: Shouldn't this rather check for correct DateTime behavior, so the test can start working when they fix DateTime?
Otherwise this will never be enabled, even when their DateTime is updated, possibly leading to bugs.

jmikola: In that case, it'd be better for Derick to clone this into the HHVM driver and mark it with XFAIL. Either way, I think we should periodically revisit skipped tests as HHVM improves to become more consistent with PHP.

derickr: I've cloned this one, and adapted it for HHVM already. Looks like I had forgotten to commit it though, but I have now: https://github.com/mongodb/mongo-hhvm-driver/blob/master/tests/bson-utcdatetime-int-size.phpt

PR #221, opened 2016-02-05
PHPC-553: Suppress PHP warnings during stream init

jmikola:
https://jira.mongodb.org/browse/PHPC-553

PR #232, opened 2016-02-12
PHPC-582: Manager::selectServer() should select exception class based on bson_error_t

jmikola:
https://jira.mongodb.org/browse/PHPC-582

Depends on #221, so please limit review to the final commit.


derickr:
LGTM

PR #234, opened 2016-02-12
PHPC-567: Upgrade libbson and libmongoc to 1.3.3

jmikola:
https://jira.mongodb.org/browse/PHPC-567


bjori:
I notice these bumps never bump the minimum dependency in config.m4.
Is that intentional?

jmikola:
> Is that intentional?

Certainly not. I'm just a noob.

PR #241, opened 2016-02-24
PHPC-544, PHPC-592: 64-bit platform detection and integer truncation

jmikola:
https://jira.mongodb.org/browse/PHPC-544
https://jira.mongodb.org/browse/PHPC-592

### DIFF
--- a/phongo_compat.h
+++ b/phongo_compat.h
@@ -92,34 +92,6 @@
 #  define ARG_UNUSED
 #endif
 
-#if SIZEOF_LONG == 4
-# define ADD_INDEX_INT64(zval, index, value) \
-	if (value > LONG_MAX || value < LONG_MIN) { \
-		char *tmp; \
-		int tmp_len; \
-		mongoc_log(MONGOC_LOG_LEVEL_WARNING, MONGOC_LOG_DOMAIN, "Integer overflow detected on your platform: %lld", value); \
-		tmp_len = spprintf(&tmp, 0, "%lld", value); \
-		ADD_INDEX_STRINGL(zval, index, tmp, tmp_len); \
-		efree(tmp); \
-	} else { \
-		add_index_long(zval, index, val); \
-	}
-# define ADD_ASSOC_INT64(zval, key, value) \
-	if (value > LONG_MAX || value < LONG_MIN) { \
-		char *tmp; \
-		int tmp_len; \
-		mongoc_log(MONGOC_LOG_LEVEL_WARNING, MONGOC_LOG_DOMAIN, "Integer overflow detected on your platform: %lld", value); \
-		tmp_len = spprintf(&tmp, 0, "%lld", value); \
-		ADD_ASSOC_STRING_EX(zval, key, strlen(key), tmp, tmp_len); \
-		efree(tmp); \
-	} else { \
-		add_assoc_long(zval, key, value); \
-	}
-#else
-# define ADD_INDEX_INT64(zval, index, value) add_index_long(zval, index, value)
-# define ADD_ASSOC_INT64(zval, key, value) add_assoc_long(zval, key, value);
-#endif
-
 #ifdef HAVE_ATOLL
 # define STRTOLL(s) atoll(s)
 #else
@@ -151,6 +123,8 @@
 # define phongo_char zend_string
 # define phongo_char_pdup(str) pestrdup(filename->val, 1)
 # define phongo_char_free(str) zend_string_release(str)
+# define phongo_long zend_long
+# define SIZEOF_PHONGO_LONG SIZEOF_ZEND_LONG
 # define phongo_str(str) str->val
 # define phongo_create_object_retval zend_object*
 # define PHONGO_ALLOC_OBJECT_T(_obj_t, _class_type) (_obj_t *)ecalloc(1, sizeof(_obj_t)+zend_object_properties_size(_class_type))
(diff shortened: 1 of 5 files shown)


phongo_compat.h
```diff
@@ -92,9 +92,17 @@
...
 
-#if SIZEOF_LONG == 4
+#if PHP_VERSION_ID < 70000
+typedef long zend_long;
+# define SIZEOF_ZEND_LONG SIZEOF_LONG
+#endif
+
+#if SIZEOF_ZEND_LONG == 8
+# define ADD_INDEX_INT64(zval, index, value) add_index_long(zval, index, value)
+# define ADD_ASSOC_INT64(zval, key, value) add_assoc_long(zval, key, value)
```
jmikola: These definitions were moved up to be consistent with `bson.c`, where we check for 64-bit platforms first. Additionally, I removed the trailing semi-colon from `ADD_ASSOC_INT64()`'s definition for consistency.

phongo_compat.h
```diff
@@ -92,9 +92,17 @@
 #  define ARG_UNUSED
 #endif
 
-#if SIZEOF_LONG == 4
+#if PHP_VERSION_ID < 70000
+typedef long zend_long;
+# define SIZEOF_ZEND_LONG SIZEOF_LONG
```
jmikola: The intention here is to allow use of `zend_long` for `zend_parse_parameters()` and simplify our architecture checks later, where we might otherwise have to check `SIZEOF_LONG` and `SIZEOF_ZEND_LONG` for PHP 5 and 7, respectively.

phongo_compat.h
```diff
@@ -92,9 +92,17 @@
...
+#endif
+
+#if SIZEOF_ZEND_LONG == 8
+# define ADD_INDEX_INT64(zval, index, value) add_index_long(zval, index, value)
+# define ADD_ASSOC_INT64(zval, key, value) add_assoc_long(zval, key, value)
+#elif SIZEOF_ZEND_LONG == 4
 # define ADD_INDEX_INT64(zval, index, value) \
-	if (value > LONG_MAX || value < LONG_MIN) { \
+	if (value > INT32_MAX || value < INT32_MIN) { \
```
jmikola: I found that these macros were inconsistent with those in `bson.c`, which used `INT_MAX` and `INT_MIN`. The `INT32` constants are more explicit.

src/bson.c
```diff
@@ -41,20 +41,18 @@
 #include "php_phongo.h"
 #include "php_bson.h"
 
-
-#define BSON_APPEND_INT32(b,key,val) \
-	bson_append_int32 (b, key, (int) strlen (key), val)
```
jmikola: This macro wasn't used at all, and it was actually already defined in libbson.

src/bson.c
```diff
@@ -41,20 +41,18 @@
...
 		bson_append_int64(b, key, keylen, val); \
 	} else { \
 		bson_append_int32(b, key, keylen, val); \
 	}
-#elif SIZEOF_LONG == 4
-#	define BSON_APPEND_INT(b, key, keylen, val) \
-	bson_append_int32(b, key, keylen, val);
+#elif SIZEOF_ZEND_LONG == 4
+# define BSON_APPEND_INT(b, key, keylen, val) \
+	bson_append_int32(b, key, keylen, val)
```
jmikola: Trailing semi-colon removed as the calling code is already careful to include it. This is consistent with how we define and use `ADD_ASSOC_INT64()` and `ADD_INDEX_INT64()` above.

tests/bson/bug0592.phpt
```diff
@@ -0,0 +1,94 @@
...
+}
+
+Test { "x": { "$numberLong": "2147483647" }}
+object(stdClass)#%d (%d) {
+  ["x"]=>
+  int(2147483647)
+}
+
+Test { "x": { "$numberLong": "4294967295" }}
+[%s]     PHONGO-BSON: WARNING   > ENTRY: Integer overflow detected on your platform: 4294967295
```
jmikola: I wasn't able to create a 64-bit Windows executable due to some DLL issues, so I can't confirm wjether the expected output for this test is accurate. We may need to revise.

phongo_compat.h
```diff
@@ -206,6 +182,35 @@
 # define PHP_STREAM_CONTEXT(stream) ((php_stream_context*) (stream)->context)
 #endif
 
+#if SIZEOF_PHONGO_LONG == 8
+# define ADD_INDEX_INT64(zval, index, value) add_index_long(zval, index, value)
+# define ADD_ASSOC_INT64(zval, key, value) add_assoc_long(zval, key, value)
```
jmikola: These definitions were moved up to be consistent with `bson.c`, where we check for 64-bit platforms first. Additionally, I removed the trailing semi-colon from `ADD_ASSOC_INT64()`'s definition for consistency.

phongo_compat.h
```diff
@@ -206,6 +182,35 @@
...
+#elif SIZEOF_PHONGO_LONG == 4
+# define ADD_INDEX_INT64(zval, index, value) \
+    if (value > INT32_MAX || value < INT32_MIN) { \
```
jmikola: I found that these macros were inconsistent with those in `bson.c`, which used `INT_MAX` and `INT_MIN`. The `INT32` constants are more explicit.

derickr:
LGTM - tests work for HHVM too.

PR #242, opened 2016-02-25
PHPC-414: Combine 32-bit and 64-bit UTCDateTime debug handler tests

jmikola:
https://jira.mongodb.org/browse/PHPC-414

This removes a SKIPIF from the 32-bit test but leaves the 64-bit test in place, since it tests construction from a 64-bit integer type instead of a string.


derickr:
LGTM - tests also pass for HHVM

PR #243, opened 2016-02-26
PHPC-593, PHPC-594, PHPC-595: Range checks for WriteConcern, Binary, and Timestamp

jmikola:
This PR depends on #241 (i.e. [PHPC-544](https://jira.mongodb.org/browse/PHPC-544)).

https://jira.mongodb.org/browse/PHPC-593
https://jira.mongodb.org/browse/PHPC-594
https://jira.mongodb.org/browse/PHPC-595

### DIFF
--- a/tests/bson/bson-timestamp_error-004.phpt
+++ b/tests/bson/bson-timestamp_error-004.phpt
@@ -1,0 +1,39 @@
+--TEST--
+BSON BSON\Timestamp constructor requires positive unsigned 32-bit integers
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+use MongoDB\BSON as BSON;
+
+echo throws(function() {
+    new BSON\Timestamp(-1, 0);
+}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
+
+echo throws(function() {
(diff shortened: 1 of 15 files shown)


tests/bson/bson-timestamp_error-004.phpt
```diff
@@ -0,0 +1,40 @@
+--TEST--
+BSON BSON\Timestamp constructor requires positive 64-bit integers <= UINT32_MAX
```
derickr: why does the title say positive 64-bit integers <= UNIT32_MAX? Surely, you could just say:

    requires positive unsigned 32-bit integers

derickr:
Besides the comment about the test case, this LGTM. The HHVM equivalent PR is https://github.com/mongodb/mongo-hhvm-driver/pull/69

tests/bson/bson-timestamp_error-004.phpt
jmikola: I believe I wanted to differentiate here because we do allow negative 32-bit integers (on 32-bit platforms), since we cast them to unsigned. The BSON timestamp type's two components are unsigned 32-bit integers. If we didn't cast 32-bit integers, we'd limit input to 2147483647 instead of 4294967295. I realize this is tricky and requires some documentation on PHP.net, but it seemed most flexible.

Anywho, only 64-bit integers are required to be positive, since we leave them as signed. How about:

```
requires 64-bit integers to be positive unsigned 32-bit integers
```

PR #248, opened 2016-03-01
Fixed PHPC-506: Use more descriptive messages in WriteExceptions

derickr:


php_phongo.c
```diff
@@ -522,6 +522,56 @@ mongoc_bulk_operation_t *phongo_bulkwrite_init(zend_bool ordered) { /* {{{ */
...
+		}
+
+		bson_iter_document(&iter, &len, &data);
+
+		if (!bson_init_static(&cbson, data, len)) {
+			continue;
+		}
+
+		if (bson_iter_init_find(&inner_iter, &cbson, "errmsg") && BSON_ITER_HOLDS_UTF8(&inner_iter)) {
+			char *tmp_errmsg = bson_iter_dup_utf8(&inner_iter, NULL);
```
jmikola: Any reason we can't simply use `bson_iter_utf8()` here, since `tmp_errmsg` is not going to be modified and we're just going to copy it into `tmp_msg`?

derickr: No reason whatsoever - I didn't know it existed :-)

PR #257, opened 2016-03-08
PHPC-564, PHPC-626: WriteResult should use libmongoc's public API

jmikola:
* https://jira.mongodb.org/browse/PHPC-564
 * https://jira.mongodb.org/browse/PHPC-626


tests/manager/manager-executeBulkWrite_error-002.phpt
```diff
@@ -32,7 +32,7 @@ var_dump(iterator_to_array($cursor));
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
-BulkWriteException: BulkWrite error :: E11000 duplicate key error %s: phongo.manager_manager_executeBulkWrite_error_002%sdup key: { : 1 } :: E11000 duplicate key error %s: phongo.manager_manager_executeBulkWrite_error_002%sdup key: { : 2 }
+BulkWriteException: Multiple write errors: "E11000 duplicate key error %s: phongo.manager_manager_executeBulkWrite_error_002%sdup key: { : 1 }", "E11000 duplicate key error %s: phongo.manager_manager_executeBulkWrite_error_002%sdup key: { : 2 }"
```
jmikola: @derickr: This is an example of the format change. libmongoc prepends "Multiple write errors" and then lists the messages wrapped in quotes and comma-delimited. The only other different I'm aware of is that the message will be capped at 504 bytes including the null terminator (since `bson_error_t` is statically allocated). I don't think that's a serious limitation, as the full messages will be available by inspecting the attached WriteResult object.

derickr: Static allocation? Seriously? :-/ PHP caps the message in warnings/errors to 1024 as well though... (not a problem for Exception messages though).

Do you think we can get them to change this value? Or increase it to 1024 or so?

jmikola: It's a compromise implemented for performance reasons, I imagine. We can certainly ask, but I don't think it's a deal breaker, since the real messages are in the WriteResult object. 500+ bytes is actually quite large when it comes to logging an exception message, and that would easily fit a handful of error messages before truncation.

jmikola:
@derickr: I believe I sorted out the PHP 5 memory issues, so this is ready for review.

src/MongoDB/WriteResult.c
```diff
@@ -97,17 +102,16 @@ PHP_METHOD(WriteResult, getModifiedCount)
 		return;
 	}
 
-	if (intern->write_result.omit_nModified) {
-		RETURN_NULL();
+	if (bson_iter_init_find(&iter, intern->reply, "nModified") && BSON_ITER_HOLDS_INT32(&iter)) {
+		RETURN_LONG(bson_iter_int32(&iter));
```
derickr: Not sure whether this is a like for like replacement. The new code doesn't use omit_nModified.

src/MongoDB/WriteResult.c
```diff
@@ -136,8 +142,9 @@ PHP_METHOD(WriteResult, getUpsertedCount)
 		return;
 	}
 
-
-	RETURN_LONG(intern->write_result.nUpserted);
+	if (bson_iter_init_find(&iter, intern->reply, "nUpserted") && BSON_ITER_HOLDS_INT32(&iter)) {
+		RETURN_LONG(bson_iter_int32(&iter));
+	}
```
derickr: The code with this iter_find, and RETURN_LONG is repeat 4/5 times. Wouldn't it make more sense to make a function (or macro) for this?

src/MongoDB/WriteResult.c
```diff
@@ -466,50 +465,111 @@ HashTable *php_phongo_writeresult_get_debug_info(zval *object, int *is_temp TSRM
...
+		ADD_ASSOC_ZVAL_EX(&retval, "writeErrors", writeErrors);
 #endif
+	}
 
-	if (!bson_empty0(&intern->write_result.writeConcernErrors) &&
-			bson_iter_init(&iter, &intern->write_result.writeConcernErrors) &&
-			bson_iter_next(&iter) &&
-			BSON_ITER_HOLDS_DOCUMENT(&iter)) {
+	if (bson_iter_init_find(&iter, intern->reply, "writeConcernErrors") && BSON_ITER_HOLDS_ARRAY(&iter) &&
+	    bson_iter_recurse(&iter, &child) && bson_iter_next(&child) && BSON_ITER_HOLDS_DOCUMENT(&child)) {
```
derickr: This "recursion" loop looks very similar to the code for getWriteConcern errors. Is it possible to make a private function for this so that code is not duplicated? The same can be said for the writeErrors and upsertedIds above.

tests/manager/manager-executeBulkWrite_error-002.phpt
derickr: OK

src/MongoDB/WriteResult.c
jmikola: libmongoc decides to include an `nModified` field in the `bson_t` based on `omit_nModified`. See: https://github.com/mongodb/mongo-c-driver/blob/d538a9c6fcd44b4fdf8e112b9223362a32e6e168/src/mongoc/mongoc-write-command.c#L1930

src/MongoDB/WriteResult.c
jmikola: I thought about it, we had other macros (e.g. `SCP()`) for copying over like fields, but wasn't sure if nested macros were kosher.

jmikola: Added macros in a subsequent commit.

src/MongoDB/WriteResult.c
jmikola: I don't think it's worth it, as the only commonality is the first level of BSON iteration, which fits within an `if` condition.

In the debug handler, we are just checking if "upsertedIds" and "writeErrors" are an array and then convert them to PHP using `phongo_bson_to_zval_ex()`. For "writeConcernErrors", we check if it's an array and then dive one more level in (checking for a document) before calling `phongo_bson_to_zval_ex()`.

In the getter methods, we create static `bson_t` buffers and use that to initialize objects, which are either returned immediately (`getWriteConcernError()`) or collected in an array (`getWriteErrors()` and `getUpsertedIds()`).

src/MongoDB/WriteResult.c
derickr: OK

src/MongoDB/WriteResult.c
derickr: Nested macros aren't great - some old Windows compilers don't like it, but I don't think we need to care about these any more.

src/MongoDB/WriteResult.c
derickr: OK

derickr:
LGTM

jmikola:
Merged to `v1.1` in https://github.com/mongodb/mongo-php-driver/commit/e6af431828acdee919ede8ce7bc7de99269a6b2a.

PR #261, opened 2016-03-11
[WIP] PHPC-542: Remove dependency on mongoc private symbols

jmikola:
* https://jira.mongodb.org/browse/PHPC-542
 * https://jira.mongodb.org/browse/PHPC-607
 * https://jira.mongodb.org/browse/PHPC-629
 * https://jira.mongodb.org/browse/PHPC-630
 * https://jira.mongodb.org/browse/PHPC-659

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -384,14 +384,26 @@ PHP_ARG_WITH(libmongoc, whether to use system libmongoc,
   else
     CPPFLAGS="$CPPFLAGS -DBSON_COMPILATION -DMONGOC_COMPILATION -DMONGOC_TRACE"
 
-    PHP_ADD_SOURCES_X(PHP_EXT_DIR(mongodb)[src/libmongoc/src/mongoc], $MONGOC_SOURCES,      [$STD_CFLAGS], shared_objects_mongodb, yes)
-    PHP_ADD_SOURCES_X(PHP_EXT_DIR(mongodb)[src/libmongoc/src/mongoc], $MONGOC_SOURCES_SSL,  [$STD_CFLAGS], shared_objects_mongodb, yes)
-    PHP_ADD_SOURCES_X(PHP_EXT_DIR(mongodb)[src/libmongoc/src/mongoc], $MONGOC_SOURCES_SASL, [$STD_CFLAGS], shared_objects_mongodb, yes)
+    PHP_ADD_SOURCES_X(PHP_EXT_DIR(mongodb)[src/libmongoc/src/mongoc], $MONGOC_SOURCES,                  [$STD_CFLAGS], shared_objects_mongodb, yes)
+    PHP_ADD_SOURCES_X(PHP_EXT_DIR(mongodb)[src/libmongoc/src/mongoc], $MONGOC_SOURCES_CRYPTO,           [$STD_CFLAGS], shared_objects_mongodb, yes)
+    PHP_ADD_SOURCES_X(PHP_EXT_DIR(mongodb)[src/libmongoc/src/mongoc], $MONGOC_SOURCES_SSL,              [$STD_CFLAGS], shared_objects_mongodb, yes)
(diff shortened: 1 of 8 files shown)


config.m4
```diff
@@ -390,6 +415,62 @@ if test "$PHP_PCRE_DIR" != "no"; then
   fi
 fi
 
+PHP_ARG_WITH(mongodb-openssl, for OpenSSL support,
+[  --with-mongodb-openssl[=DIR]     mongodb: Include OpenSSL support], auto, no)
```
bjori: The extension already had openssl support without this.
And it was configurable through --with-openssl-dir. This doesn't look right at all

src/MongoDB/Manager.c
```diff
@@ -279,7 +279,9 @@ PHP_METHOD(Manager, selectServer)
 	}
 
 	readPreference = phongo_read_preference_from_zval(zreadPreference TSRMLS_CC);
-	selected_server = mongoc_topology_select(intern->client->topology, MONGOC_SS_READ, readPreference, MONGOC_SS_DEFAULT_LOCAL_THRESHOLD_MS, &error);
+	/* TODO: Remove (mongoc_read_prefs_t *) casting once
+	 * https://github.com/mongodb/mongo-c-driver/pull/315 is merged. */
+	selected_server = mongoc_client_select_server(intern->client, true, (mongoc_read_prefs_t *) readPreference, &error);
```
derickr: When I implement a similar thing for HHVM, I get tests failing. For example "tests/functional/cursor-001.phpt" fails with:

```
Fatal error: Uncaught exception 'MongoDB\Driver\Exception\ConnectionTimeoutException' with message 'Cannot use read preferences with for_writes = true' in /home/derick/dev/php/derickr-mongo-php-driver/tests/utils/tools.php:106
Stack trace:
#0 /home/derick/dev/php/derickr-mongo-php-driver/tests/utils/tools.php(106): MongoDB\Driver\Manager->selectServer()
#1 /home/derick/dev/php/derickr-mongo-php-driver/tests/functional/cursor-001.phpt(5): LOAD()
```

The "true" is the second argument to client_select_server, so I don't think this is the right fix. I am curious as to why this does not fail for you though.

derickr: I've just checked the libmongoc docs as well, and it states:

```
      <tr><td><p>for_writes</p></td><td><p>Whether to choose a server suitable for writes or reads.</p></td></tr>
      <tr><td><p>prefs</p></td><td><p>An optional <code xref="mongoc_read_prefs_t">mongoc_read_prefs_t</code>. If <code>for_writes</code> is true, <code>prefs</code> must be NULL. Otherwi
se, use <code>prefs</code> to customize server selection, or pass NULL to use the read preference configured on the client.</p></td></tr>
```

As we don't know yet whether people want to read and/or write, I'm not sure what to do here. The original topology_select uses MONGOC_SS_READ, which I guess means a non-writeable server perhaps?

We could set the "true" depending on whether the readPreference is RP_PRIMARY, and only then? But then an RP of PRIMARY_PREFERED or SECONDARY_PREFERED, which ends up selecting a non-master server wouldn't allow for writes...

jmikola: My mistake. I should have been passing `false` for `for_writes`. Aside from the error checking (`for_writes=true` requires that no read prefs are passed`), this ultimately gets used in `mongoc_topology_description_suitable_servers()` and merely determines whether read prefs get applied for selection. And when a writable server is required, it can skip read pref checking (and also not require a primary-only read pref).

jmikola:
Superseded by:

 * https://github.com/mongodb/mongo-php-driver/pull/276
 * https://github.com/mongodb/mongo-php-driver/pull/277
 * https://github.com/mongodb/mongo-php-driver/pull/278
 * https://github.com/mongodb/mongo-php-driver/pull/279
 * https://github.com/mongodb/mongo-php-driver/pull/280

PR #265, opened 2016-03-16
PHPC-633: Add SSL/SASL support to phpinfo() output

derickr:


php_phongo.c
```diff
@@ -2583,6 +2583,16 @@ PHP_MINFO_FUNCTION(mongodb)
 #else
 	/* Bundled libraries, buildtime = runtime */
 	php_info_print_table_row(2, "libmongoc version", MONGOC_VERSION_S);
+# if HAVE_MONGOC_SSL
```
bjori: I'd suggest using the mongoc defines for this instead: MONGOC_ENABLE_SSL, MONGOC_ENABLE_SASL.
This should be updated later to print out with ssl lib it is using, maybe add a note to PHPC-542 ?

derickr: These constants were not defined in the .h files that I needed it to be in:

```
src/libmongoc/src/mongoc/mongoc-config.h:#define MONGOC_ENABLE_SSL 1
win32/mongoc-config.h:#define MONGOC_ENABLE_SSL 1
```

And I don't want to expose vendored library's defines outside of it. Having specific defines for them makes for a cleaner separation.
```

bjori: So when compiling against system mongoc these well say no sasl and ssl support.
That doesn't look right.

jmikola:
@derickr: Was [PHPC-663](https://jira.mongodb.org/browse/PHPC-663) a typo? That issue looks unrelated to this PR.

I'd like to revisit this once https://github.com/mongodb/mongo-php-driver/pull/361 is merged, since we're now relying on libmongoc for TLS.

derickr:
Looks like I meant PHPC-633

jmikola:
See https://github.com/mongodb/mongo-php-driver/pull/498.

PR #267, opened 2016-03-23
Fix PHPC-631

asp24:
Fixed  timelib_time.fraction init int php_phongo_new_datetime_from_utcdatetime 
according to [PHPC-631](https://jira.mongodb.org/browse/PHPC-631)


jmikola:
Thanks for the PR. I've incorporated this patch into https://github.com/mongodb/mongo-php-driver/pull/338, where I've also added additional regression tests.

PR #276, opened 2016-04-01
PHPC-629: Upgrade libbson and libmongoc to 1.4.0

jmikola:
https://jira.mongodb.org/browse/PHPC-629

This PR upgrades the libbson and libmongoc submodules and fixes compilation. Tests failures are still expected. [PHPC-629](https://jira.mongodb.org/browse/PHPC-629) depends on the following issues, which are included in this PR:

 * https://jira.mongodb.org/browse/PHPC-607
 * https://jira.mongodb.org/browse/PHPC-610
 * https://jira.mongodb.org/browse/PHPC-611
 * https://jira.mongodb.org/browse/PHPC-630
 * https://jira.mongodb.org/browse/PHPC-659
 * https://jira.mongodb.org/browse/PHPC-662

Note: tests currently have a slew of memory leaks due to [CDRIVER-1192](https://jira.mongodb.org/browse/CDRIVER-1192). This is apparent when the find command is supported, as it affects many test cases.


config.m4
```diff
@@ -496,6 +533,7 @@ dnl }}}
 
   BSON_EXTRA_ALIGN=0
   AC_SUBST(BSON_EXTRA_ALIGN)
+  AC_SUBST(BSON_HAVE_DECIMAL128, 0)
```
derickr: We need to set a flag to get DECIMAL128 compiled into libbson?

src/bson.c
```diff
@@ -574,6 +574,8 @@ static const bson_visitor_t php_bson_visitors = {
    php_phongo_bson_visit_int64,
    php_phongo_bson_visit_maxkey,
    php_phongo_bson_visit_minkey,
+   NULL /*php_phongo_bson_visit_unsupported_type*/,
+   NULL /*php_phongo_bson_visit_decimal128*/,
```
derickr: Because you set this BSON_DECIMAL128 flag above, would it make sense to do a conditional define for it here anyway too?

derickr:
LGTM, after we:

 - [ ] Spoke about the BSON flag for decimal128

src/bson.c
jmikola: The visitor struct includes this regardless of the flag, so it's safe as-is.

config.m4
jmikola: Correct, and we need to at least define it as 0.

bjori: Just FYI: That flag isn't about BSON Decimal128 support though.

BSON Decimal128 is always available in libbson 1.4.0.
You'll be able to get a bson_decimal128_t from a string by calling
```
bson_decimal128_t decimal128;
if (bson_decimal128_from_string("0.1234", &decimal128)) {
    /* Conversion successful */
} else {
    /* Thats not a valid Decimal128 representation */
}
```

And you can convert that decimal128 to a string with
```
char string[BSON_DECIMAL128_STRING];
bson_decimal128_to_string (&decimal128, &string)
```

This is always available in libbson 1.4

`BSON_HAVE_DECIMAL128` is a check for `_Decimal128` native type.

This allows you to get a native `_Decimal128` type from `bson_decimal128_t` to use with math libraries that support that. And, to convert a `_Decimal128` into `bson_decimal128_t` for storage in MongoDB.

Similarly, you'll want on your `BSON\Decimal128` class to provide, when available:

```
$decimal128->getAsGmp();
$decimal128->getAsBCMath();
```

jmikola: Thanks for clarifying `BSON_HAVE_DECIMAL128`. I misunderstood the purpose of the constant.

> Similarly, you'll want on your BSON\Decimal128 class to provide, when available: `getAsGmp()`, `getAsBCMath()`

Cross-referencing with existing discussion in [PHPC-640](https://jira.mongodb.org/browse/PHPC-640).

PR #277, opened 2016-04-01
Consolidate shell variable assignments with AC_SUBST()

jmikola:
See: https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Setting-Output-Variables.html


derickr:
LGTM

PR #280, opened 2016-04-01
PHPC-664: Use mongoc_write_concern_get_w()

jmikola:
https://jira.mongodb.org/browse/PHPC-664

### DIFF
--- a/src/MongoDB/WriteConcern.c
+++ b/src/MongoDB/WriteConcern.c
@@ -137,7 +137,7 @@ PHP_METHOD(WriteConcern, getW)
 		PHONGO_RETURN_STRING(PHONGO_WRITE_CONCERN_W_MAJORITY);
 	}
 
-	if (intern->write_concern->w != MONGOC_WRITE_CONCERN_W_DEFAULT) {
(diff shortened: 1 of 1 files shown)


src/MongoDB/WriteConcern.c
```diff
@@ -137,7 +137,7 @@ PHP_METHOD(WriteConcern, getW)
 		PHONGO_RETURN_STRING(PHONGO_WRITE_CONCERN_W_MAJORITY);
 	}
 
-	if (intern->write_concern->w != MONGOC_WRITE_CONCERN_W_DEFAULT) {
+	if (mongoc_write_concern_get_w(intern->write_concern) != MONGOC_WRITE_CONCERN_W_DEFAULT) {
```
derickr: Isn't there an API to check whether it's the default value now too? Or an isset one? I remember seeing that somewhere.

jmikola: Journal has an "is_set" method because the public API only returns a boolean; however, w's states were already defined as public API constants, so no method was needed.

Issue #281, opened 2016-04-05
Inserting a document with a field "_ids" does not return the generated ObjectID


csizmazia:
From the documentation, BulkWrite::insert($document) should:

> If the document did not have an _id, a MongoDB\BSON\ObjectID will be generated and returned; otherwise, no value is returned.

However, when i want to insert a document which does not have a `_id` field, but does have a `_ids` field, this does not follow the above rule. `NULL` is returned instead of the expected generated ObjectID. The insert itself (performed by a Manager) works, though.

This happens with the latest version 1.1.5 from PECL.

jmikola:
Confirmed as a bug, which is being tracked in [PHPC-667](https://jira.mongodb.org/browse/PHPC-667). Thanks for the report!

jmikola:
https://github.com/mongodb/mongo-php-driver/pull/284 is merged to v1.1 and will be released in 1.1.6. Please track the JIRA ticket above for updates.

PR #284, opened 2016-04-05
PHPC-667: Fix check for existing "_id" field during BulkWrite::insert()

jmikola:
https://jira.mongodb.org/browse/PHPC-667 and https://github.com/mongodb/mongo-php-driver/issues/281


derickr:
LGTM

PR #290, opened 2016-04-07
PHPC-600: Use mongoc_client_get_server_descriptions() to get Manager's Servers

jmikola:
https://jira.mongodb.org/browse/PHPC-600

### DIFF
--- a/src/MongoDB/Manager.c
+++ b/src/MongoDB/Manager.c
@@ -29,7 +29,6 @@
 #include <mongoc.h>
 #include <mongoc-client-private.h>
 #include <mongoc-host-list-private.h>
-#include <mongoc-server-description-private.h>
 
 /* PHP Core stuff */
 #include <php.h>
@@ -201,8 +200,8 @@ PHP_METHOD(Manager, getReadPreference)
 PHP_METHOD(Manager, getServers)
 {
 	php_phongo_manager_t         *intern;
-	mongoc_set_t                 *set;
-	size_t                        i;
+	mongoc_server_description_t **sds;
+	size_t                        i, n = 0;
 	SUPPRESS_UNUSED_WARNING(return_value_ptr) SUPPRESS_UNUSED_WARNING(return_value_used)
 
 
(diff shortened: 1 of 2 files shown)


derickr:
I don't think this is enough. There is another ```intern->client->topology->description.servers``` in the debug info that needs to use mongoc_client_get_server_descriptions() too. Just like I changed for https://jira.mongodb.org/browse/HHVM-196 (which incidentally, has the other location in the title): https://github.com/mongodb/mongo-hhvm-driver/pull/78/files#diff-dcd03e4e31145855552adc282c2fb08cL474

After make that change as well, you will see ```tests/manager/manager-var-dump-001.phpt``` fail, as there is a dummy broken server in the cluster output.

src/MongoDB/Manager.c
```diff
@@ -212,31 +212,25 @@ PHP_METHOD(Manager, getServers)
...
+		zval *obj = NULL;
 
-		phongo_server_init(obj, intern->client, sd->id TSRMLS_CC);
+		MAKE_STD_ZVAL(obj);
+		phongo_server_init(obj, intern->client, mongoc_server_description_id(sds[i]) TSRMLS_CC);
 		add_next_index_zval(return_value, obj);
 #endif
 	}
+
+	mongoc_server_descriptions_destroy_all(sds, n);
```
jmikola: @derickr: I believe https://github.com/mongodb/mongo-hhvm-driver/pull/78 was missing this. Can you confirm if you're leaking memory via the array returned from `mongoc_client_get_server_descriptions()`?

derickr: Correct: https://github.com/mongodb/mongo-hhvm-driver/pull/90

src/MongoDB/Manager.c
```diff
@@ -212,31 +212,25 @@ PHP_METHOD(Manager, getServers)
...
-		phongo_server_init(&obj, intern->client, ((mongoc_server_description_t *)set->items[i].item)->id TSRMLS_CC);
+		phongo_server_init(&obj, intern->client, mongoc_server_description_id(sds[i]) TSRMLS_CC);
 		add_next_index_zval(return_value, &obj);
 #else
-		MAKE_STD_ZVAL(obj);
+		zval *obj = NULL;
 
-		phongo_server_init(obj, intern->client, sd->id TSRMLS_CC);
+		MAKE_STD_ZVAL(obj);
+		phongo_server_init(obj, intern->client, mongoc_server_description_id(sds[i]) TSRMLS_CC);
```
derickr: Do you really need the method call ```mongoc_server_description_id(sds[i]) ``` here? I believe, ```sds[i]->id``` should be safe too? As far as I know, the ```server_description_t``` type is public.

jmikola: `mongoc-server-description.h` only defines:

```
typedef struct _mongoc_server_description_t mongoc_server_description_t;
```

`_mongoc_server_description_t` is defined in `mongoc-server-description-private.h`.

derickr:
LGTM

derickr:
LGTM

PR #291, opened 2016-04-07
PHPC-606: Use mongoc_server_description_t public API in Server methods

jmikola:
https://jira.mongodb.org/browse/PHPC-606

Possible refactoring in the debug handler: serialize "last_is_master" first, and then copy the already-serialized "tags" field. Alternatively, we can decide to remove the duplicate, top-level "tags" field.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -81,6 +81,19 @@ ZEND_DECLARE_MODULE_GLOBALS(mongodb)
 #endif
 #endif
 
+php_phongo_server_description_type_map_t
+php_phongo_server_description_type_map[PHONGO_SERVER_DESCRIPTION_TYPES] = {
+	{ PHONGO_SERVER_UNKNOWN, "Unknown" },
+	{ PHONGO_SERVER_STANDALONE, "Standalone" },
+	{ PHONGO_SERVER_MONGOS, "Mongos" },
+	{ PHONGO_SERVER_POSSIBLE_PRIMARY, "PossiblePrimary" },
+	{ PHONGO_SERVER_RS_PRIMARY, "RSPrimary" },
+	{ PHONGO_SERVER_RS_SECONDARY, "RSSecondary" },
+	{ PHONGO_SERVER_RS_ARBITER, "RSArbiter" },
+	{ PHONGO_SERVER_RS_OTHER, "RSOther" },
+	{ PHONGO_SERVER_RS_GHOST, "RSGhost" },
+};
+
 /* {{{ phongo_std_object_handlers */
 zend_object_handlers phongo_std_object_handlers;
 
@@ -1275,6 +1288,20 @@ void php_phongo_objectid_new_from_oid(zval *object, const bson_oid_t *oid TSRMLS
 	bson_oid_to_string(oid, intern->oid);
 } /* }}} */
 
+php_phongo_server_description_type_t php_phongo_server_description_type(mongoc_server_description_t *sd)
+{
+	const char* name = mongoc_server_description_type(sd);
+	int i;
+
+	for (i = 0; i < PHONGO_SERVER_DESCRIPTION_TYPES; i++) {
+		if (!strcmp(name, php_phongo_server_description_type_map[i].name)) {
+			return php_phongo_server_description_type_map[i].type;
+		}
+	}
+
+	return PHONGO_SERVER_UNKNOWN;
+}
+
 void php_phongo_server_to_zval(zval *retval, mongoc_server_description_t *sd) /* {{{ */
 {
 	mongoc_host_list_t *host = mongoc_server_description_host(sd);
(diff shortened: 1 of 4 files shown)


src/MongoDB/Server.c
```diff
@@ -236,11 +252,13 @@ PHP_METHOD(Server, getLatency)
 		return;
 	}
 
-	if ((sd = mongoc_topology_description_server_by_id(&intern->client->topology->description, intern->server_id, &error))) {
-		RETURN_LONG(sd->round_trip_time);
+	if ((sd = mongoc_client_get_server_description(intern->client, intern->server_id))) {
+		RETVAL_LONG((phongo_long) mongoc_server_description_round_trip_time(sd));
```
jmikola: `mongoc_server_description_round_trip_time()` returns an `int64_t`. It's unlikely that a round trip time would overflow a 32-bit integer, but are we OK with a simple cast for that edge case? This also applies to the Server's debug handler.

php_phongo.c
```diff
@@ -81,6 +81,19 @@ ZEND_DECLARE_MODULE_GLOBALS(mongodb)
 #endif
 #endif
 
+php_phongo_server_description_type_map_t
+php_phongo_server_description_type_map[PHONGO_SERVER_DESCRIPTION_TYPES] = {
+	{ PHONGO_SERVER_UNKNOWN, "Unknown" },
+	{ PHONGO_SERVER_STANDALONE, "Standalone" },
```
derickr: As the map is from string to int, shouldn't the order be the other way around here? I.e., first string, and then the int constant?

php_phongo.c
```diff
@@ -1275,55 +1288,71 @@ void php_phongo_objectid_new_from_oid(zval *object, const bson_oid_t *oid TSRMLS
...
-	ADD_ASSOC_BOOL_EX(retval, "is_arbiter", sd->type == MONGOC_SERVER_RS_ARBITER);
-	{
-		bson_iter_t iter;
-		zend_bool b = bson_iter_init_find_case(&iter, &sd->last_is_master, "hidden") && bson_iter_as_bool(&iter);
+	const char* name = mongoc_server_description_type(sd);
+	int i;
 
-		ADD_ASSOC_BOOL_EX(retval, "is_hidden", b);
+	for (i = 0; i < PHONGO_SERVER_DESCRIPTION_TYPES; i++) {
+		if (!strcmp(name, php_phongo_server_description_type_map[i].name)) {
```
derickr: How often, and when is this run? A whole loop is going to be slow. If this runs a lot, we should re-order the items in the map, so that the most likely show up first.

php_phongo.h
```diff
@@ -130,7 +153,9 @@ const mongoc_read_prefs_t*    phongo_read_preference_from_zval(zval *zread_prefe
 const mongoc_write_concern_t* phongo_write_concern_from_zval  (zval *zwrite_concern TSRMLS_DC);
 const php_phongo_query_t*     phongo_query_from_zval          (zval *zquery TSRMLS_DC);
 
-void php_phongo_server_to_zval(zval *retval, const mongoc_server_description_t *sd);
+php_phongo_server_description_type_t php_phongo_server_description_type(mongoc_server_description_t *sd);
+
+void php_phongo_server_to_zval(zval *retval, mongoc_server_description_t *sd);
```
derickr: You ended up removing the "const" in front of the argument ```*sd```. Was that on purpose?

jmikola: `mongoc_client_get_server_description()` doesn't return a const pointer, and the various `mongoc_server_description_X()` functions we now call in the debug handler also don't take a const pointer. I opted to just relax the argument rather than force a non-const cast within this function. That said, I don't believe they mutate the server description in any way.

php_phongo.c
jmikola: This is only called in two places:

* `Server::getType()`
* Reporting "type" in the Server's debug info

Note that keeping the map order in sync with the enum allows us to do quick lookups on strings for the `Server::isX()` methods (and debug handler).

```
ADD_ASSOC_BOOL_EX(retval, "is_primary", !strcmp(mongoc_server_description_type(sd), php_phongo_server_description_type_map[PHONGO_SERVER_RS_PRIMARY].name));
ADD_ASSOC_BOOL_EX(retval, "is_secondary", !strcmp(mongoc_server_description_type(sd), php_phongo_server_description_type_map[PHONGO_SERVER_RS_SECONDARY].name));
ADD_ASSOC_BOOL_EX(retval, "is_arbiter", !strcmp(mongoc_server_description_type(sd), php_phongo_server_description_type_map[PHONGO_SERVER_RS_ARBITER].name));
```

I don't think either case is sensitive enough that we need to change this, though.

php_phongo.c
jmikola: I didn't think the field order was important. Each element is a two-field struct with `type` and `name` fields. The array is technically a map from the enum integer to a string. We can technically use the array offset in lieu of storing the `type` field and make this an array of strings, but I figured the two-field struct was more readable.

derickr:
LGTM

src/MongoDB/Server.c
derickr: Yes

PR #297, opened 2016-04-13
PHPC-672: Fix zend_str_tolower usage

remicollet:
https://jira.mongodb.org/browse/PHPC-672

See the reproducer:

     $args = [ 'id' => 'T123456' ];
     $_id = $args['id'];
     try {
          var_dump(new \MongoDB\BSON\ObjectID($_id));
     } catch (\MongoDB\Driver\Exception\InvalidArgumentException $e) {
          var_dump($e);
     }
     var_dump($args, $_id);

Output

    array(1) {
      ["id"]=>
      string(7) "t123456"
    }
    string(7) "t123456"

I really think input should not be altered....

### DIFF
--- a/src/BSON/ObjectID.c
+++ b/src/BSON/ObjectID.c
@@ -53,7 +53,7 @@ PHP_METHOD(ObjectID, __construct)
 {
 	php_phongo_objectid_t    *intern;
 	zend_error_handling       error_handling;
-	char                     *id = NULL;
+	char                     *id = NULL, *tid;
 	phongo_zpp_char_len       id_len;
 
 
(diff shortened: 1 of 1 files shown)


remicollet:
FYI @trasher

src/BSON/ObjectID.c
```diff
@@ -67,15 +67,17 @@ PHP_METHOD(ObjectID, __construct)
...
-		if (bson_oid_is_valid(id, id_len)) {
+		tid = zend_str_tolower_dup(id, id_len);
+		if (bson_oid_is_valid(tid, id_len)) {
 			bson_oid_t oid;
 
-			bson_oid_init_from_string(&oid, id);
+			bson_oid_init_from_string(&oid, tid);
 			bson_oid_to_string(&oid, intern->oid);
 		} else {
+			efree(tid);
```
jmikola: Is this a double free, given that you're doing it again a couple of lines below?

remicollet: Right!
I was thinking throw_exception doesn't return...
Fixed

derickr: That's ok @remicollet - we get confused with this as well, as in HHVM it *does* return ;-)

jmikola:
Manually merged in https://github.com/mongodb/mongo-php-driver/commit/3f8f9742f6edb56f99ad393c41b6ffc551c97bc5. I also ported the reproduction steps above to a test case and bundled it with your commit. We should have this included in the next 1.1.x release. Thanks!

remicollet:
Thanks

PR #303, opened 2016-04-20
PHPC-550: Remove XFAIL ODS tests for removed functionality

jmikola:
https://jira.mongodb.org/browse/PHPC-550

These tests relied on returning atomic modifiers via Persistable::bsonSerialize(), which is not supported.


derickr:
LGTM

PR #306, opened 2016-04-22
PHPC-554: Rephrase unsupported/corrupt BSON messages

derickr:
https://jira.mongodb.org/browse/PHPC-554

### DIFF
--- a/src/bson.c
+++ b/src/bson.c
@@ -183,7 +183,13 @@ void php_phongo_bson_visit_corrupt(const bson_iter_t *iter ARG_UNUSED, void *dat
 {
 	mongoc_log(MONGOC_LOG_LEVEL_TRACE, MONGOC_LOG_DOMAIN, "Corrupt BSON data detected!");
 }
-/* }}} */
+
+void php_phongo_bson_visit_unsupported_type(const bson_iter_t *iter ARG_UNUSED, const char *key, uint32_t v_type_code, void *data ARG_UNUSED) /* {{{ */
+{
+	TSRMLS_FETCH();
+	phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Detected unknown BSON type 0x%02hhx for fieldname \"%s\". Are you using the latest driver?", v_type_code, key);
+}
+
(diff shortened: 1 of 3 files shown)


src/bson.c
```diff
@@ -1401,8 +1406,12 @@ PHONGO_API int phongo_bson_to_zval_ex(const unsigned char *data, int data_len, p
 	if (bson_iter_visit_all(&iter, &php_bson_visitors, state) || iter.err_off) {
 		/* Iteration stopped prematurely due to corruption or a failed visitor.
 		 * While we free the reader, state->zchild should be left as-is, since
-		 * the calling code may want to zval_ptr_dtor() it. */
-		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Could not convert BSON document to a PHP variable");
+		 * the calling code may want to zval_ptr_dtor() it. If an exception has
+		 * been thrown already (due to an unsupported BSON type for example,
+		 * don't overwrite with a generic exception message. */
+		if (!EG(exception)) {
+			phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Detected corrupt BSON data.");
```
jmikola: I realize you used the same message in https://github.com/mongodb/mongo-hhvm-driver/commit/1d51fe96e2fdc24596f4a47affbfbbfb1f34c239, but should there really be a period after this message? We typically don't use that, and only did so in the corruption message since it consists of two sentences.

jmikola:
LGTM apart from the question on the exception message format.

tests/bson/bson-toPHP_error-004.phpt
```diff
@@ -35,13 +35,13 @@ foreach ($tests as $bson) {
 <?php exit(0); ?>
 --EXPECTF--
 OK: Got MongoDB\Driver\Exception\UnexpectedValueException
-Could not convert BSON document to a PHP variable
+Detected corrupt BSON data.
```
jmikola: I believe you forgot to remove the trailing period in these expected exception messages. Fixed in https://github.com/mongodb/mongo-php-driver/commit/e6deaeb59ae24120edcf15f9dbd6ad7357c164fc.

PR #311, opened 2016-04-26
PHPC-680: BulkWrite::count() should return number of operations

jmikola:
https://jira.mongodb.org/browse/PHPC-680


jmikola:
Note: test failures are due to [PHPC-673](https://jira.mongodb.org/browse/PHPC-673).

PR #315, opened 2016-04-26
PHPC-682, PHPC-684: BulkWrite debug handler improvements

jmikola:
* https://jira.mongodb.org/browse/PHPC-682
 * https://jira.mongodb.org/browse/PHPC-684

I've grouped these issues into the same PR since PHPC-684 depends on PHPC-682's changes and the combined diff is still concise.

Combined with https://github.com/mongodb/mongo-php-driver/pull/311, this should resolve issues with depending on libmongoc's private API in BulkWrite.


jmikola:
Note: test failures may be due to [PHPC-673](https://jira.mongodb.org/browse/PHPC-673).

PR #316, opened 2016-04-28
PHPC-689: Use mongoc_cursor_set_hint() for query/command execution

jmikola:
https://jira.mongodb.org/browse/PHPC-689

We currently have two test failures:

 * tests/server/server-executeCommand_error-001.phpt
 * tests/server/server-executeQuery_error-001.phpt

Both of these tests expected RuntimeExceptions to be thrown due to "not master and slaveOk=false"; however, `mongoc_cursor_set_hint()` now forces the `slaveOk` flag (see: [CDRIVER-903](https://jira.mongodb.org/browse/CDRIVER-903)). With that change, the queries are now sent to the secondary server:

 * `executeCommand()` fails with a different RuntimeException message: "Failed to decode document from the server." originating from [`_mongoc_rpc_parse_error()`](https://github.com/mongodb/mongo-c-driver/blob/9a31e4c9adc8159ce64d43942b86999eab0eb940/src/mongoc/mongoc-rpc.c#L819). This will require some investigation (via `mongodb.debug` tracing) to see what is actually being sent to the secondary. It is unclear if the error is within PHPC or libmongoc.
 * `executeQuery()` simply executes and returns a cursor.

The Server method documentation has the following notes about the `$readPreference` parameter:

> The readPreference parameter does not control the server to which the driver issues the command; the command will always be executed on this server object. Instead, it may be used when issuing the command to a secondary (from a replica set connection, not standalone) or mongos node to ensure that the driver sets the wire protocol accordingly or adds the read preference to the command document, respectively.

The bit about "setting the wire protocol accordingly" is no longer relevant, since `slaveOk` will always be set. I would only expect the read preference to be used when communicating with a mongos node. We should create new tests to verify that behavior (tracking in [PHPC-691](https://jira.mongodb.org/browse/PHPC-691)).

In the meantime, the aforementioned tests should be revised to no longer expect exceptions and we should make necessary changes to PHPC or libmongoc to ensure that the `executeCommand()` test passes.


derickr:
FYI, if I make the changes to use mongoc_cursor_set_hint(), I am getting the exact same tests failing, with the same reasons. executeQuery_error-001 misses the exception, and executeCommand_error-001 gives a "Failed to decode document from the server."

Also, see: https://github.com/mongodb/mongo-hhvm-driver/pull/99

jmikola:
@derickr: The weird exception was due to passing a namespace instead of a database name to `Server::executeCommand()`. I suppose this was always a problem for the server, but `slaveOk` previously being omitted prevented the server from getting that far. I fixed the test in one commit and noted that it then started passing (just like the `Server::executeQuery()` test). The second commit renames and revises both of those tests, as expected.

derickr:
LGTM

PR #317, opened 2016-04-28
PHPC-693: Unacknowledged write results report null counts

jmikola:
https://jira.mongodb.org/browse/PHPC-693


derickr:
LGTM - looks like I need no code change for HHVM.

PR #320, opened 2016-05-05
PHPC-688: Change Cursor debug handler to use libmongoc public API

jmikola:
https://jira.mongodb.org/browse/PHPC-688

This significantly alters the structure of the Cursor's debug output, since we not longer have access to private fields within mongoc_cursor_t.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -103,6 +103,9 @@ PHONGO_API zend_object_handlers *phongo_get_std_object_handlers(void)
 }
 /* }}} */
 
+/* Forward declarations */
+static bool phongo_split_namespace(const char *namespace, char **dbname, char **cname);
+
 /* {{{ Error reporting and logging */
 zend_class_entry* phongo_exception_from_phongo_domain(php_phongo_error_domain_t domain)
 {
@@ -235,7 +238,7 @@ static void php_phongo_log(mongoc_log_level_t log_level, const char *log_domain,
 /* }}} */
 
 /* {{{ Init objects */
-static void phongo_cursor_init(zval *return_value, mongoc_cursor_t *cursor, mongoc_client_t *client TSRMLS_DC) /* {{{ */
+static void phongo_cursor_init(zval *return_value, mongoc_cursor_t *cursor, mongoc_client_t *client, zval *readPreference TSRMLS_DC) /* {{{ */
 {
 	php_phongo_cursor_t *intern;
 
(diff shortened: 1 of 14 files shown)


php_phongo.c
```diff
@@ -234,8 +234,27 @@ static void php_phongo_log(mongoc_log_level_t log_level, const char *log_domain,
 
 /* }}} */
 
+/* Splits a namespace name into the database and collection names, allocated with estrdup. */
+static bool phongo_split_namespace(const char *namespace, char **dbname, char **cname) /* {{{ */
```
jmikola: This was moved up since `phongo_cursor_init_for_query()` depends on it. We don't have any other early declarations in `php_phongo.c`, and I opted not to start just for this.

php_phongo.c
```diff
@@ -245,6 +264,50 @@ static void phongo_cursor_init(zval *return_value, mongoc_cursor_t *cursor, mong
...
+
+static void phongo_cursor_init_for_query(zval *return_value, mongoc_cursor_t *cursor, mongoc_client_t *client, const char *namespace, zval *zquery, zval *zreadPreference TSRMLS_DC) /* {{{ */
+{
+	php_phongo_cursor_t *intern;
+
+	phongo_cursor_init(return_value, cursor, client, zreadPreference TSRMLS_CC);
+	intern = Z_CURSOR_OBJ_P(return_value);
+
+	/* namespace has already been validated by phongo_execute_query() */
+	phongo_split_namespace(namespace, &intern->database, &intern->collection);
```
jmikola: While we could pass in `database` and `collection` strings here, that would require copying the `efree()` lines for each of the 4-5 early return statements in `phongo_execute_query()`. It seemed easier to just call `phongo_split_namespace()` a second time here.

php_phongo.c
```diff
@@ -1269,13 +1318,6 @@ const mongoc_read_prefs_t* phongo_read_preference_from_zval(zval *zread_preferen
 
 	return NULL;
 } /* }}} */
-
-const php_phongo_query_t* phongo_query_from_zval(zval *zquery TSRMLS_DC) /* {{{ */
```
jmikola: No longer needed since `phongo_execute_query()` now takes the Query zval and runs it through `Z_QUERY_OBJ_P()`.

php_phongo_classes.h
```diff
@@ -130,7 +130,6 @@
 typedef struct {
 	zend_object_iterator   intern;
 	php_phongo_cursor_t   *cursor;
-	long                   current;
```
jmikola: This was relocated to our cursor struct so we don't need a bi-directional reference between `php_phongo_cursor_iterator` and `php_phongo_cursor_t`. Also, the current document is already stored there (in `php_phongo_bson_state visitor_data`).

tests/manager/manager-executeCommand-001.phpt
```diff
@@ -37,62 +37,36 @@ object(MongoDB\Driver\Command)#%d (1) {
...
+  NULL
+  ["isDead"]=>
+  bool(true)
+  ["currentIndex"]=>
+  int(0)
+  ["currentDocument"]=>
+  NULL
+  ["server"]=>
+  object(MongoDB\Driver\Server)#%d (%d) {
+    %a
```
jmikola: The goal of these tests was to assert on `server_id`, so I think we're OK with ignoring the Server's debug output here. The alternative is including it and ignoring `last_is_master` with `%a`. In tests where we do want to check that a cursor operated on a particular server, we do so with `Cursor::getServer()`.

php_phongo.c
derickr: I'd have preferred the forwards declaration as it would minimize changes that are not really related to the change at hand.

jmikola: I'll switch to that.

src/MongoDB/Server.c
```diff
@@ -91,19 +88,18 @@ PHP_METHOD(Server, executeQuery)
 	char                     *namespace;
 	phongo_zpp_char_len       namespace_len;
 	zval                     *zquery;
-	zval                     *readPreference = NULL;
+	zval                     *zreadPreference = NULL;
```
derickr: I'd probably have removed the "z" in front of query as well. Our coding standard doesn't generally do Hungarian Notation :-)

derickr:
LGTM

PR #322, opened 2016-05-09
PHPC-673: Fix Cursor::isDead() tests for mongoc_cursor_is_alive()

jmikola:
https://jira.mongodb.org/browse/PHPC-673

This bump includes the fix [CDRIVER-1221](https://jira.mongodb.org/browse/CDRIVER-1221), which entails changes to our `Cursor::isDead()` tests (among others where we dump cursors).

### DIFF
--- a/tests/cursor/cursor-isDead-003.phpt
+++ b/tests/cursor/cursor-isDead-003.phpt
@@ -1,12 +1,12 @@
 --TEST--
-MongoDB\Driver\Cursor::isDead() with basic iteration (find command)
+MongoDB\Driver\Cursor::isDead() with basic iteration (OP_QUERY)
 --SKIPIF--
-<?php require __DIR__ . "/../utils/basic-skipif.inc"; CLEANUP(STANDALONE) ?>
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; CLEANUP(STANDALONE_30) ?>
 --FILE--
 <?php
 require_once __DIR__ . "/../utils/basic.inc";
 
-$manager = new MongoDB\Driver\Manager(STANDALONE);
+$manager = new MongoDB\Driver\Manager(STANDALONE_30);
 
 $bulk = new MongoDB\Driver\BulkWrite();
 $bulk->insert(['_id' => 1]);
--- a/tests/manager/manager-executeCommand-001.phpt
+++ b/tests/manager/manager-executeCommand-001.phpt
@@ -54,7 +54,7 @@ object(MongoDB\Driver\Cursor)#%d (%d) {
   ["readPreference"]=>
   NULL
   ["isDead"]=>
-  bool(true)
+  bool(false)
   ["currentIndex"]=>
   int(0)
   ["currentDocument"]=>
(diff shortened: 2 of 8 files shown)


tests/cursor/cursor-isDead-001.phpt
```diff
@@ -9,22 +9,25 @@ require_once __DIR__ . "/../utils/basic.inc";
...
 
+var_dump($cursor->isDead());
+
 ?>
 ===DONE===
 <?php exit(0); ?>
 --EXPECT--
 bool(false)
 bool(false)
+bool(false)
```
jmikola: Note: In libmongoc 1.3.x, `mongoc_cursor_is_alive()` would return false once the cursor entered the final batch. Since this behavior was fixed in [CDRIVER-1221](https://jira.mongodb.org/browse/CDRIVER-1221), our expectation needed to change.

tests/manager/manager-executeCommand-001.phpt
```diff
@@ -54,7 +54,7 @@ object(MongoDB\Driver\Cursor)#%d (%d) {
   ["readPreference"]=>
   NULL
   ["isDead"]=>
-  bool(true)
+  bool(false)
```
jmikola: Per https://github.com/mongodb/mongo-php-driver/pull/322/files#r63595485, the cursor is no longer dead because we've yet to iterate to the first and only result document of the command.

tests/manager/manager-executeQuery-001.phpt
```diff
@@ -71,7 +71,7 @@ object(MongoDB\Driver\Cursor)#%d (%d) {
   ["readPreference"]=>
   NULL
   ["isDead"]=>
-  bool(true)
+  bool(false)
```
jmikola: Per https://github.com/mongodb/mongo-php-driver/pull/322/files#r63595485, the cursor is no longer dead because we've yet to iterate to any of the documents in the single-batch reply for this query.

tests/cursor/cursor-isDead-001.phpt
jmikola: It's interesting to note that this same test would fail on PHPC 1.1.x. If we extend the test to insert 8 documents and query with a batch size of 3, we can see that `mongoc_cursor_is_alive()` starts return false once the 7th document is reached (i.e. we've entered the final batch, even though we still have more documents to iterate).

tests/cursor/cursor-isDead-003.phpt
```diff
@@ -0,0 +1,33 @@
...
+
+$bulk = new MongoDB\Driver\BulkWrite();
+$bulk->insert(['_id' => 1]);
+$bulk->insert(['_id' => 2]);
+$bulk->insert(['_id' => 3]);
+$manager->executeBulkWrite(NS, $bulk);
+
+$cursor = $manager->executeQuery(NS, new MongoDB\Driver\Query([], ['batchSize' => 2]));
+
+foreach ($cursor as $_) {
```
derickr: The ``$_`` looks like perl ;-)

derickr:
LGTM

PR #324, opened 2016-05-10
PHPC-698, PHPC-699: Check HAVE_OPENSSL_EXT before calling php-ssl.c functions

jmikola:
* https://jira.mongodb.org/browse/PHPC-698
 * https://jira.mongodb.org/browse/PHPC-699

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1012,10 +1012,15 @@ int php_phongo_peer_verify(php_stream *stream, X509 *cert, const char *hostname,
 			peer = hostname;
 		}
 
+#ifdef HAVE_OPENSSL_EXT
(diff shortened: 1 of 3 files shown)


derickr:
LGTM

bjori:
Wouldn't you rather want to error out much earlier, say before calling php_phongo_ssl_verify (), if there are ssl context options set, while ssl is not enabled/available?

jmikola:
@bjori: This PR is specific to the user compiling the extension without SSL but loading in a PHP runtime with SSL. I'm fine with keeping the error messages narrow around the code that would actually rely on these symbols. The user needn't be bothered if the stream options are disabled.

The inverse case is if SSL isn't available at runtime. For that, the user would already get an exception during stream initialization and before we attempt to verify any certs or hostnames.

PR #330, opened 2016-05-31
PHPC-671: Avoid mongoc_client_t use-after-free by Cursor and Server

jmikola:
https://jira.mongodb.org/browse/PHPC-671

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -224,7 +224,7 @@ static void php_phongo_log(mongoc_log_level_t log_level, const char *log_domain,
 /* }}} */
 
 /* {{{ Init objects */
-static void phongo_cursor_init(zval *return_value, mongoc_cursor_t *cursor, mongoc_client_t *client TSRMLS_DC) /* {{{ */
+static void phongo_cursor_init(zval *return_value, zval *manager, mongoc_cursor_t *cursor TSRMLS_DC) /* {{{ */
 {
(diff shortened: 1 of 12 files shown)


php_phongo.c
```diff
@@ -233,18 +233,32 @@ static void phongo_cursor_init(zval *return_value, mongoc_cursor_t *cursor, mong
 	intern = Z_CURSOR_OBJ_P(return_value);
 	intern->cursor = cursor;
 	intern->server_id = mongoc_cursor_get_hint(cursor);
-	intern->client = client;
+	intern->client = Z_MANAGER_OBJ_P(manager)->client;
+
+#if PHP_VERSION_ID >= 70000
+	ZVAL_COPY(&intern->manager, manager);
```
derickr: Are you sure this just doesn't create a shallow copy of the ZVAL - and hence doesn't really affect the ref counting of the manager structs stored in it?

derickr:
LGTM

php_phongo.c
jmikola: Per the [`ZVAL_COPY()` definition](http://lxr.php.net/xref/PHP_7_0/Zend/zend_types.h#857), the ref count is incremented if the type is considered to be ref-counted.

This is effectively the same as `ZVAL_ZVAL(dst, src, 1, 0)` (yes copy, no dtor).

PR #336, opened 2016-06-21
PHPC-536: UTCDateTime construction from DateTime or current time

jmikola:
https://jira.mongodb.org/browse/PHPC-536

### DIFF
--- a/src/BSON/UTCDateTime.c
+++ b/src/BSON/UTCDateTime.c
@@ -33,10 +33,15 @@
 #include <php_ini.h>
 #include <ext/standard/info.h>
 #include <Zend/zend_interfaces.h>
+#include <ext/date/php_date.h>
 #include <ext/spl/spl_iterators.h>
 #include <ext/date/php_date.h>
 /* Our Compatability header */
 #include "phongo_compat.h"
+
+#ifdef PHP_WIN32
+#include "win32/time.h"
+#endif
 
 /* Our stuffz */
 #include "php_phongo.h"
@@ -47,16 +52,59 @@ PHONGO_API zend_class_entry *php_phongo_utcdatetime_ce;
 
 zend_object_handlers php_phongo_handler_utcdatetime;
 
-/* {{{ proto BSON\UTCDateTime UTCDateTime::__construct(integer $milliseconds)
+static void php_phongo_utcdatetime_init_from_current_time(php_phongo_utcdatetime_t *intern)
+{
+	int64_t        sec, usec;
+	struct timeval cur_time;
+
+	gettimeofday(&cur_time, NULL);
+	sec = cur_time.tv_sec;
+	usec = cur_time.tv_usec;
+
+	intern->milliseconds = (sec * 1000) + (usec / 1000);
+}
+
+static void php_phongo_utcdatetime_init_from_date(php_phongo_utcdatetime_t *intern, php_date_obj *datetime_obj)
+{
+	int64_t sec, usec;
+
+	/* The following assignments use the same logic as date_format() in php_date.c */
+	sec = datetime_obj->time->sse;
+	usec = (int64_t) floor(datetime_obj->time->f * 1000000 + 0.5);
+
+	intern->milliseconds = (sec * 1000) + (usec / 1000);
+}
+
+/* {{{ proto BSON\UTCDateTime UTCDateTime::__construct([integer|DateTimeInterface $milliseconds = null])
    Construct a new UTCDateTime */
 PHP_METHOD(UTCDateTime, __construct)
 {
 	php_phongo_utcdatetime_t    *intern;
-	zend_error_handling       error_handling;
-
+	zend_error_handling          error_handling;
+	zval                        *datetime = NULL;
 
 	zend_replace_error_handling(EH_THROW, phongo_exception_from_phongo_domain(PHONGO_ERROR_INVALID_ARGUMENT), &error_handling TSRMLS_CC);
 	intern = Z_UTCDATETIME_OBJ_P(getThis());
+
+	if (zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "|o!", &datetime) == SUCCESS) {
+		if (datetime == NULL) {
+			php_phongo_utcdatetime_init_from_current_time(intern);
+		} else if (instanceof_function(Z_OBJCE_P(datetime), php_date_get_date_ce() TSRMLS_CC)) {
+			php_phongo_utcdatetime_init_from_date(intern, Z_PHPDATE_P(datetime));
+#if PHP_VERSION_ID >= 50500
+		} else if (instanceof_function(Z_OBJCE_P(datetime), php_date_get_immutable_ce() TSRMLS_CC)) {
+			php_phongo_utcdatetime_init_from_date(intern, Z_PHPDATE_P(datetime));
+#endif
+		} else {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC,
+				"Expected instance of DateTimeInterface, %s given",
+				phongo_str(Z_OBJCE_P(datetime)->name)
+			);
+		}
+
+		zend_restore_error_handling(&error_handling TSRMLS_CC);
+		return;
+	}
 
 #if SIZEOF_PHONGO_LONG == 8
 	{
--- a/tests/bson/bson-utcdatetime-004.phpt
+++ b/tests/bson/bson-utcdatetime-004.phpt
@@ -1,0 +1,30 @@
+--TEST--
+BSON BSON\UTCDateTime constructor defaults to current time
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$tests = [
+    new MongoDB\BSON\UTCDateTime,
+    new MongoDB\BSON\UTCDateTime(null),
+];
+
+foreach ($tests as $test) {
+    var_dump($test);
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+object(%SBSON\UTCDateTime)#%d (%d) {
+  ["milliseconds"]=>
+  int(%d)
+}
+object(%SBSON\UTCDateTime)#%d (%d) {
+  ["milliseconds"]=>
+  int(%d)
+}
+===DONE===
--- a/tests/bson/bson-utcdatetime-005.phpt
+++ b/tests/bson/bson-utcdatetime-005.phpt
@@ -1,0 +1,40 @@
+--TEST--
+BSON BSON\UTCDateTime construction from DateTime
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$tests = [
+    new DateTime(),
+    new DateTime('@1215282385'),
+    new DateTime('2011-01-01T15:03:01.012345Z'),
+    new DateTime('2050-11-12T13:14:15.999999Z'),
+];
+
+foreach ($tests as $test) {
+    var_dump(new MongoDB\BSON\UTCDateTime($test));
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+object(%SBSON\UTCDateTime)#%d (%d) {
+  ["milliseconds"]=>
+  int(%d000)
+}
+object(%SBSON\UTCDateTime)#%d (%d) {
+  ["milliseconds"]=>
+  int(1215282385000)
+}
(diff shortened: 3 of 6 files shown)


moderndeveloperllc:
Sorry for my programming ignorance, but I noticed that you were using `DateTime` in the code and tests. Is this code `DateTimeInterface`-compatible? It may also be advisable to throw in a pre-1970 date in the tests just to make sure that negative numbers are playing nicely. I'm looking forward to this landing soon!

src/BSON/UTCDateTime.c
```diff
@@ -46,17 +51,43 @@ PHONGO_API zend_class_entry *php_phongo_utcdatetime_ce;
...
+			php_date_obj *datetime_obj;
+
+			datetime_obj = Z_PHPDATE_P(datetime);
+			sec = datetime_obj->time->sse;
+			usec = (int64_t) floor(datetime_obj->time->f * 1000000 + 0.5);
+		} else {
+			/* Initialize UTCDateTime from current time */
+			struct timeval cur_time;
+
+			gettimeofday(&cur_time, NULL);
```
derickr: Windows doesn't have ```gettimeofday```.

tests/bson/bson-utcdatetime-005.phpt
```diff
@@ -0,0 +1,40 @@
...
+
+$tests = [
+    new DateTime(),
+    new DateTime('@1215282385'),
+    new DateTime('2011-01-01T15:03:01.012345Z'),
+    new DateTime('2050-11-12T13:14:15.999999Z'),
+];
+
+foreach ($tests as $test) {
+    var_dump(new MongoDB\BSON\UTCDateTime($test));
```
derickr: I think you should try to round trip them back with ``toDateTime`` as well.

src/BSON/UTCDateTime.c
```diff
@@ -46,17 +51,43 @@ PHONGO_API zend_class_entry *php_phongo_utcdatetime_ce;
...
+		if (datetime) {
+			/* Initialize UTCDateTime from DateTime object. The following
+			 * assignments use the same logic as date_format() in php_date.c */
+			php_date_obj *datetime_obj;
+
+			datetime_obj = Z_PHPDATE_P(datetime);
+			sec = datetime_obj->time->sse;
+			usec = (int64_t) floor(datetime_obj->time->f * 1000000 + 0.5);
+		} else {
+			/* Initialize UTCDateTime from current time */
```
derickr: Shouldn't this be a different ```zend_parse_parameters_ex```? I think I'd like a case for all three options? But then again, this does work too...

tests/bson/bson-utcdatetime-004.phpt
```diff
@@ -0,0 +1,30 @@
...
+BSON BSON\UTCDateTime constructor defaults to current time
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$tests = [
+    new MongoDB\BSON\UTCDateTime,
+    new MongoDB\BSON\UTCDateTime(null),
```
derickr: I think "null" should just be coerced to ``0``, and not it being empty. Hence, the milliseconds should end up being 0 (e.g., the Epoch)

src/BSON/UTCDateTime.c
jmikola: I considered the options in http://stackoverflow.com/q/10905892/162228, but then I found `gettimeofday` and the necessary structs defined in `win32/time.h` (going back to at least PHP 5.4). I also found evidence of that implementation being used in several core extensions.

tests/bson/bson-utcdatetime-004.phpt
jmikola: The current behavior is consistent with your [HHVM implementation](https://github.com/mongodb/mongo-hhvm-driver/blob/73bc3daa7691184a6e091585ad42c01cb965fe41/ext_mongodb.php#L829). Note that DateTime's constructor also defaults to the current time if `null` is provided. I say we keep this behavior and document it.

tests/bson/bson-utcdatetime-005.phpt
jmikola: Sounds like a good idea. Let's get https://github.com/mongodb/mongo-php-driver/pull/338 merged first in that case, since it fixes `toDateTime()` and the expected output.

src/BSON/UTCDateTime.c
jmikola: `zend_parse_parameters()` doesn't have a pattern to pick up `null`, which is equivalent to no argument at all, so this seemed easiest. Otherwise, I would need to change the original `zend_parse_parameters()` calls below (for string/integer parsing) to each allow for `null`.

jmikola:
> Sorry for my programming ignorance, but I noticed that you were using `DateTime` in the code and tests. Is this code `DateTimeInterface`-compatible?

DateTimeInterface only exists in PHP 5.5, but we should be able to use that if it's available and fall back to DateTime for PHP 5.4.

@derick: I noticed that `php_date.h` only exports class entries for DateTime and DateTimeImmutable:

```
PHPAPI zend_class_entry *php_date_get_date_ce(void);
PHPAPI zend_class_entry *php_date_get_immutable_ce(void);
```

Can we get away with defining `extern zend_class_entry * date_ce_interface` and relying on variable in `php_date.c` (it's not declared static). Alternatively, do we need to use `zend_fetch_class()` and resolve the interface name string to a class entry?

> It may also be advisable to throw in a pre-1970 date in the tests just to make sure that negative numbers are playing nicely. I'm looking forward to this landing soon!

Great idea. I'll work on getting these in and try to come up with some tests that exercise the range limits of PHP and BSON dates.

jmikola:
@derickr: Regarding the Travis failure, that appears to be `bson-utcdatetime_error-001.phpt`, which was asserting that `UTCDateTime::__construct()` threw when no argument was passed. Since that test case is no longer valid, I propose we delete it and decrement the other "-002" and "-003" tests.

moderndeveloperllc:
> DateTimeInterface only exists in PHP 5.5, but we should be able to use that if it's available and fall back to DateTime for PHP 5.4.

Sorry. That's just me forgetting that the driver is compatible back to 5.4. Being able to pass in `DateTimeInterface` classes would be great regardless.

tests/bson/bson-utcdatetime-004.phpt
derickr: OK

src/BSON/UTCDateTime.c
derickr: OK

tests/bson/bson-utcdatetime-005.phpt
derickr: OK

src/BSON/UTCDateTime.c
derickr: OK

derickr:
> Can we get away with defining extern zend_class_entry * date_ce_interface and relying on variable in php_date.c (it's not declared static).

No. This works on Linux, but won't work on Windows.

> Alternatively, do we need to use zend_fetch_class() and resolve the interface name string to a class entry?

That sounds kinda crappy... I'd rather export the interface in PHP (from PHP 7.1).

jmikola:
>> Alternatively, do we need to use zend_fetch_class() and resolve the interface name string to a class entry?
> That sounds kinda crappy... I'd rather export the interface in PHP (from PHP 7.1).

Of course it makes sense to start doing that in PHP 7.1, but `zend_fetch_class()` seems necessary to get this working today.

src/BSON/UTCDateTime.c
```diff
@@ -47,17 +52,58 @@ PHONGO_API zend_class_entry *php_phongo_utcdatetime_ce;
...
+
+	#if PHP_VERSION_ID >= 70000
+		zend_string *zs_classname = zend_string_init("DateTimeInterface", sizeof("DateTimeInterface")-1, 0);
+		found_ce = zend_fetch_class(zs_classname, ZEND_FETCH_CLASS_AUTO|ZEND_FETCH_CLASS_NO_AUTOLOAD|ZEND_FETCH_CLASS_SILENT);
+		zend_string_free(zs_classname);
+	#else
+		found_ce = zend_fetch_class("DateTimeInterface", sizeof("DateTimeInterface")-1, ZEND_FETCH_CLASS_AUTO|ZEND_FETCH_CLASS_NO_AUTOLOAD|ZEND_FETCH_CLASS_SILENT TSRMLS_CC);
+	#endif
+
+	return found_ce ? found_ce : php_date_get_date_ce();
```
jmikola: @derickr: Would it make sense to memoize this function by saving `found_ce` in a static global? I also thought about initializing the constructors `date_ce` in UTCDateTime's MINIT, but that assumes we are always initialized after PHP's date extension. I'd like to respect the edge case where that may not be true.

src/BSON/UTCDateTime.c
```diff
@@ -47,17 +52,58 @@ PHONGO_API zend_class_entry *php_phongo_utcdatetime_ce;
 
 zend_object_handlers php_phongo_handler_utcdatetime;
 
-/* {{{ proto BSON\UTCDateTime UTCDateTime::__construct(integer $milliseconds)
+static zend_class_entry *php_phongo_utcdatetime_get_date_ce(TSRMLS_D)
+{
+	zend_class_entry *found_ce = NULL;
+
+	#if PHP_VERSION_ID >= 70000
```
derickr: The # needs to be in the first column.

src/BSON/UTCDateTime.c
derickr: You can add a dependency on it, and that should make PHP initialise datetime first. Instead of MINIT, I would put it in RINIT.

src/BSON/UTCDateTime.c
```diff
@@ -47,17 +52,60 @@ PHONGO_API zend_class_entry *php_phongo_utcdatetime_ce;
...
 	intern = Z_UTCDATETIME_OBJ_P(getThis());
 
+#if PHP_VERSION_ID >= 50500
+	if (zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "O", &datetime, php_date_get_immutable_ce()) == SUCCESS) {
+		php_phongo_utcdatetime_init_from_date(intern, Z_PHPDATE_P(datetime));
+
+		zend_restore_error_handling(&error_handling TSRMLS_CC);
+		return;
+	}
+#endif
```
jmikola: I ran into problems calling `zend_fetch_class()` to get the class entry for DateTimeInterface during MINIT. This ultimately seemed an easier solution, which uses nothing more than the date extension's public API.

Meanwhile, I've opened [PHPC-718](https://jira.mongodb.org/browse/PHPC-718) to ensure that we start listing our extension dependencies.

src/BSON/UTCDateTime.c
```diff
@@ -47,17 +52,60 @@ PHONGO_API zend_class_entry *php_phongo_utcdatetime_ce;
...
+		if (datetime) {
+			php_phongo_utcdatetime_init_from_date(intern, Z_PHPDATE_P(datetime));
+		} else {
+			php_phongo_utcdatetime_init_from_current_time(intern);
+		}
+
+		zend_restore_error_handling(&error_handling TSRMLS_CC);
+		return;
+	}
+
```
derickr: I think if you now pass it in as a non datetime class, it will not throw an error/exception.

jmikola: We still end up with an InvalidArgumentException:

> MongoDB\BSON\UTCDateTime::__construct() expects parameter 1 to be long, object given

In the HHVM driver, you end up casting the argument to an integer if it's neither `null` nor a DateTime instance. Would you like to intercept other objects and throw a special message?

derickr: Yes.

jmikola: I went with "Expected instance of DateTimeInterface, $classname given". We technically check for an instance of DateTime and DateTimeImmutable (on 5.5+); however, DateTimeInterface cannot be implemented by userland classes so the message is still accurate.

tests/bson/bson-utcdatetime_error-001.phpt
```diff
@@ -1,22 +1,19 @@
 --TEST--
-BSON BSON\UTCDateTime #001 error
---INI--
-date.timezone=America/Los_Angeles
+BSON BSON\UTCDateTime requires object argument to implement DateTimeInterface
 --SKIPIF--
-<?php if (defined("HHVM_VERSION_ID")) exit("skip HHVM handles parameter parsing differently"); ?>
```
derickr: We might have to put this one back in I suppose, if we find that it's still different with HHVM.

derickr:
LGTM

tests/bson/bson-utcdatetime_error-001.phpt
jmikola: I don't think it will be a problem. Your constructor accepts a mixed type and all validation is left to you, so it should be trivial to customize the exception message here.

PR #338, opened 2016-06-21
PHPC-631: Fix timelib_time.f assignment in UTCDateTime::toDateTime()

jmikola:
https://jira.mongodb.org/browse/PHPC-631

This includes the patch from #267 and adds additional tests.

PR #339, opened 2016-06-21
Move php_phongo_new_datetime_from_utcdatetime() into UTCDateTime.c

jmikola:
This function was only used by UTCDateTime::toDateTime(), so it does not need to be in php_phongo.c. Additionally, this commit removes the platform-specific spprintf patterns in favor of the portable PRId64 pattern.

This PR depends on https://github.com/mongodb/mongo-php-driver/pull/338.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -2068,25 +2068,6 @@ void php_phongo_new_utcdatetime_from_epoch(zval *object, int64_t msec_since_epoc
 	intern->milliseconds = msec_since_epoch;
 } /* }}} */
 
-void php_phongo_new_datetime_from_utcdatetime(zval *object, int64_t milliseconds TSRMLS_DC) /* {{{ */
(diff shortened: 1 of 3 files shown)


derickr:
I don't mind the function being moved, but I do not agree to the inlining of its contents. And, from what I understood, PRId64, is C99 only. I am not sure if that's what PHP's base line is.

jmikola:
> I don't mind the function being moved, but I do not agree to the inlining of its contents.

I didn't see the point for just a few lines of code that is only used in one method; however, I'll move this to a static function inside UTCDateTime.c.

> And, from what I understood, PRId64, is C99 only. I am not sure if that's what PHP's base line is.

We already use `PRId64` elsewhere in the extension. AFAIK, libmongoc is C99.

derickr:
LGTM

PR #340, opened 2016-06-23
PHPC-460: BSON class support for serialization and var_export

jmikola:
https://jira.mongodb.org/browse/PHPC-460


src/BSON/Decimal128.c
```diff
@@ -148,24 +218,38 @@ phongo_create_object_retval php_phongo_decimal128_create_object(zend_class_entry
...
+	HashTable               *props;
+	char                     outbuf[BSON_DECIMAL128_STRING] = "";
 
 	intern = Z_DECIMAL128_OBJ_P(object);
-	*is_temp = 1;
-	array_init_size(&retval, 1);
+	props = zend_std_get_properties(object TSRMLS_CC);
+
+	/* TODO: track if intern is uninitialized and return early. This may require
+	 * creating an extra field, since intern->decimal is not a pointer */
```
jmikola: @derickr: Let's decide how best to handle this. I left similar todo items in the Timestamp and UTCDateTime classes as well.

src/BSON/Javascript.c
```diff
@@ -46,6 +46,49 @@ PHONGO_API zend_class_entry *php_phongo_javascript_ce;
...
 
+/* Initialize the object from a string and return whether it was successful. */
+static bool php_phongo_javascript_initialize(php_phongo_javascript_t *intern, const char *javascript, phongo_zpp_char_len javascript_len, zval *scope TSRMLS_DC)
+{
+	if (scope && Z_TYPE_P(scope) != IS_OBJECT && Z_TYPE_P(scope) != IS_ARRAY && Z_TYPE_P(scope) != IS_NULL) {
+		return false;
+	}
+
+	intern->javascript = estrndup(javascript, javascript_len);
+	intern->javascript_len = javascript_len;
```
jmikola: Side note: we may simply get rid of this field in the future (see: [PHPC-721](https://jira.mongodb.org/browse/PHPC-721)).

src/BSON/Regex.c
```diff
@@ -46,6 +46,38 @@ PHONGO_API zend_class_entry *php_phongo_regex_ce;
 
 zend_object_handlers php_phongo_handler_regex;
 
+/* Initialize the object from a string and return whether it was successful. */
+static bool php_phongo_regex_initialize(php_phongo_regex_t *intern, const char *pattern, phongo_zpp_char_len pattern_len, const char *flags, phongo_zpp_char_len flags_len)
+{
+	intern->pattern = estrndup(pattern, pattern_len);
+	intern->pattern_len = pattern_len;
+	intern->flags = estrndup(flags, flags_len);
+	intern->flags_len = flags_len;
```
jmikola: Both regex strings (pattern and flags) are cstring types in the BSON spec, so the length fields are likely not necessary here as well. Nothing we need to remove in this PR, though.

src/BSON/Timestamp.c
```diff
@@ -46,6 +46,36 @@ PHONGO_API zend_class_entry *php_phongo_timestamp_ce;
 
 zend_object_handlers php_phongo_handler_timestamp;
 
+/* Initialize the object from a string and return whether it was successful. */
+static bool php_phongo_timestamp_initialize(php_phongo_timestamp_t *intern, phongo_long increment, phongo_long timestamp)
```
jmikola: The range check to ensure these numbers are 4-byte integers (i.e. `(x < 0 || x > UINT32_MAX)`) only takes place in the constructor. I think a test should be added to ensure we cannot pass larger values through `unserialize()` and `__set_state()` on 64-bit, even though they will be truncated in the assignments below.

tests/bson/bson-binary-set-state-001.phpt
```diff
@@ -0,0 +1,56 @@
...
+))
+
+MongoDB\BSON\Binary::__set_state(array(
+   'data' => '>Eg�ӤVBfUD' . "\0" . '' . "\0" . '',
+   'type' => 4,
+))
+
+MongoDB\BSON\Binary::__set_state(array(
+   'data' => '8X�"0�<�_0
fC�?',
```
jmikola: GitHub is not showing `bson-binary-002.phpt` because it contains binary data. The `get_debug_info` handler previously printed the raw data and I opted to do the same with `get_properties`. PHP's documentation mentioned that users should be prepared for serialization data to be a binary string (possibly containing null bytes and other control characters), and I don't see a limitation with `var_export()`. Could this be a problem for `var_dump()` output?

As-is, this makes `__set_state()` consistent with the constructor in accepting raw data. If you would rather change `get_properties` to return hex-encoded data, we can then convert data from hex to binary during `__set_state()` and `__wakeup()`. Let's create a separate ticket for that if desired.

tests/bson/bson-timestamp-006.phpt
```diff
@@ -0,0 +1,56 @@
+--TEST--
+MongoDB\BSON\Timestamp serialization (64-bit)
```
jmikola: On 32-bit, it's possible for Timestamp to serialize its increment or timestamp as a negative number. Such a Timestamp could not currently be created from the constructor (zpp parses as a long), but may be created by decoding valid BSON. I believe this only presents a problem if a 64-bit platform were to unserialize data from a 32-bit platform.

On a related note, 32-bit systems cannot create valid Timestamp BSON if either value exceeds 2147483647 (largest 32-bit signed integer). Should we switch to parsing arguments as strings for 32-bit, like we do with UTCDateTime? From there, we can use `STRTOLL()` and always check the valid range appropriately. It would also be easy to support cross-platform serialization (much like UTCDateTime's initialization, which can pick up long and string values from either platform).

src/BSON/Javascript.c
derickr: If you mean just the length bit? I disagree - as it's handy when strnduping it to return to the user. And the value is "free".

src/BSON/Javascript.c
```diff
@@ -84,9 +165,17 @@ ZEND_BEGIN_ARG_INFO_EX(ai_Javascript___construct, 0, 0, 1)
...
+ZEND_END_ARG_INFO();
+
+ZEND_BEGIN_ARG_INFO_EX(ai_Javascript_void, 0, 0, 0)
+ZEND_END_ARG_INFO();
+
 static zend_function_entry php_phongo_javascript_me[] = {
 	PHP_ME(Javascript, __construct, ai_Javascript___construct, ZEND_ACC_PUBLIC|ZEND_ACC_FINAL)
-	PHP_ME(Manager, __wakeUp, NULL, ZEND_ACC_PUBLIC)
+	PHP_ME(Javascript, __set_state, ai_Javascript___set_state, ZEND_ACC_PUBLIC|ZEND_ACC_STATIC)
+	PHP_ME(Javascript, __wakeup, ai_Javascript_void, ZEND_ACC_PUBLIC)
```
derickr: I guess it's missing a __toString then...

src/BSON/Regex.c
derickr: Again, I think we should stay \0 safe - especially because it's so cheap for us to do.

src/BSON/Timestamp.c
derickr: OK

src/BSON/Javascript.c
jmikola: Alrighty, then it stays.

src/BSON/Javascript.c
jmikola: Javascript and Binary are both missing a `__toString()` implementation ([PHPC-729](https://jira.mongodb.org/browse/PHPC-729)).

Javascript also has no getters for code and scope ([PHPC-730](https://jira.mongodb.org/browse/PHPC-730)).

src/BSON/Regex.c
jmikola: As with Javascript, the length fields can stay.

tests/bson/bson-binary-set-state-001.phpt
derickr: Binary is fine.

tests/bson/bson-timestamp-006.phpt
jmikola: Related: https://jira.mongodb.org/browse/PHPC-726

src/BSON/Decimal128.c
jmikola: I added `initialized` fields to ObjectID, Decimal128, Timestamp, and UTCDateTime.

PR #342, opened 2016-06-23
Consolidate PHP 5 and 7 struct declarations

jmikola:


jmikola:
@derickr: If I remove pecl-compat from this, are we good to merge? I can then rebase #340 atop this.

derickr:
Yeah.

PR #346, opened 2016-07-13
PHPC-741: Consistent exceptions for BSON init methods

jmikola:
https://jira.mongodb.org/browse/PHPC-741

Due to overlap, I've also included the PRs for disallowing null bytes in Javascript's code and Regex's pattern and flags fields:

 * https://jira.mongodb.org/browse/PHPC-721
 * https://jira.mongodb.org/browse/PHPC-724


src/BSON/UTCDateTime.c
```diff
@@ -61,8 +61,42 @@ static bool php_phongo_utcdatetime_init(php_phongo_utcdatetime_t *intern, int64_
...
+	milliseconds = bson_ascii_strtoll(s_milliseconds, &endptr, 10);
+#endif
+
+	/* errno will set errno if conversion fails; however, we do not need to
+	 * specify the type of error.
+	 *
+	 * Note: bson_ascii_strtoll() does not properly detect out-of-range values
+	 * (see: CDRIVER-1377). strtoll() would be preferable, but it is not
+	 * available on all platforms (e.g. HP-UX), and atoll() provides no error
+	 * reporting at all. */
```
jmikola: The original `STRTOLL()` macro we had uses three possible functions:

 * `_atoi64()` on Windows, which supports error reporting (may not detect extra data)
 * `atoll()` if `HAVE_ATOLL` is defined, which doesn't support error reporting at all
 * `strtoll()` otherwise, which supports error reporting and `endptr` for detecting extra data

Support for `atoll()` appears to be a concession for HP-UX's broken `strtoll()` implementation (see: [PHP bug #33433](https://bugs.php.net/bug.php?id=33433)). If we care about HP-UX portability, I suggest we use `bson_ascii_strtoll()`, which is a `strtoll()` polyfill in libbson. It does have an issue with detecting out-of-range integers (see: [CDRIVER-1377](https://jira.mongodb.org/browse/CDRIVER-1377)), but I think this is an improvement over the current use of `atoll()` and reporting no errors at all.

src/BSON/Binary.c
```diff
@@ -46,10 +46,12 @@ PHONGO_API zend_class_entry *php_phongo_binary_ce;
...
 zend_object_handlers php_phongo_handler_binary;
 
-/* Initialize the object from a string and return whether it was successful. */
-static bool php_phongo_binary_init(php_phongo_binary_t *intern, const char *data, phongo_zpp_char_len data_len, phongo_long type)
+/* Initialize the object and return whether it was successful. An exception will
+ * be thrown on error. */
+static bool php_phongo_binary_init(php_phongo_binary_t *intern, const char *data, phongo_zpp_char_len data_len, phongo_long type TSRMLS_DC)
 {
 	if (type < 0 || type > UINT8_MAX) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected type to be an unsigned 8-bit integer, %" PHONGO_LONG_FORMAT " given", type);
```
jmikola: This message was previously only used in the constructor; however, the exception may now be thrown from `__set_state()` or `unserialize()`. What do you think about putting the class name in the message?

I've started doing this with other new messages (e.g. Javascript now allowing null bytes in its code argument) for added clarity, but did not want to change existing messages before discussing.

src/BSON/Decimal128.c
```diff
@@ -46,34 +46,39 @@ PHONGO_API zend_class_entry *php_phongo_decimal128_ce;
...
+/* Initialize the object and return whether it was successful. An exception will
+ * be thrown on error. */
+static bool php_phongo_decimal128_init(php_phongo_decimal128_t *intern, const char *value TSRMLS_DC)
 {
-	if (bson_decimal128_from_string(value, &intern->decimal)) {
-		intern->initialized = true;
-
-		return true;
+	if (!bson_decimal128_from_string(value, &intern->decimal)) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error parsing decimal string: %s", value);
```
jmikola: This is also a message that might benefit from adding the class name.

On a side note, this differs from HHVM's exception message (see: [here](https://github.com/mongodb/mongo-hhvm-driver/blob/73bc3daa7691184a6e091585ad42c01cb965fe41/src/MongoDB/BSON/Decimal128.cpp#L39)). We must have missed the different when originally reviewing [PHPC-619](https://jira.mongodb.org/browse/PHPC-619)'s PR.

Let's take this opportunity to settle on a consistent format.

src/BSON/Javascript.c
```diff
@@ -46,10 +46,17 @@ PHONGO_API zend_class_entry *php_phongo_javascript_ce;
 
 zend_object_handlers php_phongo_handler_javascript;
 
-/* Initialize the object from a string and return whether it was successful. */
+/* Initialize the object and return whether it was successful. An exception will
+ * be thrown on error. */
 static bool php_phongo_javascript_init(php_phongo_javascript_t *intern, const char *code, phongo_zpp_char_len code_len, zval *scope TSRMLS_DC)
 {
 	if (scope && Z_TYPE_P(scope) != IS_OBJECT && Z_TYPE_P(scope) != IS_ARRAY && Z_TYPE_P(scope) != IS_NULL) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected %s scope to be array or object, %s given", ZSTR_VAL(php_phongo_javascript_ce->name), zend_get_type_by_const(Z_TYPE_P(scope)));
```
jmikola: Note: this exception can only be thrown from `__set_state()` or `unserialize()`, since `__construct()` validates the type with zpp.

src/BSON/Binary.c
jmikola: To clarify, my main concern was with `unserialize()`, as I expected that would be the only function in the stack trace. After further testing, it appears that the appropriate `__wakeup()` method is included in the backtrace as "[internal function]". I tested this by unserializing a Javascript object with an invalid Regex in its scope document, just to see how nesting is handled.

```
<?php

unserialize('O:23:"MongoDB\BSON\Javascript":2:{s:4:"code";s:33:"function foo(bar) { return bar; }";s:5:"scope";O:8:"stdClass":1:{s:1:"r";O:18:"MongoDB\BSON\Regex":2:{s:7:"pattern";s:6:"regexp";s:5:"flags";s:1:"'. "\0" .'";}}}');
```

Output:

```
Fatal error: Uncaught exception 'MongoDB\Driver\Exception\InvalidArgumentException' with message 'MongoDB\BSON\Regex flags should not contain null bytes' in tests/bson/bson-regex-serialization_error-002.php:15
Stack trace:
#0 [internal function]: MongoDB\BSON\Regex->__wakeup()
#1 tests/bson/bson-regex-serialization_error-002.php(15): unserialize('O:23:"MongoDB\\B...')
#2 {main}
  thrown in tests/bson/bson-regex-serialization_error-002.php on line 15
```

src/BSON/ObjectID.c
```diff
@@ -75,28 +77,33 @@ static bool php_phongo_objectid_init_from_hex_string(php_phongo_objectid_t *inte
 		return true;
 	}
 
+	phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Invalid BSON ID provided");
```
jmikola: This is the same error message used in HHVM (see: [here](https://github.com/mongodb/mongo-hhvm-driver/blob/73bc3daa7691184a6e091585ad42c01cb965fe41/src/MongoDB/BSON/ObjectID.cpp#L42)), but I think there's room for improvement. We should at least show the offending value.

Let's discuss and create a new ticket if needed.

phongo_compat.h
```diff
@@ -147,7 +136,7 @@
 # define phongo_long long
 # define PHONGO_LONG_FORMAT "ld"
 # define SIZEOF_PHONGO_LONG SIZEOF_LONG
-# define phongo_str(str) str
+# define ZSTR_VAL(str) str
```
jmikola: A number of the exception messages for PHPC-741 involved reporting `zend_class_entry->name`. On PHP 7, this required accessing `zend_string->val`. Digging a little deeper, I found that we were inconsistently using either the `phongo_str()` macro or using `#if` blocks to conditionally use `ZSTR_VAL()` for PHP 7. Simply backporting `ZSTR_VAL()` to PHP 5.x allowed me to remove our custom macro and delete those `#if` blocks.

derickr:
LGTM

PR #351, opened 2016-07-15
PHPC-730: Javascript::getCode() and Javascript::getScope()

jmikola:
https://jira.mongodb.org/browse/PHPC-730


tests/bson/bson-javascript-getScope-001.phpt
```diff
@@ -0,0 +1,38 @@
+--TEST--
+MongoDB\BSON\Javascript::getCode()
```
derickr: This is the getScope() test, and not getCode().

derickr:
The getScope tests fail with PHP 5.5 and PHP 5.6 too

PR #352, opened 2016-07-15
[WIP] PHPC-749: Update win32 libbson and libmongoc config files

jmikola:
https://jira.mongodb.org/browse/PHPC-749

Note: do not review.


jmikola:
See: https://github.com/mongodb/mongo-php-driver/pull/365

PR #357, opened 2016-07-25
PHPC-754 replace libmongoc-priv with libmongoc

puppyofkosh:
https://jira.mongodb.org/browse/PHPC-754

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -365,11 +365,11 @@ PHP_ARG_WITH(libmongoc, whether to use system libmongoc,
   if test "$PHP_LIBMONGOC" != "no"; then
     AC_PATH_PROG(PKG_CONFIG, pkg-config, no)
     AC_MSG_CHECKING(for libmongoc)
-    if test -x "$PKG_CONFIG" && $PKG_CONFIG --exists libmongoc-1.0 && $PKG_CONFIG --exists libmongoc-priv; then
+    if test -x "$PKG_CONFIG" && $PKG_CONFIG --exists libmongoc-1.0; then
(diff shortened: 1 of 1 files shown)


config.m4
```diff
@@ -365,11 +365,11 @@ PHP_ARG_WITH(libmongoc, whether to use system libmongoc,
   if test "$PHP_LIBMONGOC" != "no"; then
     AC_PATH_PROG(PKG_CONFIG, pkg-config, no)
     AC_MSG_CHECKING(for libmongoc)
-    if test -x "$PKG_CONFIG" && $PKG_CONFIG --exists libmongoc-1.0 && $PKG_CONFIG --exists libmongoc-priv; then
+    if test -x "$PKG_CONFIG" && $PKG_CONFIG --exists libmongoc-1.0; then
       if $PKG_CONFIG libmongoc-1.0 --atleast-version 1.3.3; then
-        LIBMONGOC_INC=`$PKG_CONFIG libmongoc-priv --cflags`
```
jmikola: Wouldn't `--atleast-version 1.3.3` need to be changed to `1.4.0`?

Are there similar changes required for the libbson package dependency? I expect they need to be kept in sync.

bjori: libbson never had `-priv`, so no, thats not needed.
Good spot on `--atleast-version 1.3.3` should be bumped to `1.4.0` if you decide on making that the minimum version.

We did not try to compile current php master against `1.3.3` as it currently declares as its minimum. I guess that is not correct today before or after this patch.
Before this patch, you cannot build against system mongoc 1.4, even though you do seem to require it.

jmikola:
Folding this into https://github.com/mongodb/mongo-php-driver/pull/362.

PR #359, opened 2016-07-25
PHPC-753 Sets driver metadata at startup

puppyofkosh:
https://jira.mongodb.org/browse/PHPC-753

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -2445,6 +2445,12 @@ PHP_MINIT_FUNCTION(mongodb)
 
 	/* Initialize libmongoc */
 	mongoc_init();
+
+#ifdef MONGOC_EXPERIMENTAL_FEATURES
+    /* Set metadata strings BEFORE changing allocation routines */
+    mongoc_metadata_append("phongo", MONGODB_VERSION_S, NULL);
+#endif
+
 	/* Initialize libbson */
 	bson_mem_set_vtable(&MONGODB_G(bsonMemVTable));
 
--- a/config.m4
+++ b/config.m4
@@ -255,6 +255,7 @@ if test "$MONGODB" != "no"; then
     mongoc-matcher-op.c \
     mongoc-matcher.c \
     mongoc-memcmp.c \
+dnl    mongoc-metadata.c \
     mongoc-opcode.c \
     mongoc-queue.c \
     mongoc-read-concern.c \
@@ -567,6 +568,9 @@ dnl }}}
   fi
 
   if test "$PHP_LIBMONGOC" == "no"; then
+    dnl Disable experimental features for now
+    AC_SUBST(MONGOC_EXPERIMENTAL_FEATURES, 0)
+
     backup_srcdir=${srcdir}
     srcdir=${srcdir}/src/libmongoc/
     m4_include(src/libmongoc/build/autotools/Versions.m4)


php_phongo.c
```diff
@@ -2445,6 +2445,12 @@ PHP_MINIT_FUNCTION(mongodb)
 
 	/* Initialize libmongoc */
 	mongoc_init();
+
+#ifdef MONGOC_EXPERIMENTAL_FEATURES
```
derickr: I think it would be greatly beneficial if there was not a single "MONGOC_EXPERIMENTAL_FEATURES" flag. I can imagine, this specific flag meaning different things with different versions of mongo-c-driver (in the future). As this is somewhat of a feature flag, there should be one for each specific thing that we can turn on and off. In this case, I would suggest to have something akin to MONGOC_HAVE_METADATA. 

It also seems a bit strange to have a patch that introduces a feature, but *also* disables that same feature through a config.m4 flag.

bjori: mongoc 1.4.0 and libbson 1.4.0 have experimental support for MongoDB 3.4 features that we are working on.
This experimental support is experimental until the first release of MongoDB 3.4 since they can be removed or changed at any time.
We have no intentions of actually supporting MongoDB 3.4 features in our 1.4.x series, but we are developing with them in mind and actively trying out MongoDB 3.4 things.
This is one of such feature. All of the MongoDB 3.4 experimental features and lumped together under this flag.

We decided to go this approach rather then branching as it simplified maintenance for us.


Note that this feature hasn't actually been committed to MongoDB master, yet :)

Its up to you if you want to leave this PR hanging here for a while.
The main point for @puppyofkosh was to ensure the mongoc implementation of this accounted for how the PHP & C++11 drivers.
See also https://github.com/mongodb/mongo-cxx-driver/pull/509

jmikola: How is this supposed to work with Decimal128? I expect all of the Decimal128 code would need to be wrapped in `MONGOC_EXPERIMENTAL_FEATURES` in that case.

@derickr and I spoke about this earlier in the week and I don't see any reason to release 1.2.0 until these features are properly available in libmongoc 1.5.0, which I expect you will need to have ready to go when MongoDB 3.2 ships.

bjori: Yes, Decimal128 is behind `BSON_EXPERIMENTAL_FEATURES`.

1.5 is expected to be released around October I imagine, week before MongoDB 3.4 as usual.

1.4 meanwhile, is a huge release, with incredible amount of work dedicated for the PHP drivers and TLS improvements. We had no reason to delay this release another 3months, as jam packed as it already is.

If you want to skip this release, then that is completely up to you. It might proof problematic for distros as there is no libmongoc-priv anymore, so you'd have to change the requirements to max 1.3. Other then then, I can't think of anything bad happening.

bjori: It seems a bit weird to talk about these things in a unrelated pull request.

I'd be more then happy to schedule a call to talk about this. Unfortunately, I cannot do 7am on Monday mornings like has been previously suggested, but normal working hours I can move things around to accommodate.

jmikola: > If you want to skip this release, then that is completely up to you. It might proof problematic for distros as there is no libmongoc-priv anymore, so you'd have to change the requirements to max 1.3. Other then then, I can't think of anything bad happening.

I imagine that would be a problem the moment libmongoc 1.4.0 lands in a distro, unless they had a new version of PHPC to publish at the same time.

/cc @oerdnj @remicollet

jmikola:
After discussing with @derickr, I don't believe we will plan to use `MONGOC_EXPERIMENTAL_FEATURES` at all. We're going to discuss on Monday whether to cut the 1.2.0 release earlier just to get the libmongoc-priv situation resolved. In that case, we would disable our existing Decimal128 implementations in PHPC and HHVM. I'm making every attempt to get the libmongoc stream migration and topology persistence improvements into 1.2.0.

I believe the original plan was for 1.2.0 to be the 3.4 compatibility release; however, this would require users to stick with 1.1.x for three more months (at least).

PR #361, opened 2016-07-29
PHPC-605: Rely on libmongoc for socket handling and TLS

jmikola:
https://jira.mongodb.org/browse/PHPC-605


php_phongo.c
```diff
@@ -1879,17 +1333,94 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t *uri, bson_t *option
...
+	if (php_array_existsc(sslOptions, "passphrase")) {
+		ssl_opt.pem_pwd = php_array_fetchc_string(sslOptions, "passphrase", &plen, &free_pem_pwd);
+	}
+
+	if (php_array_existsc(sslOptions, "cafile")) {
+		ssl_opt.ca_file = php_array_fetchc_string(sslOptions, "cafile", &plen, &free_ca_file);
+	}
+
+	if (php_array_existsc(sslOptions, "capath")) {
+		ssl_opt.ca_dir = php_array_fetchc_string(sslOptions, "capath", &plen, &free_ca_dir);
```
bjori: FYI: When using the Native Darwin or Windows TLS libraries, this will result in an MONGOC_EROR() when used.

That might only be problem if you start falling back on the `openssl.capath` php.ini settings, but though I'd call it out

jmikola: Am I correct in assuming that only different here is customizing the entire CA path? All three will allow a custom CA file, which I suppose most users can stick to if they have a special cert.

Would the default CA path for each generally be sufficient? Since we won't be using PHP's OpenSSL extension at all, I'm on the fence about whether we should even default to `openssl.capath`.

bjori: That is correct. Most people using their own authority would be providing that cert/bundle as `cafile` or include it in the OS default chain, and not provide anything.

The only reason we are supporting `capath` is because `why not?` -- which then turned into `err, we shouldn't have`.

php_phongo.c
```diff
@@ -1878,17 +1264,98 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t *uri, bson_t *option
...
+#endif
+
+		if (!zoptions) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Stream-Context resource does not contain \"ssl\" options array");
+			return false;
+		}
+	}
+
+	if (php_array_existsc(zoptions, "verify_hostname")) {
+		ssl_opt.allow_invalid_hostname = !php_array_fetchc_bool(zoptions, "verify_hostname");
```
jmikola: @derickr: HHVM doesn't appear to support this at all.

Also, PHP's own [SSL context options](http://php.net/manual/en/context.ssl.php) don't have anything called `verify_hostname`, so I propose we just name our option `allow_invalid_hostname`.

php_phongo.c
```diff
@@ -1878,17 +1264,98 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t *uri, bson_t *option
...
+	}
+
+	if (php_array_existsc(zoptions, "allow_self_signed")) {
```
jmikola: For BC, we should still support `allow_self_signed` as an alias of `weak_cert_validation`, but I propose we support the canonical name as well (i.e. allow `weak_cert_validation` to be used, and give that priority).

php_phongo.c
```diff
@@ -1878,17 +1264,98 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t *uri, bson_t *option
...
+	if (php_array_existsc(zoptions, "passphrase")) {
+		ssl_opt.pem_pwd = php_array_fetchc_string(zoptions, "passphrase", &plen, &free_pem_pwd);
+	}
+
+	if (php_array_existsc(zoptions, "cafile")) {
+		ssl_opt.ca_file = php_array_fetchc_string(zoptions, "cafile", &plen, &free_ca_file);
+	}
+
+	if (php_array_existsc(zoptions, "capath")) {
+		ssl_opt.ca_dir = php_array_fetchc_string(zoptions, "capath", &plen, &free_ca_dir);
```
jmikola: Likewise for these four options, I'd like to support and document the struct fields but still support the old names for BC. Allow the new names to override.

tests/connect/standalone-ssl-0004.phpt
```diff
@@ -43,8 +38,8 @@ printf("Inserted: %d\n", $retval->getInsertedCount());
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
-OK: Got MongoDB\Driver\Exception\SSLConnectionException thrown from executeBulkWrite
-%s
+OK: Got MongoDB\Driver\Exception\ConnectionTimeoutException thrown from executeBulkWrite
+No suitable servers found (`serverSelectionTryOnce` set): [%s calling ismaster on '%s:%d']
```
jmikola: SSLConnectionException will no longer be used, since libmongoc only exposes SSL errors through server selection failures (i.e. ConnectionTimeoutException). This is a regrettable BC break for folks that were catching SSLConnectionException, but it's outside of our control.

I propose we document that SSLConnectionException is no longer useful and only applied to PHPC < 1.2.0.

php_phongo.c
jmikola: I can't find any record of `verify_hostname` being used in the old code or PHP core. It's not a documented SSL context option. Perhaps we should just remove it entirely and stick with `allow_invalid_hostname`.

bjori: See: https://jira.mongodb.org/browse/PHPC-192?focusedCommentId=861601&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-861601

Looks like I was planning on ditching this approach, in favor of custom mortal people readble options instead.
Doesn't look like that materialized, but the accidental NOOP context option remained in the code

bjori:
this lgtm. so much removed code \o/

php_phongo.c
```diff
@@ -1878,17 +1264,98 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t *uri, bson_t *option
...
+	zoptions = zdriverOptions;
+
+	if (php_array_existsc(zdriverOptions, "context")) {
+		php_stream_context *context;
+		zval *zcontext;
+
+		zcontext = php_array_fetchc(zdriverOptions, "context");
+
+		/* TODO: Do we want to check global stream options, too? */
+		context = php_stream_context_from_zval(zcontext, 1);
```
jmikola: I've opted to ignore the global stream context. Leaving this code as-is, since the second argument means we return `NULL` instead of defaulting to the global context.

jmikola:
@derickr: Ready for a final review.

php_phongo.c
```diff
@@ -1878,17 +1264,130 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t *uri, bson_t *option
...
+		zoptions = php_array_fetchc_array(context->options, "ssl");
+#endif
+
+		if (!zoptions) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Stream-Context resource does not contain \"ssl\" options array");
+			return false;
+		}
+	}
+
+	/* Check canonical option names first and fall back to SSL context options
```
derickr: I will need to do the same - are you tests covering all of this?

tests/connect/standalone-ssl-verify_cert-error-001.phpt
```diff
@@ -0,0 +1,29 @@
+--TEST--
+Connect to MongoDB with SSL and cert verification error
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; NEEDS("STANDALONE_SSL"); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$SSL_DIR = realpath(__DIR__ . '/../../scripts/ssl/');
```
derickr: You don't seem to use $SSL_DIR in here.

tests/connect/standalone-ssl-verify_cert-error-002.phpt
```diff
@@ -0,0 +1,33 @@
+--TEST--
+Connect to MongoDB with SSL and cert verification error (context options)
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; NEEDS("STANDALONE_SSL"); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$SSL_DIR = realpath(__DIR__ . '/../../scripts/ssl/');
```
derickr: Or in this one.

tests/connect/standalone-x509-auth-001.phpt
```diff
@@ -0,0 +1,33 @@
...
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; NEEDS("STANDALONE_X509"); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$SSL_DIR = realpath(__DIR__ . '/../../scripts/ssl/');
+
+$driverOptions = [
+    // libmongoc does not allow the hostname to be overridden as "server"
```
derickr: I thought you were planning on recreating certificates so that we *could* test with the valid hostname checks.

tests/connect/standalone-x509-extract_username-001.phpt
```diff
@@ -0,0 +1,37 @@
...
+
+$SSL_DIR = realpath(__DIR__ . '/../../scripts/ssl/');
+
+$driverOptions = [
+    // libmongoc does not allow the hostname to be overridden as "server"
+    'allow_invalid_hostname' => true,
+    'weak_cert_validation' => false,
+    'ca_file' => $SSL_DIR . '/ca.pem',
+    'pem_file' => $SSL_DIR . '/client.pem',
+    // TODO: this doesn't appear to have any effect. Does the PEM file not have a password?
```
derickr: Still a TODO in here.

tests/connect/standalone-x509-extract_username-001.phpt
```diff
@@ -0,0 +1,37 @@
...
+    'pem_pwd' => 'qwerty',
+];
+
+$parsed = parse_url(STANDALONE_X509);
+// TODO: authMechanism cannot be parsed from URI options array (PHPC-772)
```
derickr: Be careful to remove this once PHPC-772 is fixed.

derickr:
Looks mostly good - I had a few comments though.

php_phongo.c
jmikola: I did not bother using both options (canonical and alias) in a single test. All of the tests are just split for using a `$driverOptions` array with canonical options and a stream context with the BC aliases.

tests/connect/standalone-x509-auth-001.phpt
jmikola: I still need to look into that. Are you OK with deferring it to a future ticket?

php_phongo.c
derickr: OK

tests/connect/standalone-x509-auth-001.phpt
derickr: Yes, sure.

tests/connect/standalone-ssl-verify_cert-error-001.phpt
jmikola: Will remove.

tests/connect/standalone-ssl-verify_cert-error-002.phpt
jmikola: Will remove.

tests/connect/standalone-x509-auth-001.phpt
jmikola: Created [PHPC-774](https://jira.mongodb.org/browse/PHPC-774).

tests/connect/standalone-x509-extract_username-001.phpt
jmikola: Looking through [`mongoc_ssl_extract_subject()`](https://github.com/mongodb/mongo-c-driver/blob/1.4.0/src/mongoc/mongoc-ssl.c#L57), only `_mongoc_secure_transport_extract_subject()` appears to use the password. OpenSSL and Secure Channel implementations seem to ignore the parameter.

If I attempt to read the `client.pem` file with `ssh-keygen -y -f client.pem` (as documented [here](http://serverfault.com/a/628924)), I can see its contents without a password prompt. Therefore, I assume it was never created with a password. I'm going to remove the `pem_pwd` option from both of these tests.

@bjori: do you concur? Furthermore, can you confirm whether or not OpenSSL and Secure Channel support passwords on the PEM file?

bjori: Secure Channel does not support password protected PEM files.
Secure Transport, OpenSSL, LibreSSL and libtls all do.

The password is not needed to extract the subject name from a certificate though.

None of the certificates in the phongo test suite use the password protected private keys, so you don't have to provide one for client authentication. Ignoring password for unencrypted private keys is implementation dependent anyway, I don't know if all of them do, so I wouldn't provide bogus password

tests/connect/standalone-x509-auth-001.phpt
```diff
@@ -0,0 +1,33 @@
...
+    'allow_invalid_hostname' => true,
+    'weak_cert_validation' => false,
+    'ca_file' => $SSL_DIR . '/ca.pem',
+    'pem_file' => $SSL_DIR . '/client.pem',
+    // TODO: this doesn't appear to have any effect. Does the PEM file not have a password?
+    'pem_pwd' => 'qwerty',
```
jmikola: Forgot to remove this before merging. Will be addressed in https://github.com/mongodb/mongo-php-driver/pull/376.

tests/connect/standalone-x509-auth-002.phpt
```diff
@@ -0,0 +1,37 @@
...
+$driverOptions = [
+    'context' => stream_context_create([
+        'ssl' => [
+            // libmongoc does not allow the hostname to be overridden as "server"
+            'allow_invalid_hostname' => true,
+            'allow_self_signed' => false, // "weak_cert_validation" alias
+            'cafile' => $SSL_DIR . '/ca.pem', // "ca_file" alias
+            'local_cert' => $SSL_DIR . '/client.pem', // "pem_file" alias
+            // TODO: this doesn't appear to have any effect. Does the PEM file not have a password?
+            'passphrase' => 'qwerty', // "pem_pwd" alias
```
jmikola: Forgot to remove this before merging. Will be addressed in https://github.com/mongodb/mongo-php-driver/pull/376.

php_phongo.c
```diff
@@ -1878,17 +1264,130 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t *uri, bson_t *option
...
+	}
+
+	valid_options = true;
+
+#ifdef MONGOC_ENABLE_SSL_OPENSSL
+	if (ssl_opt.ca_dir) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "\"ca_dir\" or \"capath\" options may only be used with OpenSSL");
+		valid_options = false;
+	}
+#endif
```
jmikola: This logic was actually incorrect. We should error if  `MONGOC_ENABLE_SSL_OPENSSL` was not defined.

Since libmongoc 1.5.0-dev now has LibreSSL support, a more correct approach would be to error based on whether Secure Channel or Secure Transport are being used. I'll address this in https://github.com/mongodb/mongo-php-driver/pull/376.


 Forgot to remove this before merging. Will be addressed in https://github.com/mongodb/mongo-php-driver/pull/376.

PR #365, opened 2016-08-01
PHPC-447: Generate libmongoc and libbson headers in config.w32

jmikola:
https://jira.mongodb.org/browse/PHPC-447

This removes the old config and version headers, which required manual updates and were often out-of-date as a result.

Additionally, this syncs the config header replacements with 1.4.0-beta1.


derickr:
LGTM

PR #367, opened 2016-08-02
Remove make target for generating release notes from Git logs

jmikola:
While releasing 1.2.0-alpha1, I noticed that our Makefile was only generating release notes from the previous 1.1.8 tag, which missed a number of issues. I'm opting to remove all of this behavior and document that pull release notes from JIRA, as we did with the legacy driver. This will also ensure that release notes report issues in categories (e.g. bug fixes, tasks, new features).


derickr:
LGTM

PR #368, opened 2016-08-04
BSON test cleanup

jmikola:
This removes unnecessary `SKIPIF` blocks from BSON tests that do not depend on a server. Additionally, this includes some small fixes to UTCDateTime tests that assumed a 64-bit environment.


jmikola:
@derickr: [tests/bson/bson-utcdatetime-int-size.php](https://github.com/mongodb/mongo-php-driver/blob/master/tests/bson/bson-utcdatetime-int-size.phpt) (see: [history](https://github.com/mongodb/mongo-php-driver/commits/master/tests/bson/bson-utcdatetime-int-size.phpt)) looks like it has a problem with 32-bit platforms, since it uses a 64-bit integer literal. I personally think the test is redundant in light of other files, but the least I can do is split it up into two files to separate the bit that requires a 64-bit platform.

derickr:
LGTM

PR #369, opened 2016-08-04
PHPC-761: Remove BSON_NAMESPACE constant

jmikola:
https://jira.mongodb.org/browse/PHPC-761

Note: this is not a BC break, since we never documented the constant and it was never defined by HHVM.

Depends on https://github.com/mongodb/mongo-php-driver/pull/368.


derickr:
LGTM

PR #371, opened 2016-08-11
PHPC-629: Bump libbson and libmongoc to 1.4.0

jmikola:
https://jira.mongodb.org/browse/PHPC-629


derickr:
LGTM - I'll do the same on Monday for the v1.2 branch (and see what breaks).

PR #376, opened 2016-08-29
PHPC-433: Persist topology state between requests

jmikola:
https://jira.mongodb.org/browse/PHPC-433
https://jira.mongodb.org/browse/PHPC-605
https://jira.mongodb.org/browse/PHPC-704

Depends on #361.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1281,123 +1281,104 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t *uri, bson_t *option
 	return true;
 } /* }}} */
 
-static bool php_phongo_apply_ssl_opts(mongoc_client_t *client, zval *zdriverOptions TSRMLS_DC)
+static inline char *php_phongo_fetch_ssl_opt_string(zval *zoptions, const char *key, int key_len)
 {
-	mongoc_ssl_opt_t ssl_opt = {0};
-	zval *zoptions;
-	zend_bool free_pem_file = 0, free_pem_pwd = 0, free_ca_file = 0, free_ca_dir = 0, free_crl_file = 0;
-	int plen;
-	bool valid_options;
+	int        plen;
+	zend_bool  pfree;
+	char      *pval, *value;
 
-	if (!zdriverOptions) {
-		return false;
+	pval = php_array_fetchl_string(zoptions, key, key_len, &plen, &pfree);
+	value = pfree ? pval : estrndup(pval, plen);
+
+	return value;
+}
+
+static mongoc_ssl_opt_t *php_phongo_make_ssl_opt(zval *zoptions TSRMLS_DC)
+{
+	mongoc_ssl_opt_t *ssl_opt;
+
+	if (!zoptions) {
+		return NULL;
 	}
 
-	zoptions = zdriverOptions;
-
-	/* Use SSL context options if provided for backwards compatibility */
-	if (php_array_existsc(zdriverOptions, "context")) {
-		php_stream_context *context;
-		zval *zcontext;
-
-		zcontext = php_array_fetchc(zdriverOptions, "context");
-
-		context = php_stream_context_from_zval(zcontext, 1);
-
-		if (!context) {
-			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "\"context\" driver option is not a valid Stream-Context resource");
-			return false;
-		}
-
-#if PHP_VERSION_ID >= 70000
-		zoptions = php_array_fetchc_array(&context->options, "ssl");
-#else
-		zoptions = php_array_fetchc_array(context->options, "ssl");
+#if defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL) || defined(MONGOC_ENABLE_SSL_SECURE_TRANSPORT)
+	if (php_array_existsc(zoptions, "ca_dir") || php_array_existsc(zoptions, "capath")) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "\"ca_dir\" and \"capath\" options are not supported by Secure Channel and Secure Transport");
+		return NULL;
+	}
 #endif
 
-		if (!zoptions) {
-			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Stream-Context resource does not contain \"ssl\" options array");
-			return false;
-		}
-	}
+	ssl_opt = ecalloc(1, sizeof(mongoc_ssl_opt_t));
 
 	/* Check canonical option names first and fall back to SSL context options
 	 * for backwards compatibility. */
 	if (php_array_existsc(zoptions, "allow_invalid_hostname")) {
-		ssl_opt.allow_invalid_hostname = php_array_fetchc_bool(zoptions, "allow_invalid_hostname");
+		ssl_opt->allow_invalid_hostname = php_array_fetchc_bool(zoptions, "allow_invalid_hostname");
 	}
 
 	if (php_array_existsc(zoptions, "weak_cert_validation")) {
-		ssl_opt.weak_cert_validation = php_array_fetchc_bool(zoptions, "weak_cert_validation");
+		ssl_opt->weak_cert_validation = php_array_fetchc_bool(zoptions, "weak_cert_validation");
 	} else if (php_array_existsc(zoptions, "allow_self_signed")) {
-		ssl_opt.weak_cert_validation = php_array_fetchc_bool(zoptions, "allow_self_signed");
+		ssl_opt->weak_cert_validation = php_array_fetchc_bool(zoptions, "allow_self_signed");
 	}
 
 	if (php_array_existsc(zoptions, "pem_file")) {
-		ssl_opt.pem_file = php_array_fetchc_string(zoptions, "pem_file", &plen, &free_pem_file);
+		ssl_opt->pem_file = php_phongo_fetch_ssl_opt_string(zoptions, ZEND_STRL("pem_file"));
 	} else if (php_array_existsc(zoptions, "local_cert")) {
-		ssl_opt.pem_file = php_array_fetchc_string(zoptions, "local_cert", &plen, &free_pem_file);
+		ssl_opt->pem_file = php_phongo_fetch_ssl_opt_string(zoptions, ZEND_STRL("local_cert"));
 	}
 
 	if (php_array_existsc(zoptions, "pem_pwd")) {
-		ssl_opt.pem_pwd = php_array_fetchc_string(zoptions, "pem_pwd", &plen, &free_pem_pwd);
+		ssl_opt->pem_pwd = php_phongo_fetch_ssl_opt_string(zoptions, ZEND_STRL("pem_pwd"));
 	} else if (php_array_existsc(zoptions, "passphrase")) {
-		ssl_opt.pem_pwd = php_array_fetchc_string(zoptions, "passphrase", &plen, &free_pem_pwd);
+		ssl_opt->pem_pwd = php_phongo_fetch_ssl_opt_string(zoptions, ZEND_STRL("passphrase"));
 	}
 
 	if (php_array_existsc(zoptions, "ca_file")) {
-		ssl_opt.ca_file = php_array_fetchc_string(zoptions, "ca_file", &plen, &free_ca_file);
+		ssl_opt->ca_file = php_phongo_fetch_ssl_opt_string(zoptions, ZEND_STRL("ca_file"));
 	} else if (php_array_existsc(zoptions, "cafile")) {
-		ssl_opt.ca_file = php_array_fetchc_string(zoptions, "cafile", &plen, &free_ca_file);
+		ssl_opt->ca_file = php_phongo_fetch_ssl_opt_string(zoptions, ZEND_STRL("cafile"));
 	}
 
 	if (php_array_existsc(zoptions, "ca_dir")) {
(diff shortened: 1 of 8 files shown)


php_phongo.c
```diff
@@ -507,7 +518,10 @@ zend_bool phongo_writeerror_init(zval *return_value, bson_t *bson TSRMLS_DC) /*
 		writeerror->code = bson_iter_int32(&iter);
 	}
 	if (bson_iter_init_find(&iter, bson, "errmsg") && BSON_ITER_HOLDS_UTF8(&iter)) {
-		writeerror->message = bson_iter_dup_utf8(&iter, NULL);
+		uint32_t errmsg_len;
+		const char *err_msg = bson_iter_utf8(&iter, &errmsg_len);
+
+		writeerror->message = estrndup(err_msg, errmsg_len);
```
jmikola: The `writeconcernerror->message` and `writeerror->message` assignments were longstanding bugs, which weren't apparent until I tested with libbson's default memory allocation. These strings are both freed with `efree()` in the respective class dtors.

php_phongo.c
```diff
@@ -1742,22 +1856,22 @@ zend_object_iterator *php_phongo_cursor_get_iterator(zend_class_entry *ce, zval
 /* {{{ Memory allocation wrappers */
 static void* php_phongo_malloc(size_t num_bytes) /* {{{ */
 {
-	return emalloc(num_bytes);
+	return pemalloc(num_bytes, 1);
```
jmikola: Persistent memory allocation is required since clients are persisted between requests. This means that all BSON allocation will also use `malloc()` instead of `emalloc()`, even though much of that is only relevant inside a single request. Provided libmongoc and libbson don't have their own memory leaks, this shouldn't be an issue, though.

php_phongo.h
```diff
@@ -44,6 +44,7 @@ ZEND_BEGIN_MODULE_GLOBALS(mongodb)
 	char *debug;
 	FILE *debug_fd;
 	bson_mem_vtable_t bsonMemVTable;
+	HashTable clients;
```
jmikola: The [Internals Book](http://www.phpinternalsbook.com/hashtables/hashtable_api.html) talks about using `ALLOC_HASHTABLE()`, but that is only useful for non-persistent hash tables. Allocating the value inside of our global struct seemed kosher.

php_phongo_structs.h
```diff
@@ -81,7 +81,6 @@ typedef struct {
 typedef struct {
 	PHONGO_ZEND_OBJECT_PRE
 	mongoc_client_t    *client;
-	char               *pem_file;
```
jmikola: I forgot to remove this field in the PR for PHPC-605 when we dropped PHP streams.

tests/manager/manager-debug-001.phpt
```diff
@@ -10,16 +10,18 @@ require_once __DIR__ . "/../utils/basic.inc";
 $name = tempnam(sys_get_temp_dir(), "PHONGO");
 unlink($name);
 mkdir($name);
-$manager = new MongoDB\Driver\Manager(STANDALONE, array(), array("debug" => $name));
```
jmikola: This was changed for [PHPC-704](https://jira.mongodb.org/browse/PHPC-704), since we no longer want to support INI modifications via a constructor argument. As a result, I need to manually enable/disable the INI setting below.

tests/manager/manager-debug-001.phpt
```diff
@@ -10,16 +10,18 @@ require_once __DIR__ . "/../utils/basic.inc";
...
 $bulk->insert(array('_id' => 1, 'x' => 1));
 $result = $manager->executeBulkWrite(NS, $bulk);
+ini_set('mongodb.debug', 'off');
 foreach(glob($name."/*") as $file);
 $content = file($file);
 unlink($file);
 rmdir($name);
 
-echo $content[0];
+echo $content[0], $content[1];
```
jmikola: Previously, the `debug` option was applied after URI parsing, which mean that the first debugging output would have been the "Creating Manager" line. Since we need to alter the INI setting before invoking the constructor, the first line is now "Connection string". I've elected to output both for the test.

tests/manager/manager-debug-002.phpt
```diff
@@ -14,7 +14,6 @@ $bulk = new MongoDB\Driver\BulkWrite();
 $bulk->insert(array('_id' => 1, 'x' => 1));
 $result = $manager->executeBulkWrite(NS, $bulk);
 
-ini_set("mongodb.debug", "off");
```
jmikola: Best I can tell, the request reverts to the original INI setting (`mongodb.debug=stderr` from the ``--INI--`` block for this test). Disabling debugging inside the request here does not prevent further output from cleaning up the persistent clients during MSHUTDOWN.

jmikola:
I'll note that I manually tested this PR with PHP's built-in web server (`php -S localhost:8080`) and watched as multiple requests utilized the same mongod connection. The `index.php` file I used was quite basic:

```php
$manager = new MongoDB\Driver\Manager;
$cursor = $manager->executeCommand('test', new MongoDB\Driver\Command(['ping' => 1]));
var_dump($cursor->toArray()[0]);
```

This at least demonstrates that connections are persisted beyond RSHUTDOWN. When killing the web server, I can observe the connection being closed in the mongod server logs.

php_phongo.c
bjori: I don't understand the bug... nor the fix.


http://mongoc.org/libbson/current/bson_iter_dup_utf8.html

Yes, this string is supposed to be free()d with bson_free (), but that is the same as efree() because we set bson allocators to the same functions.

Wouldn't it be easier to keep this as is, and use bson_free () correctly, rather then bulky code to copy it?

php_phongo.c
```diff
@@ -1281,148 +1295,186 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t *uri, bson_t *option
...
+		return NULL;
+	}
 
-#if PHP_VERSION_ID >= 70000
-		zoptions = php_array_fetchc_array(&context->options, "ssl");
-#else
-		zoptions = php_array_fetchc_array(context->options, "ssl");
+#if defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL) || defined(MONGOC_ENABLE_SSL_SECURE_TRANSPORT)
+	if (php_array_existsc(zoptions, "ca_dir") || php_array_existsc(zoptions, "capath")) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "\"ca_dir\" and \"capath\" options are not supported by Secure Channel and Secure Transport");
```
bjori: Is it intentional that the errormessage is so broad, rather then mentioning the exact option and the exact implementation?

bjori: Note that crl_file is also not supported by secure transport, nor libressl's libtls

php_phongo.c
```diff
@@ -1450,42 +1502,104 @@ static mongoc_client_t *php_phongo_make_mongo_client(const mongoc_uri_t *uri, zv
 		return NULL;
 	}
 
-	if (mongoc_uri_get_ssl(uri) && driverOptions) {
-		if (!php_phongo_apply_ssl_opts(client, driverOptions TSRMLS_CC)) {
-			mongoc_client_destroy(client);
-			return NULL;
-		}
+	if (mongoc_uri_get_ssl(uri) && ssl_opt) {
+		mongoc_client_set_ssl_opts(client, ssl_opt);
```
bjori: You might as well call php_phongo_free_ssl_opt() right away here, the entire struct is deep copied to the client here.

php_phongo.c
bjori: Is there a benefit in still registering these functions?

derickr: I'm not sure either. I don't think we ought to be setting the libbson memory functions anymore.

php_phongo.c
```diff
@@ -1450,42 +1502,104 @@ static mongoc_client_t *php_phongo_make_mongo_client(const mongoc_uri_t *uri, zv
...
 		phongo_throw_exception(PHONGO_ERROR_RUNTIME TSRMLS_CC, "Failed to create Manager from URI: '%s'", uri_string);
-		return false;
+		goto cleanup;
 	}
 
-	return true;
+	MONGOC_DEBUG("Created client hash: %s\n", hash);
+#if PHP_VERSION_ID >= 70000
+	ZVAL_PTR(&new_client_ptr, manager->client);
+	zend_hash_str_update(&MONGODB_G(clients), hash, hash_len, &new_client_ptr);
```
derickr: Travis says that hash_len might be able to use uninitialized: https://travis-ci.org/mongodb/mongo-php-driver/jobs/158304272#L1220

php_phongo.c
jmikola: The bug was using inconsistent `strdup()` and `free()` functions, and the alternative fix was to use `bson_free()` in the class dtors. For consistency with other class structs, where the strings are always allocated by PHP, I chose to modify the allocation to match and leave the `efree()` in place.

Of course, the issue only presented itself when I disabled the custom BSON memory allocators, but I was uncomfortable with having this bit of unrelated code depend on that implementation detail.

php_phongo.c
jmikola: AFAIK, the only benefit would be an error message from `_zend_mm_alloc_int()` when using Zend's memory manager and an OOM error is hit. Looking through the PHP 5.x code, it looks like PHP keeps heap memory in reserve to print such errors. I'd say that's preferable to the [alternatives](http://stackoverflow.com/a/6132375/162228).

php_phongo.c
jmikola: > Is it intentional that the errormessage is so broad, rather then mentioning the exact option and the exact implementation?

I didn't want to handle every permutation and figured this would get the point across just the same. We planned to document the context option names as deprecated aliases anyway.

> Note that crl_file is also not supported by secure transport, nor libressl's libtls

Thanks for noting LibreSSL's lack of support.

Is there a reason that mongoc-stream-libressl.c has a [commented-out function call](https://github.com/mongodb/mongo-c-driver/blob/13ce2f7f269bccb1e5bff69ab24333f343c5026f/src/mongoc/mongoc-stream-tls-libressl.c#L479) instead of `MONGOC_ERROR()` calls as found in [Secure Transport's implementation](https://github.com/mongodb/mongo-c-driver/blob/13ce2f7f269bccb1e5bff69ab24333f343c5026f/src/mongoc/mongoc-stream-tls-secure-transport.c#L475)?

In your opinion, should we simply skip this checking and allow libmongoc to log its error at runtime. This would result in an exception being thrown from whatever method was responsible for initializing the stream (e.g. `executeCommand()` for a ping in most of our SSL tests).

I'll note that our libmongoc log handler, [`php_phongo_log()`](https://github.com/mongodb/mongo-php-driver/blob/0fa224607270721dc9fa3e25ec165aed8b8f3fef/php_phongo.c#L202), is only used when `mongodb.debug` is enabled. That function has logic to throw exceptions for ERROR and CRITICAL level log messages and log everything else to the debug file descriptor. Perhaps we should have another log handler to use when `mongodb.debug` is disabled so we can still provide exceptions for ERROR and CRITICAL messages while ignoring lesser logs. Thoughts?

/cc @derickr

php_phongo.c
jmikola: I wanted to keep the freeing of `ssl_opt` local to the same function where it was allocated (i.e. `phongo_manager_init()`) for readability.

php_phongo.c
jmikola: > Perhaps we should have another log handler to use when mongodb.debug is disabled so we can still provide exceptions for ERROR and CRITICAL messages while ignoring lesser logs.

@derickr made a good point that this would simply cause existing code to start throwing new or different exceptions out of the blue. Grepping the libmongoc code, SSL option parsing appears to be one of the few places where an ERROR is logged (to report an option being ignored) and the code continues doing its thing without immediately returning. I suppose we'll stick with checking for these during PHPC option parsing.

php_phongo.c
jmikola: Should be fixed by initializing `hash_len` to zero in `phongo_manager_init()`.

php_phongo.c
jmikola: @bjori: I've split up the exception messages for `capath` and `ca_dir` in a later commit and also added a separate exception for the unsupported `crl_file` option.

bjori: > 
> 
> Perhaps we should have another log handler to use when mongodb.debug is disabled so we can still provide exceptions for ERROR and CRITICAL messages while ignoring lesser logs. Thoughts?


I think there might be a ticket for that already.

MONGOC_WARNING and MONGOC_ERROR should always result in exceptions.

> @derickr made a good point that this would simply cause existing code to start throwing new or different exceptions out of the blue. 

You do want to do that. You can't silently ignore ssl configuration like this.
If the application is relying on providing its custom `capath` (highly unlikely) then it should error as it doesn't work, and they'll have a hellofaconfusing time debuging why not, unless you throw exception explaining it.

jmikola: > You do want to do that. You can't silently ignore ssl configuration like this.

Isn't that what libmongoc is doing by merely logging an error and allowing SSL initialization to continue? We generally do check for errors after calling libmongoc functions and throw accordingly, so also throwing on ERROR and CRITICAL logs will likely just result in two exceptions being thrown in most cases (with the first getting stuffed under the most recently thrown).

Anywho, we are not catching the unsupported options at the time of URI and option processing, so I think we're OK here.

bjori: > Isn't that what libmongoc is doing by merely logging an error and allowing SSL initialization to continue?

Looks like that was indeed what we did in couple of places. I've filed a ticket and CR for it:
https://jira.mongodb.org/browse/CDRIVER-1532

derickr:
LGTM

Issue #375, opened 2016-08-24
Got unsupported type 0 'unknown'


sandipmavani:
Got Error While Pass From Recursive Function(Pass By Reference).

here $object Pass From Function 
$bulk->insert($object);
$result = $mng->executeBulkWrite("test.coll", $bulk,$writeConcern);

Got unsupported type 0 'unknown'

derickr:
Can you share a short, but *complete* script that we can run ourselves that exhibits this behaviour? And also share the full exception message (and class) that you get?

sandipmavani:
Hi @derickr 
when I invoke following function on data before inserting to database i got this error.
but without passing from this function it's work fine and successfully inserted.

function recursiveUnsetExplicitType(&$array)
{
	if(is_object($array))
		unset($array->_explicitType);
	if(is_array($array))
		unset($array['_explicitType']);

	if(is_array($array) || is_object($array))
		foreach($array as $key => &$value)
		{
			if(is_array($value) || is_object($value))
			{
				recursiveUnsetExplicitType($value);
			}
		}
}

derickr:
Sorry, but I need a *complete* script. One that I can run, to try and reproduce this.

However, I do think this might be related to opcache - can you try disabling this too?

sandipmavani:
hi @derickr and @jmikola 
There are one case remain for

PR #377, opened 2016-08-30
Update bson.c

sandipmavani:
| Q                   | A
| ------------------- | ---
| Target branch?      | "master" for new features / vX.Y for fixes
| Fixed tickets       | GitHub or JIRA issue(s) fixed by this PR
| Doc changes needed? | yes/no

Bug Fix
Got unsupported type 0 'unknown' #375

### DIFF
--- a/src/bson.c
+++ b/src/bson.c
@@ -1101,6 +1101,9 @@ try_again:
 		case IS_REFERENCE:
 			ZVAL_DEREF(entry);
 			goto try_again;
+		case IS_UNDEF:
+        		  bson_append_undefined(bson, key, key_len);
+             		break;	
 #endif
 
 		default:


src/bson.c
```diff
@@ -1101,6 +1101,9 @@ static void phongo_bson_append(bson_t *bson, php_phongo_bson_flags_t flags, cons
 		case IS_REFERENCE:
 			ZVAL_DEREF(entry);
 			goto try_again;
+		case IS_UNDEF:
+        		  bson_append_undefined(bson, key, key_len);
```
jmikola: This appends deprecated type 0x06, which correlates with an `undefined` value in JavaScript.

The `IS_UNDEF` constant here indicates an uninitialized or unset zval. To quote [this article](https://nikic.github.io/2015/05/05/Internal-value-representation-in-PHP-7-part-1.html) for PHP 7:

> The `IS_UNDEF` type is used in places where previously a `NULL` zval pointer (not to be confused with an `IS_NULL` zval) was used. For example, in the refcounting examples above the `IS_UNDEF` type is set for variables that have been unset.

@derickr: Based on https://github.com/mongodb/mongo-php-driver/issues/375, I think we should consider adding the field name to the exception message.

@sandipmavani: Can you attempt improving the exception message a few lines below for the `default:` case?

```
phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Got unsupported type %d '%s for field %s'", Z_TYPE_P(entry), zend_get_type_by_const(Z_TYPE_P(entry), key));
```

With that in place, you should have more knowledge of what field in the PHP array/object is uninitialized. I think this is likely a case of memory corruption. Given the field name, you may be able to conclude that it should always have a value in your application, which would further confirm the corruption hypothesis.

PR #378, opened 2016-08-31
PHPC-757: Implement The MongoDB Handshake Protocol

derickr:
https://jira.mongodb.org/browse/PHPC-757


php_phongo.c
```diff
@@ -1390,8 +1390,6 @@ static mongoc_client_t *php_phongo_make_mongo_client(const mongoc_uri_t *uri, zv
 	const char                *mongoc_version, *bson_version;
 	mongoc_client_t           *client;
 
-	ENTRY;
```
jmikola: I assume `php_phongo_make_mongo_client()` was the only use of libmongoc tracing in PHPC. Is that correct?

php_phongo.c
```diff
@@ -1880,11 +1878,17 @@ PHP_MINIT_FUNCTION(mongodb)
 {
 	(void)type; /* We don't care if we are loaded via dl() or extension= */
 
+	char *php_version_string = malloc(4 + sizeof(MONGODB_VERSION_S) + 1);
```
jmikola: Why `malloc()` instead of `emalloc()`?

php_phongo.c
```diff
@@ -1880,11 +1878,17 @@ PHP_MINIT_FUNCTION(mongodb)
...
 
 	REGISTER_INI_ENTRIES();
 
 	/* Initialize libmongoc */
 	mongoc_init();
+
+	/* Set handshake options */
+	snprintf(php_version_string, 4 + sizeof(MONGODB_VERSION_S) + 1, "PHP %s", PHP_VERSION);
```
jmikola: `php_version_string`'s size is the length of PHPC's version constant. Should it have been calculated based on `PHP_VERSION`'s size instead?

tests/manager/manager-debug-003.phpt
```diff
@@ -1,26 +0,0 @@
...
-$result = $manager->executeBulkWrite(NS, $bulk);
-
-ini_set("mongodb.debug", "off");
-?>
-===DONE===
-<?php exit(0); ?>
---EXPECTF--
-%a
-[%s]     PHONGO: TRACE   > ENTRY: php_phongo_make_mongo_client():%d
-[%s]     PHONGO: DEBUG   > Creating Manager, phongo-1.%d.%d%S[%s] - mongoc-1.%s(%s), libbson-1.%s(%s), php-%s
```
jmikola: I see you didn't remove the `MONGOC_DEBUG()` macro call in `php_phongo_make_mongo_client()`. Should we keep this test and simply remove the TRACE line above?

tests/manager/manager-var-dump-001.phpt
```diff
@@ -68,6 +68,7 @@ object(MongoDB\Driver\Manager)#%d (%d) {
         int(%d)
         ["minWireVersion"]=>
         int(0)
+        %a
```
jmikola: Just saw https://github.com/mongodb/mongo-hhvm-driver/pull/127/files#r76989497. Makes sense.

tests/bson/bug0623.phpt
```diff
@@ -2,8 +2,6 @@
 PHPC-623: Numeric keys limited to unsigned 32-bit integer
 --SKIPIF--
 <?php if (8 !== PHP_INT_SIZE) { die('skip Only for 64-bit platform'); } ?>
---INI--
-mongodb.debug=stderr
```
jmikola: Not sure how this ended up here and in `bug0544.phpt`, but thanks for removing.

php_phongo.c
derickr: ```mongoc_handshake_data_append()``` frees it. This is why the handshake is added *before* the libbson memory vtable is set-up too.

php_phongo.c
derickr: Good catch!

tests/manager/manager-debug-003.phpt
derickr: I didn't see the MONGOC_DEBUG() call... I don't really care about this test.

php_phongo.c
derickr: Yes. As this was the only file that failed to compile now @bjori made it not leak.

tests/manager/manager-debug-003.phpt
jmikola: Just confirmed that `MONGOC_DEBUG()` is from `mongoc-log.c` and unrelated to the tracing stuff. I'm OK with killing this test.

php_phongo.c
jmikola: Looking at [this code](https://github.com/mongodb/mongo-c-driver/blob/bc116fe08857837980a64b36bce8beac7d9c4874/src/mongoc/mongoc-handshake.c#L473), the arguments are `const char *`. I don't see any indication that they're freed.

derickr: _append_and_truncate may free then.

With emalloc, I got a segfault *or* after moving the call to after the libson vtable init I got a segfault. It *does* free it in some way.

https://github.com/mongodb/mongo-c-driver/blob/bc116fe08857837980a64b36bce8beac7d9c4874/src/mongoc/mongoc-handshake.c#L459

jmikola: `old_str` in the code that you linked is the libmongoc global. The string you're creating here goes into that function as a `const char *` as well under the `suffix` param (see: [code](https://github.com/mongodb/mongo-c-driver/blob/8fd25223acba9e63fe81c1d4712890cad96163b8/src/mongoc/mongoc-handshake.c#L430)).

If I had to guess, `emalloc()` segfaults because you're creating this outside of a request. I recall reading something about persistent allocations using `malloc()` instead of `emalloc()` in the PHP Internals Book ([here](http://www.phpinternalsbook.com/zvals/memory_management.html#allocating-and-initializing-zvals)). In this case, I think you should still `free()` after you append the value.

PR #385, opened 2016-09-05
PHPC-552: Implement ObjectID::getTimestamp() method

derickr:


src/BSON/ObjectID.c
```diff
@@ -134,6 +134,24 @@ PHP_METHOD(ObjectID, __construct)
 }
 /* }}} */
 
+/* {{{ proto string ObjectID::getTimestamp()
```
jmikola: This method returns an integer.

jmikola:
LGTM with the proto revision.

PR #386, opened 2016-09-05
PHPC-726: Allow cross-platform serialization of Timestamp and UTCDateTime

derickr:
I've opted to make Timestamp either accept strings for both increment and timestamp, or an integer for both. You can not mix a string increment and an integer timestamp, or v.v.

### DIFF
--- a/src/BSON/Timestamp.c
+++ b/src/BSON/Timestamp.c
@@ -67,6 +67,48 @@ static bool php_phongo_timestamp_init(php_phongo_timestamp_t *intern, phongo_lon
 	return true;
 }
 
+/* Initialize the object from numeric strings and return whether it was
+ * successful. An exception will be thrown on error. */
+static bool php_phongo_timestamp_init_from_string(php_phongo_timestamp_t *intern, const char *s_increment, phongo_zpp_char_len s_increment_len, const char *s_timestamp, phongo_zpp_char_len s_timestamp_len TSRMLS_DC)
+{
+	int64_t increment, timestamp;
+	char *endptr = NULL;
+
+	errno = 0;
+
+	/* errno will set errno if conversion fails; however, we do not need to
+	 * specify the type of error.
+	 *
+	 * Note: bson_ascii_strtoll() does not properly detect out-of-range values
+	 * (see: CDRIVER-1377). strtoll() would be preferable, but it is not
+	 * available on all platforms (e.g. HP-UX), and atoll() provides no error
+	 * reporting at all. */
+
+#if defined(PHP_WIN32)
+	increment = _atoi64(s_increment);
+#else
+	increment = bson_ascii_strtoll(s_increment, &endptr, 10);
+#endif
+
+	if (errno || (endptr && endptr != ((const char *)s_increment + s_increment_len))) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error parsing \"%s\" as 64-bit integer increment for %s initialization", s_increment, ZSTR_VAL(php_phongo_timestamp_ce->name));
+		return false;
+	}
+
+#if defined(PHP_WIN32)
+	timestamp = _atoi64(s_timestamp);
+#else
+	timestamp = bson_ascii_strtoll(s_timestamp, &endptr, 10);
+#endif
+
+	if (errno || (endptr && endptr != ((const char *)s_timestamp + s_timestamp_len))) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error parsing \"%s\" as 64-bit integer timestamp for %s initialization", s_timestamp, ZSTR_VAL(php_phongo_timestamp_ce->name));
+		return false;
+	}
+
+	return php_phongo_timestamp_init(intern, increment, timestamp TSRMLS_CC);
+}
+
 /* Initialize the object from a HashTable and return whether it was successful.
  * An exception will be thrown on error. */
 static bool php_phongo_timestamp_init_from_hash(php_phongo_timestamp_t *intern, HashTable *props TSRMLS_DC)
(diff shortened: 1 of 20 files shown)


src/BSON/Timestamp.c
```diff
@@ -67,6 +67,48 @@ static bool php_phongo_timestamp_init(php_phongo_timestamp_t *intern, phongo_lon
...
+/* Initialize the object from numeric strings and return whether it was
+ * successful. An exception will be thrown on error. */
+static bool php_phongo_timestamp_init_from_string(php_phongo_timestamp_t *intern, const char *s_increment, phongo_zpp_char_len s_increment_len, const char *s_timestamp, phongo_zpp_char_len s_timestamp_len TSRMLS_DC)
+{
+	int64_t increment, timestamp;
+	char *endptr = NULL;
+
+	errno = 0;
+
+	/* errno will set errno if conversion fails; however, we do not need to
```
jmikola: UTCDateTime.c had this comment block above the `if (errno || ...` lines below. Any reason you decided to move it up?

src/BSON/Timestamp.c
```diff
@@ -67,6 +67,48 @@ static bool php_phongo_timestamp_init(php_phongo_timestamp_t *intern, phongo_lon
...
+	 * reporting at all. */
+
+#if defined(PHP_WIN32)
+	increment = _atoi64(s_increment);
+#else
+	increment = bson_ascii_strtoll(s_increment, &endptr, 10);
+#endif
+
+	if (errno || (endptr && endptr != ((const char *)s_increment + s_increment_len))) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error parsing \"%s\" as 64-bit integer for %s initialization", s_increment, ZSTR_VAL(php_phongo_utcdatetime_ce->name));
```
jmikola: `ZSTR_VAL(php_phongo_utcdatetime_ce->name)` should be changed.

src/BSON/Timestamp.c
```diff
@@ -67,6 +67,48 @@ static bool php_phongo_timestamp_init(php_phongo_timestamp_t *intern, phongo_lon
...
+	}
+
+#if defined(PHP_WIN32)
+	timestamp = _atoi64(s_timestamp);
+#else
+	timestamp = bson_ascii_strtoll(s_timestamp, &endptr, 10);
+#endif
+
+	if (errno || (endptr && endptr != ((const char *)s_timestamp + s_timestamp_len))) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error parsing \"%s\" as 64-bit integer for %s initialization", s_timestamp, ZSTR_VAL(php_phongo_utcdatetime_ce->name));
```
jmikola: This also uses `ZSTR_VAL(php_phongo_utcdatetime_ce->name)`.

I think we should also specify whether the increment or timestamp argument is failing in these messages. We can either add something like "(timestamp)" and "(increment)" to the end of the messages, or include the field between the classname and "initialization". What do you think?

src/BSON/Timestamp.c
derickr: Yes, because it belongs to *both* increment and timestamp here, and adding it in the middle looked weird.

src/BSON/Timestamp.c
derickr: Sorry?

jmikola: It's towards the end of this line. You're currently reporting the UTCDateTime class name instead of Timestamp.

derickr: Ah! OK. That was unclear :-)

src/BSON/Timestamp.c
```diff
@@ -285,10 +337,12 @@ HashTable *php_phongo_timestamp_get_properties(zval *object TSRMLS_DC) /* {{{ */
 
 		MAKE_STD_ZVAL(increment);
 		ZVAL_LONG(increment, intern->increment);
+		convert_to_string(increment);
```
jmikola: Since the values are unsigned 32-bit integers, don't we still have an issue if `SIZEOF_ZEND_LONG` is 4?

src/BSON/UTCDateTime.c
```diff
@@ -370,43 +355,28 @@ HashTable *php_phongo_utcdatetime_get_properties(zval *object TSRMLS_DC) /* {{{
...
 	{
 		zval *milliseconds;
 
 		MAKE_STD_ZVAL(milliseconds);
-#if SIZEOF_LONG == 4
-		ZVAL_STRINGL(milliseconds, s_milliseconds, s_milliseconds_len, 1);
-#else
 		ZVAL_LONG(milliseconds, intern->milliseconds);
-#endif
+		convert_to_string(milliseconds);
```
jmikola: Same as with Timestamp, do we have an issue if `SIZEOF_ZEND_LONG` is 4?

tests/bson/bson-utcdatetime-004.phpt
```diff
@@ -18,10 +18,10 @@ foreach ($tests as $test) {
 --EXPECTF--
 object(MongoDB\BSON\UTCDateTime)#%d (%d) {
   ["milliseconds"]=>
-  int(%d)
+  string(13) "%d"
 }
 object(MongoDB\BSON\UTCDateTime)#%d (%d) {
   ["milliseconds"]=>
-  int(%d)
+  string(13) "%d"
```
jmikola: Consider using `%d` for the string length for all places where you also accept a `%d` in the pattern.

tests/bson/bson-utcdatetime-set_state-002.phpt
```diff
@@ -23,15 +23,15 @@ foreach ($tests as $milliseconds) {
 <?php exit(0); ?>
 --EXPECT--
 MongoDB\BSON\UTCDateTime::__set_state(array(
-   'milliseconds' => 0,
+   'milliseconds' => '0',
```
jmikola: If `var_export()` is always using strings, you should update `bson-utcdatetime-set_state-001.phpt` as well. It no longer needs the EXPECTF and regex patterns to conditionally accept single quotes around these values.

src/BSON/Timestamp.c
jmikola: To elaborate: since timestamp and increment are `uint32_t`, they could exceed the range of signed 32-bit ints. ZVAL_LONG is being used to stuff this unsigned value into a 32-bit integer (within the zval struct), which can obviously hold the value but it might be interpreted as a negative value later on. My concern is that `convert_to_string()` might render `4294967295` as `"-2147483648"`.

tests/bson/bson-utcdatetime-004.phpt
derickr: I don't agree with that one. The string length should always be 13.

jmikola:
LGTM.

Will the last two commits be rebased into the one tagged with the issue number? Otherwise, they should probably have the issue number as well.

derickr:
You mean squash the three commits? I can do, but I wasn't going to.

derickr:
Squashed :-)

PR #388, opened 2016-09-06
PHPC-742: Regex constructor should default flags arg to empty string

derickr:


tests/bson/bson-regex-003.phpt
```diff
@@ -0,0 +1,36 @@
...
+$regexp = new MongoDB\BSON\Regex("regexp");
+printf("Pattern: %s\n", $regexp->getPattern());
+printf("Flags: %s\n", $regexp->getFlags());
+printf("String representation: %s\n", $regexp);
+
+$tests = array(
+    array("regex" => $regexp),
+);
+
+foreach($tests as $n => $test) {
```
jmikola: Do you have intention of adding additional test cases? If not, we can do away with this `foreach`.

tests/bson/bson-regex-serialization-002.phpt
```diff
@@ -0,0 +1,27 @@
+--TEST--
+MongoDB\BSON\Regex serialization #2
```
jmikola: Can we replace "#2" with "with flags omitted"?

tests/bson/bson-regex-003.phpt
```diff
@@ -0,0 +1,36 @@
+--TEST--
+MongoDB\BSON\Regex #003
```
jmikola: Can we replace "#003" with "construction with flags omitted"?

tests/bson/bson-regex-004.phpt
```diff
@@ -0,0 +1,20 @@
+--TEST--
+MongoDB\BSON\Regex debug handler
```
jmikola: Can we add "with flags omitted" as a suffix?

Alternatively, I have no qualms about adding this to `bson-regex-002.phpt`, which is the existing debug handler test. The test itself is basically a one-liner if we do away with the variable assignment.

jmikola:
LGTM with above comments.

I'll understand if you don't want to combine the omitted flag variations in the original test cases; however, I would like more descriptive test titles if we're keeping them split.

tests/bson/bson-regex-003.phpt
derickr: It's a copy of #001, which had the same thing. I didn't want to deviate (or change an existing test).

derickr:
Will update the test descriptions and merge.

tests/bson/bson-regex-003.phpt
jmikola: If you're merging this as-is, don't object if I go back and refactor this at some point. It's dead code that doesn't add any value to the test.

derickr: I don't mind!

PR #389, opened 2016-09-06
PHPC-731: Parse Timestamp argument as strings to accept large integers

derickr:
Depends on PHPC-726/PR #386

### DIFF
--- a/src/BSON/Timestamp.c
+++ b/src/BSON/Timestamp.c
@@ -141,27 +141,28 @@ static bool php_phongo_timestamp_init_from_hash(php_phongo_timestamp_t *intern, 
 	return false;
 }
 
-/* {{{ proto void Timestamp::__construct(integer $increment, integer $timestamp)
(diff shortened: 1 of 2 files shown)


tests/bson/bson-timestamp-006.phpt
```diff
@@ -0,0 +1,39 @@
+--TEST--
+MongoDB\BSON\Timestamp constructor requires 64-bit integers to be positive unsigned 32-bit integers (as string)
```
jmikola: Unlike the 003 and 004 tests, these are only using string types. It doesn't make sense to refer to 64-bit integers or restrict the test to 64-bit platforms.

Both of these test cases should just be added into `bson-timestamp-005.phpt`, especially because we do want to test that the full range of 32-bit unsigned integers can be expressed on a 32-bit platform.

derickr:
All updated.

jmikola:
LGTM.

PR #391, opened 2016-09-06
PHPC-779: Switch serialization from O type to C type

derickr:


src/BSON/Binary.c
```diff
@@ -199,6 +205,94 @@ PHP_METHOD(Binary, getType)
...
+	php_var_serialize(&buf, &retval, &var_hash);
+	smart_str_0(&buf);
+	PHP_VAR_SERIALIZE_DESTROY(var_hash);
+
+#if PHP_VERSION_ID >= 70000
+	PHONGO_RETVAL_STRINGL(ZSTR_VAL(buf.s), ZSTR_LEN(buf.s));
+	zend_string_release(buf.s);
+#else
+	PHONGO_RETVAL_STRINGL(buf.c, buf.len);
+	efree(buf.c);
```
jmikola: `zend_string_release()` and `efree()` can be replaced with `smart_str_free()`, which will work for both versions of PHP.

src/BSON/Binary.c
```diff
@@ -199,6 +205,94 @@ PHP_METHOD(Binary, getType)
...
+#else
+	ALLOC_INIT_ZVAL(retval);
+	array_init_size(retval, 2);
+	ADD_ASSOC_STRINGL(retval, "data", intern->data, intern->data_len);
+	ADD_ASSOC_LONG_EX(retval, "type", intern->type);
+#endif
+
+	PHP_VAR_SERIALIZE_INIT(var_hash);
+	php_var_serialize(&buf, &retval, &var_hash);
+	smart_str_0(&buf);
```
jmikola: Is the null byte really necessary if you're returning the string by length? Since this is primarily a function for userland, where strings are not null-terminated (i.e. null bytes aren't exposed to the users), I expect you would normally just return whatever the inner call to `serialize()` returned.

src/BSON/Binary.c
```diff
@@ -199,6 +205,94 @@ PHP_METHOD(Binary, getType)
...
+	PHP_VAR_SERIALIZE_DESTROY(var_hash);
+
+#if PHP_VERSION_ID >= 70000
+	PHONGO_RETVAL_STRINGL(ZSTR_VAL(buf.s), ZSTR_LEN(buf.s));
+	zend_string_release(buf.s);
+#else
+	PHONGO_RETVAL_STRINGL(buf.c, buf.len);
```
jmikola: Since the `smart_str` freeing already has a compatibility function, I would also suggest defining a new `PHONGO_RETVAL_SMART_STR()` macro in `phongo_compat.h`, which would let you do away with this entire `#if` block.

src/BSON/Binary.c
derickr: Probably not, but it's safer to do. It's a single if + memset, so I am inclined to leave it.

jmikola:
Would you like to change `bson-binary-serialization-001.phpt` to replace the binary content with EXPECTF patterns? I believe you did that for HHVM.

src/BSON/Binary.c
```diff
@@ -199,6 +205,96 @@ PHP_METHOD(Binary, getType)
...
+	}
+	zend_restore_error_handling(&error_handling TSRMLS_CC);
+
+#if PHP_VERSION_ID < 70000
+	ALLOC_INIT_ZVAL(props);
+#endif
+	PHP_VAR_UNSERIALIZE_INIT(var_hash);
+	if (!php_var_unserialize(&props, (const unsigned char**) &serialized, (unsigned char *) serialized + serialized_len, &var_hash TSRMLS_CC)) {
+		zval_ptr_dtor(&props);
+		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "%s deserialization failed", ZSTR_VAL(php_phongo_binary_ce->name));
```
jmikola: I'd use "unserialization" based on the method name.

src/BSON/Binary.c
```diff
@@ -199,6 +205,96 @@ PHP_METHOD(Binary, getType)
...
+
+		PHP_VAR_UNSERIALIZE_DESTROY(var_hash);
+		RETURN_FALSE;
```
jmikola: [`Serializable::unserialize()`](http://php.net/manual/en/serializable.unserialize.php) has a void return type, so I think we can simply use `return` here.

src/BSON/Javascript.c
```diff
@@ -227,6 +233,116 @@ PHP_METHOD(Javascript, getScope)
...
+
+	intern = Z_JAVASCRIPT_OBJ_P(getThis());
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+#if PHP_VERSION_ID >= 70000
+	if (intern->scope && intern->scope->len) {
+		phongo_bson_to_zval_ex(bson_get_data(intern->scope), intern->scope->len, &state);
```
jmikola: This function can throw an exception and return zero. 

If you look at `php_phongo_javascript_get_properties()`, it stores null for the "scope" field (as if it didn't exist at all) and allows the function to return. That works out because nothing later in the function would throw. In this case, it might make more sense to return early instead of continuing to call `php_var_serialize()` below.

src/BSON/Decimal128.c
```diff
@@ -163,6 +169,96 @@ PHP_METHOD(Decimal128, __wakeup)
...
+	}
+	zend_restore_error_handling(&error_handling TSRMLS_CC);
+
+#if PHP_VERSION_ID < 70000
+	ALLOC_INIT_ZVAL(props);
+#endif
+	PHP_VAR_UNSERIALIZE_INIT(var_hash);
+	if (!php_var_unserialize(&props, (const unsigned char**) &serialized, (unsigned char *) serialized + serialized_len, &var_hash TSRMLS_CC)) {
+		zval_ptr_dtor(&props);
+		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "%s unserialization failed", ZSTR_VAL(php_phongo_binary_ce->name));
```
jmikola: `php_phongo_binary_ce->name` should be `php_phongo_decimal128_ce->name`.

src/BSON/Timestamp.c
```diff
@@ -226,6 +232,102 @@ PHP_METHOD(Timestamp, __wakeup)
...
+	int                       s_timestamp_len;
+
+	intern = Z_TIMESTAMP_OBJ_P(getThis());
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	s_increment_len = snprintf(s_increment, sizeof(s_increment), "%" PRId64, intern->increment);
+	s_timestamp_len = snprintf(s_timestamp, sizeof(s_timestamp), "%" PRId64, intern->timestamp);
```
jmikola: These are `uint32_t` values, so `PRIu32` would be more correct.

src/BSON/Timestamp.c
```diff
@@ -226,6 +232,102 @@ PHP_METHOD(Timestamp, __wakeup)
...
+	php_phongo_timestamp_t   *intern;
+#if PHP_VERSION_ID >= 70000
+	zval                      retval;
+#else
+	zval                     *retval;
+#endif
+	php_serialize_data_t      var_hash;
+	smart_str                 buf = { 0 };
+	char                      s_increment[24];
+	char                      s_timestamp[24];
```
jmikola: If `PRIu32` is used for these unsigned, 32-bit values, the largest string representation would be "4294967295" (11 characters including the null byte). I think 12 is sufficient here.

24 was fine for 64-bit values, which could be 21 characters including the null byte (signed or unsigned).

src/BSON/Timestamp.c
jmikola: I just realized that `php_phongo_timestamp_init_from_string()` is using `int64_t` for these, which is incorrect. I must have missed that when reviewing https://github.com/mongodb/mongo-php-driver/pull/386/files#diff-3598851e9e7e7e1da5e874e6bbfdb449R72.

jmikola:
I noticed that the `__wakeup()` method was still defined for these BSON classes. AFAIK, PHP checks if the class implements Serializable first, but the old method should likely be removed with the addition of `unserialize()`.

src/BSON/Timestamp.c
jmikola: On second glance, `php_phongo_timestamp_init_from_string()` is using `int64_t` due to the string conversion functions. It relies on `php_phongo_timestamp_init()`, which does the range checking.

Perhaps `php_phongo_timestamp_init()` has an issue on 32-bit platforms, since its arguments are `phongo_long` types, which would be a signed 32-bit value on some platforms. In any event, that's clearly outside the scope of this PR.

My first comment in this thread still stands, though.

src/BSON/Javascript.c
derickr: The return value of this functions is almost never checked though... it should.

src/BSON/Timestamp.c
derickr: Things are definitely messed up on 32bit systems:

```
=====================================================================
FAILED TEST SUMMARY
---------------------------------------------------------------------
MongoDB\BSON\Timestamp debug handler [tests/bson/bson-timestamp-002.phpt]
MongoDB\BSON\Timestamp constructor requires positive unsigned 32-bit integers [tests/bson/bson-timestamp-003.phpt]
MongoDB\BSON\Timestamp constructor requires positive unsigned 32-bit integers (as string) [tests/bson/bson-timestamp-005.phpt]
MongoDB\BSON\Timestamp serialization [tests/bson/bson-timestamp-serialization-001.phpt]
MongoDB\BSON\Timestamp::__set_state() [tests/bson/bson-timestamp-set_state-001.phpt]
MongoDB\BSON\Timestamp::__set_state() requires positive unsigned 32-bit integers [tests/bson/bson-timestamp-set_state_error-002.phpt]
PHPC-592: Property name corrupted when unserializing 64-bit integer on 32-bit platform [tests/bson/bug0592.phpt]
Sorting single field, ascending, using the Cursor Iterator [tests/functional/cursor-001.phpt]
Sorting single field, ascending, using the Cursor Iterator [tests/functional/cursorid-001.phpt]
Sorting single field, ascending [tests/functional/query-sort-001.phpt]
Sorting single field, descending [tests/functional/query-sort-002.phpt]
Sorting single field, ascending, using the Cursor Iterator [tests/functional/query-sort-003.phpt]
=====================================================================
```

Ticket → https://jira.mongodb.org/browse/PHPC-788

derickr:
New ticket for that → https://jira.mongodb.org/browse/PHPC-789

derickr:
Fixed all the things.

jmikola:
LGTM!

PR #399, opened 2016-09-12
PHPC-424: Validate that RP tag set is an array of documents

jmikola:
https://jira.mongodb.org/browse/PHPC-424

This adds common validation for read preference tag sets when specified through either the Manager constructor's URI options array or ReadPreference constructor.

An additional test case for a malformed tag set has been added to the Manager::__construct() error test for read preference options. Additionally, the ReadPreference::__construct() error test has been split up to test for mode and tagSet errors separately.

Note: we cannot test for the exceptions for bson_init_static() and mongoc_read_prefs_is_valid(), since those points will never be hit in normal operation.

### DIFF
--- a/src/MongoDB/ReadPreference.c
+++ b/src/MongoDB/ReadPreference.c
@@ -74,22 +74,39 @@ PHP_METHOD(ReadPreference, __construct)
 		case MONGOC_READ_SECONDARY_PREFERRED:
 		case MONGOC_READ_NEAREST:
 			intern->read_preference = mongoc_read_prefs_new(mode);
-
-			if (tagSets) {
-				bson_t *tags = bson_new();
-
-				phongo_zval_to_bson(tagSets, PHONGO_BSON_NONE, (bson_t *)tags, NULL TSRMLS_CC);
-				mongoc_read_prefs_set_tags(intern->read_preference, tags);
-				bson_destroy(tags);
-				if (!mongoc_read_prefs_is_valid(intern->read_preference)) {
-					phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Invalid tagSets");
-					return;
-				}
-			}
 			break;
 		default:
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Invalid mode: %" PHONGO_LONG_FORMAT, mode);
 			return;
+	}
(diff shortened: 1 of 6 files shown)


src/MongoDB/ReadPreference.c
```diff
@@ -74,22 +74,39 @@ PHP_METHOD(ReadPreference, __construct)
 		case MONGOC_READ_SECONDARY_PREFERRED:
 		case MONGOC_READ_NEAREST:
 			intern->read_preference = mongoc_read_prefs_new(mode);
+			break;
+		default:
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Invalid mode: %" PHONGO_LONG_FORMAT, mode);
+			return;
+	}
 
+	switch(ZEND_NUM_ARGS()) {
```
jmikola: This single-case `switch` should make sense once you see https://github.com/mongodb/mongo-php-driver/pull/400. Since I was refactoring the tagSet validation already, I opted to restructure this early. It should also make the diff for PHPC-752's commit easier to understand.

tests/manager/manager-ctor_error-003.phpt
```diff
@@ -28,4 +32,6 @@ OK: Got MongoDB\Driver\Exception\InvalidArgumentException
 Unsupported readPreference value: 'nothing'
 OK: Got MongoDB\Driver\Exception\InvalidArgumentException
 Primary read preference mode conflicts with tags
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Read preference tags must be an array of zero or more documents
```
derickr: I think this should say "Primary read preference mode conflicts with tags". It doesn't matter whether they're valid or not yet.

jmikola: This is more consistent with the ReadPreference constructor and I think it's more helpful to the user. We're validating individual values before we check the combined state of the read preference.

On that note, I think there's an oversight in `php_phongo_apply_rp_options_to_uri()` where we're never checking if "maxstalenessms" is a positive value. As-is, I think a negative value would only trigger a generic "Read preference is not valid" error.

jmikola: I added another case to the ReadPreference ctor test to demonstrate that it also reports invalid tag set structure first (before the primary conflict).

PR #380, opened 2016-08-31
PHPC-752: Allow users to set a limit on acceptable staleness

derickr:
https://jira.mongodb.org/browse/PHPC-752

PR #400, opened 2016-09-12
PHPC-752: Add maxStalenessMS to ReadPreference class

jmikola:
https://jira.mongodb.org/browse/PHPC-752

I was premature in signing off on https://github.com/mongodb/mongo-php-driver/pull/380, as it was lacking support for maxStalenessMS on the ReadPreference object. This adds a new constructor argument, getter method, and adds "maxStalenessMS" to the ReadPreference debug output.


tests/readPreference/readpreference-ctor-001.phpt
```diff
@@ -7,8 +7,9 @@ MongoDB\Driver\ReadPreference construction
 require_once __DIR__ . "/../utils/basic.inc";
 
 var_dump(new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_PRIMARY));
-var_dump(new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_SECONDARY, array(array("tag" => "one"))));
-var_dump(new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_PRIMARY, array()));
+var_dump(new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_SECONDARY, [['tag' => 'one']]));
+var_dump(new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_PRIMARY, []));
```
jmikola: Just `array()` to `[]` refactoring here. Since this is the only ReadPreference constructor test, I added a new case for "maxStalenessMS" below; however, it's probably redundant in light of the `ReadPreference::getMaxStalenessMS()` test.

src/MongoDB/ReadPreference.c
```diff
@@ -46,21 +46,22 @@ PHONGO_API zend_class_entry *php_phongo_readpreference_ce;
 
 zend_object_handlers php_phongo_handler_readpreference;
 
-/* {{{ proto void ReadPreference::__construct(integer $mode[, array $tagSets = array()])
+/* {{{ proto void ReadPreference::__construct(integer $mode[, array $tagSets = array()[, integer $maxStalenessMS]])
```
derickr: I would much rather see this being an options array, as it is possible that other options might be added in the future.

tests/readPreference/readpreference-getTagSets-001.phpt
```diff
@@ -1,5 +1,5 @@
 --TEST--
-MongoDB\Driver\ReadPreference::getMode()
+MongoDB\Driver\ReadPreference::getTagSets()
```
derickr: This commit doesn't really belong here, perhaps in #399.

jmikola: It's a stand-alone commit akin to any typo fix. The PR should make no difference.

jmikola: Moved to https://github.com/mongodb/mongo-php-driver/pull/408.

src/MongoDB/ReadPreference.c
jmikola: I'll refactor this to expect an options array and check a "maxStalenessMS" value.

src/MongoDB/ReadPreference.c
```diff
@@ -80,28 +83,46 @@ PHP_METHOD(ReadPreference, __construct)
 			return;
 	}
 
-	switch(ZEND_NUM_ARGS()) {
-		case 2:
-			if (tagSets) {
-				bson_t *tags = bson_new();
+	if (tagSets) {
```
jmikola: Since I removed the `switch` statement and am processing `tagSets` first, I added a test that tag set exceptions are thrown before maxStalenessMS.

tests/manager/manager-ctor-read_preference-003.phpt
```diff
@@ -0,0 +1,54 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): read preference options of the wrong type are ignored
```
jmikola: We may want to change this behavior in the future, but this is how processing of Manager options operates today. We can revise this test if we decide to change this in the future.

I'll note that `mongoc-uri.c` is also quite lax about type errors. For instance, options that are expected to be integers are run through `strtol()` with no error handling.

tests/manager/manager-maxstalenessms-001.phpt
```diff
@@ -1,13 +0,0 @@
---TEST--
-MongoDB\Driver\Manager: maxStalenessMS
```
jmikola: This was essentially renamed to `manager-ctor-read_preference-002.phpt`, where I added an additional test case and assert the state of `Manager::getReadPreference()`.

tests/manager/manager-maxstalenessms_error-001.phpt
```diff
@@ -1,33 +0,0 @@
---TEST--
-MongoDB\Driver\Manager: maxStalenessMS
```
jmikola: This was renamed to `manager-ctor-read_preference-error-002.phpt`, where I added an additional error case for an out-of-range value.

tests/manager/manager-ctor-read_preference-001.phpt
```diff
@@ -0,0 +1,96 @@
...
+    var_dump($manager->getReadPreference());
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+object(MongoDB\Driver\ReadPreference)#%d (%d) {
+  ["mode"]=>
+  int(1)
```
derickr: This is going to fail/create conflicts after the PHPC-498 changes (ReadPreference, ReadConcern, and WriteConcern value).

php_phongo.c
```diff
@@ -982,6 +982,8 @@ void php_phongo_read_preference_to_zval(zval *retval, const mongoc_read_prefs_t
 	} else {
 		ADD_ASSOC_NULL_EX(retval, "tags");
 	}
+
+	ADD_ASSOC_LONG_EX(retval, "maxStalenessMS", mongoc_read_prefs_get_max_staleness_ms(read_prefs));
```
derickr: I don't think this should be added if maxstaleness is the default (0), just like we're no longer to do that for "journal", "w", etc.

tests/manager/manager-ctor-read_preference-001.phpt
```diff
@@ -0,0 +1,96 @@
...
+  ["tags"]=>
+  array(0) {
+  }
+  ["maxStalenessMS"]=>
+  int(0)
```
derickr: I think this should be excluded here (As mentioned in an earlier comments).

tests/manager/manager-ctor-read_preference-001.phpt
jmikola: Is that because you're changing it to display the string name? If so, you can easily update this when PHPC-489 is implemented.

I have *absolutely no reservations whatever* about going back and modifying tests.

tests/manager/manager-ctor-read_preference-001.phpt
jmikola: Will do.

php_phongo.c
jmikola: Updated.

tests/manager/manager-ctor-read_preference-001.phpt
derickr: Yes, it turns into a string. I'll do the PHPC-489 tomorrow (Monday), should be ready for you to review when you wake up.

php_phongo.c
```diff
@@ -963,7 +963,7 @@ void php_phongo_read_preference_to_zval(zval *retval, const mongoc_read_prefs_t
 {
 	const bson_t *tags = mongoc_read_prefs_get_tags(read_prefs);
 
-	array_init_size(retval, 2);
+	array_init_size(retval, 3);
```
derickr: I guess using 3 when there are only 2 items is fine, right?

jmikola: This just controls the initial HashTable size, so it should be fine. The buckets are still empty until we fill them.

PR #402, opened 2016-09-13
Fix Windows build for libmongoc

jmikola:


config.w32
```diff
@@ -140,6 +151,10 @@ if (PHP_MONGODB != "no") {
     WARNING("mongodb libsasl support not enabled, libs not found");
   }
 
+  mongoc_opts.MONGOC_CC = COMPILER_NAME;
+  mongoc_opts.MONGOC_USER_SET_CFLAGS = get_define("CFLAGS_MONGODB");
+  mongoc_opts.MONGOC_USER_SET_LDFLAGS = get_define("LDFLAGS_MONGODB");
```
jmikola: This will result in something similar to:

```
/* MONGOC_USER_SET_CFLAGS is set from config based on what compiler flags were
 * used to compile mongoc */
#define MONGOC_USER_SET_CFLAGS "/D BSON_COMPILATION /D MONGOC_COMPILATION /D MONGOC_TRACE /I..\pecl\mongodb /I..\pecl\mongodb/src/libmongoc/src/mongoc /I..\pecl\mongodb/src/libbson/src/bson/ /I..\pecl\mongodb/src/libbson/src /I..\pecl\mongodb/src/contrib /I \"C:\php-sdk\phpdev\vc14\x86\deps\include\" /DHAVE_OPENSSL_SSL_H=1"

#define MONGOC_USER_SET_LDFLAGS "/libpath:\"..\deps\lib\" /libpath:\"..\deps\lib\""

/* MONGOC_CC is used to determine what C compiler was used to compile mongoc */
#define MONGOC_CC "MSVC14 (Visual C++ 2015)"
```

The `MONGOC_CC` constant is certainly useful, but I don't think CFLAGS and LDFLAGS are helpful. In libmongoc's `configure.ac`, the CFLAGS and LDFLAGS are set before the library's own configuration. Do we even expect users to have custom CFLAGS for Windows builds?

Also, the `LIBS_MONGODB` environment variable may be more useful, as it actually lists the linked libraries. `LDFLAGS_MONGODB` is only reporting the paths.

config.w32
```diff
@@ -115,11 +122,15 @@ if (PHP_MONGODB != "no") {
...
     MONGOC_NO_AUTOMATIC_GLOBALS: 1,
+    MONGOC_CC: "",
+    MONGOC_USER_SET_CFLAGS: "",
+    MONGOC_USER_SET_LDFLAGS: ""
   };
 
   if (CHECK_LIB("ssleay32.lib", "mongodb", PHP_MONGODB) &&
       CHECK_LIB("libeay32.lib", "mongodb", PHP_MONGODB) &&
-      CHECK_HEADER_ADD_INCLUDE("openssl/ssl.h", "CFLAGS_MONGOC")) {
+      CHECK_LIB("crypt32.lib", "mongodb", PHP_MONGODB) &&
```
jmikola: This is necessary for both OpenSSL (currently used) and Secure Channel (will be used in the future).

config.w32
```diff
@@ -140,6 +151,14 @@ if (PHP_MONGODB != "no") {
     WARNING("mongodb libsasl support not enabled, libs not found");
   }
 
+  if (typeof COMPILER_NAME === 'string') {
+    mongoc_opts.MONGOC_CC = COMPILER_NAME;
+  } else if (typeof VC_VERSIONS[VCVERS] === 'string') {
+    mongoc_opts.MONGOC_CC = VC_VERSIONS[VCVERS];
```
jmikola: PHP 7 defines COMPILER_NAME in `confutils.js`, but we need this for PHP 5.x.

config.w32
```diff
@@ -140,6 +151,14 @@ if (PHP_MONGODB != "no") {
...
+  }
+  mongoc_opts.MONGOC_USER_SET_CFLAGS = get_define("CFLAGS_MONGODB");
+  mongoc_opts.MONGOC_USER_SET_LDFLAGS = get_define("LDFLAGS_MONGODB");
```
jmikola: This will result in something similar to:

```
/* MONGOC_USER_SET_CFLAGS is set from config based on what compiler flags were
 * used to compile mongoc */
#define MONGOC_USER_SET_CFLAGS "/D BSON_COMPILATION /D MONGOC_COMPILATION /D MONGOC_TRACE /I..\pecl\mongodb /I..\pecl\mongodb/src/libmongoc/src/mongoc /I..\pecl\mongodb/src/libbson/src/bson/ /I..\pecl\mongodb/src/libbson/src /I..\pecl\mongodb/src/contrib /I \"C:\php-sdk\phpdev\vc14\x86\deps\include\" /DHAVE_OPENSSL_SSL_H=1"

#define MONGOC_USER_SET_LDFLAGS "/libpath:\"..\deps\lib\" /libpath:\"..\deps\lib\""

/* MONGOC_CC is used to determine what C compiler was used to compile mongoc */
#define MONGOC_CC "MSVC14 (Visual C++ 2015)"
```

The `MONGOC_CC` constant is certainly useful, but I don't think CFLAGS and LDFLAGS are helpful. In libmongoc's `configure.ac`, the CFLAGS and LDFLAGS are set before the library's own configuration. Do we even expect users to have custom CFLAGS for Windows builds?

Also, the `LIBS_MONGODB` environment variable may be more useful, as it actually lists the linked libraries. `LDFLAGS_MONGODB` is only reporting the paths.

jmikola:
Per @bjori's suggestion, I've left MONGOC_USER_SET_CFLAGS and MONGOC_USER_SET_LDFLAGS blank. We don't expect these to be customized at build time and it's not very useful to repeat what we already setup earlier in `config.w32`.

config.w32
```diff
@@ -80,7 +86,7 @@ if (PHP_MONGODB != "no") {
     BSON_HAVE_TIMESPEC: 0,
     BSON_EXTRA_ALIGN: 0,
     BSON_HAVE_SYSCALL_TID: 0,
-    BSON_HAVE_DECIMAL128: 0,
```
derickr: Does the Windows compiler not like this at all?

config.w32
derickr: What does this to the output? I'd rather have it being consistent, than "proper". So if VC_VERSIONS works for everything, then we should use that (unless it does nothing to the result of course).

jmikola: * [VC_VERSIONS in PHP 7.0](https://github.com/php/php-src/blob/php-7.0.10/win32/build/confutils.js#L56)
 * [VC_VERSIONS in PHP 5.x](https://github.com/php/php-src/blob/php-5.6.25/win32/build/confutils.js#L43)

They are consistent for VC, but PHP 7.0 also supports CLANG and ICC. `COMPILER_NAME` is initialized from [`toolset_get_compiler_name()`](https://github.com/php/php-src/blob/php-7.0.10/win32/build/confutils.js#L2583) if you'd like to see the logic.

config.w32
jmikola: The trailing comma did raise an error. I'm not sure why this wasn't an issue before, but I noticed it after updating my php-sdk sources and deps (now 5.6.25 and 7.0.10).

derickr:
LGTM

PR #403, opened 2016-09-13
PHPC-757: Set compiler name and user CFLAGS/LDFLAGS earlier

jmikola:
https://jira.mongodb.org/browse/PHPC-757

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -6,6 +6,23 @@ PHP_ARG_WITH(openssl-dir, OpenSSL dir for mongodb,
 PHP_ARG_WITH(system-ciphers, whether to use system default cipher list instead of hardcoded value,
 [  --with-system-ciphers   OPENSSL: Use system default cipher list instead of hardcoded value], no, no)
 
+dnl borrowed from libmongoc configure.ac
+# AS_VAR_COPY is available in AC 2.64 and on, but we only require 2.60.
+# If we're on an older version, we define it ourselves:
+m4_ifndef([AS_VAR_COPY],
+          [m4_define([AS_VAR_COPY],
+          [AS_LITERAL_IF([$1[]$2], [$1=$$2], [eval $1=\$$2])])])
+
+# Get "user-set cflags" here, before we've added the flags we use by default
+AS_VAR_COPY(MONGOC_USER_SET_CFLAGS, [CFLAGS])
+AC_SUBST(MONGOC_USER_SET_CFLAGS)
+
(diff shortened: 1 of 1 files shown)


config.m4
```diff
@@ -6,6 +6,23 @@ PHP_ARG_WITH(openssl-dir, OpenSSL dir for mongodb,
...
 [  --with-system-ciphers   OPENSSL: Use system default cipher list instead of hardcoded value], no, no)
 
+dnl borrowed from libmongoc configure.ac
+# AS_VAR_COPY is available in AC 2.64 and on, but we only require 2.60.
+# If we're on an older version, we define it ourselves:
+m4_ifndef([AS_VAR_COPY],
+          [m4_define([AS_VAR_COPY],
+          [AS_LITERAL_IF([$1[]$2], [$1=$$2], [eval $1=\$$2])])])
+
+# Get "user-set cflags" here, before we've added the flags we use by default
```
jmikola: @derickr: I think this might address your concern with these strings being too long. In my local builds, LDFLAGS were empty and CFLAGS was a modest `"-g  -O0"`.

config.m4
```diff
@@ -6,6 +6,23 @@ PHP_ARG_WITH(openssl-dir, OpenSSL dir for mongodb,
...
+AS_VAR_COPY(MONGOC_USER_SET_CFLAGS, [CFLAGS])
+AC_SUBST(MONGOC_USER_SET_CFLAGS)
+
+AS_VAR_COPY(MONGOC_USER_SET_LDFLAGS, [LDFLAGS])
+AC_SUBST(MONGOC_USER_SET_LDFLAGS)
+
+AS_VAR_COPY(MONGOC_CC, [CC])
+AC_SUBST(MONGOC_CC)
```
jmikola: This was never added in https://github.com/mongodb/mongo-php-driver/commit/5a0514826b2a05ef61676dbedd8150a2eeee77ef.

config.m4
derickr: That's what it was in mine as well.

config.m4
derickr: I never spotted a missing MONGOC_CC bit t.b.h. Maybe that got added later?

derickr:
LGTM

config.m4
jmikola: Possibly. Although this would not have failed like other numeric values, since `"@MONGOC_CC@"` appears as a string in the `mongoc-config.h.in` template and isn't a syntax error when left as-is.

PR #404, opened 2016-09-14
Allow compiling without SSL

bjori:


php_phongo.c
```diff
@@ -59,9 +59,6 @@
 /* PHP array helpers */
 #include "php_array_api.h"
 
-/* For our stream verifications */
-#include <openssl/x509.h>
```
jmikola: Good catch! This made me realize that I never cleaned up the `php-ssl.c` contrib code.

Removing that in https://github.com/mongodb/mongo-php-driver/pull/405.

jmikola:
See: #549

PR #409, opened 2016-09-15
Use string literal php_array macros where possible

jmikola:
This updates a few cases where we were already passing string literals to the php_array macros. The "c" variant expands to `lit, sizeof(lit) - 1` instead of `lit, strlen(lit)`.


derickr:
LGTM

PR #410, opened 2016-09-16
Fix required version of system installed mongoc and libbson

bjori:


derickr:
LGTM

jmikola:
Manually merged in https://github.com/mongodb/mongo-php-driver/commit/0c62e64228cbeaa3151cfac80775ce1ae1ec6db7.

I believe you missed updating the libmongoc pgk-config line, so I amended that to your commit.

bjori:
I did indeed miss it, only noticed the ones that said `1.3.3`, while the libmongoc one had been updated to `1.4`. Thanks!

PR #411, opened 2016-09-16
PHPC-783: Use mongoc_collection_find_with_opts() for Query exec

jmikola:
https://jira.mongodb.org/browse/PHPC-734
https://jira.mongodb.org/browse/PHPC-783
https://jira.mongodb.org/browse/PHPC-791


src/MongoDB/Query.c
```diff
@@ -50,64 +50,244 @@ PHONGO_API zend_class_entry *php_phongo_query_ce;
...
+		if (!php_phongo_query_opts_append_string(intern->opts, (opt), (zarr), (key) TSRMLS_CC)) { \
+			return false; \
+		} \
+	}
+
+	PHONGO_QUERY_OPT_BOOL("allowPartialResults", options, "allowPartialResults");
+	PHONGO_QUERY_OPT_BOOL("awaitData", options, "awaitData");
+	PHONGO_QUERY_OPT_INT64("batchSize", options, "batchSize");
+	PHONGO_QUERY_OPT_DOCUMENT("collation", options, "collation");
+	PHONGO_QUERY_OPT_STRING("comment", options, "comment") else PHONGO_QUERY_OPT_STRING("comment", modifiers, "$comment");
```
derickr: I am no keen on this style with just the random `else`. That looks really confusing, especially because it's "just tacked onto the end". Perhaps try to come up with a clever style of aligning here?

src/MongoDB/Query.c
```diff
@@ -50,64 +50,244 @@ PHONGO_API zend_class_entry *php_phongo_query_ce;
...
-			if (Z_TYPE_P(zread_concern) != IS_OBJECT || !instanceof_function(Z_OBJCE_P(zread_concern), php_phongo_readconcern_ce TSRMLS_CC)) {
-				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected \"readConcern\" option to be %s, %s given", ZSTR_VAL(php_phongo_readconcern_ce->name), zend_get_type_by_const(Z_TYPE_P(zread_concern)));
-				bson_clear(&intern->query);
-				return;
+	/* Note: handling of integer options will depend on SIZEOF_ZEND_LONG and we
+	 * are not converting strings to 64-bit integers for 32-bit platforms. */
+
+#define PHONGO_QUERY_OPT_INT64(opt, zarr, key) \
+	if (zarr && php_array_existsc(zarr, key)) { \
+		if (!BSON_APPEND_INT64(intern->opts, (opt), php_array_fetchc_long(zarr, key))) { \
```
derickr: These exceptions here are really just defensive coding? I can't imagine these appends would fail for any real reason. I also can't think of a way to reproduce this in HHVM.

src/MongoDB/Query.c
```diff
@@ -50,64 +50,244 @@ PHONGO_API zend_class_entry *php_phongo_query_ce;
...
+		} \
+	}
+
+	PHONGO_QUERY_OPT_BOOL("allowPartialResults", options, "allowPartialResults");
+	PHONGO_QUERY_OPT_BOOL("awaitData", options, "awaitData");
+	PHONGO_QUERY_OPT_INT64("batchSize", options, "batchSize");
+	PHONGO_QUERY_OPT_DOCUMENT("collation", options, "collation");
+	PHONGO_QUERY_OPT_STRING("comment", options, "comment") else PHONGO_QUERY_OPT_STRING("comment", modifiers, "$comment");
+	PHONGO_QUERY_OPT_BOOL("exhaust", options, "exhaust");
+	PHONGO_QUERY_OPT_DOCUMENT("max", options, "max") else PHONGO_QUERY_OPT_DOCUMENT("max", modifiers, "$max");
```
derickr: What's the reason of this duality between `max` and `$max` anyway?

src/MongoDB/Query.c
```diff
@@ -183,49 +363,44 @@ HashTable *php_phongo_query_get_debug_info(zval *object, int *is_temp TSRMLS_DC)
 	*is_temp = 1;
 	intern = Z_QUERY_OBJ_P(object);
 
-	array_init_size(&retval, 6);
+	array_init_size(&retval, 3);
 
-	/* Avoid using PHONGO_TYPEMAP_NATIVE_ARRAY for decoding query and selector
+	/* Avoid using PHONGO_TYPEMAP_NATIVE_ARRAY for decoding filter and opts
```
derickr: How does this avoid it?

src/MongoDB/Query.c
```diff
@@ -50,64 +50,244 @@ PHONGO_API zend_class_entry *php_phongo_query_ce;
...
+	PHONGO_QUERY_OPT_DOCUMENT("min", options, "min") else PHONGO_QUERY_OPT_DOCUMENT("min", modifiers, "$min");
+	PHONGO_QUERY_OPT_BOOL("noCursorTimeout", options, "noCursorTimeout");
+	PHONGO_QUERY_OPT_BOOL("oplogReplay", options, "oplogReplay");
+	PHONGO_QUERY_OPT_DOCUMENT("projection", options, "projection");
+	PHONGO_QUERY_OPT_BOOL("returnKey", options, "returnKey") else PHONGO_QUERY_OPT_BOOL("returnKey", modifiers, "$returnKey");
+	PHONGO_QUERY_OPT_BOOL("showRecordId", options, "showRecordId") else PHONGO_QUERY_OPT_BOOL("showRecordId", modifiers, "$showDiskLoc");
+	PHONGO_QUERY_OPT_INT64("skip", options, "skip");
+	PHONGO_QUERY_OPT_DOCUMENT("sort", options, "sort") else PHONGO_QUERY_OPT_DOCUMENT("sort", modifiers, "$orderby");
+	PHONGO_QUERY_OPT_BOOL("snapshot", options, "snapshot") else PHONGO_QUERY_OPT_BOOL("snapshot", modifiers, "$snapshot");
+	PHONGO_QUERY_OPT_BOOL("tailable", options, "tailable");
```
derickr: Because this is a hard coded list, it would be necessary to add code if new options are going to be supported by libmongoc/the server. Do we want to add a fall back mechanism for this?

src/MongoDB/Query.c
```diff
@@ -183,49 +363,44 @@ HashTable *php_phongo_query_get_debug_info(zval *object, int *is_temp TSRMLS_DC)
...
 		zval zv;
 #else
 		zval *zv;
 #endif
 
-		phongo_bson_to_zval(bson_get_data(intern->selector), intern->selector->len, &zv);
+		phongo_bson_to_zval(bson_get_data(intern->opts), intern->opts->len, &zv);
 #if PHP_VERSION_ID >= 70000
-		ADD_ASSOC_ZVAL_EX(&retval, "selector", &zv);
+		ADD_ASSOC_ZVAL_EX(&retval, "opts", &zv);
```
derickr: Did you use "opts" because libmongoc uses that? If not, I believe we should use "options".

tests/query/query-ctor_error-002.phpt
```diff
@@ -0,0 +1,90 @@
...
+echo throws(function() {
+    new MongoDB\Driver\Query([], ['modifiers' => ['$orderby' => 0]]);
+}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected "modifiers" option to be array, integer given
```
derickr: I find this a little hard to read. Any chance you can make it `"\n\n"` everywhere?

src/MongoDB/Query.c
```diff
@@ -50,64 +50,244 @@ PHONGO_API zend_class_entry *php_phongo_query_ce;
...
-	zend_error_handling       error_handling;
-	zval                     *zfilter;
-	zval                     *zoptions = NULL;
-	bson_t                    bfilter;
-	bson_t                    boptions = BSON_INITIALIZER;
-	SUPPRESS_UNUSED_WARNING(return_value_ptr) SUPPRESS_UNUSED_WARNING(return_value) SUPPRESS_UNUSED_WARNING(return_value_used)
+	zval *value = php_array_fetch(zarr, zarr_key);
 
+	if (Z_TYPE_P(value) != IS_STRING) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected \"%s\" %s to be string, %s given", zarr_key, zarr_key[0] == '$' ? "modifier" : "option", zend_get_type_by_const(Z_TYPE_P(value)));
```
derickr: Do you have tests for this case? (All `php_phongo_query_opts_append_*` functions)

src/MongoDB/Query.c
```diff
@@ -50,64 +50,244 @@ PHONGO_API zend_class_entry *php_phongo_query_ce;
...
+
+	/* The "$explain" modifier should be converted to an "explain" option, which
+	 * libmongoc will later convert back to a modifier for the OP_QUERY code
+	 * path. This modifier will be ignored for the find command code path. */
+	PHONGO_QUERY_OPT_BOOL("explain", modifiers, "$explain");
+
+	/* mongoc_collection_find_with_opts() requires a non-negative limit. For
+	 * backwards compatibility, a negative limit should be set as a positive
+	 * value and default singleBatch to true. */
+	if (php_array_existsc(options, "limit") && php_array_fetchc_long(options, "limit") < 0) {
```
derickr: To make this function smaller, I would probable recommend to split the special cases out into a function on their own.

src/MongoDB/Query.c
```diff
@@ -50,64 +50,244 @@ PHONGO_API zend_class_entry *php_phongo_query_ce;
...
+	if (php_array_existsc(options, "modifiers")) {
+		modifiers = php_array_fetchc(options, "modifiers");
+
+		if (Z_TYPE_P(modifiers) != IS_ARRAY) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected \"modifiers\" option to be array, %s given", zend_get_type_by_const(Z_TYPE_P(modifiers)));
+			return false;
+		}
+	}
+
+#define PHONGO_QUERY_OPT_BOOL(opt, zarr, key) \
```
derickr: I'd like to see the defines made *before* the function body.

src/MongoDB/Query.c
```diff
@@ -50,64 +50,244 @@ PHONGO_API zend_class_entry *php_phongo_query_ce;
...
+			return false;
 		}
 
-		phongo_zval_to_bson(zoptions, PHONGO_BSON_NONE, &boptions, NULL TSRMLS_CC);
+		intern->read_concern = mongoc_read_concern_copy(phongo_read_concern_from_zval(read_concern TSRMLS_CC));
 	}
 
-	if (!phongo_query_init(intern, &bfilter, &boptions TSRMLS_CC)) {
-		bson_clear(&intern->query);
+#undef PHONGO_QUERY_OPT_BOOL
```
derickr: Just like the `#define`s before, the `#undef`s should be (just) below.

src/MongoDB/Query.c
jmikola: Purely defensive coding. It can theoretically happen if `_bson_grow()` fails or `_bson_append()` detects that appending the field would cause the document length to overflow a 32-bit signed integer.

src/MongoDB/Query.c
jmikola: If the first macro attempts to set the `max` option from `$options['max']`, while the second attempts to set `max` from `$options['modifiers']['$max']`. In this case, the modifier and option names differ by only a "$" prefix, but that is not always the case (e.g. `$showDiskLoc` and `showRecordId`).

src/MongoDB/Query.c
jmikola: I don't believe we do this for any other options arrays (e.g. BulkWrite). We certainly don't need to do this for modifiers, but it may be worthwhile for top-level options. I'll create a separate tracking ticket for it and triage in 1.2.0, as it will require heavy modification to this PR.

jmikola: Tracking ticket: [PHPC-793](https://jira.mongodb.org/browse/PHPC-793)

src/MongoDB/Query.c
jmikola: This contrasts with other debug functions, which typically have:

```
/* Use native arrays for debugging output */
state.map.root_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
state.map.document_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
```

This debug handler uses `phongo_bson_to_zval()`, which defaults to the `PHONGO_TYPEMAP_NONE` via `PHONGO_BSON_STATE_INITIALIZER`.

src/MongoDB/Query.c
jmikola: This was based on the struct field, but I'll rename to "options".

tests/query/query-ctor_error-002.phpt
jmikola: Will do.

src/MongoDB/Query.c
jmikola: `query-ctor_error-002.phpt` covers all of the string and document (i.e. array or object) type checks for options and modifiers.

src/MongoDB/Query.c
jmikola: Will relocate.

src/MongoDB/Query.c
jmikola: I'll prefix each `else` with a newline and tab for readability.

src/MongoDB/Query.c
jmikola: Will relocate.

src/MongoDB/Query.c
jmikola: Will relocate.

tests/query/query-ctor-002.phpt
```diff
@@ -0,0 +1,169 @@
...
+  ["options"]=>
+  object(stdClass)#%d (%d) {
+    ["allowPartialResults"]=>
+    bool(false)
+    ["awaitData"]=>
+    bool(false)
+    ["batchSize"]=>
+    int(10)
+    ["collation"]=>
+    object(stdClass)#3 (1) {
```
derickr: This misses a %d (and hence my test against it fails ;-) ).

jmikola: Fixed in https://github.com/mongodb/mongo-php-driver/pull/421

PR #413, opened 2016-09-19
PHPC-498: ReadPreference, ReadConcern, and WriteConcern should serialize to BSON

derickr:
https://jira.mongodb.org/browse/PHPC-498

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -931,8 +931,6 @@ void php_phongo_read_concern_to_zval(zval *retval, const mongoc_read_concern_t *
 
 	if (level) {
 		ADD_ASSOC_STRING(retval, "level", (char *)level);
-	} else {
-		ADD_ASSOC_NULL_EX(retval, "level");
 	}
 } /* }}} */
 
@@ -962,12 +960,21 @@ bool php_phongo_read_preference_tags_are_valid(const bson_t *tags) /* {{{ */
 void php_phongo_read_preference_to_zval(zval *retval, const mongoc_read_prefs_t *read_prefs) /* {{{ */
 {
 	const bson_t *tags = mongoc_read_prefs_get_tags(read_prefs);
+	mongoc_read_mode_t mode = mongoc_read_prefs_get_mode(read_prefs);
 
 	array_init_size(retval, 2);
 
-	ADD_ASSOC_LONG_EX(retval, "mode", mongoc_read_prefs_get_mode(read_prefs));
+	switch (mode) {
+		case MONGOC_READ_PRIMARY: ADD_ASSOC_STRING(retval, "mode", "primary"); break;
+		case MONGOC_READ_PRIMARY_PREFERRED: ADD_ASSOC_STRING(retval, "mode", "primaryPreferred"); break;
+		case MONGOC_READ_SECONDARY: ADD_ASSOC_STRING(retval, "mode", "secondary"); break;
+		case MONGOC_READ_SECONDARY_PREFERRED: ADD_ASSOC_STRING(retval, "mode", "secondaryPreferred"); break;
+		case MONGOC_READ_NEAREST: ADD_ASSOC_STRING(retval, "mode", "nearest"); break;
+		default: /* Do nothing */
+			break;
+	}
 
-	if (tags->len) {
+	if (!bson_empty0(tags)) {
 		php_phongo_bson_state  state = PHONGO_BSON_STATE_INITIALIZER;
 		/* Use native arrays for debugging output */
 		state.map.root_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
@@ -979,8 +986,6 @@ void php_phongo_read_preference_to_zval(zval *retval, const mongoc_read_prefs_t 
 #else
 		ADD_ASSOC_ZVAL_EX(retval, "tags", state.zchild);
 #endif
-	} else {
-		ADD_ASSOC_NULL_EX(retval, "tags");
 	}
 } /* }}} */
 
@@ -988,6 +993,7 @@ void php_phongo_write_concern_to_zval(zval *retval, const mongoc_write_concern_t
 {
 	const char *wtag = mongoc_write_concern_get_wtag(write_concern);
 	const int32_t w = mongoc_write_concern_get_w(write_concern);
+	const int32_t wtimeout = mongoc_write_concern_get_wtimeout(write_concern);
 
 	array_init_size(retval, 4);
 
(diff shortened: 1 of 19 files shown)


php_phongo.c
```diff
@@ -979,15 +986,14 @@ void php_phongo_read_preference_to_zval(zval *retval, const mongoc_read_prefs_t
...
-	} else {
-		ADD_ASSOC_NULL_EX(retval, "tags");
 	}
 } /* }}} */
 
 void php_phongo_write_concern_to_zval(zval *retval, const mongoc_write_concern_t *write_concern) /* {{{ */
 {
 	const char *wtag = mongoc_write_concern_get_wtag(write_concern);
 	const int32_t w = mongoc_write_concern_get_w(write_concern);
+	const int32_t wtimeout = mongoc_write_concern_get_wtimeout(write_concern);
```
jmikola: Does this count as mixing code and declarations? I actually would have thought the compiler would yell about the existing `*wtag` and `w` lines.

src/MongoDB/ReadConcern.c
```diff
@@ -101,6 +101,16 @@ PHP_METHOD(ReadConcern, getLevel)
 }
 /* }}} */
 
+/* {{{ proto array ReadConcern::bsonSerialize()
+   */
+PHP_METHOD(ReadConcern, bsonSerialize)
+{
+	const mongoc_read_concern_t *read_concern = phongo_read_concern_from_zval(getThis() TSRMLS_CC);
```
jmikola: For consistency with our other functions that take no arguments, I'd consider adding this:

```
	return;
}
```

Also, did we need the following suppressions for PHP 5.x?

```
SUPPRESS_UNUSED_WARNING(return_value_ptr) SUPPRESS_UNUSED_WARNING(return_value_used)
```

If these warnings didn't trigger for you, you can always omit this change. I'll add them back if I see them later.

src/MongoDB/WriteConcern.c
```diff
@@ -182,6 +182,16 @@ PHP_METHOD(WriteConcern, getJournal)
 }
 /* }}} */
 
+/* {{{ proto array WriteConcern::bsonSerialize()
+   */
+PHP_METHOD(WriteConcern, bsonSerialize)
+{
+	const mongoc_write_concern_t *write_concern = phongo_write_concern_from_zval(getThis() TSRMLS_CC);
```
jmikola: Same changes as `ReadConcern::bsonSerialize()`.

src/MongoDB/ReadPreference.c
```diff
@@ -162,6 +162,16 @@ PHP_METHOD(ReadPreference, getTagSets)
 }
 /* }}} */
 
+/* {{{ proto array ReadPreference::bsonSerialize()
+   */
+PHP_METHOD(ReadPreference, bsonSerialize)
+{
+	const mongoc_read_prefs_t *read_preference = phongo_read_preference_from_zval(getThis() TSRMLS_CC);
```
jmikola: Same changes as `ReadConcern::bsonSerialize()`.

php_phongo.c
derickr: I don't think that's a problem for initializers.... but a good point. But we do this in many other places too:

```
 858 php_phongo_server_description_type_t php_phongo_server_description_type(mongoc_server_description_t *sd)
 859 {
 860     const char* name = mongoc_server_description_type(sd);
 861     int i;
 862 
```
```
 872 void php_phongo_server_to_zval(zval *retval, mongoc_server_description_t *sd) /* {{{ */
 873 {
 874     mongoc_host_list_t *host = mongoc_server_description_host(sd);
 875     const bson_t       *is_master = mongoc_server_description_ismaster(sd);
 876     bson_iter_t         iter;
 877 
```

in the same file for example.

src/MongoDB/ReadConcern.c
derickr: These warnings are IMO superfluous. I have never noticed this being necessary, and I have never seen this in other extensions either.

bjori: If you configure phongo with `--enable-developer-flags` it'll add `-Wunused-parameter`, which would issue a warning for arguments that aren't used.

The PHP_METHOD() macro defines a c function that has all sorts of arguments, including `return_value_ptr` and `return_value_used` which are rarely used, which is why those macros meant to suppress them, so the flag can still be helpful for avoid accidentally declaring superfluous arguments in other functions

jmikola: It appears that `-Wno-unused-parameter` was introduced in https://github.com/mongodb/mongo-php-driver/commit/e3e3c9d552e52484f504c31a7cb0f496b437f76a#diff-788d457a20b110cc38e571dec9ddc68cR126, builds have been ignoring unused variables since then.

php_phongo.c
jmikola: Confirming that this doesn't appear to be a problem with my local builds.

php_phongo.c
```diff
@@ -962,12 +960,21 @@ bool php_phongo_read_preference_tags_are_valid(const bson_t *tags) /* {{{ */
...
 
 	array_init_size(retval, 2);
 
-	ADD_ASSOC_LONG_EX(retval, "mode", mongoc_read_prefs_get_mode(read_prefs));
+	switch (mode) {
+		case MONGOC_READ_PRIMARY: ADD_ASSOC_STRING(retval, "mode", "primary"); break;
+		case MONGOC_READ_PRIMARY_PREFERRED: ADD_ASSOC_STRING(retval, "mode", "primaryPreferred"); break;
+		case MONGOC_READ_SECONDARY: ADD_ASSOC_STRING(retval, "mode", "secondary"); break;
+		case MONGOC_READ_SECONDARY_PREFERRED: ADD_ASSOC_STRING(retval, "mode", "secondaryPreferred"); break;
+		case MONGOC_READ_NEAREST: ADD_ASSOC_STRING(retval, "mode", "nearest"); break;
```
jmikola: `add_assoc_string_ex()` requires a `char *` for the value string, so each of these will need `(char *)` casting. Alternatively, I suppose we can just add that in the macro definition.

tests/writeResult/writeresult-debug-001.phpt
```diff
@@ -63,15 +63,7 @@ object(MongoDB\Driver\WriteResult)#%d (%d) {
...
-  array(4) {
-    ["w"]=>
-    NULL
-    ["wmajority"]=>
-    bool(false)
-    ["wtimeout"]=>
-    int(0)
-    ["journal"]=>
-    NULL
+  array(%d) {
```
jmikola: Since you're expecting an empty array here, `array(0)` would be preferred.

jmikola:
LGTM. I'll add the `(char *)` casts to `ADD_ASSOC_STRING()` on my own after this is merged.

php_phongo.c
derickr: I *would* like to have a proper definitive answer on this somewhere. Can't be arsed to read the spec though :-)

src/MongoDB/ReadConcern.c
derickr: I'm fine with not having warnings for unused function arguments though. Catching locally undefined vars is important though.

PR #424, opened 2016-09-23
PHPC-359 and PHPC-801: RC, RP, and WC BSON serialization

jmikola:
https://jira.mongodb.org/browse/PHPC-359
https://jira.mongodb.org/browse/PHPC-801

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -823,6 +823,49 @@ void php_phongo_read_concern_to_zval(zval *retval, const mongoc_read_concern_t *
 	}
 } /* }}} */
 
+/* Prepare tagSets for BSON encoding by converting each array in the set to an
+ * object. This ensures that empty arrays will serialize as empty documents.
+ *
+ * php_phongo_read_preference_tags_are_valid() handles actual validation of the
+ * tag set structure. */
+void php_phongo_read_preference_prep_tagsets(zval *tagSets TSRMLS_DC) /* {{{ */
+{
+	HashTable     *ht_data;
+
+	if (Z_TYPE_P(tagSets) != IS_ARRAY) {
+		return;
+	}
+
+	ht_data = HASH_OF(tagSets);
+
+#if PHP_VERSION_ID >= 70000
+	{
+		zval *tagSet;
+
+		ZEND_HASH_FOREACH_VAL(ht_data, tagSet) {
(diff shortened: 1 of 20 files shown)


php_phongo.c
```diff
@@ -864,10 +907,10 @@ void php_phongo_read_preference_to_zval(zval *retval, const mongoc_read_prefs_t
 	}
 
 	if (!bson_empty0(tags)) {
+		/* Use PHONGO_TYPEMAP_NATIVE_ARRAY for the root type since tags is an
+		 * array; however, inner documents and arrays can use the default. */
```
derickr: Can we be explicit about the document_type? i.e., by setting it to something like:

```
state.map.document_type = PHONGO_TYPEMAP_DEFAULT;
```

php_phongo.c
```diff
@@ -823,6 +823,49 @@ void php_phongo_read_concern_to_zval(zval *retval, const mongoc_read_concern_t *
...
+	}
+#else
+	{
+		HashPosition   pos;
+		zval         **tagSet;
+
+		for (zend_hash_internal_pointer_reset_ex(ht_data, &pos);
+		     zend_hash_get_current_data_ex(ht_data, (void **) &tagSet, &pos) == SUCCESS;
+		     zend_hash_move_forward_ex(ht_data, &pos)) {
+			if (Z_TYPE_PP(tagSet) == IS_ARRAY) {
```
derickr: Looks like whitespace issues here.

src/MongoDB/Manager.c
```diff
@@ -100,6 +100,62 @@ static bool php_phongo_manager_merge_context_options(zval *zdriverOptions TSRMLS
 	return true;
 }
 
+static void php_phongo_manager_prep_tagsets(zval *options TSRMLS_DC)
```
derickr: I'd like the comment that you used for RPs here too.

php_phongo.c
jmikola: `PHONGO_BSON_STATE_INITIALIZER` is already the default, which we use for the `state` intialization. This should suffice.

php_phongo.c
jmikola: This is correct. The `for` components are all indented five spaces and the `if` (and lines below) are indented with a tab, since they're within the `for` block.

PR #432, opened 2016-10-12
Update .gitmodules to use fork of libmongoc

stolenegg:


stolenegg:
created in error, apologies

PR #435, opened 2016-10-13
PHPC-788: Timestamp on 32-bit platforms breaks

derickr:


src/BSON/Timestamp.c
```diff
@@ -54,7 +54,7 @@ zend_object_handlers php_phongo_handler_timestamp;
 
 /* Initialize the object and return whether it was successful. An exception will
  * be thrown on error. */
-static bool php_phongo_timestamp_init(php_phongo_timestamp_t *intern, phongo_long increment, phongo_long timestamp TSRMLS_DC)
+static bool php_phongo_timestamp_init(php_phongo_timestamp_t *intern, int64_t increment, int64_t timestamp TSRMLS_DC)
```
derickr: This needs to be int64_t, as that's what the original arguments that were passed in here are as well. phongo_long is also not 64-bit on 32-bit platforms, and hence the comparison with UINT32_MIN would fail for some values.

src/BSON/Timestamp.c
```diff
@@ -413,8 +413,8 @@ HashTable *php_phongo_timestamp_get_properties(zval *object TSRMLS_DC) /* {{{ */
 		return props;
 	}
 
-	s_increment_len = snprintf(s_increment, sizeof(s_increment), "%" PRId64, intern->increment);
-	s_timestamp_len = snprintf(s_timestamp, sizeof(s_timestamp), "%" PRId64, intern->timestamp);
+	s_increment_len = snprintf(s_increment, sizeof(s_increment), "%" PRIu32, intern->increment);
+	s_timestamp_len = snprintf(s_timestamp, sizeof(s_timestamp), "%" PRIu32, intern->timestamp);
```
derickr: The value in "increment" and "timestamp" is 32bit unsigned, so using a 64bit formatting would add a random 32bit value to the start of each formatted string, resulting in test failures.

tests/bson/bson-timestamp-set_state_error-002.phpt
```diff
@@ -10,15 +10,15 @@ echo throws(function() {
 }, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
 
 echo throws(function() {
-    MongoDB\BSON\Timestamp::__set_state(['increment' => -2147483648, 'timestamp' => 5678]);
+    MongoDB\BSON\Timestamp::__set_state(['increment' => -2147483647, 'timestamp' => 5678]);
```
derickr: PHP's parser parses "-" and "2147483648" separately. And as "2147483648" is over the int-32 limit, it gets converted to a double. -2147483647 is therefore the highest 32bit number PHP itself can represent.

tests/bson/bug0592.phpt
```diff
@@ -44,15 +45,22 @@ object(stdClass)#%d (%d) {
   int(2147483647)
 }
 
+Test { "x": { "$numberLong": "4294967294" }}
+[%s] PHONGO-BSON: WARNING > Integer overflow detected on your platform: 4294967294
+object(stdClass)#%d (%d) {
+  ["x"]=>
+  string(10) "4294967294"
+}
+
```
derickr: This test had missing output sections, and the warning formatting was wrong too. This test was always broken.

src/BSON/Timestamp.c
jmikola: I assume you mean the `< 0` comparisons would fail?

Since `php_phongo_timestamp_init_from_string()` was always parsing the strings into `int64_t` values, I suppose there was possible degredation there. The other code path is `php_phongo_timestamp_init_from_hash()` picking up `IS_LONG` values. In that case, would a negative `int32_t` value expand to a negative `int64_t` value (essentially the sign bit relocates) and still yield an exception?

jmikola: I assume you mean the `< 0` comparisons would fail?

Since `php_phongo_timestamp_init_from_string()` was always parsing the strings into `int64_t` values, I suppose there was possible degredation there. The other code path is `php_phongo_timestamp_init_from_hash()` picking up `IS_LONG` values. In that case, would a negative `int32_t` value expand to a negative `int64_t` value (essentially the sign bit relocates) and still yield an exception?

src/BSON/Timestamp.c
```diff
@@ -66,8 +66,8 @@ static bool php_phongo_timestamp_init(php_phongo_timestamp_t *intern, phongo_lon
 		return false;
 	}
 
-	intern->increment = increment;
-	intern->timestamp = timestamp;
+	intern->increment = (uint32_t) increment;
+	intern->timestamp = (uint32_t) timestamp;
```
jmikola: Do we have to worry about endianness here?

tests/bson/bson-timestamp-set_state_error-002.phpt
jmikola: In that case, this isn't really testing a boundary condition, is it? I was testing `-2147483648` because it should have been equal to `4294967295` in unsigned form.

I'd be OK with remove these test cases unless you really want to keep them.

tests/bson/bug0592.phpt
```diff
@@ -32,6 +32,7 @@ foreach ($tests as $json) {
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
+%a
```
jmikola: What is being printed before the first test case? It may make more sense to simply enable and disable `mongodb.debug=stderr` around the `foreach` that executes all of the test cases, instead of having the INI setting on for the entire process. We really only want to check for the overflow warnings.

src/BSON/Timestamp.c
derickr: No.

src/BSON/Timestamp.c
derickr: Yes, the ```< 0``` would fail. And a sign bit doesn't change when you cast.

tests/bson/bug0592.phpt
derickr: I'll make the debug=stderr changes, but the other PR will just remove them again...

PR #454, opened 2016-11-07
PHPC-459: BSON objects should implement JsonSerializable

jmikola:
https://jira.mongodb.org/browse/PHPC-459

### DIFF
--- a/src/BSON/Binary.c
+++ b/src/BSON/Binary.c
@@ -31,6 +31,8 @@
 /* PHP Core stuff */
 #include <php.h>
 #include <php_ini.h>
+#include <ext/json/php_json.h>
+#include <ext/standard/base64.h>
 #include <ext/standard/info.h>
 #include <Zend/zend_interfaces.h>
 #include <ext/spl/spl_iterators.h>
@@ -187,6 +189,42 @@ PHP_METHOD(Binary, getType)
 }
 /* }}} */
 
+/* {{{ proto array Binary::jsonSerialize()
+*/
+PHP_METHOD(Binary, jsonSerialize)
+{
+	php_phongo_binary_t *intern;
+	char                 type[3];
+	int                  type_len;
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	intern = Z_BINARY_OBJ_P(getThis());
+
+	array_init_size(return_value, 2);
+
+#if PHP_VERSION_ID >= 70000
+	{
+		zend_string *data = php_base64_encode((unsigned char *)intern->data, intern->data_len);
+		ADD_ASSOC_STRINGL(return_value, "$binary", ZSTR_VAL(data), ZSTR_LEN(data));
+		zend_string_free(data);
+	}
+#else
+	{
+		int data_len = 0;
+		unsigned char *data = php_base64_encode((unsigned char *)intern->data, intern->data_len, &data_len);
+		ADD_ASSOC_STRINGL(return_value, "$binary", (char *)data, data_len);
+		efree(data);
+	}
+#endif
+
+	type_len = snprintf(type, sizeof(type), "%02x", intern->type);
+	ADD_ASSOC_STRINGL(return_value, "$type", type, type_len);
+}
+/* }}} */
+
 /* {{{ proto string Binary::serialize()
 */
 PHP_METHOD(Binary, serialize)
@@ -299,6 +337,7 @@ static zend_function_entry php_phongo_binary_me[] = {
 	PHP_ME(Binary, __construct, ai_Binary___construct, ZEND_ACC_PUBLIC|ZEND_ACC_FINAL)
 	PHP_ME(Binary, __set_state, ai_Binary___set_state, ZEND_ACC_PUBLIC|ZEND_ACC_STATIC)
 	PHP_ME(Binary, __toString, ai_Binary_void, ZEND_ACC_PUBLIC|ZEND_ACC_FINAL)
+	PHP_ME(Binary, jsonSerialize, ai_Binary_void, ZEND_ACC_PUBLIC|ZEND_ACC_FINAL)
 	PHP_ME(Binary, serialize, ai_Binary_void, ZEND_ACC_PUBLIC|ZEND_ACC_FINAL)
 	PHP_ME(Binary, unserialize, ai_Binary_unserialize, ZEND_ACC_PUBLIC|ZEND_ACC_FINAL)
 	PHP_ME(Binary, getData, ai_Binary_void, ZEND_ACC_PUBLIC|ZEND_ACC_FINAL)
@@ -399,6 +438,7 @@ PHP_MINIT_FUNCTION(Binary)
 	php_phongo_binary_ce->create_object = php_phongo_binary_create_object;
 	PHONGO_CE_FINAL(php_phongo_binary_ce);
 
+	zend_class_implements(php_phongo_binary_ce TSRMLS_CC, 1, php_json_serializable_ce);
 	zend_class_implements(php_phongo_binary_ce TSRMLS_CC, 1, php_phongo_type_ce);
 	zend_class_implements(php_phongo_binary_ce TSRMLS_CC, 1, zend_ce_serializable);
 
--- a/src/BSON/Decimal128.c
+++ b/src/BSON/Decimal128.c
@@ -31,6 +31,7 @@
 /* PHP Core stuff */
 #include <php.h>
 #include <php_ini.h>
+#include <ext/json/php_json.h>
 #include <ext/standard/info.h>
 #include <Zend/zend_interfaces.h>
 #include <ext/spl/spl_iterators.h>
@@ -151,6 +152,25 @@ PHP_METHOD(Decimal128, __toString)
 }
 /* }}} */
 
+/* {{{ proto array Decimal128::jsonSerialize()
+*/
+PHP_METHOD(Decimal128, jsonSerialize)
+{
+	php_phongo_decimal128_t *intern;
+	char                     outbuf[BSON_DECIMAL128_STRING] = "";
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	intern = Z_DECIMAL128_OBJ_P(getThis());
+
+	array_init_size(return_value, 1);
+	bson_decimal128_to_string(&intern->decimal, outbuf);
+	ADD_ASSOC_STRING(return_value, "$numberDecimal", outbuf);
+}
+/* }}} */
+
 /* {{{ proto string Decimal128::serialize()
 */
 PHP_METHOD(Decimal128, serialize)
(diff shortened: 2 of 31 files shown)


src/BSON/Javascript.c
```diff
@@ -215,6 +216,41 @@ PHP_METHOD(Javascript, getScope)
...
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	intern = Z_JAVASCRIPT_OBJ_P(getThis());
+
+	has_scope = (intern->scope && intern->scope->len);
+	array_init_size(return_value, has_scope ? 2 : 1);
+	ADD_ASSOC_STRINGL(return_value, "$code", intern->code, intern->code_len);
```
jmikola: This implementation is consistent with the current Extended JSON spec. As noted in the tests, libbson currently converts Javascript types to JSON by returning its code field as a string value (see: [CDRIVER-1335](https://jira.mongodb.org/browse/CDRIVER-1335)). It also does not properly parse `$code` and `$scope` fields when converting JSON to BSON, which causes the JSON we produce here to be read by libbson as a normal embedded document.

The [HHVM implementation](https://github.com/mongodb/mongo-hhvm-driver/pull/148) currently mimics libbson (see: [HHVM-156](https://jira.mongodb.org/browse/HHVM-158)); however, I think we should change its behavior to comply with the spec rather than libbson, which should itself be fixed in 1.6.0.

jmikola: I opened [HHVM-266](https://jira.mongodb.org/browse/HHVM-266) to track this.

src/BSON/Regex.c
```diff
@@ -206,6 +207,24 @@ PHP_METHOD(Regex, __toString)
...
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	intern = Z_REGEX_OBJ_P(getThis());
+
+	array_init_size(return_value, 2);
+	ADD_ASSOC_STRINGL(return_value, "$regex", intern->pattern, intern->pattern_len);
+	ADD_ASSOC_STRINGL(return_value, "$options", intern->flags, intern->flags_len);
```
jmikola: I left a comment in [DRIVERS-331](https://jira.mongodb.org/browse/DRIVERS-331?focusedCommentId=1428306&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1428306) asking whether we should alphabetically order the regex flag characters (as should be done for raw BSON to support binary comparisons). The implementation here leaves the flags as-is.

If we do start alphabetically ordering the characters in the future, we'll likely do so when the Regex object is initialized and this code shouldn't need to change.

src/BSON/Timestamp.c
```diff
@@ -214,6 +215,43 @@ PHP_METHOD(Timestamp, __toString)
...
+
+	array_init_size(return_value, 1);
+
+#if PHP_VERSION_ID >= 70000
+	{
+		zval ts;
+
+		array_init_size(&ts, 2);
+		ADD_ASSOC_LONG_EX(&ts, "t", intern->timestamp);
+		ADD_ASSOC_LONG_EX(&ts, "i", intern->increment);
```
jmikola: The Extended JSON spec currently states that a Timestamp should yield:

```
{"$timestamp": {"t": integer, "i": integer}}
```

We're still discussing whether these values should be single strings or perhaps `{"$numberLong": string}` objects. We cannot guarantee portability with integers, since the values may fall outside the range of 32-bit signed integers (both values are unsigned 32-bit integers).

jmikola: I've opened [PHPC-828](https://jira.mongodb.org/browse/PHPC-828) to track this.

src/BSON/UTCDateTime.c
```diff
@@ -275,6 +276,45 @@ PHP_METHOD(UTCDateTime, toDateTime)
...
+
+	array_init_size(return_value, 1);
+
+#if PHP_VERSION_ID >= 70000
+	{
+		zval udt;
+
+		array_init_size(&udt, 1);
+		ADD_ASSOC_STRINGL(&udt, "$numberLong", s_milliseconds, s_milliseconds_len);
+		ADD_ASSOC_ZVAL_EX(return_value, "$date", &udt);
```
jmikola: The Extended JSON spec currently states that a UTCDateTime should yield:

```
{"$date": {"$numberLong": "<ms since epoch as an integer>"}}

Alternative, optional format:

{"$date": {"$isoDate": <ISO-8601 date string in UTC as a string>}}
```

I'm opting to use the first format as ISO-8601 has issues representing the full range of dates supported by the BSON date type (as discussed during the previous driver team meeting). Additionally, the `$numberLong` format is much easier to generate.

HHVM currently yields the `$isoDate` format, which I think we should change. Discussion is ongoing in the spec about whether drivers should always yield `$numberLong` and optionally provide a `$isoDate` purely for user readability purposes (i.e. it will never be parsed for the authoritative value).

jmikola: I've opened [HHVM-267](https://jira.mongodb.org/browse/HHVM-267) to track this.

src/BSON/Regex.c
jmikola: The relevant PHPC ticket for flag ordering is [PHPC-829](https://jira.mongodb.org/browse/PHPC-829).

src/BSON/Javascript.c
derickr: Works for me, I'll make a PR for [HHVM-266](https://jira.mongodb.org/browse/HHVM-266) soon.

src/BSON/Regex.c
derickr: And HHVM is at [HHVM-264](https://jira.mongodb.org/browse/HHVM-264).

src/BSON/Timestamp.c
derickr: I think HHVM does the same format too:

```
        return [
            '$timestamp' => [
                't' => (int) $this->timestamp,
                'i' => (int) $this->increment,
            ]
        ];
```

So nothing we should change there - right?

tests/bson/bson-binary-jsonserialize-001.phpt
```diff
@@ -0,0 +1,19 @@
+--TEST--
+MongoDB\BSON\Binary::jsonSerialize() return value
```
derickr: Did you just copy all of my tests and added these ```==DONE==``` things for all of them?

src/BSON/Binary.c
```diff
@@ -401,6 +440,7 @@ PHP_MINIT_FUNCTION(Binary)
 
 	zend_class_implements(php_phongo_binary_ce TSRMLS_CC, 1, php_phongo_type_ce);
 	zend_class_implements(php_phongo_binary_ce TSRMLS_CC, 1, zend_ce_serializable);
+	zend_class_implements(php_phongo_binary_ce TSRMLS_CC, 1, php_json_serializable_ce);
```
derickr: The pedant in me would like this ordered alphabetically ;-)

src/BSON/Decimal128.c
```diff
@@ -151,6 +152,25 @@ PHP_METHOD(Decimal128, __toString)
 }
 /* }}} */
 
+/* {{{ proto array Decimal128::jsonSerialize()
+*/
+PHP_METHOD(Decimal128, jsonSerialize)
+{
+	php_phongo_decimal128_t *intern;
+	char                     outbuf[BSON_DECIMAL128_STRING] = "";
```
derickr: What's the point of initialising this with a empty string in the code segment?

src/BSON/Javascript.c
derickr: I am not sure what we should do here. In my opinion, we need to pick something now, and stick with it. Change the format would be an API change, and I'd be reluctant to change that before a 2.0 release.

src/BSON/Javascript.c
```diff
@@ -215,6 +216,41 @@ PHP_METHOD(Javascript, getScope)
...
+	bool has_scope;
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	intern = Z_JAVASCRIPT_OBJ_P(getThis());
+
+	has_scope = (intern->scope && intern->scope->len);
+	array_init_size(return_value, has_scope ? 2 : 1);
```
derickr: I wouldn't bother with the trinary here. Just make it 2 - it's merely a hint to the underlying algorithm anyway. And in that case, you can get rid of the ```has_scope``` variable, and just do the test in the ```if()``` below.

src/BSON/Regex.c
derickr: I think that we should order the flags in the constructor/initialiser only. That should be all what we need to do about this I believe.

The HHVM ticket is [HHVM-264](https://jira.mongodb.org/browse/HHVM-264).

src/BSON/Timestamp.c
derickr: Just like with JavaScript, we need to make a decision now. We won't be able to change this until 2.0 otherwise due to BC.

src/BSON/UTCDateTime.c
derickr: I disagree that we need to change this now, as we had the exact same discussion in the HHVM PR, where we decided to do it that way.

The team meeting notes only say:

> Our JsonSerializable output should be in a format that libbson and read and convert back into BSON. This means that we may not be able to use formats in the Extended JSON spec just yet.

tests/bson/bson-binary-jsonserialize-001.phpt
jmikola: These "return value" tests are new. The "json_encode()" tests are based on your own (with different code formatting and `===DONE===` blocks.

src/BSON/Regex.c
jmikola: > I think that we should order the flags in the constructor/initialiser only

I concur. Nothing to do here and we have our respective tickets.

src/BSON/Javascript.c
jmikola: Waiting on a response from the C driver team. I imagine they may keep their non-compliant output format (i.e. code as a string and ignoring scope) for BC until 2.0 but allow the `$code`/`$scope` format to be parsed properly when they address the ticket in 1.6.0.

I'm all in favor of sticking to the spec and returning a compliant value today, even if libbson cannot parse it yet.

src/BSON/Timestamp.c
jmikola: Will follow up with @llvtt in chat later and ask for this to be address in the spec ASAP.

src/BSON/Decimal128.c
jmikola: This was previously done in `php_phongo_decimal128_get_properties()` as insurance that the string was null-terminated even if `bson_decimal128_to_string()` decided not to touch it. Based on my reading of `bson_decimal128_to_string()`, does always assign a null-terminated string to its output buffer, but it wasn't documented as doing so and I wanted to be defensive. The alternative was a `memset()` on the array, but I didn't think that was necessary.

`__toString()` and `serialize()` currently use `char outbuf[BSON_DECIMAL128_STRING]` without any assignment, but I wanted to update those later.

src/BSON/Javascript.c
jmikola: I'll note that libbson cannot currently roundtrip Javascript types anyway (they'd be parsed back as a BSON string), so there is no risk by sticking with the proper `$code`/`$scope` format.

src/BSON/Timestamp.c
jmikola: After speaking with @llvtt, it's clear that the spec is not far enough along to make an assumption about the Timestamp format.

@derickr: Would you rather we throw an exception for out-of-range values on 32-bit systems or instead allow the unsigned values to be printed in their signed representation? I can re-purpose [PHPC-828](https://jira.mongodb.org/browse/PHPC-828) if necessary.

derickr: I don't think an exception is great here, as that means that some random data can suddenly abort a script. I think for now, showing their signed representation is "good enough". We can fix that, if the ext-json spec has been finalised.

src/BSON/UTCDateTime.c
derickr: OK.

tests/bson/bson-binary-jsonserialize-001.phpt
derickr: OK

PR #455, opened 2016-11-07
PHPC-718: Define extension dependencies

jmikola:
https://jira.mongodb.org/browse/PHPC-718

Note: https://github.com/mongodb/mongo-php-driver/pull/454 (for [PHPC-459](https://jira.mongodb.org/browse/PHPC-459)) adds a dependency on the "json" extension.

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -523,7 +523,10 @@ fi
 dnl }}}
 
   PHP_NEW_EXTENSION(mongodb,    $MONGODB_ROOT, $ext_shared,, [$STD_CFLAGS $MAINTAINER_CFLAGS $COVERAGE_CFLAGS])
+  PHP_ADD_EXTENSION_DEP(mongodb, date)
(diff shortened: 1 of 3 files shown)


jmikola:
> LGTM, but shouldn't it rather be the other way around? PR 454 depending on this one - with this change happening first?

Perhaps, although it doesn't make much difference. Seemed odd to list the dep before we introduce code to use it.

Is it worth scouting through our headers to see if there are other extensions we use? I know that we depend on class entries for date, json, and spl, and there are a fair number of standard functions that we use, but I didn't do an exhaustive search.

jmikola:
> Is it worth scouting through our headers to see if there are other extensions we use? I know that we depend on class entries for date, json, and spl, and there are a fair number of standard functions that we use, but I didn't do an exhaustive search.

Just confirmed that all date, json, spl, and standard are the only extensions with headers that we include.

derickr:
LGTM

PR #463, opened 2016-11-21
PHPC-829: BSON Regex flags must be alphabetically ordered

jmikola:
https://jira.mongodb.org/browse/PHPC-829

We do not test that MongoDB\BSON\fromJSON() alphabetizes the flags, as that will be handled by CDRIVER-1883.


derickr:
LGTM

PR #466, opened 2016-11-22
PHPC-349: Implement APM specification

derickr:


php_phongo.c
```diff
@@ -1355,6 +1356,203 @@ static void php_phongo_free_ssl_opt(mongoc_ssl_opt_t *ssl_opt)
...
+		 * constant names, but not with parameterized ones as it does
+		 * "sizeof(char*)" in that case. */
+		zend_call_method(value, NULL, NULL, name, strlen(name), NULL, 1, z_event, NULL TSRMLS_CC);
+	}
+#endif
+}
+
+static void php_phongo_command_started(const mongoc_apm_command_started_t *event)
+{
+	mongoc_client_t *client = mongoc_apm_command_started_get_context(event);
```
jmikola: Should function calls and statements appear after variable declarations in the same scope? IIRC, this is why `TSRMLS_FETCH()` can't precede declarations.

derickr: Not needed as far as I know. Our compile flags also create a warning if we do this wrong (through ```-Werror=declaration-after-statement```).

From what I understand, you can define a variable and initialise this, but C90 does not allow to have "just code" before further variable declarations/initialisations.

src/MongoDB/Monitoring.c
```diff
@@ -0,0 +1,99 @@
...
+	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "O", &zSubscriber, php_phongo_subscriber_ce) == FAILURE) {
+		return;
+	}
+
+	hash = php_phongo_make_subscriber_hash(zSubscriber TSRMLS_CC);
+
+#if PHP_VERSION_ID >= 70000
+	zend_hash_str_del(&MONGODB_G(subscribers), hash, strlen(hash));
+#else
+	zend_hash_del(&MONGODB_G(subscribers), hash, strlen(hash));
```
jmikola: If `zend_hash_update()` doesn't add a reference, are you sure that this removes a reference on the zval after deleting it from the hash?

jmikola: I suppose if there was a leak, the `removeSubscriber()` tests would catch it. I found no issues running Valgrind locally for 5.6-zts and 7.1.

jmikola:
Most of the minor comments (e.g. formatting, alphabetizing) are addressed in https://github.com/mongodb/mongo-php-driver/pull/568.

config.m4
```diff
@@ -192,6 +192,12 @@ if test "$MONGODB" != "no"; then
     src/MongoDB/Exception/SSLConnectionException.c \
     src/MongoDB/Exception/UnexpectedValueException.c \
     src/MongoDB/Exception/WriteException.c \
+    src/MongoDB/Monitoring.c \
+    src/MongoDB/Monitoring/Subscriber.c \
+    src/MongoDB/Monitoring/CommandSubscriber.c \
+    src/MongoDB/Monitoring/CommandStartedEvent.c \
+    src/MongoDB/Monitoring/CommandSucceededEvent.c \
+    src/MongoDB/Monitoring/CommandFailedEvent.c
```
derickr: Yes - although, they're now ordered logically.

config.w32
```diff
@@ -85,8 +85,9 @@ if (PHP_MONGODB != "no") {
...
-  ADD_SOURCES(configure_module_dirname + "/src/MongoDB", "BulkWrite.c Command.c Cursor.c CursorId.c Manager.c Query.c ReadConcern.c ReadPreference.c Server.c WriteConcern.c WriteConcernError.c WriteError.c WriteResult.c", "mongodb");
+  ADD_SOURCES(configure_module_dirname + "/src/MongoDB", "BulkWrite.c Command.c Cursor.c CursorId.c Manager.c Monitoring.c Query.c ReadConcern.c ReadPreference.c Server.c WriteConcern.c WriteConcernError.c WriteError.c WriteResult.c", "mongodb");
   ADD_SOURCES(configure_module_dirname + "/src/MongoDB/Exception", "AuthenticationException.c BulkWriteException.c ConnectionException.c ConnectionTimeoutException.c Exception.c ExecutionTimeoutException.c InvalidArgumentException.c LogicException.c RuntimeException.c SSLConnectionException.c UnexpectedValueException.c WriteException.c", "mongodb");
+  ADD_SOURCES(configure_module_dirname + "/src/MongoDB/Monitoring", "Subscriber.c CommandSubscriber.c CommandStartedEvent.c CommandSucceededEvent.c CommandFailedEvent.c", "mongodb");
```
derickr: Yes, although they're now ordered logically.

php_phongo.c
```diff
@@ -1355,6 +1356,203 @@ static void php_phongo_free_ssl_opt(mongoc_ssl_opt_t *ssl_opt)
...
+#if PHP_VERSION_ID >= 70000
+	object_init_ex(&z_event, php_phongo_commandstartedevent_ce);
+	p_event = Z_COMMANDSTARTEDEVENT_OBJ_P(&z_event);
+#else
+	MAKE_STD_ZVAL(z_event);
+	object_init_ex(z_event, php_phongo_commandstartedevent_ce);
+	p_event = Z_COMMANDSTARTEDEVENT_OBJ_P(z_event);
+#endif
+
+	p_event->client = client;
```
derickr: I suppose so :-)

php_phongo.c
```diff
@@ -1355,6 +1356,203 @@ static void php_phongo_free_ssl_opt(mongoc_ssl_opt_t *ssl_opt)
...
+	php_phongo_dispatch_handlers("commandStarted", &z_event);
+#else
+	php_phongo_dispatch_handlers("commandStarted", z_event);
+#endif
+	zval_ptr_dtor(&z_event);
+}
+
+static void php_phongo_command_succeeded(const mongoc_apm_command_succeeded_t *event)
+{
+	mongoc_client_t *client = mongoc_apm_command_succeeded_get_context(event);
```
derickr: OK

php_phongo.c
```diff
@@ -1355,6 +1356,203 @@ static void php_phongo_free_ssl_opt(mongoc_ssl_opt_t *ssl_opt)
...
+	php_phongo_dispatch_handlers("commandSucceeded", &z_event);
+#else
+	php_phongo_dispatch_handlers("commandSucceeded", z_event);
+#endif
+	zval_ptr_dtor(&z_event);
+}
+
+static void php_phongo_command_failed(const mongoc_apm_command_failed_t *event)
+{
+	mongoc_client_t *client = mongoc_apm_command_failed_get_context(event);
```
derickr: OK

php_phongo.c
```diff
@@ -1355,6 +1356,203 @@ static void php_phongo_free_ssl_opt(mongoc_ssl_opt_t *ssl_opt)
...
+	{
+#if PHP_VERSION_ID < 70000
+		MAKE_STD_ZVAL(p_event->z_error);
+		object_init_ex(p_event->z_error, phongo_exception_from_mongoc_domain(tmp_error.domain, tmp_error.code));
+		zend_update_property_string(default_exception_ce, p_event->z_error, "message", sizeof("message")-1, tmp_error.message TSRMLS_CC);
+		zend_update_property_long(default_exception_ce, p_event->z_error, "code", sizeof("code")-1, tmp_error.code TSRMLS_CC);
+#else
+		object_init_ex(&p_event->z_error, phongo_exception_from_mongoc_domain(tmp_error.domain, tmp_error.code));
+		zend_update_property_string(default_exception_ce, &p_event->z_error, "message", sizeof("message")-1, tmp_error.message TSRMLS_CC);
+		zend_update_property_long(default_exception_ce, &p_event->z_error, "code", sizeof("code")-1, tmp_error.code TSRMLS_CC);
```
derickr: I copied this from somewhere else though? Perhaps that was before you fixed things up.

The whole point of making the exception here, is that it gets the right trace information. THe original in HHVM showed that it needed to be in this spot as well. Sadly, the overview.phpt test no longer shows it, as you wanted that caught by a ``%a``. We should manually check this.

php_phongo.c
```diff
@@ -1825,6 +2023,16 @@ static void php_phongo_pclient_dtor(void *pp)
 }
 #endif
 
+/* {{{ PHP_RINIT_FUNCTION */
+PHP_RINIT_FUNCTION(mongodb)
+{
+	/* Initialize HashTable for persistent clients */
```
derickr: OK

php_phongo.c
```diff
@@ -1961,6 +2170,12 @@ PHP_MINIT_FUNCTION(mongodb)
 	php_phongo_sslconnectionexception_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_unexpectedvalueexception_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 
+	php_phongo_subscriber_init_ce(INIT_FUNC_ARGS_PASSTHRU);
+	php_phongo_commandsubscriber_init_ce(INIT_FUNC_ARGS_PASSTHRU);
+	php_phongo_commandstartedevent_init_ce(INIT_FUNC_ARGS_PASSTHRU);
+	php_phongo_commandsucceededevent_init_ce(INIT_FUNC_ARGS_PASSTHRU);
+	php_phongo_commandfailedevent_init_ce(INIT_FUNC_ARGS_PASSTHRU);
```
derickr: OK

src/MongoDB/Manager.c
```diff
@@ -29,6 +29,8 @@
 
 #define PHONGO_MANAGER_URI_DEFAULT "mongodb://127.0.0.1/"
 
+ZEND_EXTERN_MODULE_GLOBALS(mongodb)
```
derickr: It was definitely needed when I wrote the code. If any code in this file uses MONGODB_G, it needs to be there. Seems that we removed all of that code though.

src/MongoDB/Manager.c
```diff
@@ -191,6 +195,10 @@ static PHP_METHOD(Manager, __construct)
 	}
 
 	phongo_manager_init(intern, uri_string ? uri_string : PHONGO_MANAGER_URI_DEFAULT, options, driverOptions TSRMLS_CC);
+
+	if (intern->client) {
+		php_phongo_set_monitoring_callbacks(intern->client);
+	}
```
derickr: Just as my comment on the HHVM PR, in earlier discussions we came to the conclusion that we didn't want to remove the apm callbacks at request shutdown, and to be safe, always set them upon Manager creation.

src/MongoDB/Manager.c
```diff
@@ -385,6 +393,34 @@ static PHP_METHOD(Manager, selectServer)
...
+{
+	SUPPRESS_UNUSED_WARNING(return_value_ptr) SUPPRESS_UNUSED_WARNING(return_value_used) SUPPRESS_UNUSED_WARNING(return_value) SUPPRESS_UNUSED_WARNING(this_ptr)
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	phongo_throw_exception(PHONGO_ERROR_RUNTIME TSRMLS_CC, "%s", "MongoDB\\Driver objects cannot be serialized");
+}
+/* }}} */
```
derickr: It's not new code, probably comes from a mess while rebasing.

src/MongoDB/Manager.c
```diff
@@ -385,6 +393,34 @@ static PHP_METHOD(Manager, selectServer)
...
+ * Typically, users will connect to a cluster using a URI, and the Manager will
+ * perform tasks such as replica set discovery and create the necessary Server
+ * objects. That said, it is also possible to create a Manager with an arbitrary
+ * collection of Server objects using the static factory method (this can be
+ * useful for testing or administration).
+ *
+ * Operation methods do not take socket-level options (e.g. socketTimeoutMS).
+ * Those options should be specified during construction.
+ */
+/* {{{ MongoDB\Driver\Manager */
```
derickr: It's not new code, probably comes from a mess while rebasing.

src/MongoDB/Monitoring.c
derickr: No, as the hash dtor uses zval_ptr_dtor on the element, it shouldn't be needed. AFAIK, the dtor decreases the refcount and frees when the refcount hits 0.

config.m4
jmikola: I did keep the logical order when initializing the class entries in #568.

php_phongo.c
jmikola: Would it make sense to create a tracking ticket to implement such a test later?

src/MongoDB/Manager.c
jmikola: In that case, I think we can still move this to `phongo_manager_init()` just to keep this logic out of the constructor. I'd like to revisit down the line with some testing over multiple requests to see if re-assignment is really necessary. Since the callbacks are fixed functions, I have a feeling this has no effect.

php_phongo.c
derickr: Yes

PR #467, opened 2016-11-22
PHPC-790: Accept integer and floats in Timestamp and UTCDateTime constructors

jmikola:
https://jira.mongodb.org/browse/PHPC-790
https://jira.mongodb.org/browse/PHPC-804

### DIFF
--- a/src/BSON/Timestamp.c
+++ b/src/BSON/Timestamp.c
@@ -140,28 +140,48 @@ static bool php_phongo_timestamp_init_from_hash(php_phongo_timestamp_t *intern, 
 	return false;
 }
 
-/* {{{ proto void Timestamp::__construct(string $increment, string $timestamp)
+/* {{{ proto void Timestamp::__construct(int|string $increment, int|string $timestamp)
    Construct a new BSON timestamp type, which consists of a 4-byte increment and
    4-byte timestamp. */
 PHP_METHOD(Timestamp, __construct)
 {
-	php_phongo_timestamp_t    *intern;
-	zend_error_handling       error_handling;
-	char                      *s_increment;
-	phongo_zpp_char_len       s_increment_len;
-	char                      *s_timestamp;
-	phongo_zpp_char_len       s_timestamp_len;
+	php_phongo_timestamp_t *intern;
+	zend_error_handling     error_handling;
+	zval                   *increment = NULL, *timestamp = NULL;
 
 	zend_replace_error_handling(EH_THROW, phongo_exception_from_phongo_domain(PHONGO_ERROR_INVALID_ARGUMENT), &error_handling TSRMLS_CC);
 	intern = Z_TIMESTAMP_OBJ_P(getThis());
 
-	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "ss", &s_increment, &s_increment_len, &s_timestamp, &s_timestamp_len) == FAILURE) {
+	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "zz", &increment, &timestamp) == FAILURE) {
 		zend_restore_error_handling(&error_handling TSRMLS_CC);
 		return;
 	}
 	zend_restore_error_handling(&error_handling TSRMLS_CC);
 
-	php_phongo_timestamp_init_from_string(intern, s_increment, s_increment_len, s_timestamp, s_timestamp_len TSRMLS_CC);
+	if (Z_TYPE_P(increment) == IS_LONG && Z_TYPE_P(timestamp) == IS_LONG) {
+		php_phongo_timestamp_init(intern, Z_LVAL_P(increment), Z_LVAL_P(timestamp) TSRMLS_CC);
+		return;
+	}
+
+	if (Z_TYPE_P(increment) == IS_LONG) {
+		convert_to_string(increment);
+	}
+
+	if (Z_TYPE_P(increment) != IS_STRING) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected increment to be an unsigned 32-bit integer or string, %s given", zend_get_type_by_const(Z_TYPE_P(increment)));
+		return;
+	}
+
+	if (Z_TYPE_P(timestamp) == IS_LONG) {
+		convert_to_string(timestamp);
+	}
+
+	if (Z_TYPE_P(timestamp) != IS_STRING) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected timestamp to be an unsigned 32-bit integer or string, %s given", zend_get_type_by_const(Z_TYPE_P(timestamp)));
+		return;
+	}
+
+	php_phongo_timestamp_init_from_string(intern, Z_STRVAL_P(increment), Z_STRLEN_P(increment), Z_STRVAL_P(timestamp), Z_STRLEN_P(timestamp) TSRMLS_CC);
 }
 /* }}} */
 
--- a/src/BSON/Timestamp.c
+++ b/src/BSON/Timestamp.c
@@ -91,22 +91,14 @@ static bool php_phongo_timestamp_init_from_string(php_phongo_timestamp_t *intern
 	 * available on all platforms (e.g. HP-UX), and atoll() provides no error
 	 * reporting at all. */
 
-#if defined(PHP_WIN32)
-	increment = _atoi64(s_increment);
-#else
 	increment = bson_ascii_strtoll(s_increment, &endptr, 10);
-#endif
 
 	if (errno || (endptr && endptr != ((const char *)s_increment + s_increment_len))) {
 		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error parsing \"%s\" as 64-bit integer increment for %s initialization", s_increment, ZSTR_VAL(php_phongo_timestamp_ce->name));
 		return false;
 	}
 
-#if defined(PHP_WIN32)
-	timestamp = _atoi64(s_timestamp);
-#else
 	timestamp = bson_ascii_strtoll(s_timestamp, &endptr, 10);
-#endif
 
 	if (errno || (endptr && endptr != ((const char *)s_timestamp + s_timestamp_len))) {
 		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error parsing \"%s\" as 64-bit integer timestamp for %s initialization", s_timestamp, ZSTR_VAL(php_phongo_timestamp_ce->name));
--- a/src/BSON/UTCDateTime.c
+++ b/src/BSON/UTCDateTime.c
@@ -77,11 +77,7 @@ static bool php_phongo_utcdatetime_init_from_string(php_phongo_utcdatetime_t *in
 
 	errno = 0;
 
-#if defined(PHP_WIN32)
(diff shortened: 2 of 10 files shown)


src/BSON/Timestamp.c
```diff
@@ -148,28 +140,48 @@ static bool php_phongo_timestamp_init_from_hash(php_phongo_timestamp_t *intern,
...
+		php_phongo_timestamp_init(intern, Z_LVAL_P(increment), Z_LVAL_P(timestamp) TSRMLS_CC);
+		return;
+	}
+
+	if (Z_TYPE_P(increment) == IS_LONG) {
+		convert_to_string(increment);
+	}
+
+	if (Z_TYPE_P(increment) != IS_STRING) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected increment to be integer or string, %s given", zend_get_type_by_const(Z_TYPE_P(increment)));
```
jmikola: HHVM will throw "Expected timestamp to be an unsigned 32-bit integer or string" on type errors (see: [here](https://github.com/mongodb/mongo-hhvm-driver/blob/d36a92128a782ff2b678127ff745bcbfc78a05a9/ext_mongodb.php#L1640)). I'm happy to submit a PR to revise that once this is approved.

src/BSON/UTCDateTime.c
```diff
@@ -161,49 +157,58 @@ static bool php_phongo_utcdatetime_init_from_date(php_phongo_utcdatetime_t *inte
...
 
-		php_phongo_utcdatetime_init_from_string(intern, s_milliseconds, s_milliseconds_len TSRMLS_CC);
+	if (Z_TYPE_P(milliseconds) == IS_LONG) {
+		php_phongo_utcdatetime_init(intern, Z_LVAL_P(milliseconds));
+		return;
 	}
 
-	zend_restore_error_handling(&error_handling TSRMLS_CC);
+	if (Z_TYPE_P(milliseconds) != IS_STRING) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected integer or string, %s given", zend_get_type_by_const(Z_TYPE_P(milliseconds)));
```
jmikola: HHVM only throws "Expected instance of DateTimeInterface" and "Error parsing %s as 64-bit integer" exceptions, so it may need a PR for this (see [here](https://github.com/mongodb/mongo-hhvm-driver/blob/d36a92128a782ff2b678127ff745bcbfc78a05a9/ext_mongodb.php#L1721)).

tests/bson/bson-timestamp-set_state_error-004.phpt
```diff
@@ -8,19 +8,19 @@ MongoDB\BSON\Timestamp::__set_state() with broken numeric strings
 require_once __DIR__ . '/../utils/tools.php';
 
 echo throws(function() {
-    MongoDB\BSON\Timestamp::__set_state(['increment' => 'broken', 'timestamp' => '5678']);
+    MongoDB\BSON\Timestamp::__set_state(['increment' => '1.23', 'timestamp' => '5678']);
```
jmikola: This test would have been problematic for Windows since `_atoi64()` stops parsing at the first invalid character without raising an error; however, `bson_ascii_strtoll()` is more strict.

tests/bson/bson-timestamp_error-006.phpt
```diff
@@ -0,0 +1,46 @@
...
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected increment to be integer or string, %r(null|NULL)%r given
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected increment to be integer or string, %r(double|float)%r given
```
jmikola: Pattern is necessary for PHP 5 and 7.

tests/bson/bson-timestamp_error-006.phpt
```diff
@@ -0,0 +1,46 @@
...
+        new MongoDB\BSON\Timestamp(0, $invalidValue);
+    }, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected increment to be integer or string, %r(null|NULL)%r given
```
jmikola: Pattern is necessary for PHP and HHVM.

tests/bson/bson-timestamp-set_state_error-004.phpt
```diff
@@ -1,26 +1,24 @@
 --TEST--
-MongoDB\BSON\Timestamp::__set_state() with broken numeric strings
---SKIPIF--
-<?php if (8 !== PHP_INT_SIZE) { die('skip Only for 64-bit platform'); } ?>
```
jmikola: I believe this was erroneously copy-pasted when the test was created in d25e565f647d824ced02146bd41d8ef17ff93cb9.

src/BSON/Timestamp.c
derickr: But the HHVM warning is not wrong? It has to be an unsigned 32-bit integer. I don't see you test for this here, but it's in php_phongo_timestamp_init I believe.

src/BSON/UTCDateTime.c
```diff
@@ -161,49 +157,58 @@ static bool php_phongo_utcdatetime_init_from_date(php_phongo_utcdatetime_t *inte
...
 
-	{
-		char                *s_milliseconds;
-		phongo_zpp_char_len  s_milliseconds_len;
-
-		if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "s", &s_milliseconds, &s_milliseconds_len) == FAILURE) {
-			zend_restore_error_handling(&error_handling TSRMLS_CC);
-			return;
-		}
+	if (Z_TYPE_P(milliseconds) == IS_DOUBLE) {
```
derickr: This could cause data loss, and I'm not keen about that :-/

src/BSON/UTCDateTime.c
derickr: I think the HHVM message is much clearer as an API point of view. The "error parsing x as 64-bit integer" is what people are interested in. Whether it's an "integer", "float", or "string" is irrelevant. The current message in your code also misses "float".

tests/bson/bson-utcdatetime-serialization_error-002.phpt
```diff
@@ -6,7 +6,7 @@ MongoDB\BSON\UTCDateTime unserialization requires "milliseconds" string to parse
 require_once __DIR__ . '/../utils/tools.php';
 
 echo throws(function() {
-    unserialize('C:24:"MongoDB\BSON\UTCDateTime":40:{a:1:{s:12:"milliseconds";s:7:"INVALID";}}');
+    unserialize('C:24:"MongoDB\BSON\UTCDateTime":42:{a:1:{s:12:"milliseconds";s:9:"1234.5678";}}');
```
derickr: I would have kept  the length of the argument the same, as to not have to mess with lengths in hardcoded serialized strings. But that's done now.

src/BSON/UTCDateTime.c
jmikola: I omitted "float" from the message since the cast to a long is just for the user's convenience. I wasn't planning on documenting that we accept a float, since it is being truncated. This is mainly to skirt around scalar type hint errors.

In HHVM's case, the invalid argument is cast to a string value for the exception message. That leads to some odd output for null (empty string), boolean ("1" and "0"), and would likely emit some notice on an array. The type error I was using here was consistent with what we already did for ReadPreference, WriteConcern, etc.

src/BSON/Timestamp.c
jmikola: `php_phongo_timestamp_init()` performs range checking and will throw the "Expected increment to be an unsigned 32-bit integer, %d given" exceptions. I can change this to "Expected timestamp to be an unsigned 32-bit integer or string" and remove the "%s given" bit. Sharing the unexpected type was for consistency with our other exceptions for type errors.

src/BSON/UTCDateTime.c
jmikola: I admittedly has some trouble reading through all of the `zend_dval_to_lval()` cases to see if this would be an issue. Would you agree that the best solution is to convert this to a string via `"%f.0"`?

src/BSON/Timestamp.c
derickr: I think that for HHVM, we should add the "%s given" bit, and you change the message to "Expected increment to be an unsigned 32-bit integer or string, %s given" — does that make sense? (note, in your reply you used **timestamp** and not **increment**).

src/BSON/UTCDateTime.c
derickr: No, I don't think that's needed. It's just that we ignore a user's sub-milliseconds, because they passed in a float, without notifying them.

src/BSON/UTCDateTime.c
derickr: OK - add some tests for these additional types, so I can test HHVM against it as well? (If you haven't done so)

src/BSON/Timestamp.c
jmikola: Works for me. The timestamp/increment was indeed a typo on my part. I'll take care of the HHVM PR.

src/BSON/UTCDateTime.c
jmikola: If this isn't a matter of stuffing a double into a 32-bit integer, I don't understand the concern. UTCDateTime is documented as only representing milliseconds and the purpose of this change was because users previously had to use `microtime(true) * 1000` to create a UTCDateTime with the current time. This is no different than a userland method using an `(int)` cast to ensure it's working with an integer (before scalar type hinting was available).

That said, I think the 32-bit issue is still a concern worth addressing. `microtime(true) * 1000` will easily result in values that exceed `INT32_MAX`.

derickr: Yes, I agree. That should be handled.

src/BSON/UTCDateTime.c
```diff
@@ -161,49 +158,64 @@ static bool php_phongo_utcdatetime_init_from_date(php_phongo_utcdatetime_t *inte
...
-		phongo_zpp_char_len  s_milliseconds_len;
+	if (Z_TYPE_P(milliseconds) == IS_DOUBLE) {
+		char tmp[24];
+		int tmp_len;
 
-		if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "s", &s_milliseconds, &s_milliseconds_len) == FAILURE) {
-			zend_restore_error_handling(&error_handling TSRMLS_CC);
-			return;
-		}
+		tmp_len = snprintf(tmp, sizeof(tmp), "%.0f", Z_DVAL_P(milliseconds) > 0 ? floor(Z_DVAL_P(milliseconds)) : ceil(Z_DVAL_P(milliseconds)));
```
jmikola: `trunc()` is available in C99, but I didn't want to risk any Windows build issues for older PHP issues.

derickr: Wouldn't it make more sense to convert to a long straight away here?

jmikola: I originally used `convert_to_long()`, but you mentioned it would lose data. Although you were referring to precision after milliseconds, which I later explained we don't care about, that solution did have issues on 32-bit platforms. This method ensures that values larger than INT32_MAX in a float will convert properly on 32-bit.

PR #472, opened 2016-11-28
PHPC-635: WriteResult debug handler should return objects

jmikola:
https://jira.mongodb.org/browse/PHPC-635

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -354,35 +354,35 @@ void phongo_writeconcern_init(zval *return_value, const mongoc_write_concern_t *
 zend_bool phongo_writeconcernerror_init(zval *return_value, bson_t *bson TSRMLS_DC) /* {{{ */
 {
 	bson_iter_t iter;
-	php_phongo_writeconcernerror_t *writeconcernerror;
+	php_phongo_writeconcernerror_t *intern;
 
-	writeconcernerror = Z_WRITECONCERNERROR_OBJ_P(return_value);
+	object_init_ex(return_value, php_phongo_writeconcernerror_ce);
+
+	intern = Z_WRITECONCERNERROR_OBJ_P(return_value);
 
 	if (bson_iter_init_find(&iter, bson, "code") && BSON_ITER_HOLDS_INT32(&iter)) {
-		writeconcernerror->code = bson_iter_int32(&iter);
+		intern->code = bson_iter_int32(&iter);
 	}
+
 	if (bson_iter_init_find(&iter, bson, "errmsg") && BSON_ITER_HOLDS_UTF8(&iter)) {
 		uint32_t errmsg_len;
 		const char *err_msg = bson_iter_utf8(&iter, &errmsg_len);
 
-		writeconcernerror->message = estrndup(err_msg, errmsg_len);
+		intern->message = estrndup(err_msg, errmsg_len);
 	}
+
 	if (bson_iter_init_find(&iter, bson, "errInfo") && BSON_ITER_HOLDS_DOCUMENT(&iter)) {
-		uint32_t               len;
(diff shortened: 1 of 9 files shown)


php_phongo.c
```diff
@@ -395,38 +395,43 @@ zend_bool phongo_writeconcernerror_init(zval *return_value, bson_t *bson TSRMLS_
...
-	if (bson_iter_init_find(&iter, bson, "errInfo")) {
-		bson_t                 info;
 
-		bson_init(&info);
-		bson_append_iter(&info, NULL, 0, &iter);
+	if (bson_iter_init_find(&iter, bson, "errInfo") && BSON_ITER_HOLDS_DOCUMENT(&iter)) {
+		uint32_t       len;
+		const uint8_t *data = NULL;
+
+		bson_iter_document(&iter, &len, &data);
```
jmikola: `BSON_ITER_HOLDS_DOCUMENT()` and `bson_iter_document()` is consistent with what we were already doing in `phongo_writeconcernerror_init()` for a very similar routine. This replaces the previous `bson_append_iter()` code path, which is much less common.

php_phongo.c
```diff
@@ -354,35 +354,35 @@ void phongo_writeconcern_init(zval *return_value, const mongoc_write_concern_t *
...
-		uint32_t               len;
-		const uint8_t         *data;
+		uint32_t       len;
+		const uint8_t *data = NULL;
 
 		bson_iter_document(&iter, &len, &data);
 
-		if (!data) {
-			return false;
-		}
```
jmikola: With `BSON_ITER_HOLDS_DOCUMENT()` guarding this block, `bson_iter_document()` will always populate the document's data. This removes an odd case where we might have returned false without throwing. If the data is somehow null after this (I can't imagine how), `phongo_bson_to_zval()` will catch and throw an error when initializing the BSON reader.

src/MongoDB/WriteResult.c
```diff
@@ -50,6 +50,108 @@ PHONGO_API zend_class_entry *php_phongo_writeresult_ce;
...
+			ZVAL_ZVAL(return_value, &writeconcernerror, 1, 1);
+#else
+			MAKE_STD_ZVAL(writeconcernerror);
+
+			if (!phongo_writeconcernerror_init(writeconcernerror, &cbson TSRMLS_CC)) {
+				zval_ptr_dtor(&writeconcernerror);
+				return false;
+			}
+
+			ZVAL_ZVAL(return_value, writeconcernerror, 1, 1);
```
jmikola: I ran into trouble trying to use `zval_ptr_dtor()` on the input argument so it seemed more straightforward to create my own zval in this function and copy it into the return value on success. For any failure case, we clean up our internally allocated zval and leave the return value initialized to null.

src/MongoDB/WriteResult.c
```diff
@@ -50,6 +50,108 @@ PHONGO_API zend_class_entry *php_phongo_writeresult_ce;
 
 zend_object_handlers php_phongo_handler_writeresult;
 
+static bool php_phongo_writeresult_get_writeconcernerror(php_phongo_writeresult_t *intern, zval *return_value TSRMLS_DC)
```
jmikola: Although these two functions return a boolean indicating their success, we don't actually check for it later. The return value is always initialized to something usable (e.g. empty array or null), so we add it to the debug info's hash regardless. If an exception was thrown, it will be caught above the debug handler (technically get_properties).

jmikola:
@derickr: We didn't want to stuff this into 1.2.0 at the last minute, but I believe it's worth revisiting for inclusion in a future release. Please review when you get a chance.

Issue #478, opened 2016-12-04
Driver on Debian 8


thibaultfalque:
Hello, 
### Description

I try to install the php mongodb driver in my server. But I have this error 
PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php5/20131226/mongodb.so' - /usr/lib/php5/20131226/mongodb.so: undefined symbol: php_json_serializable_ce in Unknown on line 0
I use Symfony and must have the the mondb.so and not mongo.so. 
I have install this driver with pecl: 
pecl install mongodb

I create this file: /etc/php5/mods-available/mongodb.ini with this content: 
extension=mongodb.so
I execute this command: 
sudo ln -sv /etc/php5/mods-available/mongodb.ini /etc/php5/cli/conf.d/20-mongodb.ini
sudo ln -sv /etc/php5/mods-available/mongodb.ini /etc/php5/apache2/conf.d/20-mongodb.ini

### Environment
$php - i| grep mongodb
PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php5/20131226/mongodb.so' - /usr/lib/php5/20131226/mongodb.so: undefined symbol: php_json_serializable_ce in Unknown on line 0

/etc/php5/cli/conf.d/20-mongodb.ini,
mongodb
mongodb support => enabled
mongodb version => 1.2.0
mongodb stability => stable
mongodb.debug => no value => no value

$php -m 
PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php5/20131226/mongodb.so' - /usr/lib/php5/20131226/mongodb.so: undefined symbol: php_json_serializable_ce in Unknown on line 0
[PHP Modules]
bcmath
bz2
calendar
Core
ctype
curl
date
dba
dom
ereg
exif
fileinfo
filter
ftp
gettext
hash
iconv
intl
json
libxml
mbstring
mhash
mongodb
mysql
mysqli
openssl
pcntl
pcre
PDO
pdo_mysql
Phar
posix
readline
Reflection
session
shmop
SimpleXML
soap
sockets
SPL
standard
sysvmsg
sysvsem
sysvshm
tokenizer
wddx
xml
xmlreader
xmlwriter
Zend OPcache
zip
zlib
[Zend Modules]
Zend OPcache

Thank you for your help.

Visualq:
Hi, read https://derickrethans.nl/

thibaultfalque:
Hi,
[...]

Issue #477, opened 2016-11-30
Unable to load dynamic library after upgrade to 1.2.0 "undefined symbol: php_json_serializable_ce"


stephadz:
### Description
After upgrade driver to release 1.2.0, i've got an error at apache startup and cli : 

PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib64/php/5.6/modules/mongodb.so' - /usr/lib64/php/5.6/modules/mongodb.so: undefined symbol: php_json_serializable_ce in Unknown on line 0

No problem when i downgrade to 1.1.9

### Environment

We use preconfigured AWS environnement : 64bit Amazon Linux 2016.09 v2.2.0 running PHP 5.6
PHP 5.6.26
Apache 2.4.23

### Tests / debug

On release 1.2.0 
```
$ php -i | grep -E 'mongodb|libmongoc|libbson'
PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib64/php/5.6/modules/mongodb.so' - /usr/lib64/php/5.6/modules/mongodb.so: undefined symbol: php_json_serializable_ce in Unknown on line 0

```

After downgrading to 1.1.9
```
$ php -i | grep -E 'mongodb|libmongoc|libbson'
mongodb
mongodb support => enabled
mongodb version => 1.1.9
mongodb stability => stable
libmongoc version => 1.3.6
libbson version => 1.3.6
mongodb.debug => no value => no value

```

i hope i have provided you with sufficient information

Thanks

jmikola:
This is likely a duplicate of https://github.com/mongodb/mongo-php-driver/issues/475. @derickr is in the process of writing a blog post to discuss the issue in more detail.

This is related to a new dependency on the `json` extension in 1.2.0, as our BSON classes now implement JsonSerializable ([PHPC-459](https://jira.mongodb.org/browse/PHPC-459)). We likewise declare dependencies on the `date` and `spl` extensions for UTCDateTime and our exceptions, respectively; however, those are almost always compiled into PHP. The `json` extension is sometimes loaded dynamically by different distributions. Those distributions may also elect to resolve symbols at execution time instead of lazily during runtime (PHP's default behavior), which can cause this error.

Please confirm whether the `json` extension is compiled into PHP or if you're loading it a `json.so` file dynamically in your PHP configuration. Assuming the latter, you need to ensure that the extension is loaded before `mongodb.so`.

derickr:
[...]

Issue #475, opened 2016-11-29
Undefined Symbol php_json_serializable_ce in Unknown on Line 0 (1.2.0 Release)


MineSQL:
### Description

When installing mongodb from pecl on debian jessie, the extension seems to install fine but is never initialized due to this error. I regularly build from pecl, but right after today's update all my build scripts broke for all my enviroments. Installing the actual extension works fine, but the extension is never loaded due to the error below.

### Environment

Debootstrap Debian Jessie running PHP 5.6.27 

```
root@b0277afcbe10:/# php -m                                                                                                                                                                                                                 
PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php5/20131226/mongodb.so' - /usr/lib/php5/20131226/mongodb.so: undefined symbol: php_json_serializable_ce in Unknown on line 0
[PHP Modules]
bcmath
bz2
calendar
Core
ctype
curl
date
dba
dom
ereg
exif
fileinfo
filter
ftp
gettext
hash
iconv
json
libxml
mbstring
mhash
mysql
mysqli
openssl
pcntl
pcre
PDO
pdo_mysql
Phar
posix
readline
Reflection
session
shmop
SimpleXML
soap
sockets
SPL
standard
sysvmsg
sysvsem
sysvshm
tokenizer
wddx
xdebug
xml
xmlreader
xmlwriter
xsl
Zend OPcache
zip
zlib

[Zend Modules]
Xdebug
Zend OPcache

```


```
root@b0277afcbe10:/# php -i | grep -E 'mongodb|libmongoc|libbson'
PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php5/20131226/mongodb.so' - /usr/lib/php5/20131226/mongodb.so: undefined symbol: php_json_serializable_ce in Unknown on line 0
```


### Test Script

When trying to use the mongodb/mongodb package from Composer it throws: 

( ! ) Fatal error: Class 'MongoDB\Driver\Manager' not found in /var/www/html/vendor/mongodb/mongodb/src/Client.php on line 61

derickr:
Thanks for your query, could you please answer the following questions:

How are you loading the json extension?

Through a module?

Does it work if you load the json extension through the php.ini system before the mongodb extension?

What's the output of php --ini

MineSQL:
After grinding the entire day, I've came up with a temporary fix if anyone else is having this issue.

[...]

PR #480, opened 2016-12-05
Print a custom error message on platforms with ext/json as shared extension

bjori:
Before:
```
$ php -n -dextension=modules/mongodb.so -r 'echo phpversion("mongodb"), "\n";'
PHP Warning:  PHP Startup: Unable to load dynamic library 'modules/mongodb.so' - modules/mongodb.so: undefined symbol: php_json_serializable_ce in Unknown on line 0
```
After:
```
$ php -n -dextension=modules/mongodb.so -r 'echo phpversion("mongodb"), "\n";'
Please make sure json.so is loaded before mongodb.so
PHP Warning:  Cannot load module 'mongodb' because required module 'json' is not loaded in Unknown on line 0
```

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -2016,6 +2016,18 @@ static void php_phongo_client_dtor(void *client)
 }
 #endif
 
+static void _phongo_check_json (void) __attribute__((constructor));
+static void _phongo_check_json (void)
+{
+	DL_HANDLE handle = DL_LOAD(NULL);
+	zend_class_entry *json = (zend_class_entry *)DL_FETCH_SYMBOL (handle, "php_json_serializable_ce");
+
+	if (json == NULL) {
+		fprintf (stderr, "Please make sure json.so is loaded before mongodb.so\n");
+		fflush (stderr);
+	}
+}
+
 /* {{{ PHP_MINIT_FUNCTION */
 PHP_MINIT_FUNCTION(mongodb)
 {
--- a/src/BSON/Binary.c
+++ b/src/BSON/Binary.c
@@ -452,6 +452,7 @@ HashTable *php_phongo_binary_get_properties(zval *object TSRMLS_DC) /* {{{ */
 PHP_MINIT_FUNCTION(Binary)
 {
 	zend_class_entry ce;
+	zend_string *zstr = zend_string_init("jsonserializable", sizeof("jsonserializable") - 1, 0);
 	(void)type;(void)module_number;
 
 	INIT_NS_CLASS_ENTRY(ce, "MongoDB\\BSON", "Binary", php_phongo_binary_me);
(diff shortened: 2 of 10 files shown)


bjori:
The extra warning there is of course not printed when things go fine:
```
php -dextension=modules/mongodb.so -m                                                                                                                                                                        <<<
[PHP Modules]
calendar
Core
ctype
date
dom
exif
fileinfo
filter
ftp
gettext
hash
iconv
json
libxml
mongodb
openssl
pcntl
pcre
PDO
Phar
posix
readline
Reflection
session
shmop
SimpleXML
sockets
SPL
standard
sysvmsg
sysvsem
sysvshm
tokenizer
wddx
xml
xmlreader
xmlwriter
xsl
Zend OPcache
zlib

[Zend Modules]
Zend OPcache

```

bjori:
Related issues:
- https://github.com/mongodb/mongo-php-driver/issues/478
- https://github.com/mongodb/mongo-php-driver/issues/477
- https://github.com/mongodb/mongo-php-driver/issues/475

php_phongo.c
```diff
@@ -2016,6 +2016,18 @@ static void php_phongo_client_dtor(void *client)
 }
 #endif
 
+static void _phongo_check_json (void) __attribute__((constructor));
+static void _phongo_check_json (void)
```
derickr: This does not seem to be called from anywhere?

php_phongo.c
```diff
@@ -2016,6 +2016,18 @@ static void php_phongo_client_dtor(void *client)
...
+{
+	DL_HANDLE handle = DL_LOAD(NULL);
+	zend_class_entry *json = (zend_class_entry *)DL_FETCH_SYMBOL (handle, "php_json_serializable_ce");
```
derickr: Some OSses prefix the symbols with a ``_``, and no check for that is done here. 

See: https://github.com/php/php-src/blob/master/ext/standard/dl.c#L142

src/BSON/Binary.c
```diff
@@ -452,14 +452,16 @@ HashTable *php_phongo_binary_get_properties(zval *object TSRMLS_DC) /* {{{ */
 PHP_MINIT_FUNCTION(Binary)
 {
 	zend_class_entry ce;
+	zend_string *zstr = zend_string_init("jsonserializable", sizeof("jsonserializable") - 1, 0);
```
derickr: zend_string_init is new in PHP 7.0, so this needs an #ifdef.

php_phongo.c
bjori: Its called automatically during `dlopen`

bjori:
This PR actually does two things.
1) Removes the need for that symbol, so PHPs error message is properly displayed.
2) Augments the PHP error message with custom error to guide the user

I consider this to be significant improvement to the user experience and even it is not the mongodb drivers fault, we can still improve the experience for the users

php_phongo.c
derickr: That's a problem then, as it would mean that PHP's normal loading reordering will no longer work. Basically, making the off case work (where distributions mangle PHP's original itents), but not the standard case where PHP itself manages the loading/symbol fetching order (the patch actively prevents this from working now).

jmikola: Does `__attribute__((constructor))` result in this preempting PHP's own extension dependency resolution based on the list in our zend_module_entry?

derickr: PHP's mechanism is as follows:

- ``dlopen()`` all extensions in their loading order (order as in .ini file). This works normally, as ``LTDL_LAZY`` is being used and no symbols are being resolved yet.
- create dependency graph
- go initialize extensions in the right order as to not have undefined symbols

The new constructor is run on ``dlopen()``, and hence, breaks the original mechanism of being to *open* extensions in a random order.

php_phongo.c
remicollet: I have doubt about this working if php_json_serializable_ce is not publicly exported.

bjori: this isn't actually needed. as long as the references to the symbol are removed and it rather dynamically loaded, like the rest of the patch.
This function is only for the purpose of augmenting the error message from php if they want to give direct instructions

jmikola:
See https://github.com/mongodb/mongo-php-driver/pull/500.

PR #481, opened 2016-12-06
PHPC-848: Consult ZEND_HASH_APPLY_PROTECTION() in PHP7

jmikola:
https://jira.mongodb.org/browse/PHPC-848

When a numeric array was encoded to BSON, the hash's apply counter was incremented and decremented irrespective of whether apply protection was enabled. This was not compatible with immutable arrays, which may be created by OPcache. We now consult ZEND_HASH_APPLY_PROTECTION() whenever an apply counter might be modified.

Previously, object_to_bson() incremented the apply counter when converting a MongoDB\BSON\Type instance. Associative arrays and other objects were not protected. This patch adds recursion checking for those types.

Since PHP7 already has macros for working with the apply counter, this patch adds equivalent macros for PHP 5.x.

Related issues:
 * https://github.com/mongodb/mongo-php-library/issues/216
 * https://github.com/jbboehr/php-mustache/pull/29

### DIFF
--- a/phongo_compat.h
+++ b/phongo_compat.h
@@ -167,7 +167,10 @@
 # define Z_ISUNDEF(x) !x
 # define phongo_free_object_arg void
 # define phongo_zpp_char_len int
-# define ZEND_HASH_APPLY_COUNT(ht) (ht)->nApplyCount
+# define ZEND_HASH_APPLY_PROTECTION(ht) true
+# define ZEND_HASH_GET_APPLY_COUNT(ht) ((ht)->nApplyCount)
+# define ZEND_HASH_DEC_APPLY_COUNT(ht) ((ht)->nApplyCount -= 1)
+# define ZEND_HASH_INC_APPLY_COUNT(ht) ((ht)->nApplyCount += 1)
 # define PHONGO_RETVAL_STRINGL(s, slen) RETVAL_STRINGL(s, slen, 1)
 # define PHONGO_RETURN_STRINGL(s, slen) RETURN_STRINGL(s, slen, 1)
 # define PHONGO_RETURN_STRING(s) RETURN_STRING(s, 1)
--- a/src/bson.c
+++ b/src/bson.c
@@ -862,8 +862,8 @@ void object_to_bson(zval *object, php_phongo_bson_flags_t flags, const char *key
 			tmp_ht = HASH_OF(obj_data);
 #endif
 
-			if (tmp_ht) {
-				ZEND_HASH_APPLY_COUNT(tmp_ht)++;
+			if (tmp_ht && ZEND_HASH_APPLY_PROTECTION(tmp_ht)) {
+				ZEND_HASH_INC_APPLY_COUNT(tmp_ht);
 			}
 
 			/* Persistable objects must always be serialized as BSON documents;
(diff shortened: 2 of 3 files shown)


phongo_compat.h
```diff
@@ -167,7 +167,10 @@
 # define Z_ISUNDEF(x) !x
 # define phongo_free_object_arg void
 # define phongo_zpp_char_len int
-# define ZEND_HASH_APPLY_COUNT(ht) (ht)->nApplyCount
+# define ZEND_HASH_APPLY_PROTECTION(ht) true
+# define ZEND_HASH_GET_APPLY_COUNT(ht) ((ht)->nApplyCount)
+# define ZEND_HASH_DEC_APPLY_COUNT(ht) ((ht)->nApplyCount -= 1)
+# define ZEND_HASH_INC_APPLY_COUNT(ht) ((ht)->nApplyCount += 1)
```
derickr: I'm pretty sure the Zend API has something for this in PHP < 7 too?

src/bson.c
```diff
@@ -862,8 +862,8 @@ void object_to_bson(zval *object, php_phongo_bson_flags_t flags, const char *key
 			tmp_ht = HASH_OF(obj_data);
 #endif
 
-			if (tmp_ht) {
-				ZEND_HASH_APPLY_COUNT(tmp_ht)++;
+			if (tmp_ht && ZEND_HASH_APPLY_PROTECTION(tmp_ht)) {
+				ZEND_HASH_INC_APPLY_COUNT(tmp_ht);
```
derickr: This looks scary - I would check with @dstogov whether this is correct. What happens when a "static" array has recursive loops now?

phongo_compat.h
jmikola: I only see [`nApplyCount` declared in `zend_hash.h`](https://github.com/php/php-src/blob/PHP-5.6/Zend/zend_hash.h#L78) on PHP-5.6. There are no other macros that refer to it. Prior to this PR, we had a single macro to access the struct field.

src/bson.c
jmikola: If we look at PHP's `json_encode()`, the apply count is always incremented when iterating on arrays and objects (compare [PHP-5.6](https://github.com/php/php-src/blob/PHP-5.6/ext/json/json.c#L253) and [PHP-7.0](https://github.com/php/php-src/blob/PHP-7.0/ext/json/json_encoder.c#L131)).

The `ZEND_HASH_APPLY_PROTECTION()` macros back to @dstogov's original commit in https://github.com/php/php-src/commit/b3b616cf7ee0702aca2f6fdb022dc7e28e015a22.

jmikola: > What happens when a "static" array has recursive loops now?

I asked Dmitry in IRC and he said, "opcache never creates immutable circular arrays." I suppose if it started doing so, PHP would have a lot more problems on its hands.

jmikola:
I added a regression test that tests encoding PHP values with cyclic references. It segfaults quite easily without the patch in place for associative arrays and objects, since `object_to_bson()` was not incrementing the apply count for such values. Packed arrays do not segfault pre-patch, since we were always incrementing the apply count there.

PR #485, opened 2016-12-09
PHPC-854: Client persistence crashes with ZTS (e.g. Apache)

jmikola:
https://jira.mongodb.org/browse/PHPC-854

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1511,8 +1511,7 @@ void phongo_manager_init(php_phongo_manager_t *manager, const char *uri_string, 
 	bson_iter_t       iter;
 
 #if PHP_VERSION_ID >= 70000
-	zval             *client_ptr;
-	zval              new_client_ptr;
+	mongoc_client_t  *client_ptr;
 #else
 	mongoc_client_t **client_ptr;
 #endif
@@ -1523,9 +1522,9 @@ void phongo_manager_init(php_phongo_manager_t *manager, const char *uri_string, 
 	}
 
 #if PHP_VERSION_ID >= 70000
-	if ((client_ptr = zend_hash_str_find(&MONGODB_G(clients), hash, hash_len)) && Z_TYPE_P(client_ptr) == IS_PTR) {
+	if ((client_ptr = zend_hash_str_find_ptr(&MONGODB_G(clients), hash, hash_len)) != NULL) {
 		MONGOC_DEBUG("Found client for hash: %s\n", hash);
-		manager->client = (mongoc_client_t *)Z_PTR_P(client_ptr);
+		manager->client = client_ptr;
 		goto cleanup;
 	}
 #else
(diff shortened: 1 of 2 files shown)


php_phongo.c
```diff
@@ -1998,24 +2008,13 @@ PHP_GINIT_FUNCTION(mongodb)
 	ZEND_TSRMLS_CACHE_UPDATE();
 #endif
 #endif
-	mongodb_globals->debug_fd = NULL;
+	memset(mongodb_globals, 0, sizeof(zend_mongodb_globals));
 	mongodb_globals->bsonMemVTable = bsonMemVTable;
-
+	/* Initialize HashTable for persistent clients */
+	zend_hash_init_ex(&mongodb_globals->clients, 0, NULL, php_phongo_client_dtor, 1, 0);
```
jmikola: It's possible that the root cause was not GINIT vs. MINIT, but rather previously fetching `mongodb_globals->clients` through our `MONGODB_G()`, since that uses [`ZEND_MODULE_GLOBALS_ACCESSOR()`](https://github.com/php/php-src/blob/php-7.0.14/Zend/zend_API.h#L162) on ZTS. For non-ZTS builds, the macro directly accesses the struct field, as we're doing here.

That said, other extensions initialize their persistent hashes in GINIT (see: [pgsql](https://github.com/php/php-src/blob/php-7.0.14/ext/pgsql/pgsql.c#L1098)), so I'd consider it good practice.

php_phongo.h
```diff
@@ -47,16 +47,21 @@ ZEND_BEGIN_MODULE_GLOBALS(mongodb)
 	HashTable clients;
 ZEND_END_MODULE_GLOBALS(mongodb)
 
+ZEND_EXTERN_MODULE_GLOBALS(mongodb)
+
 #if PHP_VERSION_ID >= 70000
-#   define MONGODB_G(v) ZEND_MODULE_GLOBALS_ACCESSOR(mongodb, v)
+# define MONGODB_G(v) ZEND_MODULE_GLOBALS_ACCESSOR(mongodb, v)
+# if defined(ZTS) && defined(COMPILE_DL_MONGODB)
+ZEND_TSRMLS_CACHE_EXTERN()
```
derickr: You don't really need the "ZEND_EXTERN_MODULE_GLOBALS(mongodb)" though, as you've just declared the globals the few lines above. It is only there to put a forwards declaration in in places that normally can't "see" the globals.

I have never actually seen that PHP 7 needs a different thing here. I don't do that in Xdebug either...

php_phongo.c
derickr: I'm going to bet the main issue was the zval mangling in the previous block...

php_phongo.h
jmikola: Good point. I removed `ZEND_EXTERN_MODULE_GLOBALS(mongodb)`. The only time I saw that used in other extensions was in various C files, which may not have included the declaring header.

PR #488, opened 2016-12-13
PHPC-851: Separate argument zvals before modification

jmikola:
https://jira.mongodb.org/browse/PHPC-851

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -837,7 +837,9 @@ void php_phongo_read_preference_prep_tagsets(zval *tagSets TSRMLS_DC) /* {{{ */
 		zval *tagSet;
 
 		ZEND_HASH_FOREACH_VAL(ht_data, tagSet) {
+			ZVAL_DEREF(tagSet);
 			if (Z_TYPE_P(tagSet) == IS_ARRAY) {
+				SEPARATE_ZVAL_NOREF(tagSet);
 				convert_to_object(tagSet);
 			}
 		} ZEND_HASH_FOREACH_END();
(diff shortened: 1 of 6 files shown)


php_phongo.c
```diff
@@ -837,7 +837,9 @@ void php_phongo_read_preference_prep_tagsets(zval *tagSets TSRMLS_DC) /* {{{ */
 		zval *tagSet;
 
 		ZEND_HASH_FOREACH_VAL(ht_data, tagSet) {
+			ZVAL_DEREF(tagSet);
```
derickr: Why do you do the deref as well?

bashofmann:
Looks good, with this patch I can't reproduce the segmentation fault on setting read preferences anymore

php_phongo.c
jmikola: Basing this on iteration examples I found in core (e.g. [`php_zval_filter_recursive()`](https://github.com/php/php-src/blob/php-7.0.14/ext/filter/filter.c#L508)).

We currently handle dereferencing in `phongo_bson_append()`, which is the only other time we iterate apart from preparing tag sets. BSON encoding also resolves indirect references, which I could do with the `ZEND_HASH_FOREACH_VAL_IND()` macro, but I wasn't sure if that was necessary (even after reading up on indirect zvals in [Nikita's blog](https://nikic.github.io/2015/06/19/Internal-value-representation-in-PHP-7-part-2.html)).

jmikola: I opened [PHPC-870](https://jira.mongodb.org/browse/PHPC-870) to track future iteration changes.

PR #498, opened 2016-12-19
PHPC-633: Include libmongoc SSL, crypto, and SASL details in phpinfo()

jmikola:
https://jira.mongodb.org/browse/PHPC-633

Supersedes https://github.com/mongodb/mongo-php-driver/pull/265

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1982,6 +1982,49 @@ PHP_MINFO_FUNCTION(mongodb)
 	php_info_print_table_row(2, "libmongoc bundled version", MONGOC_VERSION_S);
 #endif
 
+#ifdef MONGOC_ENABLE_SSL
+	php_info_print_table_row(2, "libmongoc SSL", "enabled");
+# if defined(MONGOC_ENABLE_SSL_OPENSSL)
+	php_info_print_table_row(2, "libmongoc SSL library", "OpenSSL");
+# elif defined(MONGOC_ENABLE_SSL_LIBRESSL)
+	php_info_print_table_row(2, "libmongoc SSL library", "LibreSSL");
+# elif defined(MONGOC_ENABLE_SSL_SECURE_TRANSPORT)
+	php_info_print_table_row(2, "libmongoc SSL library", "Secure Transport");
+# elif defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+	php_info_print_table_row(2, "libmongoc SSL library", "Secure Channel");
+# else
+	php_info_print_table_row(2, "libmongoc SSL library", "unknown");
+# endif
+#else
+	php_info_print_table_row(2, "libmongoc SSL", "disabled");
+#endif
+
+#ifdef MONGOC_ENABLE_CRYPTO
+	php_info_print_table_row(2, "libmongoc crypto", "enabled");
+# if defined(MONGOC_ENABLE_CRYPTO_LIBCRYPTO)
+	php_info_print_table_row(2, "libmongoc crypto library", "libcrypto");
+# elif defined(MONGOC_ENABLE_CRYPTO_COMMON_CRYPTO)
+	php_info_print_table_row(2, "libmongoc crypto library", "Common Crypto");
+# elif defined(MONGOC_ENABLE_CRYPTO_CNG)
+	php_info_print_table_row(2, "libmongoc crypto library", "CNG");
+# else
+	php_info_print_table_row(2, "libmongoc crypto library", "unknown");
+# endif
+# ifdef MONGOC_ENABLE_CRYPTO_SYSTEM_PROFILE
+	php_info_print_table_row(2, "libmongoc crypto system profile", "enabled");
+# else
+	php_info_print_table_row(2, "libmongoc crypto system profile", "disabled");
+# endif
+#else
+	php_info_print_table_row(2, "libmongoc crypto", "disabled");
+#endif
+
+#ifdef MONGOC_ENABLE_SASL
+	php_info_print_table_row(2, "libmongoc SASL", "enabled");
+#else
+	php_info_print_table_row(2, "libmongoc SASL", "disabled");
+#endif
+
 	php_info_print_table_end();
 
 	DISPLAY_INI_ENTRIES();
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1966,6 +1966,14 @@ PHP_MINFO_FUNCTION(mongodb)
 	php_info_print_table_header(2, "MongoDB support", "enabled");
 	php_info_print_table_row(2, "MongoDB extension version", MONGODB_VERSION_S);
 	php_info_print_table_row(2, "MongoDB extension stability", MONGODB_STABILITY_S);
+
+#ifdef HAVE_SYSTEM_LIBBSON
+	php_info_print_table_row(2, "libbson headers version", BSON_VERSION_S);
+	php_info_print_table_row(2, "libbson library version", bson_get_version());
+#else
+	php_info_print_table_row(2, "libbson bundled version", BSON_VERSION_S);
+#endif
+
 #ifdef HAVE_SYSTEM_LIBMONGOC
 	php_info_print_table_row(2, "libmongoc headers version", MONGOC_VERSION_S);
 	php_info_print_table_row(2, "libmongoc library version", mongoc_get_version());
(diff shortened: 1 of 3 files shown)


php_phongo.c
```diff
@@ -1963,22 +1963,65 @@ PHP_GSHUTDOWN_FUNCTION(mongodb)
...
-	php_info_print_table_row(2, "mongodb stability", MONGODB_STABILITY_S);
+	php_info_print_table_header(2, "MongoDB support", "enabled");
+	php_info_print_table_row(2, "MongoDB extension version", MONGODB_VERSION_S);
+	php_info_print_table_row(2, "MongoDB extension stability", MONGODB_STABILITY_S);
+
+#ifdef HAVE_SYSTEM_LIBBSON
+	php_info_print_table_row(2, "libbson headers version", BSON_VERSION_S);
+	php_info_print_table_row(2, "libbson library version", bson_get_version());
+#else
+	php_info_print_table_row(2, "libbson bundled version", BSON_VERSION_S);
```
jmikola: Including "bundled" here was more clear than simply saying "libbson version". I also moved the libbson output above libmongoc in advance of adding the additional SSL, crypto, and SASL information.

php_phongo.c
```diff
@@ -1963,22 +1963,65 @@ PHP_GSHUTDOWN_FUNCTION(mongodb)
...
+# elif defined(MONGOC_ENABLE_SSL_LIBRESSL)
+	php_info_print_table_row(2, "libmongoc SSL library", "LibreSSL");
+# elif defined(MONGOC_ENABLE_SSL_SECURE_TRANSPORT)
+	php_info_print_table_row(2, "libmongoc SSL library", "Secure Transport");
+# elif defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+	php_info_print_table_row(2, "libmongoc SSL library", "Secure Channel");
+# endif
 #else
-	php_info_print_table_row(2, "libbson version", BSON_VERSION_S);
+	php_info_print_table_row(2, "libmongoc SSL", "disabled");
```
jmikola: I realize we don't support building without SSL support, but this is for completeness' sake. I also wanted to keep binary enabled/disabled output for "libmongoc SSL", to match "libmongoc SASL".

php_phongo.c
```diff
@@ -1963,22 +1963,65 @@ PHP_GSHUTDOWN_FUNCTION(mongodb)
...
 #endif
+
+#ifdef MONGOC_ENABLE_CRYPTO
+	php_info_print_table_row(2, "libmongoc crypto", "enabled");
+# if defined(MONGOC_ENABLE_CRYPTO_LIBCRYPTO)
+	php_info_print_table_row(2, "libmongoc crypto library", "libcrypto");
+# elif defined(MONGOC_ENABLE_CRYPTO_COMMON_CRYPTO)
+	php_info_print_table_row(2, "libmongoc crypto library", "Common Crypto");
+# elif defined(MONGOC_ENABLE_CRYPTO_CNG)
+	php_info_print_table_row(2, "libmongoc crypto library", "CNG");
```
jmikola: These values can technically be inferred from the SSL library, but I didn't see the harm in including them.

php_phongo.c
```diff
@@ -1963,22 +1963,65 @@ PHP_GSHUTDOWN_FUNCTION(mongodb)
...
+#ifdef MONGOC_ENABLE_SSL
+	php_info_print_table_row(2, "libmongoc SSL", "enabled");
+# if defined(MONGOC_ENABLE_SSL_OPENSSL)
+	php_info_print_table_row(2, "libmongoc SSL library", "OpenSSL");
+# elif defined(MONGOC_ENABLE_SSL_LIBRESSL)
+	php_info_print_table_row(2, "libmongoc SSL library", "LibreSSL");
+# elif defined(MONGOC_ENABLE_SSL_SECURE_TRANSPORT)
+	php_info_print_table_row(2, "libmongoc SSL library", "Secure Transport");
+# elif defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+	php_info_print_table_row(2, "libmongoc SSL library", "Secure Channel");
```
bjori: this looks like something that could easily break in the future.
Should mongoc provide a function to get this information?
Otherwise, when we start supporting gnuTLS for example, this will break

php_phongo.c
```diff
@@ -1963,22 +1963,64 @@ PHP_GSHUTDOWN_FUNCTION(mongodb)
...
+	php_info_print_table_row(2, "libmongoc crypto library", "libcrypto");
+# elif defined(MONGOC_ENABLE_CRYPTO_COMMON_CRYPTO)
+	php_info_print_table_row(2, "libmongoc crypto library", "Common Crypto");
+# elif defined(MONGOC_ENABLE_CRYPTO_CNG)
+	php_info_print_table_row(2, "libmongoc crypto library", "CNG");
+# endif
+# ifdef MONGOC_ENABLE_CRYPTO_SYSTEM_PROFILE
+	php_info_print_table_row(2, "libmongoc crypto system profile", "enabled");
+# else
+	php_info_print_table_row(2, "libmongoc crypto system profile", "disabled");
```
jmikola: AFAIK, this is currently only relevant to OpenSSL, which uses libcrypto. In any event, I liked including it at all times (if crypto is enabled) for consistency.

@bjori: Am I correct in assuming this will never apply to LibreSSL, despite the fact that it also uses libcrypto?

php_phongo.c
```diff
@@ -1963,22 +1963,64 @@ PHP_GSHUTDOWN_FUNCTION(mongodb)
...
 #else
-	php_info_print_table_row(2, "libbson version", BSON_VERSION_S);
+	php_info_print_table_row(2, "libmongoc crypto", "disabled");
 #endif
+
+#ifdef MONGOC_ENABLE_SASL
+	php_info_print_table_row(2, "libmongoc SASL", "enabled");
+#else
+	php_info_print_table_row(2, "libmongoc SASL", "disabled");
+#endif
```
jmikola: I don't think we care about reporting `MONGOC_HAVE_SASL_CLIENT_DONE` here.

tests/functional/phpinfo-1.phpt
```diff
@@ -16,11 +16,17 @@ phpinfo();
...
-mongodb support => enabled
-mongodb version => 1.%d.%d%S
-mongodb stability => %s
-libmongoc version => 1.%s
-libbson version => 1.%s
+MongoDB support => enabled
+MongoDB extension version => 1.%d.%d%S
+MongoDB extension stability => %s
+libbson bundled version => 1.%d.%d%S
+libmongoc bundled version => 1.%d.%d%S
```
jmikola: libbson and libmongoc follow the same version format with optional trailing strings, so I updated these patterns to match our own.

tests/functional/phpinfo-1.phpt
```diff
@@ -16,11 +16,17 @@ phpinfo();
...
+libmongoc SSL => enabled
+libmongoc SSL library => %s
+libmongoc crypto => enabled
+libmongoc crypto library => %s
+libmongoc crypto system profile => %s
+libmongoc SASL => enabled
```
jmikola: I'm making assumptions that CI will build with SSL and SASL, but make no assumptions about the SSL and cypto libraries. This should continue to work once we start supporting other libraries for Darwin and Windows.

php_phongo.c
bjori: this isn't related to libcrypto. This is related to which SSL/TLS ciphers we announce support for during SSL/TLS handshake.

The default for libressl's libtls is the string `"secure"` which currently is an alias for `TLSv1.2+AEAD+ECDHE:TLSv1.2+AEAD+DHE`.
We do not overwrite this and have no intention of exposing such option.

The default for OpenSSL is "every cipher I have ever heard of", so instead of that, we explicitly set the list of ciphers to `HIGH:!EXPORT:!aNULL@STRENGTH` which gets resolved into ciphers that noone really knows which are, but are what the "OpenSSL community" seems to recommend.
There is a way to configure OpenSSL to set a default ciphers for everything, rather then "support everything", and some distro's do this, in which case, explicitly providing a list of ciphers will overwrite the ciphers the distro wants to supports.

To be honest, I don't actually know if you can do such thing in libtls. It seems unlikely as it doesn't provide any additional value since the default configure is secure by default.

php_phongo.c
jmikola: Having libmongoc return a string would be helpful, but we won't have that available in 1.5.x anyway. I suppose we should have a final `#else` case that prints "other" or "unknown". This would only affect system library users running older versions of the PHP extension.

PR #500, opened 2016-12-21
PHPC-878, PHPC-879: Fetch JSON and Date class entries during MINIT

jmikola:
https://jira.mongodb.org/browse/PHPC-878
https://jira.mongodb.org/browse/PHPC-879

Supersedes https://github.com/mongodb/mongo-php-driver/pull/480.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -74,6 +74,7 @@ ZEND_DECLARE_MODULE_GLOBALS(mongodb)
 #endif
 
 /* Declare zend_class_entry dependencies, which are initialized in MINIT */
+zend_class_entry *php_phongo_date_immutable_ce;
 zend_class_entry *php_phongo_json_serializable_ce;
 
 php_phongo_server_description_type_map_t
@@ -1910,11 +1911,16 @@ PHP_MINIT_FUNCTION(mongodb)
 
 	/* Initialize zend_class_entry dependencies.
 	 *
+	 * Although DateTimeImmutable was introduced in PHP 5.5.0,
+	 * php_date_get_immutable_ce() is not available in PHP versions before
+	 * 5.5.24 and 5.6.8.
+	 *
 	 * Although JsonSerializable was introduced in PHP 5.4.0,
 	 * php_json_serializable_ce is not exported in PHP versions before 5.4.26
 	 * and 5.5.10. For later PHP versions, looking up the class manually also
 	 * helps with distros that disable LTDL_LAZY for dlopen() (e.g. Fedora).
 	 */
+	php_phongo_date_immutable_ce = php_phongo_fetch_internal_class(ZEND_STRL("datetimeimmutable") TSRMLS_CC);
 	php_phongo_json_serializable_ce = php_phongo_fetch_internal_class(ZEND_STRL("jsonserializable") TSRMLS_CC);
 
 	if (php_phongo_json_serializable_ce == NULL) {
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -73,6 +73,9 @@ ZEND_DECLARE_MODULE_GLOBALS(mongodb)
 #endif
 #endif
 
+/* Declare zend_class_entry dependencies, which are initialized in MINIT */
+zend_class_entry *php_phongo_json_serializable_ce;
+
 php_phongo_server_description_type_map_t
 php_phongo_server_description_type_map[PHONGO_SERVER_DESCRIPTION_TYPES] = {
 	{ PHONGO_SERVER_UNKNOWN, "Unknown" },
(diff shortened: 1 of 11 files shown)


jmikola:
I'm aware that there are test failures on PHP 5.x due to fetching the "JsonSerializable" interface.

remicollet:
1st try, Fedora 25, PHP 7.0.14

```
$ php -n -d extension=modules/mongodb.so -v
PHP Warning:  Cannot load module 'mongodb' because required module 'json' is not loaded in Unknown on line 0

$ php -n -d extension=json.so -d extension=modules/mongodb.so -v
PHP 7.0.14 (cli) (built: Dec  7 2016 09:36:47) ( NTS )

$ php -n -d extension=modules/mongodb.so -d extension=json.so -v
PHP 7.0.14 (cli) (built: Dec  7 2016 09:36:47) ( NTS )

```

remicollet:
2nd try, RHEL 7.3, PHP 5.4.16

```
$ php -n -d extension=modules/mongodb.so -v
PHP Warning:  Cannot load module 'mongodb' because required module 'json' is not loaded in Unknown on line 0
PHP 5.4.16 (cli) (built: Aug  5 2016 07:50:38) 

$ php -n -d extension=json.so -d extension=modules/mongodb.so --re mongodb | grep UTCDateTime
    Class [ <internal:mongodb> final class MongoDB\BSON\UTCDateTime implements JsonSerializable, MongoDB\BSON\Type, Serializable ] {

$ php -n -d extension=modules/mongodb.so -d extension=json.so --re mongodb | grep UTCDateTime
    Class [ <internal:mongodb> final class MongoDB\BSON\UTCDateTime implements JsonSerializable, MongoDB\BSON\Type, Serializable ] {
```

So in both case, works as expected, using any load order

+1

php_phongo.c
```diff
@@ -1879,6 +1902,25 @@ PHP_MINIT_FUNCTION(mongodb)
...
+	 * Although JsonSerializable was introduced in PHP 5.4.0,
+	 * php_json_serializable_ce is not exported in PHP versions before 5.4.26
+	 * and 5.5.10. For later PHP versions, looking up the class manually also
+	 * helps with distros that disable LTDL_LAZY for dlopen() (e.g. Fedora).
+	 */
+	php_phongo_date_immutable_ce = php_phongo_fetch_internal_class(ZEND_STRL("datetimeimmutable") TSRMLS_CC);
+	php_phongo_json_serializable_ce = php_phongo_fetch_internal_class(ZEND_STRL("jsonserializable") TSRMLS_CC);
+
+	if (php_phongo_json_serializable_ce == NULL) {
+		zend_error(E_ERROR, "JsonSerializable class not defined. Please make sure json.so is loaded before mongodb.so.");
```
derickr: You can't use ``.so`` here, as that is wrong on Windows and some other OSses. There is a constant for it: ``PHP_SHLIB_SUFFIX``.

jmikola: In addition to `PHP_SHLIB_SUFFIX`, we'd probably also have to add a "php_" prefix for DLL filenames. Since there are two filenames and fixed parameters (e.g. "%1$") are not portable here,   I'd have to repeat things twice.

Any objection to removing the filenames and simply referring to "json" and "mongodb" extensions?

jmikola:
> I would like to test this with a normal PHP, before approving this.

I'll await your LGTM on that.

php_phongo.c
```diff
@@ -1886,6 +1909,25 @@ PHP_MINIT_FUNCTION(mongodb)
...
+	 * Although JsonSerializable was introduced in PHP 5.4.0,
+	 * php_json_serializable_ce is not exported in PHP versions before 5.4.26
+	 * and 5.5.10. For later PHP versions, looking up the class manually also
+	 * helps with distros that disable LTDL_LAZY for dlopen() (e.g. Fedora).
+	 */
+	php_phongo_date_immutable_ce = php_phongo_fetch_internal_class(ZEND_STRL("datetimeimmutable") TSRMLS_CC);
+	php_phongo_json_serializable_ce = php_phongo_fetch_internal_class(ZEND_STRL("jsonserializable") TSRMLS_CC);
+
+	if (php_phongo_json_serializable_ce == NULL) {
+		zend_error(E_ERROR, "JsonSerializable class not defined. Please ensure that the json extension is loaded before the mongodb extension.");
```
jmikola: > The error message should also be fixed.

Rewritten.

derickr: How would I ever trigger this error here?

Besides this, and this is a nitpick, the normal warning says:

```
Cannot load module 'mongodb' because required module 'json' is not loaded
```

I.e., the word used is "module" and not "extension", and it encapsulates the names in '. Would it make sense to use this here too? But then again, I need to be able to trigger this error too. I've tried with 5.5.8 and here this doesn't hit either.

remicollet: @derickr I think you cannot trigger this error in MINIT which is never if dependencies are not satisfied (see my test results above)

jmikola: Assuming our MINIT is only ever called through [`zend_startup_module_ex()`](https://github.com/php/php-src/blob/PHP-7.0.0/Zend/zend_API.c#L1794), I agree that this is probably redundant. The class is always available in PHP 5.4.0+ (https://github.com/php/php-src/commit/1381b14eaaa6d0f3dcbc9b9a2c8116c2ecfac2c4), so we should never fail to find it on any supported PHP version. Should I remove the check entirely?

derickr: No, I think it should stay.

PR #504, opened 2016-12-29
PHPC-884: Do not skip public property with empty string name during BSON encoding

jmikola:
https://jira.mongodb.org/browse/PHPC-884
https://jira.mongodb.org/browse/PHPC-886
https://jira.mongodb.org/browse/PHPC-891

### DIFF
--- a/src/bson.c
+++ b/src/bson.c
@@ -1271,44 +1271,6 @@ try_again:
 	}
 }
 
-#if PHP_VERSION_ID >= 70000
-static bool is_public_property(zend_class_entry *ce, zend_string *name, zend_string **member TSRMLS_DC) /* {{{ */
-#else
-static bool is_public_property(zend_class_entry *ce, const char *prop_name, int prop_name_len TSRMLS_DC) /* {{{ */
-#endif
-{
-	zend_property_info *property_info = NULL;
-
-#if PHP_VERSION_ID >= 70000
-	if (ZSTR_VAL(name)[0] == 0) {
-		const char *prop_name,
-			 *class_name;
-		size_t prop_name_len;
-
-		zend_unmangle_property_name_ex(name,
-			&class_name, &prop_name, &prop_name_len);
-		(*member) = zend_string_init(prop_name, prop_name_len, 0);
-	} else (*member) = zend_string_copy(name);
-	property_info = zend_get_property_info(ce, (*member), 1 TSRMLS_CC);
-
-	if (!property_info) /* undefined property */
-		return true;
-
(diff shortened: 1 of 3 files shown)


tests/bson/bson-fromPHP-006.phpt
```diff
@@ -0,0 +1,50 @@
+--TEST--
+BSON\fromPHP(): PHP documents with null bytes in field name
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/tools.php';
+
+echo "\nTesting array with one leading null byte in field name\n";
+hex_dump(fromPHP(["\0" => 1]));
```
jmikola: While implementing this test, I reviewed [field name restrictions](https://docs.mongodb.com/manual/core/document/#field-names) for MongoDB documents, which prohibits `"$"` and `"."` characters in addition to null bytes. That said, the [BSON specification](http://bsonspec.org/spec.html) only prohibits null bytes on account of using C strings to encode field names, so I see no reason to enforce the other restrictions here. Validation of documents destined for MongoDB storage falls under [PHPC-712](https://jira.mongodb.org/browse/PHPC-712).

I'm also comfortable relying on `strlen()` to truncate the field name here instead of throwing an exception because the user included null bytes within a field name; however, I'd be happy to reconsider that. I expect it'd simply be a matter of comparing the string length reported by PHP's HashTable API with that of `strlen()` and throwing if there is a discrepancy.

derickr: I agree that should throw an exception. Silent data truncation is scary.

src/bson.c
```diff
@@ -1330,17 +1280,16 @@ void phongo_zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *bson
...
+				if (string_key[0] == '\0' && string_key_len > 1) {
 					continue;
 				}
-			} else {
-				/* Chop off the \0 from string lengths */
-				string_key_len -= 1;
+			}
+
+			if (strlen(string_key) != string_key_len - 1) {
+				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "BSON keys cannot contain null bytes. Unexpected null byte after \"%s\".", ZSTR_VAL(string_key));
```
jmikola: Added this exception message. This appears after we might skip object properties whose name contains a leading null byte (i.e. PHP's encoding for protected and private members). For arrays, the check happens on all string keys.

We don't check this for integer keys, since we convert those to strings ourself and don't have to worry about null bytes.

PR #506, opened 2016-12-29
PHPC-885: Alphabetize Regex flags when instantiating from BSON

jmikola:
https://jira.mongodb.org/browse/PHPC-885

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1666,6 +1666,15 @@ void php_phongo_new_decimal128(zval *object, const bson_decimal128_t *decimal TS
 	intern->initialized = true;
 } /* }}} */
 
+/* qsort() compare callback for alphabetizing regex flags upon initialization */
+static int php_phongo_regex_compare_flags(const void *f1, const void *f2) {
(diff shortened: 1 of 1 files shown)


php_phongo.c
```diff
@@ -1683,6 +1692,10 @@ void php_phongo_new_regex_from_regex_and_options(zval *object, const char *patte
 	intern->pattern = estrndup(pattern, intern->pattern_len);
 	intern->flags_len = strlen(flags);
 	intern->flags = estrndup(flags, intern->flags_len);
+
+	/* Ensure flags are alphabetized upon initialization. This may be removed
+	 * once CDRIVER-1883 is implemented. */
+	qsort((void *) intern->flags, intern->flags_len, 1, php_phongo_regex_compare_flags);
```
jmikola: I realize this duplicates code from Regex.c; however, it should only be temporary until this is handled in libbson.

I also anticipate refactoring BSON encoding and decoding down the line, which will remove all of the `php_phongo_new_xxx()` functions from this file in favor of using the init functions in the actual BSON class source files.

PR #515, opened 2017-01-23
PHPC-894: Implement get_gc handlers for BSON classes

jmikola:
https://jira.mongodb.org/browse/PHPC-894

This addresses a memory leak if gc_possible_root() were to invoke the existing get_properties handlers.

### DIFF
--- a/src/BSON/ObjectID.c
+++ b/src/BSON/ObjectID.c
@@ -373,6 +373,14 @@ static int php_phongo_objectid_compare_objects(zval *o1, zval *o2 TSRMLS_DC) /* 
 	return strcmp(intern1->oid, intern2->oid);
 } /* }}} */
 
+static HashTable *php_phongo_objectid_get_gc(zval *object, phongo_get_gc_table table, int *n TSRMLS_DC) /* {{{ */
+{
+	*table = NULL;
+	*n = 0;
+
+	return zend_std_get_properties(object TSRMLS_CC);
+} /* }}} */
+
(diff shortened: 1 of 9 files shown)


src/BSON/ObjectID.c
```diff
@@ -384,7 +392,7 @@ HashTable *php_phongo_objectid_get_properties(zval *object TSRMLS_DC) /* {{{ */
 	if (!intern->oid) {
 		return props;
 	}
-
+fprintf(stderr, "%p\n", (void *)intern);
```
bjori: typo alert

jmikola: Thanks!

tests/bson/bug0894-001.phpt
```diff
@@ -0,0 +1,43 @@
...
+    function() { return new MongoDB\BSON\UTCDateTime; },
+];
+
+foreach ($factoryMethods as $factoryMethod) {
+    $objects = [];
+
+    /* GC_ROOT_BUFFER_MAX_ENTRIES is defined as 10000 on PHP 5.x and 10001 on
+     * PHP 7.1, so use the larger value to always trigger gc_possible_root(). */
+    for ($i = 0; $i < 10001; $i++) {
+        $objects[] = $factoryMethod();
```
derickr: You probably don't have to do this. You can call it once or twice, and then call [gc_collect_cycles()](http://php.net/manual/en/function.gc-collect-cycles.php). But not important enough to change.

PR #518, opened 2017-01-24
Refactor construction of developer build flags in config.m4

jmikola:


config.m4
```diff
@@ -129,12 +124,13 @@ if test "$MONGODB" != "no"; then
     PHP_CHECK_GCC_ARG(-Wempty-body,                     _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wempty-body")
     PHP_CHECK_GCC_ARG(-Wparentheses,                    _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wparentheses")
     PHP_CHECK_GCC_ARG(-Wdeclaration-after-statement,    _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wdeclaration-after-statement")
+    PHP_CHECK_GCC_ARG(-Werror,                          _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Werror")
+    PHP_CHECK_GCC_ARG(-Wextra,                          _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wextra")
+    PHP_CHECK_GCC_ARG(-Wno-unused-parameter,            _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wno-unused-parameter")
+    PHP_CHECK_GCC_ARG(-Wno-unused-but-set-variable,     _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wno-unused-but-set-variable")
+    PHP_CHECK_GCC_ARG(-Wno-missing-field-initializers,  _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wno-missing-field-initializers")
 
-    if test "$PHP_FOUND_VERNUM" -ge "50400"; then
```
jmikola: We don't support anything below PHP 5.4.0, so this was redundant.

config.m4
```diff
@@ -85,11 +85,6 @@ AC_DEFUN([PHP_BSON_CLOCK],
 
 MONGOC_SYMBOL_SUFFIX="priv"
 
-AC_MSG_CHECKING(PHP version)
-PHP_FOUND_VERSION=`${PHP_CONFIG} --version`
-PHP_FOUND_VERNUM=`echo "${PHP_FOUND_VERSION}" | $AWK 'BEGIN { FS = "."; } { printf "%d", ([$]1 * 100 + [$]2) * 100 + [$]3;}'`
-AC_MSG_RESULT($PHP_FOUND_VERNUM)
-
```
derickr: I think we should keep this check, as it shows up in the configure output as a good hint of what's going on. Right now we might not use it, but we may be in the future again.

jmikola:
Verified that the `config.w32` changes still build successfully on Windows.

PR #524, opened 2017-02-02
PHPC-712: Validate BSON documents for insert and update

jmikola:
https://jira.mongodb.org/browse/PHPC-712
https://jira.mongodb.org/browse/PHPC-944

Ensure that insert and update/replacement documents are validated. Additionally, detect insert documents that appear to be legacy indexes and allow dots in BSON keys (via libmongoc's new `legacyIndex` option in [CDRIVER-2018](https://jira.mongodb.org/browse/CDRIVER-2018)).

We now translate `MONGOC_ERROR_COMMAND_INVALID_ARG` to an InvalidArgumentException, which preserves the expected exception for BulkWrite argument validation. This change did not affect other areas of the test suite.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -145,6 +145,9 @@ zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_d
 		case MONGOC_ERROR_CLIENT_AUTHENTICATE:
 			return php_phongo_authenticationexception_ce;
 
+		case MONGOC_ERROR_COMMAND_INVALID_ARG:
+			return php_phongo_invalidargumentexception_ce;
+
 		case MONGOC_ERROR_STREAM_INVALID_TYPE:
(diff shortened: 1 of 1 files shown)


src/MongoDB/BulkWrite.c
```diff
@@ -31,6 +31,20 @@
 
 zend_class_entry *php_phongo_bulkwrite_ce;
 
+/* Returns whether the insert document appears to be a legacy index. */
+static inline bool php_phongo_bulkwrite_insert_is_legacy_index(bson_t *bdocument) /* {{{ */
```
jmikola: It proved much easier to inspect the BSON document than the original zval argument. The only reason I considered checking the zval was because we might want to disable `_id` generation for these documents, but pre-2.6 servers don't seem to care (the `_id` field never shows up in `system.indexes` document). Also, we would fail to properly generate an `_id` if this function detected a false positive (`bulkwrite-insert-003.phpt` tests for this).

PR #528, opened 2017-02-08
PHPC-909: Spring cleaning for 1.3.0

jmikola:
https://jira.mongodb.org/browse/PHPC-909


config.m4
```diff
@@ -153,6 +153,7 @@ if test "$MONGODB" != "no"; then
     php_phongo.c \
     phongo_compat.c \
     src/bson.c \
```
derickr: Don't want to rename this to src/bson-decode.c?

src/BSON/Binary.c
```diff
@@ -90,7 +90,7 @@ static bool php_phongo_binary_init_from_hash(php_phongo_binary_t *intern, HashTa
 
 /* {{{ proto void Binary::__construct(string $data, int $type)
    Construct a new BSON binary type */
-PHP_METHOD(Binary, __construct)
+static PHP_METHOD(Binary, __construct)
```
derickr: I think this is going to break on Windows - also, as no other extension in PHP does this AFAIK, I don't think we should do this.

config.m4
jmikola: I plan on doing so, but it currently still has the generic BSON functions in it.

src/BSON/Binary.c
jmikola: I spoke with @saramg and she uses this method in all of her extensions. I also found evidence of it in core (see: [here](https://php-lxr.adamharvey.name/source/search?q=%22static+PHP_METHOD%22&defs=&refs=&path=&hist=&type=&project=PHP-5.4&project=PHP-7.0)).

PR #545, opened 2017-03-03
PHPC-302: Throw exceptions for unsupported BSON types

jmikola:
https://jira.mongodb.org/browse/PHPC-302

Previously, the undefined, DBPointer, and symbol types were ignored since they were known to libbson and no visitor function was defined.

### DIFF
--- a/src/bson.c
+++ b/src/bson.c
@@ -137,6 +137,14 @@ static bool php_phongo_bson_visit_binary(const bson_iter_t *iter ARG_UNUSED, con
 	return false;
 } /* }}} */
 
+static bool php_phongo_bson_visit_undefined(const bson_iter_t *iter, const char *key, void *data) /* {{{ */
+{
+	TSRMLS_FETCH();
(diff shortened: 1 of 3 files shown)


src/bson.c
```diff
@@ -137,6 +137,14 @@ static bool php_phongo_bson_visit_binary(const bson_iter_t *iter ARG_UNUSED, con
 	return false;
 } /* }}} */
 
+static bool php_phongo_bson_visit_undefined(const bson_iter_t *iter, const char *key, void *data) /* {{{ */
+{
+	TSRMLS_FETCH();
+	phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Detected unsupported BSON type 0x06 (undefined) for fieldname \"%s\"", key);
+
+	return true;
```
jmikola: Return `false` here so that BSON visitation can continue and we can report multiple exceptions for a single document. Subsequent exceptions will be accessible via `getPrevious()`.

PR #549, opened 2017-03-09
PHPC-875: Allow compiling without SSL

jmikola:
https://jira.mongodb.org/browse/PHPC-875

Supersedes https://github.com/mongodb/mongo-php-driver/pull/404


php_phongo.c
```diff
@@ -1289,6 +1289,7 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t *uri, bson_t *option
 	return true;
 } /* }}} */
 
+#ifdef MONGOC_ENABLE_SSL
```
jmikola: #404 used `#if MONGOC_ENABLE_SSL`, but `mongoc-config.h` will `#undef` the constant if is not equal to `1`.

php_phongo.c
```diff
@@ -1599,20 +1593,35 @@ void phongo_manager_init(php_phongo_manager_t *manager, const char *uri_string,
...
+	 * be thrown for unsupported driver options. */
 	ssl_opt = php_phongo_make_ssl_opt(driverOptions TSRMLS_CC);
 
 	/* An exception may be thrown during SSL option creation */
 	if (EG(exception)) {
 		goto cleanup;
 	}
+#else
+	if (mongoc_uri_get_ssl(uri)) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Cannot create SSL client. SSL is not enabled in this build.");
```
jmikola: Although `mongoc_client_new_from_uri()` (invoked from `php_phongo_make_mongo_client()`) will log an error and return `NULL` if the SSL option is specified but SSL is not enabled in the build, it is better to catch this preemptively. The aforementioned error logging may or may not throw a specific exception depending on whether `mongodb.error` is enabled (see [PHPC-930](https://jira.mongodb.org/browse/PHPC-930)), but we can always throw a specific exception here.

php_phongo.c
derickr: I can't see where the php_phongo_fetch_ssl_opt_string function is used though, so please check and make sure it is only called when #ifdef MONGOC_ENABLE_SSL is true as well.

derickr: OK, found it. It's just that github didn't include the code between lines 1295 and 1398.

PR #557, opened 2017-03-13
PHPC-920: Regression test for mongoc_cursor_destroy() crash

jmikola:
https://jira.mongodb.org/browse/PHPC-920

### DIFF
--- a/tests/cursor/bug0920-001.phpt
+++ b/tests/cursor/bug0920-001.phpt
@@ -1,0 +1,24 @@
+--TEST--
+PHPC-924: Using a projection with an empty field name causes a crash when destroying cursor
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; CLEANUP(STANDALONE) ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(STANDALONE);
+
+$query = new MongoDB\Driver\Query([], ['projection' => ['' => 1]]);
+
+echo throws(function() use ($manager, $query) {
+    $cursor = $manager->executeQuery(NS, $query);
+    var_dump($cursor->toArray());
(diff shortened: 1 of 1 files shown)


tests/cursor/bug0920-001.phpt
```diff
@@ -0,0 +1,24 @@
...
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(STANDALONE);
+
+$query = new MongoDB\Driver\Query([], ['projection' => ['' => 1]]);
+
+echo throws(function() use ($manager, $query) {
+    $cursor = $manager->executeQuery(NS, $query);
```
jmikola: With libmongoc 1.5.4, the segfault occurs here when the cursor is checked for an error, an exception is thrown, and the cursor is destroyed. In 1.5.5, the segfault is fixed but we still throw the exception here.

I think it might make more sense to do the same BSON validation that libmongoc does during cursor initialization in our own Query constructor, so we can instead throw an InvalidArgumentException earlier. This duplicates the validation logic, but the filter and options documents should be small enough where this isn't a concern. If this sound reasonable, I'll open a new ticket to address this for 1.3.0.

derickr: It depends a little bit how much code it involves. I'm generally sceptical about doing these in two places. In general, I do think it's better to have an InvalidArgumentException here.

jmikola: To clarify, we are currently limited to a RuntimeException when `executeQuery()` is called, because `_mongoc_cursor_new_with_opts()` reports an error with the `MONGOC_ERROR_CURSOR` domain and `MONGOC_ERROR_CURSOR_INVALID_CURSOR` code.

Allowing us to throw an InvalidArgumentException at Query construct time should require less than ten lines of code. I'll create a ticket and submit a PR soon.

jmikola: https://github.com/mongodb/mongo-php-driver/pull/558

PR #561, opened 2017-03-15
PHPC-640: Implement interfaces for userland BSON type classes

jmikola:
https://jira.mongodb.org/browse/PHPC-640
https://jira.mongodb.org/browse/PHPC-937

Related:

 * https://github.com/mongodb/mongo-hhvm-driver/blob/master/serialization.rst#scalar-types
 * https://github.com/mongodb/mongo-hhvm-driver/pull/111

### DIFF
--- a/php_bson.h
+++ b/php_bson.h
@@ -69,7 +69,7 @@ int php_phongo_bson_to_zval(const unsigned char *data, int data_len, zval *out);
 #else
 int php_phongo_bson_to_zval(const unsigned char *data, int data_len, zval **out);
 #endif
-void php_phongo_bson_typemap_to_state(zval *typemap, php_phongo_bson_typemap *map TSRMLS_DC);
+bool php_phongo_bson_typemap_to_state(zval *typemap, php_phongo_bson_typemap *map TSRMLS_DC);
 
 PHP_FUNCTION(MongoDB_BSON_toPHP);
 PHP_FUNCTION(MongoDB_BSON_fromPHP);
--- a/php_bson.h
+++ b/php_bson.h
@@ -60,11 +60,7 @@ typedef struct {
 	bool                     is_visiting_array;
 } php_phongo_bson_state;
 
-#if PHP_VERSION_ID >= 70000
-#define PHONGO_BSON_STATE_INITIALIZER  { {{ 0 }}, { PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL}, NULL, 0 }
-#else
-#define PHONGO_BSON_STATE_INITIALIZER  { NULL, { PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL}, NULL, 0 }
-#endif
+#define PHONGO_BSON_STATE_INITIALIZER  { 0 }
 
 void php_phongo_zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *bson, bson_t **bson_out TSRMLS_DC);
 int php_phongo_bson_to_zval_ex(const unsigned char *data, int data_len, php_phongo_bson_state *state);
--- a/php_bson.h
+++ b/php_bson.h
@@ -47,6 +47,15 @@ typedef struct {
 	zend_class_entry              *array;
 	php_phongo_bson_typemap_types  root_type;
 	zend_class_entry              *root;
+	zend_class_entry              *binary;
+	zend_class_entry              *decimal128;
+	zend_class_entry              *javascript;
+	zend_class_entry              *maxkey;
+	zend_class_entry              *minkey;
+	zend_class_entry              *objectid;
+	zend_class_entry              *regex;
+	zend_class_entry              *timestamp;
+	zend_class_entry              *utcdatetime;
 } php_phongo_bson_typemap;
 
 typedef struct {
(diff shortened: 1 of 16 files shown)


php_bson.h
```diff
@@ -60,11 +69,7 @@ typedef struct {
 	bool                     is_visiting_array;
 } php_phongo_bson_state;
 
-#if PHP_VERSION_ID >= 70000
-#define PHONGO_BSON_STATE_INITIALIZER  { {{ 0 }}, { PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL}, NULL, 0 }
-#else
-#define PHONGO_BSON_STATE_INITIALIZER  { NULL, { PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL}, NULL, 0 }
-#endif
+#define PHONGO_BSON_STATE_INITIALIZER  { 0 }
```
derickr: I don't believe this is right. I had some compilers whine about this change to just have a single { 0 } in Xdebug: https://github.com/xdebug/xdebug/commit/7ad9db738bfa80d69621e4c7f66a04cd361269f9

php_phongo_classes.h
```diff
@@ -242,18 +242,19 @@ extern zend_class_entry *php_phongo_connectiontimeoutexception_ce;
...
-extern zend_class_entry *php_phongo_persistable_ce;
-extern zend_class_entry *php_phongo_unserializable_ce;
-extern zend_class_entry *php_phongo_serializable_ce;
 extern zend_class_entry *php_phongo_binary_ce;
 extern zend_class_entry *php_phongo_decimal128_ce;
 extern zend_class_entry *php_phongo_javascript_ce;
 extern zend_class_entry *php_phongo_maxkey_ce;
 extern zend_class_entry *php_phongo_minkey_ce;
 extern zend_class_entry *php_phongo_objectid_ce;
+extern zend_class_entry *php_phongo_persistable_ce;
```
derickr: I don't think it makes sense to reorganise this. The specific classes' ce's should be together, but not mixed in with their interfaces. The rest of the CE's isn't alphabetised either.

src/bson-encode.c
```diff
@@ -113,7 +116,32 @@ static int php_phongo_is_array_or_document(zval *val TSRMLS_DC) /* {{{ */
...
+		zend_call_method_with_0_params(object, NULL, NULL, "toBSONType", &retval);
+#else
+		zval *retval;
+
+		zend_call_method_with_0_params(&object, NULL, NULL, "toBSONType", &retval);
+#endif
+
+		if (Z_ISUNDEF(retval)) {
+			/* zend_call_method() failed */
+			return;
```
derickr: Shouldn't that throw an exception?

src/bson.c
```diff
@@ -35,6 +35,30 @@
...
+# define PHONGO_BSON_STATE_ZCHILD(state) (&((php_phongo_bson_state *)(state))->zchild)
+#else
+# define PHONGO_BSON_STATE_ZCHILD(state) (((php_phongo_bson_state *)(state))->zchild)
+#endif
+
+#if PHP_VERSION_ID >= 70000
+# define PHONGO_BSON_TYPE_WRAPPER(wrapper_ce, zchild) \
+	if (wrapper_ce) { \
+		zval zwrapper; \
+		zend_call_method_with_1_params(NULL, wrapper_ce, NULL, "createfrombsontype", &zwrapper, &(zchild)); \
```
derickr: Do you not have a constant for "createfrombsontype" ?

src/bson-encode.c
jmikola: [`zend_call_method()`](https://github.com/php/php-src/blob/PHP-7.1.4/Zend/zend_interfaces.c#L34) uses `zend_error_noreturn()`. I suppose PHP 7 might throw an Error object from that, but I expect basic error reporting in PHP 5.x.

This is consistent with other handling we do for potential `zend_call_method()` failures (e.g. `bsonSerialize()`).

src/bson.c
jmikola: I did not create a constant like we had for `BSON_UNSERIALIZE_FUNC_NAME`, as I recall some case where `zend_call_method()` could not find the function when using `"createFromBSONType"`. I wasn't able to determine at the time why that wasn't an issue for `BSON_UNSERIALIZE_FUNC_NAME` (defined as `"bsonUnserialize"`), but I'll check again.

php_bson.h
jmikola: Do you recall what compilers were affected (e.g. clang, MS)?

I can always revert this, but I'd be curious to test it.

php_phongo_classes.h
jmikola: The `MongoDB\Driver\Exception` CEs are not alphabetized, but `MongoDB\Driver` and `MongoDB\Driver\Monitoring` CEs were, as were all externs for the init functions. I'll revert this change, though.

src/bson.c
jmikola: Confirmed that `zend_call_method()` allows `function_name` to have any case when invoking a method on an object (specifying the first `obj` arg); however, invoking a static method when only supplying `obj_ce` requires the function name to be lowercase. I assume this has something to do with differences in storing the function entries on the CE and object instances.

In any event, these strings are only used twice in these two macros, unlike the `bsonSerialize` and `bsonUnserialize` constants, so I will not create a new constant.

php_bson.h
jmikola: Note that we're already using `{0}` for initializing potentially nested structs (e.g. `smart_str` in PHP 7: https://github.com/mongodb/mongo-php-driver/blob/608ce77839e1760f9a7272aa61d1158d6bcd741c/src/BSON/ObjectID.c#L199) and we've received no complaints about build issues.

http://stackoverflow.com/a/1761873/162228 suggests that this is kosher.

Issue #562, opened 2017-03-17
Can't install with pickle


rageshkrishna:
### Description

Running `pickle install mongodb` results in the following error:

````
root@00b966884cbc:~# pickle install mongodb
  - Installing mongodb (latest-stable): Downloading (100%)


  [Exception]
  Couldn't parse the version defined in the PHP_MONGODB_VERSION macro


install [--no-convert] [--defaults] [--source] [--with-configure-options WITH-CONFIGURE-OPTIONS] [--save-logs SAVE-LOGS] [--dry-run] [--php PHP] [--ini INI] [--tmp-dir TMP-DIR] [--] [<path>]
````

I suspect pickle is looking for `PHP_MONGODB_VERSION` but mongo-php-driver doesn't define it anywhere. To be honest, I'm not sure if this is a bug in mongo-php-driver or in pickle, but it looks like pickle is able to find it on several other extensions I am installing so is it some sort of standard approach that is not adopted by mongo-php-driver?

### Environment

Ubuntu 16.04 docker container, PHP 7.1 installed with php-build and managed with phpenv.

### Test Script

Fails on install with the command above. No other test script,

### Expected and Actual Behavior

Expected the extension to be installed, but it did not.

jmikola:
`PHP_EXTNAME_VERSION` is included PHP's [skeleton header for extensions](https://github.com/php/php-src/blob/php-7.1.3/ext/skeleton/php_skeleton.h#L9). I don't recall if we used that when creating this extension, but this is easy to address. I've opened [PHPC-936](https://jira.mongodb.org/browse/PHPC-936) to add the missing macro.

----

@weltling, @pierrejoye: I believe the `PHP_EXTNAME_VERSION` convention dates back to https://github.com/php/php-src/commit/dcfd753104085d8337b8a5f4a21ce3b629ad4651, but can you explain why Pickle needs to derive a version from extension headers? I realize you need the version before building the extension, so `phpversion('extname')` is not an option, but Composer pulls version info from Git tags. Since you support both Git and HTTP sources (I assume that means PECL's index), wouldn't you simply pull version info from tags and PECL's database, respectively?

pierrejoye:
Extension version in header is a norm. Used in phpinfo() or in extension
versions constants.

We added this check to avoid releases without it.

pierrejoye:
On top of it we make sure header matches with desired versions.

jmikola:
@pierrejoye: Thanks for the quick reply.

[...]

PR #563, opened 2017-03-17
PHPC-936: Define PHP_MONGODB_VERSION and STABILITY constants

jmikola:
https://jira.mongodb.org/browse/PHPC-936

This removes the redundant MONGODB_VERSION constant renames the existing version and stability constants to be consistent with the format in PHP's extension skeleton: https://github.com/php/php-src/commit/dcfd753104085d8337b8a5f4a21ce3b629ad4651

Fixes: https://github.com/mongodb/mongo-php-driver/issues/562

### DIFF
--- a/php_phongo.h
+++ b/php_phongo.h
@@ -25,9 +25,8 @@
 extern zend_module_entry mongodb_module_entry;
 
 /* FIXME: Its annoying to bump version. Move into phongo_version.h.in */
-#define MONGODB_VERSION_S "1.2.8-dev"
-#define MONGODB_STABILITY_S "devel"
-#define MONGODB_VERSION   MONGODB_VERSION_S
+#define PHP_MONGODB_VERSION "1.2.8-dev"
+#define PHP_MONGODB_STABILITY "devel"
 
 #ifdef PHP_WIN32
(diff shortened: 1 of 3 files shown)


php_phongo.h
```diff
@@ -25,9 +25,8 @@
 extern zend_module_entry mongodb_module_entry;
 
 /* FIXME: Its annoying to bump version. Move into phongo_version.h.in */
-#define MONGODB_VERSION_S "1.2.8-dev"
-#define MONGODB_STABILITY_S "devel"
-#define MONGODB_VERSION   MONGODB_VERSION_S
+#define PHP_MONGODB_VERSION "1.2.8-dev"
+#define PHP_MONGODB_STABILITY "devel"
```
jmikola: Arguments of the [extension skeleton](https://github.com/php/php-src/blob/php-7.1.3/ext/skeleton/php_skeleton.h#L9) aside, these names are also more consistent with the `MONGODB_VERSION` and `MONGODB_STABILITY` constants we define for userland.

As for `MONGODB_VERSION`, I found no references to it apart from `zend_module_entry`. This seemed ripe for removal after updated the struct to use `PHP_MONGODB_VERSION`.

PR #568, opened 2017-03-28
PHPC-349: Fixes and code cleanup

jmikola:
Related to: https://github.com/mongodb/mongo-php-driver/pull/466

### DIFF
--- a/php_phongo_structs.h
+++ b/php_phongo_structs.h
@@ -204,6 +204,18 @@ typedef struct {
 	PHONGO_ZEND_OBJECT_PRE
 	mongoc_client_t    *client;
 	char               *command_name;
+	int                 server_id;
+	uint64_t            operation_id;
+	uint64_t            request_id;
+	uint64_t            duration_micros;
+	PHONGO_STRUCT_ZVAL  z_error;
+	PHONGO_ZEND_OBJECT_POST
+} php_phongo_commandfailedevent_t;
+
+typedef struct {
+	PHONGO_ZEND_OBJECT_PRE
+	mongoc_client_t    *client;
+	char               *command_name;
 	uint32_t            server_id;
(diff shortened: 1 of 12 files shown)


php_phongo_structs.h
```diff
@@ -204,25 +204,25 @@ typedef struct {
 	PHONGO_ZEND_OBJECT_PRE
 	mongoc_client_t    *client;
 	char               *command_name;
-	uint32_t            server_id;
+	int                 server_id;
```
derickr: This is curious. For some of the event_ts we use int, and sometimes uint32_t for this. One of them is right...

src/MongoDB/Monitoring.h
```diff
@@ -17,6 +17,8 @@
 #ifndef PHONGO_MONITORING_H
 #define PHONGO_MONITORING_H
 
+#include <php.h>
```
derickr: Why did you add this?

php_phongo_structs.h
jmikola: The server and cursor structs use `int`. I'm not sure where `uint32_t` came from.

I believe this is only a diff because I re-ordered the struct definitions to be alphabetical; however, it'd be a good idea to make them consistent.

src/MongoDB/Monitoring.h
jmikola: `php.h` is where `PHP_FUNCTION()` is defined. This wasn't an error before as you generally have `php.h` included before this file, but I considered it good practice to have headers include their dependencies.

php_phongo_structs.h
derickr: Can you fix this one in your PR so I can merge it cleanly then?

derickr:
These last two commits made Travis unhappy...

PR #570, opened 2017-03-31
PHPC-904: ReadPreference constructor should accept strings

derickr:


src/MongoDB/ReadPreference.c
```diff
@@ -45,24 +45,43 @@ static PHP_METHOD(ReadPreference, __construct)
...
+			case MONGOC_READ_SECONDARY_PREFERRED:
+			case MONGOC_READ_NEAREST:
+				intern->read_preference = mongoc_read_prefs_new(Z_LVAL_P(mode));
+				break;
+			default:
+				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Invalid mode: %" PHONGO_LONG_FORMAT, Z_LVAL_P(mode));
+				return;
+		}
+	} else if (Z_TYPE_P(mode) == IS_STRING) {
+		if (strcmp(Z_STRVAL_P(mode), "primary") == 0) {
```
jmikola: libmongoc uses [`strcasecmp()`](https://github.com/mongodb/mongo-c-driver/blob/b932ad7dd602a3b098b7a2d239475fbedffde2dd/src/mongoc/mongoc-uri.c#L998), which I think would be worth doing here as well.

jmikola: I'll note that libmongoc also uses `strcasecmp()` for WriteConcern's "majority" string. Perhaps we want to do that as well in the WriteConcern constructor? We currently have:

```
if (strcmp(Z_STRVAL_P(w), PHONGO_WRITE_CONCERN_W_MAJORITY) == 0) {
```

derickr: Happy to change that, I had just followed the spec, which says these are case sensitive I believe.

jmikola: Can you link me to the the case-sensitive requirement when you get a chance? I only found the following from the read preference spec's section on [mongos](https://github.com/mongodb/specifications/blob/master/source/driver-read-preferences.rst#communication-with-mongos):

> mode: the mode field must be present exactly once and have a lowercase string value, a valid mode in camel case ("primary", "secondary", "primaryPreferred", etc.)

derickr: Can't find that now either - but I've made the change to strcasecmp.

tests/readPreference/readpreference-ctor-002.phpt
```diff
@@ -0,0 +1,102 @@
...
+	[ MongoDB\Driver\ReadPreference::RP_SECONDARY_PREFERRED ],
+
+	[ "nearest" ],
+	[ "NeaRest" ],
+	[ MongoDB\Driver\ReadPreference::RP_NEAREST ],
+];
+
+foreach ($data as $item) {
+	$tags = NULL;
+	@list($rp, $tags) = $item;
```
jmikola: Let's delete this dead code.

src/MongoDB/ReadPreference.c
```diff
@@ -99,7 +118,7 @@ static PHP_METHOD(ReadPreference, __construct)
 				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected maxStalenessSeconds to be <= %" PRId32 ", %" PHONGO_LONG_FORMAT " given", INT32_MAX, maxStalenessSeconds);
 				return;
 			}
-			if (mode == MONGOC_READ_PRIMARY) {
+			if ((Z_TYPE_P(mode) == IS_LONG && Z_LVAL_P(mode) == MONGOC_READ_PRIMARY) || (Z_TYPE_P(mode) == IS_STRING && strcmp(Z_STRVAL_P(mode), "primary") == 0)) {
```
jmikola: Since we already assigned mode here, can we simply use `mongoc_read_prefs_get_mode(intern->read_preference)` instead of duplicating the integer and string comparison?

derickr: Updated, and also fixed the test case.

jmikola: Shouldn't this also be `mongoc_read_prefs_get_mode(intern->read_preference) == MONGOC_READ_PRIMARY`?

jmikola:
LGTM with the last change addressed (using `mongoc_read_prefs_get_mode()` during `maxStalenessSeconds` validation).

PR #575, opened 2017-04-06
PHPC-948: BSON encoding should throw on circular references

jmikola:
https://jira.mongodb.org/browse/PHPC-948


src/bson.c
```diff
@@ -1364,19 +1366,6 @@ void phongo_zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *bson
...
-#endif
-			if (!Z_ISUNDEF(obj_data)) {
-				zval_ptr_dtor(&obj_data);
-			}
-#if PHP_VERSION_ID >= 70000
-		}
-#endif
-		return;
-	}
-
```
jmikola: This block was only necessary for cleaning up after the exception when `bsonSerialize()` does not return an array or stdClass. I moved the `if (!Z_ISUNDEF(obj_data)) { zval_ptr_dtor(&obj_data); }` condition up to immediately follow that exception and changed the original `break` to a `return`.

I don't believe the Serializable `instanceof` for PHP 7 was especially relevant, as `zval *obj_data` is initialized to `NULL` on PHP 5.x and only ever assigned when we're dealing with a Serializable (it stores the return value of `bsonSerialize()`. AFAICT, this was dead code.

src/bson.c
```diff
@@ -1296,7 +1294,7 @@ void phongo_zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *bson
 
 				if (Z_ISUNDEF(obj_data)) {
 					/* zend_call_method() failed */
-					break;
+					return;
```
jmikola: With the original `break`, this would have returned in the block I removed below since `ht_data` is not initialized at this point. If `zend_call_method()` fails, we either had an exception from `bsonSerialize()` or a PHP error from `zend_call_method()` itself. In any event, the behavior is the same in that we're not throwing any additional here.

src/bson.c
```diff
@@ -1349,7 +1351,7 @@ void phongo_zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *bson
 			if (instanceof_function(Z_OBJCE_P(data), php_phongo_type_ce TSRMLS_CC)) {
 				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "%s instance %s cannot be serialized as a root element", ZSTR_VAL(php_phongo_type_ce->name), ZSTR_VAL(Z_OBJCE_P(data)->name));
 
-				break;
+				return;
```
jmikola: This also would have simply returned in the removed block below, since `ht_data` is not initialized at this point. There is also no `obj_data` to clean up.

tests/bson/bson-fromPHP-004.phpt
```diff
@@ -1,38 +0,0 @@
...
-<?php exit(0); ?>
---EXPECTF--
-Testing packed array with circular reference
-{ "x" : 1, "y" : [ [  ] ] }
-
-Testing associative array with circular reference
-{ "x" : 1, "y" : { "z" : {  } } }
-
-Testing object with circular reference
-{ "x" : 1, "y" : { "z" : {  } } }
```
jmikola: Per the [PHPC-948](https://jira.mongodb.org/browse/PHPC-948) issue description, circular references should result in exceptions. This test has been modified and moved to `bson-fromPHP_error-004.phpt`.

tests/bson/bson-fromPHP_error-004.phpt
```diff
@@ -0,0 +1,47 @@
...
+    fromPHP($document);
+}, 'MongoDB\Driver\Exception\UnexpectedValueException'), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+Testing packed array with circular reference
+OK: Got MongoDB\Driver\Exception\UnexpectedValueException
+Detected recursion for fieldname "0"
```
jmikola: Let me know if you'd prefer changing "fieldname" to "array index" for clarity. Unfortunately, we aren't tracking the full property path, but that may be something we can improve in [PHPC-922](https://jira.mongodb.org/browse/PHPC-922) (I've left a note).

src/bson.c
```diff
@@ -1060,16 +1060,6 @@ void object_to_bson(zval *object, php_phongo_bson_flags_t flags, const char *key
...
-#if PHP_VERSION_ID >= 70000
-			tmp_ht = HASH_OF(&obj_data);
-#else
-			tmp_ht = HASH_OF(obj_data);
-#endif
-
-			if (tmp_ht && ZEND_HASH_APPLY_PROTECTION(tmp_ht)) {
-				ZEND_HASH_INC_APPLY_COUNT(tmp_ht);
-			}
-
```
jmikola: `object_to_bson()` is only called through `phongo_bson_append()`'s `IS_OBJECT` case, which is where we're now handling recursion detection and in/decrementing the hash's apply count.

php_phongo.c
```diff
@@ -385,11 +385,7 @@ zend_bool phongo_writeconcernerror_init(zval *return_value, bson_t *bson TSRMLS_
 
 		if (!phongo_bson_to_zval(data, len, &intern->info)) {
 			zval_ptr_dtor(&intern->info);
-#if PHP_VERSION_ID >= 70000
 			ZVAL_UNDEF(&intern->info);
-#else
-			intern->info = NULL;
-#endif
```
jmikola: This isn't really related to PHPC-948. I'm just employing the `ZVAL_UNDEF()` compatibility macro created for the patch and cleaning up some old code.

src/bson.c
```diff
@@ -1285,6 +1283,8 @@ void phongo_zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *bson
 	 * properties, we'll need to filter them out later. */
 	bool         ht_data_from_properties = false;
 
+	ZVAL_UNDEF(&obj_data);
```
jmikola: Alternatively, I could have initialized the PHP 7 zval with `{0}` and had it work with `Z_ISUNDEF()`, but that makes assumptions about its internals and we don't do it anywhere else.

src/bson.c
derickr: Maybe add that as a comment then?

tests/bson/bson-fromPHP_error-004.phpt
derickr: I don't care for now, but I do like that per PHPC-922, we should do this for *all* error messages that involve fieldnames.

jmikola: Will leave this as-is for now, then.

src/bson.c
jmikola: Will do.

PR #577, opened 2017-04-06
PHPC-587: CursorId should serialize to BSON as 64-bit integer

jmikola:
https://jira.mongodb.org/browse/PHPC-587

### DIFF
--- a/tests/cursorid/cursorid-002.phpt
+++ b/tests/cursorid/cursorid-002.phpt
@@ -1,0 +1,40 @@
+--TEST--
+MongoDB\Driver\CursorID BSON serialization for killCursors command
+--SKIPIF--
+<?php if (8 !== PHP_INT_SIZE) { die('skip Only for 64-bit platform'); } ?>
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS('STANDALONE'); CLEANUP(STANDALONE); ?>
+--FILE--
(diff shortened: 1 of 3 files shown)


tests/cursorid/cursorid-002.phpt
```diff
@@ -0,0 +1,40 @@
...
+$cursorId = $cursor->getId();
+
+$command = new MongoDB\Driver\Command([
+    'killCursors' => COLLECTION_NAME,
+    'cursors' => [ $cursorId ],
+]);
+
+/* Since the killCursors command result includes cursor IDs as 64-bit integers,
+ * unserializing the result document requires a 64-bit platform. */
+$result = $manager->executeCommand(DATABASE_NAME, $command)->toArray()[0];
```
derickr: What would happen on a 32bit platform?

jmikola: An exception for the overflow will be thrown from the [macros in `phongo_compat.h`](https://github.com/mongodb/mongo-php-driver/blob/a2da17fadebe24f6d6ec02198489cc5f42694f41/phongo_compat.h#L151). The only way around this would be remembering that a `killCursors` command was executed and conditionally unserializing the BSON values as `MongoDB\Driver\CursorId` instances; however, that all requires some assumptions about the command's return value (cursor IDs can show up in any of several array fields).

PR #579, opened 2017-04-14
PHPC-951: CursorId is a signed 64-bit integer

jmikola:
https://jira.mongodb.org/browse/PHPC-951


jmikola:
n> why is a CursorID a signed integer?

Cursor IDs use BSON's 64-bit integer type, which is always signed. They may never be negative in practice, but they are only solicited and provided as signed values.

PR #590, opened 2017-05-05
PHPC-579: Regression test mixing update/replacement with BulkWrite::update()

jmikola:
https://jira.mongodb.org/browse/PHPC-579

CDRIVER-1341 implemented client-side detection of the invalid newObj argument.


jmikola:
LGTM from @derickr.

PR #594, opened 2017-05-16
PHPC-586: Regression tests for tailable cursor iteration

jmikola:
https://jira.mongodb.org/browse/PHPC-586

Note: iterating a tailable cursor with `foreach` is not possible because [`php_phongo_cursor_iterator_valid()`](https://github.com/mongodb/mongo-php-driver/blob/0c683809881ee9633fd61383855ff6b4ebf977d7/src/MongoDB/Cursor.c#L58) returns `false` if the current element is undefined. When [`php_phongo_cursor_iterator_move_forward()`](https://github.com/mongodb/mongo-php-driver/blob/0c683809881ee9633fd61383855ff6b4ebf977d7/src/MongoDB/Cursor.c#L102) advances the cursor, it clears the current element by calling [`php_phongo_cursor_free_current()`](https://github.com/mongodb/mongo-php-driver/blob/0c683809881ee9633fd61383855ff6b4ebf977d7/src/MongoDB/Cursor.c#L31) and will leave it undefined unless `mongoc_cursor_next()` returns `true` and we have a BSON document to convert to the current element. This means that `foreach` is only able to iterate as long as the server returns documents in result batches and will not allow the driver to issue a `getMore` and tail the collection for more results.

### DIFF
--- a/tests/cursor/cursor-tailable-001.phpt
+++ b/tests/cursor/cursor-tailable-001.phpt
@@ -1,0 +1,86 @@
+--TEST--
+MongoDB\Driver\Cursor tailable iteration
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS('STANDALONE'); CLEANUP(STANDALONE); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+function insert(MongoDB\Driver\Manager $manager, $from, $to = null)
+{
+    if ($to === null) {
+        $to = $from;
+    }
+
+    $bulkWrite = new MongoDB\Driver\BulkWrite;
+
+    for ($i = $from; $i <= $to; $i++) {
+        $bulkWrite->insert(['_id' => $i]);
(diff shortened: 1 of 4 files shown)


jmikola:
@alcaeus: This may be relevant to [mongo-php-adapter](https://github.com/alcaeus/mongo-php-adapter).

tests/cursor/cursor-tailable-001.phpt
```diff
@@ -0,0 +1,86 @@
...
+
+insert($manager, 1, 3);
+
+$cursor = $manager->executeQuery(NS, new MongoDB\Driver\Query([], ['tailable' => true]));
+$it = new IteratorIterator($cursor);
+
+$numAwaitAttempts = 0;
+$maxAwaitAttempts = 7;
+
+for ($it->rewind(); $numAwaitAttempts < $maxAwaitAttempts; $it->next()) {
```
jmikola: @alcaeus: This logic may be relevant for [mongo-php-adapter](https://github.com/alcaeus/mongo-php-adapter).

PR #596, opened 2017-05-25
PHPC-839: Support maxAwaitTimeMS Query option for tailable await cursors

jmikola:
https://jira.mongodb.org/browse/PHPC-839

### DIFF
--- a/src/MongoDB/Query.c
+++ b/src/MongoDB/Query.c
@@ -205,6 +205,32 @@ static bool php_phongo_query_init_readconcern(php_phongo_query_t *intern, zval *
 	return true;
 } /* }}} */
 
+/* Initialize the "maxAwaitTimeMS" option. Returns true on success; otherwise,
+ * false is returned and an exception is thrown.
+ *
+ * The "maxAwaitTimeMS" option is assigned to the cursor after query execution
+ * via mongoc_cursor_set_max_await_time_ms(). */
+static bool php_phongo_query_init_max_await_time_ms(php_phongo_query_t *intern, zval *options TSRMLS_DC) /* {{{ */
+{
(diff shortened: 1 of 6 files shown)


src/MongoDB/Query.c
```diff
@@ -205,6 +205,32 @@ static bool php_phongo_query_init_readconcern(php_phongo_query_t *intern, zval *
...
 } /* }}} */
 
+/* Initialize the "maxAwaitTimeMS" option. Returns true on success; otherwise,
+ * false is returned and an exception is thrown.
+ *
+ * The "maxAwaitTimeMS" option is assigned to the cursor after query execution
+ * via mongoc_cursor_set_max_await_time_ms(). */
+static bool php_phongo_query_init_max_await_time_ms(php_phongo_query_t *intern, zval *options TSRMLS_DC) /* {{{ */
+{
+	if (php_array_existsc(options, "maxAwaitTimeMS")) {
```
derickr: I would explicitly set ```intern->max_await_time_ms``` to 0 here, in case it doesn't get initialised.

jmikola: Existing code depends on all of our internal structs being calloc-ed, which is why I omitted this. Elsewhere, we have code segments that assume pointers are NULL if we haven't set them, which seems far more dangerous if the calloc behavior were ever to change.

I'm up for making a new ticket to explicitly assign all of these properties if you agree.

derickr: Nah, that's OK.

derickr: (But I'm fine if you want to do this too)

jmikola: Created [PHPC-966](https://jira.mongodb.org/browse/PHPC-966) to track that.

PR #607, opened 2017-07-05
PHPC-939: Don't modify std props in BSON get_properties handlers

jmikola:
https://jira.mongodb.org/browse/PHPC-939

### DIFF
--- a/php_phongo.h
+++ b/php_phongo.h
@@ -167,6 +167,20 @@ zend_bool phongo_writeconcernerror_init(zval *return_value, bson_t *bson TSRMLS_
     ce->unserialize = zend_class_unserialize_deny; \
 } while(0);
 
+#define PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_debug, intern, props, size) do { \
+    if (is_debug) {                                                             \
+        ALLOC_HASHTABLE(props);                                                 \
+        zend_hash_init((props), (size), NULL, ZVAL_PTR_DTOR, 0);                \
+    } else if ((intern)->properties) {                                          \
+        zend_hash_clean((intern)->properties);                                  \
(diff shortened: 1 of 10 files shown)


src/BSON/ObjectID.c
```diff
@@ -352,16 +357,27 @@ static HashTable *php_phongo_objectid_get_gc(zval *object, phongo_get_gc_table t
...
+		ALLOC_HASHTABLE(props);
+		zend_hash_init(props, 1, NULL, ZVAL_PTR_DTOR, 0);
+	} else if (intern->properties) {
+		zend_hash_clean(intern->properties);
+		props = intern->properties;
+	} else {
+		ALLOC_HASHTABLE(props);
+		zend_hash_init(props, 1, NULL, ZVAL_PTR_DTOR, 0);
+		intern->properties = props;
+	}
```
derickr: This section is done for every class. Is it not possible to extract this into its own function? It would save a lot of boiler plate.

php_phongo.h
```diff
@@ -167,6 +167,20 @@ zend_bool phongo_writeconcernerror_init(zval *return_value, bson_t *bson TSRMLS_
...
+    } else if ((intern)->properties) {                                          \
+        zend_hash_clean((intern)->properties);                                  \
+        (props) = (intern)->properties;                                         \
+    } else {                                                                    \
+        ALLOC_HASHTABLE(props);                                                 \
+        zend_hash_init((props), (size), NULL, ZVAL_PTR_DTOR, 0);                \
+        (intern)->properties = (props);                                         \
+    }                                                                           \
+} while(0);
+
```
jmikola: @derickr: Refactored to use a macro. A function wasn't feasible since the type of `intern` varies with each class.

PR #621, opened 2017-07-31
PHPC-640: Create interfaces for BSON types classes

jmikola:
https://jira.mongodb.org/browse/PHPC-640
https://jira.mongodb.org/browse/PHPC-988

These interfaces are implemented by the corresponding BSON classes and intended to be used by TypeWrapper implementations.

TODO:

 * [x] Tests to ensure that each BSON type class implements the appropriate interface
 * [x] Tests to demonstrate how we expect TypeWrapper implementations to use these interfaces

### DIFF
--- a/tests/bson/bson-decimal128interface-002.phpt
+++ b/tests/bson/bson-decimal128interface-002.phpt
@@ -1,0 +1,97 @@
+--TEST--
+MongoDB\BSON\Decimal128Interface example with MongoDB\BSON\TypeWrapper
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/tools.php';
+
+class MyDecimal128 implements MongoDB\BSON\TypeWrapper, MongoDB\BSON\Decimal128Interface
+{
+    private $decimal;
+
+    public function __construct($value)
+    {
+        $this->decimal = new MongoDB\BSON\Decimal128($value);
+    }
+
+    public static function createFromBSONType(MongoDB\BSON\Type $type)
+    {
+        if ( ! $type instanceof MongoDB\BSON\Decimal128) {
+            throw new InvalidArgumentException('Cannot create MyDecimal128 from ' . get_class($type));
+        }
+
+        return new self($type);
+    }
+
+    public function toBSONType()
+    {
+        return $this->decimal;
+    }
+
+    public function __toString()
+    {
+        return (string) $this->decimal;
+    }
+}
+
+$decimal = new MongoDB\BSON\Decimal128('1234.5678');
+$myDecimal = new MyDecimal128('1234.5678');
+
+echo "\nTesting MyDecimal128 and Decimal128 are equivalent\n";
+var_dump((string) $decimal === (string) $myDecimal);
+
+echo "\nTesting MyDecimal128 and Decimal128 produce the same BSON\n";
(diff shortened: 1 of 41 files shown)


tests/bson/bson-decimal128interface-002.phpt
```diff
@@ -0,0 +1,97 @@
...
+    {
+        return (string) $this->decimal;
+    }
+}
+
+$decimal = new MongoDB\BSON\Decimal128('1234.5678');
+$myDecimal = new MyDecimal128('1234.5678');
+
+echo "\nTesting MyDecimal128 and Decimal128 are equivalent\n";
+var_dump((string) $decimal === (string) $myDecimal);
```
jmikola: Note: `__toString()` is the only way to obtain the value of `Decimal128` and `ObjectID` objects. The method is merely a convenience for Binary, Javascript, Regex, Timestamp, and UTCDateTime (note: it is the easiest way to obtain a UTCDateTime's milliseconds). It does not apply at all for MaxKey and MinKey.

I originally decided not to add it to the type-specific interfaces, as I didn't like the idea of incorporating a magic method into an interface (one could say the same about `__debugInfo()` if [PHPC-470](https://jira.mongodb.org/browse/PHPC-470) was already merged). However, I'm not wondering if it makes sense to add it. I also purposefully omitted JsonSerializable and Serializable, am plan to simply suggest that users also implement those in their custom wrapper classes if desired.

tests/bson/bson-utcdatetimeinterface-002.phpt
```diff
@@ -0,0 +1,105 @@
...
+        $this->date = new MongoDB\BSON\UTCDateTime($milliseconds);
+    }
+
+    public static function createFromBSONType(MongoDB\BSON\Type $type)
+    {
+        if ( ! $type instanceof MongoDB\BSON\UTCDateTime) {
+            throw new InvalidArgumentException('Cannot create MyUTCDateTime from ' . get_class($type));
+        }
+
+        return new self((string) $type);
```
jmikola: A string cast is necessary to avoid `UTCDateTime::__construct()` throwing an exception because its object argument is not a DateTime instance. This issue didn't come up in the Decimal128 and ObjectID tests, as we only parse the argument as a string and PHP happily calls `__toString()` on the object argument for us.

I don't think this is anything we need to fix, as this is all consistent with UTCDateTime's documentation and the Decimal128 and ObjectID behavior is just a happy coincidence.

jmikola:
/ping @alcaeus, @malarzm: Let me know if this seems useful for ODM 2.0.

tests/bson/bson-utcdatetimeinterface-002.phpt
derickr: We discussed this while doing the HHVM versions, and this is what the conclusion was I believe: https://github.com/mongodb/mongo-hhvm-driver/blob/master/ext_mongodb.php#L1182

tests/bson/bson-decimal128interface-002.phpt
derickr: We discussed this while doing the HHVM versions, and this is what the conclusion was I believe: https://github.com/mongodb/mongo-hhvm-driver/blob/master/ext_mongodb.php#L1182

jmikola: Added `__toString()` to all interfaces except MinKey and MaxKey.

PR #622, opened 2017-07-31
PHPC-985: Timestamp::getIncrement() and getTimestamp() methods

jmikola:
https://jira.mongodb.org/browse/PHPC-985

### DIFF
--- a/src/BSON/Timestamp.c
+++ b/src/BSON/Timestamp.c
@@ -163,6 +163,36 @@ static PHP_METHOD(Timestamp, __construct)
 	php_phongo_timestamp_init_from_string(intern, Z_STRVAL_P(increment), Z_STRLEN_P(increment), Z_STRVAL_P(timestamp), Z_STRLEN_P(timestamp) TSRMLS_CC);
 } /* }}} */
 
+/* {{{ proto integer MongoDB\BSON\Timestamp::getIncrement()
+*/
+static PHP_METHOD(Timestamp, getIncrement)
+{
+	php_phongo_timestamp_t *intern;
+
+	intern = Z_TIMESTAMP_OBJ_P(getThis());
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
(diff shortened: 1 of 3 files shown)


src/BSON/Timestamp.c
```diff
@@ -163,6 +163,36 @@ static PHP_METHOD(Timestamp, __construct)
...
+{
+	php_phongo_timestamp_t *intern;
+
+	intern = Z_TIMESTAMP_OBJ_P(getThis());
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	RETVAL_LONG(intern->increment);
```
jmikola: https://github.com/mongodb/mongo-php-driver/commit/ff00668385c26823db3ac37ab614ba84a0ffa06f enhanced `Timestamp::__construct()` to accept string arguments for providing 64-bit integers on 32-bit platforms ([PHPC-731](https://jira.mongodb.org/browse/PHPC-731)). The timestamp and increment components are both unsigned 32-bit values, so that feature was needed to allow integers between `INT32_MAX` and `UINT32_MAX`.

When returning the unsigned 32-bit values via these getters, it's possible that we could be assigning such an integer into a zval's signed 32-bit value, which would appear as a negative value. That said, I don't think this is worth addressing, as we already have the same issue with the `ObjectID::getTimestamp()` method. In the ObjectID case, the 4-byte timestamp is extracted via [`bson_oid_get_time_t_unsafe()`](https://github.com/mongodb/libbson/blob/1.6.3/src/bson/bson-oid.h#L234) and we assign its return value with `RETVAL_LONG()` in the same fashion.

jmikola:
> If you are sure the order of increment and timestamp are right - I believe this has changed?

I took this into consideration. The arguments for `MongoDB\BSON\Timestamp` are increment and timestamp, in that order. [PHPC-845](https://jira.mongodb.org/browse/PHPC-845) explains how we ended up with this due to a typo in the BSON spec.

PR #623, opened 2017-08-03
PHPC-846: Throw if maxStalenessSeconds URI option is out of range

jmikola:
https://jira.mongodb.org/browse/PHPC-846

Note: Resolution of [CDRIVER-2232](https://jira.mongodb.org/browse/CDRIVER-2232) will decide if 64-bit integers are actually valid for `maxStalenessSeconds` (they appear to already be supported my libmongoc's read preference API, while the URI parsing restricts it to 32-bit values). We currently document the ReadPreference constructor argument as a "signed 32-bit integer".

If 64-bit integers are approved for `maxStalenessSeconds`, I will repurpose this issue to add support in both our URI options array and ReadPreference constructor.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1166,20 +1166,24 @@ static bool php_phongo_apply_rp_options_to_uri(mongoc_uri_t *uri, bson_t *option
 
 	/* Handle maxStalenessSeconds, and make sure it is not combined with primary
 	 * readPreference */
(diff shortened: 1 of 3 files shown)


jmikola:
@derickr: @ajdavis suggested resolving CDRIVER-2232 as "Won't Fix", since there is no practical need to support values outside the range of 32-bit integers. I'll merge this as-is.

PR #624, opened 2017-08-03
PHPC-887: Throw for unexpected types in URI options array

jmikola:
https://jira.mongodb.org/browse/PHPC-887

This PR is based on https://github.com/mongodb/mongo-php-driver/pull/623 and supersedes https://github.com/mongodb/mongo-php-driver/pull/601.

TODO:

 * [x] Error checking for special options in `php_phongo_apply_options_to_uri()`
 * [x] Update tests to expect exceptions for invalid types, which would have been previously ignored


php_phongo.c
```diff
@@ -927,6 +927,42 @@ static mongoc_uri_t *php_phongo_make_uri(const char *uri_string, bson_t *options
...
 
+static const char *php_phongo_bson_type_to_string(bson_type_t type) /* {{{ */
+{
+	switch (type) {
+		case BSON_TYPE_EOD: return "EOD";
+		case BSON_TYPE_DOUBLE: return "double";
+		case BSON_TYPE_UTF8: return "string";
+		case BSON_TYPE_DOCUMENT: return "document";
+		case BSON_TYPE_ARRAY: return "array";
+		case BSON_TYPE_BINARY: return "Binary";
```
jmikola: Since this is really only used for URI option validation, I considered replacing these strings with our BSON class names as that is really the only way that users could provide such invalid values when constructing a Manager.

That said, it would make the output more verbose and wouldn't allow us to differentiate between things like `BSON_TYPE_CODE` and `BSON_TYPE_CODEWSCOPE`. I don't see any reason we can't refine these in the future if we like, but let me know if things are OK as-is. Note that each return value is a single term except for "Javascript with scope", "32-bit integer", and "64-bit integer".

php_phongo.c
```diff
@@ -955,6 +991,9 @@ static bool php_phongo_apply_options_to_uri(mongoc_uri_t *uri, bson_t *options T
 		}
 
 		if (mongoc_uri_option_is_bool(key)) {
+			/* The option's type is not validated because bson_iter_as_bool() is
+			 * used to cast the value to a boolean. Validation may introduced in
+			 * PHPC-990. */
```
jmikola: See: https://jira.mongodb.org/browse/PHPC-990

php_phongo.c
```diff
@@ -1062,16 +1111,21 @@ static bool php_phongo_apply_rc_options_to_uri(mongoc_uri_t *uri, bson_t *option
 		return true;
 	}
 
-	if (!bson_iter_init_find_case(&iter, options, "readconcernlevel")) {
+	if (!bson_iter_init_find_case(&iter, options, MONGOC_URI_READCONCERNLEVEL)) {
```
jmikola: This change is unrelated to the PR, but I noticed while fixing `bson_iter_init_find_case()` below that we weren't using the defined constant for this option. Let me know if you'd like it split out into a separate commit.

php_phongo.c
```diff
@@ -1111,8 +1165,17 @@ static bool php_phongo_apply_rp_options_to_uri(mongoc_uri_t *uri, bson_t *option
 		mongoc_read_prefs_set_mode(new_rp, MONGOC_READ_SECONDARY_PREFERRED);
 	}
 
-	if (bson_iter_init_find_case(&iter, options, MONGOC_URI_READPREFERENCE) && BSON_ITER_HOLDS_UTF8(&iter)) {
-		const char *str = bson_iter_utf8(&iter, NULL);
+	if (bson_iter_init_find_case(&iter, options, MONGOC_URI_READPREFERENCE)) {
```
jmikola: It occurred to me while working on this PR that the driver handles duplicate options with different character casing very poorly. For instance, if the options array contained both `readpreference` and `readPreference`, this logic would only pick up the first matching key and ignore subsequent duplicates. Ideally, we would raise some exception for duplicates or simply require a single form to be used (e.g. "readPreference"). The latter would be a BC break and would serve little benefit if duplicate detection is implemented.

I believe libmongoc's handling of URI strings processes each parameter as it appears in the string so that the last value overrides any previous occurrences.

I've opened [PHPC-991](https://jira.mongodb.org/browse/PHPC-991) to track this.

tests/manager/manager-ctor-read_preference-003.phpt
```diff
@@ -1,44 +0,0 @@
---TEST--
-MongoDB\Driver\Manager::__construct(): read preference options of the wrong type are ignored
```
jmikola: This test is no longer applicable, because we now throw for unexpected types.

php_phongo.c
```diff
@@ -1807,15 +1947,6 @@ void phongo_manager_init(php_phongo_manager_t *manager, const char *uri_string,
 		goto cleanup;
 	}
 
-	if (bson_iter_init_find_case(&iter, &bson_options, MONGOC_URI_APPNAME) && BSON_ITER_HOLDS_UTF8(&iter)) {
-		const char *str = bson_iter_utf8(&iter, NULL);
-
-		if (!mongoc_uri_set_appname(uri, str)) {
-			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Invalid appname value: '%s'", str);
```
jmikola: This was moved to `php_phongo_apply_options_to_uri()`, per my earlier comment.

php_phongo.c
```diff
@@ -972,60 +1015,93 @@ static bool php_phongo_apply_options_to_uri(mongoc_uri_t *uri, bson_t *options T
...
+			if (!BSON_ITER_HOLDS_UTF8(&iter)) {
+				PHONGO_URI_INVALID_TYPE(iter, "string");
+				return false;
+			}
+
 			if (!mongoc_uri_set_option_as_utf8(uri, key, bson_iter_utf8(&iter, NULL))) {
-				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Failed to parse \"%s\" URI option", key);
+				/* Assignment uses mongoc_uri_set_appname() for the "appname"
+				 * option, which validates length in addition to UTF-8 encoding.
+				 * For BC, we report the invalid string to the user. */
```
jmikola: AFAICT, "appname" was only being handled in `phongo_manager_init()` because `mongoc_uri_set_appname()` also checked the length in addition to basic UTF-8 validation and we wanted to use a more specific error message than "Failed to parse". Since `mongoc_uri_option_is_utf8()` does consider "appname" a string option and libmongoc supports assignment with `mongoc_uri_set_option_as_utf8()`, we can handle the special error reporting here and reduce some clutter in `phongo_manager_init()`.

php_phongo.c
```diff
@@ -1107,12 +1188,30 @@ static bool php_phongo_apply_rp_options_to_uri(mongoc_uri_t *uri, bson_t *option
 
 	new_rp = mongoc_read_prefs_copy(old_rp);
 
-	if (bson_iter_init_find_case(&iter, options, MONGOC_URI_SLAVEOK) && BSON_ITER_HOLDS_BOOL(&iter) && bson_iter_bool(&iter)) {
-		mongoc_read_prefs_set_mode(new_rp, MONGOC_READ_SECONDARY_PREFERRED);
+	if (bson_iter_init_find_case(&iter, options, MONGOC_URI_SLAVEOK)) {
+		if (!BSON_ITER_HOLDS_BOOL(&iter)) {
+			PHONGO_URI_INVALID_TYPE(iter, "boolean");
```
jmikola: Note: unlike generic boolean options (e.g. "ssl"), booleans for read preferences and write concerns were previously ignored if an unexpected type. I don't consider it a BC error to start throwing an error here, since libmongoc does the same during URI string parsing. See also: "journal" and "safe" for write concerns.

tests/manager/manager-ctor-004.phpt
```diff
@@ -0,0 +1,38 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): Deprecated boolean options in URI string
```
jmikola: We previously did not have any tests for this, but now that we're testing for invalid boolean types (e.g. `"?journal=invalid"`) I think it's helpful to have some record of strings accepted by libmongoc (even if deprecated).

tests/manager/manager-ctor-read_concern-001.phpt
```diff
@@ -0,0 +1,36 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): read concern options
+--FILE--
+<?php
+
+$tests = [
+    ['mongodb://127.0.0.1/?readConcernLevel=local', []],
+    // libmongoc does not discriminate numeric strings for expected string types
+    ['mongodb://127.0.0.1/?readConcernLevel=1', []],
```
jmikola: Using a numeric value doesn't make sense and may seem like an error, but libmongoc's URI parser doesn't validate options expecting a string value beyond UTF-8 encoding.

tests/manager/manager-ctor-read_preference-004.phpt
```diff
@@ -27,9 +27,6 @@ $tests = [
     ['mongodb://127.0.0.1/?SLAVEOK=TRUE', []],
     [null, ['slaveOk' => true]],
     [null, ['SLAVEOK' => true]],
-    // Strict type checking on options array (non-booleans are ignored)
```
jmikola: This is also no longer applicable, since we now throw.

tests/manager/manager-ctor-write_concern-004.phpt
```diff
@@ -0,0 +1,72 @@
...
+    ['mongodb://127.0.0.1/?safe=true', []],
+    ['mongodb://127.0.0.1/?safe=false', []],
+    // w overrides safe if provided in the same format
+    ['mongodb://127.0.0.1/?w=1&safe=false', []],
+    ['mongodb://127.0.0.1/?w=0&safe=true', []],
+    [null, ['safe' => true]],
+    [null, ['safe' => false]],
+    [null, ['w' => 1, 'safe' => false]],
+    [null, ['w' => 0, 'safe' => true]],
+    // safe in URI options array may override w in URI string
```
jmikola: We previously had no test for this.

tests/manager/manager-ctor-write_concern-error-001.phpt
```diff
@@ -1,66 +1,31 @@
...
-    $manager = new MongoDB\Driver\Manager('mongodb://127.0.0.1/?w=-1', ['journal' => true]);
-}, "MongoDB\Driver\Exception\InvalidArgumentException"), "\n";
-
-echo throws(function() {
-    $manager = new MongoDB\Driver\Manager('mongodb://127.0.0.1/?w=0', ['journal' => true]);
-}, "MongoDB\Driver\Exception\InvalidArgumentException"), "\n";
-
-echo throws(function() {
-    $manager = new MongoDB\Driver\Manager('mongodb://127.0.0.1/?journal=true', ['w' => -1]);
-}, "MongoDB\Driver\Exception\InvalidArgumentException"), "\n";
```
jmikola: The write concern URI error tests are now organized by the option (like read preferences). All of these "journal" test cases were moved to `manager-ctor-write_concern-error-005.phpt` (git considers it a copy of `manager-ctor-write_concern-error-001.phpt` with some modifications).

php_phongo.c
```diff
@@ -963,7 +1001,12 @@ static bool php_phongo_apply_options_to_uri(mongoc_uri_t *uri, bson_t *options T
 			continue;
 		}
 
-		if (mongoc_uri_option_is_int32(key) && BSON_ITER_HOLDS_INT32(&iter)) {
+		if (mongoc_uri_option_is_int32(key)) {
+			if (!BSON_ITER_HOLDS_INT32(&iter)) {
+				PHONGO_URI_INVALID_TYPE(iter, "32-bit integer");
```
derickr: I think instead of strings, we should use constants for these instead.

tests/manager/manager-ctor_error-003.phpt
```diff
@@ -0,0 +1,266 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): invalid types in URI options arrays
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/tools.php';
+
+/* Note: generic boolean options (e.g. "ssl") are not tested because the driver
+ * uses bson_iter_as_bool() to cast the value to a boolean for assignment. This
+ * may be changed in PHPC-990.
```
derickr: I prefer not having an issue mentioned in here, as we might likely forget to remove it when PHPC-990 gets fixed—unless this test starts failing when we address PHPC-990.

php_phongo.c
jmikola: Are you suggesting I create constants for all of the return values in `php_phongo_bson_type_to_string()`?

I don't necessarily agree with using constants here, as we also have `PHONGO_URI_INVALID_TYPE(iter, "array or object");` below. These strings are merely pieces of an exception message, just as we do in the [Query constructor](https://github.com/mongodb/mongo-php-driver/blob/223d1b91786ff8aefc7bc036f894d43835679cbf/src/MongoDB/Query.c#L135).

tests/manager/manager-ctor_error-003.phpt
jmikola: I'll remove the issue reference and retain the first sentence.

php_phongo.c
derickr: OK, forget it then :-)

PR #631, opened 2017-08-14
PHPC-996: Remove support for PHP 5.4

jmikola:
https://jira.mongodb.org/browse/PHPC-996


tests/bson/bson-utcdatetimeinterface-002.phpt
```diff
@@ -1,5 +1,7 @@
 --TEST--
 MongoDB\BSON\UTCDateTimeInterface example with MongoDB\BSON\TypeWrapper
+--INI--
+date.timezone=UTC
```
jmikola: Unrelated to PHP 5.4. This is just something I noticed while testing the build against my PHP 5.5 ZTS environment, which didn't have `date.timezone` in its `php.ini`. All other test files that work with DateTime appear to define this in an `--INI--` block, so this is just something I forgot to include in https://github.com/mongodb/mongo-php-driver/commit/36487bc0a97421425ab1e067922aa7f078b888fc.

phongo_compat.h
```diff
@@ -43,28 +43,6 @@
...
-			break; \
-		case HASH_KEY_IS_STRING: \
-			ZVAL_STRINGL(z, str_key, str_key_len-1, 0); \
-			break; \
-		default: \
-			RETURN_NULL(); \
-	} \
-}
-#endif
-
```
derickr: Was this define not used at all? Don't see it referenced anywhere else in this PR?

jmikola: It was introduced in cdc6471b46838d11bcf951a860ab22a893ef6f65 for `Cursor::key()` ([PHPC-130](https://jira.mongodb.org/browse/PHPC-130)) but later removed in 34fff68d850954f59c1c736b6b7bf6fe62355f53 ([PHPC-224](https://jira.mongodb.org/browse/PHPC-224)). This was all before 1.0.0. Best I can tell, it has been unused since.

PR #633, opened 2017-08-16
PHPC-997: Handle changes in field names in timelib 2017.05beta7

derickr:


remicollet:
Looks ok at buildtime.
BTW, what will happen if user update PHP (timelib) ?
(ie build  against an old PHP verison, with previous timelib, then update php, then run again new PHP with new timelib)

derickr:
I believe results will be wrong in the fraction field. But I am not sure if there is anything we can do there?

Jan-E:
This is a fix for the mongodb extension. It does not fix the BC break in PHP core:
https://github.com/php/php-src/commit/bdd56f31078bf1f34341943603cf6aaa72e0db5c#commitcomment-23668223

src/BSON/UTCDateTime.c
```diff
@@ -142,7 +142,11 @@ static bool php_phongo_utcdatetime_init_from_date(php_phongo_utcdatetime_t *inte
 
 	/* The following assignments use the same logic as date_format() in php_date.c */
 	sec = datetime_obj->time->sse;
+#if TIMELIB_VERSION >= 201705
+	usec = (int64_t) floor(datetime_obj->time->us);
```
remicollet: us is an int64, so floor is not needed, right ?

jmikola: It should be either a `int64_t` or `signed long long` per https://github.com/php/php-src/blob/21cd57e5f2d286ab77ab44a05d2d16c16e4f15e6/ext/date/lib/timelib.h#L124

Issue #635, opened 2017-08-17
\MongoDB\BSON\UTCDateTime issue


dking3876:
### Description

Briefly describe the issue. Feel free to cross-reference your report with
another related GitHub or JIRA issue for additional context.

Before reporting an issue, check that you are using the most recent version of
the extension.

All BSON\UTCDateTime implimentations are returning MongoDB\Driver\Exception\UnexpectedValueException: Got unsupported type 0 'unknown' when i attempt to insert or update any records in a collection

I was previously using 1.2.9 of the mongodb.dll and installed the latest 1.3.0beta1 release a few days ago with the same result

```
 $write = new \MongoDB\Driver\BulkWrite;
$write->insert($document); //This is where the error is thrown
document object is as follows
object(Project)#3 (8) {
  ["projectName"]=>
  string(39) "Create Project after changing constants"
  ["description"]=>
  string(574) "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum."
  ["startDate"]=>
  object(MongoDB\BSON\UTCDateTime)#6 (1) {
    ["milliseconds"]=>
    string(13) "1503002896000"
  }
  ["lastModifiedDate"]=>
  object(MongoDB\BSON\UTCDateTime)#5 (1) {
    ["milliseconds"]=>
    string(13) "1503002896000"
  }
  ["createdBy"]=>
  object(stdClass)#4 (4) {
    ["id"]=>
    string(24) "5824b9f2d4ff1a0fc088b11d"
    ["name"]=>
    string(10) "Deryk King"
    ["email"]=>
    string(22) "deryk.king@brafton.com"
    ["userIcon"]=>
    string(59) "https://avatars3.githubusercontent.com/u/10117938?v=3&s=460"
  }
  ["subject"]=>
  string(13) "Site Redesign"
  ["projectType"]=>
  string(8) "Personal"
  ["status"]=>
  string(4) "open"
}
```

the fields causing the issue are the UTCDateTime fields (startDate, and lastModifiedDate) removing these results in successful insertion.

[...]

PR #637, opened 2017-08-25
PHPC-1001: Report field name for invalid UTF-8 and unsupported zval types

jmikola:
https://jira.mongodb.org/browse/PHPC-1001

See: https://github.com/mongodb/mongo-php-driver/issues/635#issuecomment-323367576

### DIFF
--- a/src/bson-encode.c
+++ b/src/bson-encode.c
@@ -321,7 +321,7 @@ try_again:
 			if (bson_utf8_validate(Z_STRVAL_P(entry), Z_STRLEN_P(entry), true)) {
 				bson_append_utf8(bson, key, key_len, Z_STRVAL_P(entry), Z_STRLEN_P(entry));
 			} else {
-				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Got invalid UTF-8 value serializing '%s'", Z_STRVAL_P(entry));
+				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Detected invalid UTF-8 for fieldname \"%s\": %s", key, Z_STRVAL_P(entry));
 			}
 			break;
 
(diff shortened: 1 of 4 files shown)


src/bson-encode.c
```diff
@@ -351,7 +351,7 @@ static void php_phongo_bson_append(bson_t *bson, php_phongo_bson_flags_t flags,
 			if (bson_utf8_validate(Z_STRVAL_P(entry), Z_STRLEN_P(entry), true)) {
 				bson_append_utf8(bson, key, key_len, Z_STRVAL_P(entry), Z_STRLEN_P(entry));
 			} else {
-				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Got invalid UTF-8 value serializing '%s'", Z_STRVAL_P(entry));
+				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Detected invalid UTF-8 in \"%s\" for fieldname \"%s\"", Z_STRVAL_P(entry), key);
```
derickr: I would probably suggest to put the field name first in the error message, as if the value includes for example a null char, it might get cut off.

src/bson-encode.c
```diff
@@ -410,7 +410,7 @@ static void php_phongo_bson_append(bson_t *bson, php_phongo_bson_flags_t flags,
 #endif
 
 		default:
-			phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Got unsupported type %d '%s'", Z_TYPE_P(entry), zend_get_type_by_const(Z_TYPE_P(entry)));
+			phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Detected unsupported PHP type %d (%s) for fieldname \"%s\"", Z_TYPE_P(entry), zend_get_type_by_const(Z_TYPE_P(entry)), key);
```
derickr: ditto - and also the value is not in " ... " here.

src/bson-encode.c
jmikola: The current order was taken from other error messages, but we can easily swap it for this exception.

src/bson-encode.c
jmikola: There isn't a value to report here, as the zval type is unknown.

src/bson-encode.c
jmikola: > value includes for example a null char

PHP does allow null characters in exception messages:

```
>>> throw new Exception("foo\0bar")
Exception with message 'foobar'
```

Are you more concerned with logging the messages outside of PHP?

PR #642, opened 2017-09-02
PHPC-1005: Use convert_to_object() to create stdClass from array

jmikola:
https://jira.mongodb.org/browse/PHPC-1005

This will ensure that the array symtable is properly converted to an object proptable in PHP 7.2 (e.g. integer key 0 converts to string key "0").

### DIFF
--- a/src/bson.c
+++ b/src/bson.c
@@ -580,14 +580,14 @@ static bool php_phongo_bson_visit_document(const bson_iter_t *iter ARG_UNUSED, c
 				case PHONGO_TYPEMAP_NATIVE_OBJECT:
 				default:
 #if PHP_VERSION_ID >= 70000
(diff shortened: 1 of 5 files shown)


bjori:
I don't quite understand this change.
If PHP7.2 no longer creates the properties table, isn't that a bug in PHP 7.2?

I've no memories as to why this was done this way though, it seems weird if there was no point..
maybe tsrmls weirdness or something?
I've no idea.

bjori:
Ok. Now I get it. This adds support for the new PHP 7.2 casting feature.
The change outside of the version ifdefs is just to keep the code consistent as there is no functional difference in calling one or the other in earlier versions.
lgtm

jmikola:
> there is no functional difference in calling one or the other in earlier versions

That's correct. In PHP versions before 7.2, we'll get an extra `switch` statement on the zval type, but I'm not worried about that.

PR #643, opened 2017-09-05
PHPC-1006: Do not modify Persistable::bsonSerialize() retval

jmikola:
https://jira.mongodb.org/browse/PHPC-1006

### DIFF
--- a/src/bson-encode.c
+++ b/src/bson-encode.c
@@ -401,6 +401,11 @@ void php_phongo_zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *
 	 * properties, we'll need to filter them out later. */
 	bool         ht_data_from_properties = false;
 
+	/* If the object is an instance of MongoDB\BSON\Persistable, we will need to
+	 * inject the PHP class name as a BSON key and ignore any existing key in
+	 * the return value of bsonSerialize(). */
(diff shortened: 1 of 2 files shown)


src/bson-encode.c
```diff
@@ -556,6 +566,10 @@ void php_phongo_zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *
 				goto cleanup;
 			}
 
+			if (skip_odm_field && !strcmp(string_key, PHONGO_ODM_FIELD_NAME)) {
```
bjori: Its confusing that string_key is both a char* and a zend_string.
but that might just be the context given github, which was very badly broken :|

bjori:
I've no idea why this is failing on Travis but this lgtm

src/bson-encode.c
jmikola: This diff line is for the PHP 5.x block, where `string_key` is the `char*` populated by `zend_hash_get_current_key_ex()`. In the preceding PHP 7.x block, `ZEND_HASH_FOREACH_KEY_VAL()` assigns the `zend_string`.

jmikola:
> I've no idea why this is failing on Travis but this lgtm

Fixed in https://github.com/mongodb/mongo-php-driver/pull/644

PR #650, opened 2017-09-08
[backport] PHPC-875: Allow compiling without SSL

jmikola:
https://jira.mongodb.org/browse/PHPC-875


config.m4
```diff
@@ -267,11 +267,17 @@ if test "$MONGODB" != "no"; then
...
+    AC_SUBST(MONGOC_ENABLE_SSL, 0)
+    AC_SUBST(MONGOC_ENABLE_CRYPTO_LIBCRYPTO, 0)
+    AC_SUBST(MONGOC_ENABLE_SSL_OPENSSL, 0)
+
+    PHP_SETUP_OPENSSL(MONGODB_SHARED_LIBADD, [
+      AC_SUBST(MONGOC_ENABLE_CRYPTO, 1)
+      AC_SUBST(MONGOC_ENABLE_SSL, 1)
+      AC_SUBST(MONGOC_ENABLE_CRYPTO_LIBCRYPTO, 1)
+      AC_SUBST(MONGOC_ENABLE_SSL_OPENSSL, 1)
+    ])
```
bjori: can I suggest, on failure, and os=darwin, to enable SECURE_TRANSPORT and CRYPTO_COMMON?
And same in config.w32, on openssl failure, enable SECURE_CHANNEL and CRYTPO_CNG

jmikola: That's an outstanding todo item in [PHPC-888](https://jira.mongodb.org/browse/PHPC-888) and not something that will be implemented in 1.2.x.

I should also note that the reason for backporting this was actually for Windows, as PECL's PHP 7.2 builds were not detecting OpenSSL 1.1.x (see: [PHPC-1008](https://jira.mongodb.org/browse/PHPC-1008)). 

`PHP_SETUP_OPENSSL()` actually emits an error if it fails to find OpenSSL, which aborts the configure process. This makes it useless for allowing SSL as an optional dependency. The recently added `SETUP_OPENSSL()` function in `confutils.js` is much better, as it simply returns the version of SSL that was found and lets the caller decide what to do if it isn't found. I was chatting with Remi about this shortcoming in `#php.pecl` on Friday and may end up submitting a PR to php-src to improve the autotools function.

PR #652, opened 2017-09-08
PHPC-1009: Check TIMELIB_VERSION instead of PHP_VERSION_ID

### LINKED ISSUE #649: 1.2.10 breaks Travis builds
Hello.
as1.2.10 was released today, it started breaking Travis builds using this extension from PECL.
This is caused by BC break in cbb5f9f1f2810ebf1f9a086f2d8978b37633d30f which addresses API/ABI break in PHP 7.2 RC phase introduced in https://github.com/php/php-src/commit/bdd56f31078bf1f34341943603cf6aaa72e0db5c#diff-05931d95f4541dc2f528625e662a307c.
Travis is still running on PHP 7.2.0beta3 and this extension fails to compile there.
Here's a failing job for Doctrine: https://travis-ci.org/doctrine/data-fixtures/jobs/273067384#L1141

jmikola:
https://jira.mongodb.org/browse/PHPC-1009

php/php-src@bdd56f3 bumped timelib for PHP 7.2RC1. Since all PHP 7.2 pre-release versions have the same PHP_VERSION_ID, the ABI change in `timelib_rel_time` and `timelib_time` can only be detected via TIMELIB_VERSION.

Fixes #649

### DIFF
--- a/src/BSON/UTCDateTime.c
+++ b/src/BSON/UTCDateTime.c
@@ -29,6 +29,7 @@
 #include <ext/standard/info.h>
 #include <Zend/zend_interfaces.h>
 #include <ext/date/php_date.h>
+#include <ext/date/lib/timelib.h>
 #include <ext/spl/spl_iterators.h>
 #include <ext/standard/php_var.h>
 #if PHP_VERSION_ID >= 70000
@@ -142,7 +143,7 @@ static bool php_phongo_utcdatetime_init_from_date(php_phongo_utcdatetime_t *inte
 
 	/* The following assignments use the same logic as date_format() in php_date.c */
 	sec = datetime_obj->time->sse;
-#if PHP_VERSION_ID >= 70200
+#if TIMELIB_VERSION >= 201705
 	usec = (int64_t) floor(datetime_obj->time->us);
 #else
 	usec = (int64_t) floor(datetime_obj->time->f * 1000000 + 0.5);
@@ -277,7 +278,7 @@ PHP_METHOD(UTCDateTime, toDateTime)
 	php_date_initialize(datetime_obj, sec, sec_len, NULL, NULL, 0 TSRMLS_CC);
 	efree(sec);
 
-#if PHP_VERSION_ID >= 70200
+#if TIMELIB_VERSION >= 201705
 	datetime_obj->time->us = (intern->milliseconds % 1000) * 1000;
 #else
 	datetime_obj->time->f = (double) (intern->milliseconds % 1000) / 1000;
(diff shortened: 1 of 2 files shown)


src/BSON/UTCDateTime.c
```diff
@@ -142,7 +143,7 @@ static bool php_phongo_utcdatetime_init_from_date(php_phongo_utcdatetime_t *inte
 
 	/* The following assignments use the same logic as date_format() in php_date.c */
 	sec = datetime_obj->time->sse;
-#if PHP_VERSION_ID >= 70200
+#if TIMELIB_VERSION >= 201705
```
jmikola: I did not use `TIMELIB_EXTENDED_VERSION` here, as that only seems to exist in PHP's master branch (added in https://github.com/php/php-src/commit/c0383b50399db7bd5996e87fed66be3fd7ded5d6 for what I assume is 7.3-dev).

src/BSON/UTCDateTime.c
```diff
@@ -29,6 +29,7 @@
 #include <ext/standard/info.h>
 #include <Zend/zend_interfaces.h>
 #include <ext/date/php_date.h>
+#include <ext/date/lib/timelib.h>
```
jmikola: @derickr: I realize that `ext/date/php_date.h` includes `lib/timelib.h` internally, but I reasoned it was preferable to explicitly include this header if we're going to rely on the `TIMELIB_VERSION` constant it defines.

The only downside to this approach is if there is any risk of the path changing across PHP versions. Do you have any input on that?

src/BSON/UTCDateTime.c
jmikola: The ABI change was actually introduced in timelib version 2017.05beta5 (https://github.com/derickr/timelib/commit/f04e68166b2d1fcb8156ea9c5dd424308ae4f20f). Unfortunately, all of those beta releases of timelib define the same value for, which makes this unreliable to use for detection.

src/BSON/UTCDateTime.c
jmikola: For the record, Derick confirmed offline that the path is reliable.

jmikola:
[PHPC-1009](https://jira.mongodb.org/browse/PHPC-1009) has been resolved as Won't Fix. I plan to cherry-pick https://github.com/mongodb/mongo-php-driver/pull/652/commits/9d2f6a1424f5655d607dc47cf6a849227662ab69 into a new PR, but we'll hold off on the timelib workaround and instead wait for Travis CI to be updated to 7.2 RC1.

PR #659, opened 2017-10-04
[WAIT] PHPC-1017: Add pecl option prompt for `--with-openssl-dir`

kabel:
This option is necessary for building on systems that may have outdated system openssl packages (macOS).

### DIFF
--- a/bin/package.xml.in
+++ b/bin/package.xml.in
@@ -55,5 +55,7 @@ necessary to build a fully-functional MongoDB driver.
   </required>
  </dependencies>
  <providesextension>mongodb</providesextension>
- <extsrcrelease/>
+ <extsrcrelease>
+  <configureoption name="with-openssl-dir" default="no" prompt="openssl directory"/>
+ </extsrcrelease>
 </package>


jmikola:
Thanks for the PR. I've opened [PHPC-1017](https://jira.mongodb.org/browse/PHPC-1017) to track this, as I think it'd make sense to expose other configure flags to PECL install as well.

Although PECL uses bundled versions of libbson and libmongoc by default, we do support using system libraries (mainly used by Debian and RedHat packages of the driver). As there already exists a [mongo-c-driver](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mongo-c-driver.rb) in Homebrew, I think the PHP driver's formula could also take advantage of that.

I'm not very familiar with Homebrew, but is this PR purely for PECL installs or is it also intended to address issues installing the driver through Homebrew? A quick looks at [`php72-mongodb`](https://github.com/Homebrew/homebrew-php/blob/master/Formula/php72-mongodb.rb) tells me that the driver formula only downloads sources through PECL and builds manually with `./configure --with-openssl-dir=[DIR]`.

Based on https://github.com/Homebrew/homebrew-core/pull/16067#issuecomment-333973443, it looks like @javian was installing with `pecl install mongodb`. Offhand, do you know if most Homebrew PHP users tend towards using `pecl install` over extension formula, or vice versa?

jmikola:
@derickr: I assume it's kosher to incorporate this PR into the upcoming 1.3.1 patch release? AFAIK, PECL does not support taking any of these options via the command line, so the user experience is the same (i.e. `pecl install mongodb` will still be run as-is).

According to the [`<configureoption>` documentation](https://pear.php.net/manual/en/guide.developers.package2.pecl.php), this will just cause PECL to start prompting for input during the installation process.

kabel:
You're welcome. At the moment this is just to support installs through pecl. Discussion continues about the future of pecl sourced formulas in the core of Homebrew. As it is today, I would say the majority of Homebrew users use formulae to install these extensions (the current default for the php formula does not include pecl).

derickr:
I"m closing this. Not because it's wrong (it's not!), but because we have another PR that extends this to all the other options that our ``./configure`` line takes: https://github.com/mongodb/mongo-php-driver/pull/692 — that PR also still needs some work though.

PR #668, opened 2017-11-09
PHPC-1031: Run configure check to use ASN1_STRING_get0_data over ASN1_STRING_data in bundled libmongoc

derickr:
https://jira.mongodb.org/browse/PHPC-1031


config.m4
```diff
@@ -295,6 +295,8 @@ if test "$MONGODB" != "no"; then
       AC_SUBST(MONGOC_ENABLE_SSL, 1)
       AC_SUBST(MONGOC_ENABLE_CRYPTO_LIBCRYPTO, 1)
       AC_SUBST(MONGOC_ENABLE_SSL_OPENSSL, 1)
+
+      AC_CHECK_DECLS([ASN1_STRING_get0_data], [AC_SUBST(MONGOC_HAVE_ASN1_STRING_GET0_DATA, 1)], [AC_SUBST(MONGOC_HAVE_ASN1_STRING_GET0_DATA, 1)], [[#include <openssl/asn1.h>]])
```
jmikola: Assuming this is for OpenSSL 1.1, we might need a similar check in [this part of `config.w32`](https://github.com/mongodb/mongo-php-driver/blob/0fc2923549a17effce272e266c5b643a5df389b9/config.w32#L170). That could be done by checking if [`SETUP_OPENSSL()`](https://github.com/php/php-src/blob/3a59c1e068a644effcad075b1a41e8b861f7beec/win32/build/confutils.js#L3368) returns `2`.

PR #673, opened 2017-11-20
 PHPC-1029 Support maxTimeMS getMore option for tailable command cursors 

derickr:
Waiting on PHPC-1037 / PR #672

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -654,7 +654,18 @@ int phongo_execute_command(mongoc_client_t *client, const char *db, zval *zcomma
 	/* According to mongoc_cursor_new_from_command_reply(), the reply bson_t
 	 * is ultimately destroyed on both success and failure. */
 	if (bson_iter_init_find(&iter, &reply, "cursor") && BSON_ITER_HOLDS_DOCUMENT(&iter)) {
-		cmd_cursor = mongoc_cursor_new_from_command_reply(client, &reply, selected_server_id);
+		bson_t initial_reply = BSON_INITIALIZER;
+
+		bson_copy_to(&reply, &initial_reply);
+
+		if (command->max_await_time_ms) {
+			bson_append_bool(&initial_reply, "awaitData", -1, 1);
+			bson_append_int32(&initial_reply, "maxAwaitTimeMS", -1, command->max_await_time_ms);
+			bson_append_bool(&initial_reply, "tailable", -1, 1);
+		}
+
+		cmd_cursor = mongoc_cursor_new_from_command_reply(client, &initial_reply, selected_server_id);
+		bson_destroy(&reply);
 	} else {
 		bson_t *wrapped_reply = create_wrapped_command_envelope(db, &reply);
 
@@ -663,7 +674,7 @@ int phongo_execute_command(mongoc_client_t *client, const char *db, zval *zcomma
 	}
 
 	if (!phongo_advance_cursor_and_check_for_error(cmd_cursor TSRMLS_CC)) {
-		mongoc_cursor_destroy(cmd_cursor);
+		/* If an error is found, the cmd_cursor is destroyed already */
 		return false;
 	}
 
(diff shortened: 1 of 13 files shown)


jmikola:
Looking at the diff, this change will conflict with #672 so I'll put off any review until that PR is merged and this can be rebased.

scripts/ubuntu/mongo-orchestration.sh
```diff
@@ -2,28 +2,32 @@
...
 dpkg -x mongodb-enterprise-mongos_3.2.0_amd64.deb 3.2.0
-
+dpkg -x mongodb-enterprise-server_3.6.0~rc3_amd64.deb 3.6.0rc3
 
 
 # Python stuff for mongo-orchestration
 apt-get install -y python python-dev python-pip
 
-pip install --upgrade 'git+https://github.com/10gen/mongo-orchestration.git#egg=mongo_orchestration'
+pip install --index-url=https://pypi.python.org/simple/ --upgrade 'git+https://github.com/10gen/mongo-orchestration.git#egg=mongo_orchestration'
```
derickr: The installed pip defaults to "http://pypi.python.org/simple/" which no longer works as it just returns "SSL is required"

tests/command/cursor-tailable-001.phpt
```diff
@@ -0,0 +1,71 @@
...
+];
+
+$command = new MongoDB\Driver\Command([
+	'aggregate' => COLLECTION_NAME,
+	'pipeline' => $pipeline,
+	'cursor' => ['batchSize' => 0],
+], [
+    'tailable' => true,
+    'awaitData' => true,
+    'maxAwaitTimeMS' => 100,
```
derickr: Using "10" was too flakey, so hence made it 100ms.

php_phongo.c
```diff
@@ -654,7 +654,18 @@ int phongo_execute_command(mongoc_client_t *client, const char *db, zval *zcomma
...
 	if (bson_iter_init_find(&iter, &reply, "cursor") && BSON_ITER_HOLDS_DOCUMENT(&iter)) {
-		cmd_cursor = mongoc_cursor_new_from_command_reply(client, &reply, selected_server_id);
+		bson_t initial_reply = BSON_INITIALIZER;
+
+		bson_copy_to(&reply, &initial_reply);
+
+		if (command->max_await_time_ms) {
+			bson_append_bool(&initial_reply, "awaitData", -1, 1);
+			bson_append_int32(&initial_reply, "maxAwaitTimeMS", -1, command->max_await_time_ms);
+			bson_append_bool(&initial_reply, "tailable", -1, 1);
```
jmikola: You can use `BSON_APPEND_BOOL()` and friends here, which will run `strlen()` on the string literal.

src/MongoDB/Command.c
```diff
@@ -21,35 +21,86 @@
...
+	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "A|a!", &document, &options) == FAILURE) {
 		zend_restore_error_handling(&error_handling TSRMLS_CC);
 		return;
 	}
 	zend_restore_error_handling(&error_handling TSRMLS_CC);
 
-
-	php_phongo_zval_to_bson(document, PHONGO_BSON_NONE, bson, NULL TSRMLS_CC);
-	intern->bson = bson;
+	php_phongo_command_init(intern, document, options TSRMLS_CC);
```
jmikola: The return value is never checked here. Should we free `intern->bson` on error, or call the dtor? Or are we trusting PHP to run the dtor for us, since a `false` return value will correlate with an exception being thrown?

tests/bulk/bulkwrite-update-arrayFilters-001.phpt
```diff
@@ -2,12 +2,12 @@
 MongoDB\Driver\BulkWrite::update with arrayFilters
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
-<?php NEEDS('STANDALONE'); CLEANUP(STANDALONE); ?>
+<?php START('THROWAWAY', ["version" => "36-release"]); CLEANUP(THROWAWAY); ?>
```
jmikola: This shouldn't be necessary, since 3.6 should be the default for MO (as the first version listed).

tests/command/cursor-tailable-001.phpt
```diff
@@ -0,0 +1,71 @@
+--TEST--
+MongoDB\Driver\Cursor tailable iteration with awaitData and maxAwaitTimeMS options
```
jmikola: Should this title be changed to refer to commands, or not necessary?

PR #679, opened 2017-11-28
PHPC-834: Add Read, ReadWrite, and Write command execute methods

derickr:


php_phongo.c
```diff
@@ -775,7 +849,24 @@ int phongo_execute_command(mongoc_client_t *client, const char *db, zval *zcomma
...
+			result = mongoc_client_command_with_opts(client, db, command->bson, phongo_read_preference_from_zval(zreadPreference TSRMLS_CC), opts, &reply, &error);
+			break;
+		case PHONGO_COMMAND_READ:
+			result = mongoc_client_read_command_with_opts(client, db, command->bson, phongo_read_preference_from_zval(zreadPreference TSRMLS_CC), opts, &reply, &error);
+			break;
+		case PHONGO_COMMAND_WRITE:
+			result = mongoc_client_write_command_with_opts(client, db, command->bson, opts, &reply, &error);
+			break;
+		case PHONGO_COMMAND_READ_WRITE:
+			result = mongoc_client_read_write_command_with_opts(client, db, command->bson, phongo_read_preference_from_zval(zreadPreference TSRMLS_CC), opts, &reply, &error);
```
jmikola: The read preference argument is not used for RW mode, so you should be able to pass NULL.

php_phongo.c
```diff
@@ -775,7 +849,24 @@ int phongo_execute_command(mongoc_client_t *client, const char *db, zval *zcomma
...
+			break;
+		default:
+			/* Should never happen, but if it does: abort */
+			assert(false);
```
jmikola: We should throw an exception instead of asserting here.

php_phongo.c
```diff
@@ -451,10 +451,63 @@ mongoc_bulk_operation_t *phongo_bulkwrite_init(zend_bool ordered) { /* {{{ */
 #define PHONGO_WRITECONCERN_ALLOWED   0x01
 #define PHONGO_READPREFERENCE_ALLOWED 0x02
 
-static int process_read_preference(zval *option, zval **zreadPreference TSRMLS_DC)
+static int process_read_concern(zval *option, bson_t *mongoc_opts TSRMLS_DC)
```
jmikola: `php_phongo_command_opts_append_document()` returns a `bool`, so this should, too.

php_phongo.c
```diff
@@ -451,10 +451,63 @@ mongoc_bulk_operation_t *phongo_bulkwrite_init(zend_bool ordered) { /* {{{ */
...
+			}
+
+			return false;
+		}
+	}
+
+	return selected_server_id;
+}
+
+static int process_read_preference(zval *option, bson_t *mongoc_opts, zval **zreadPreference, mongoc_client_t *client, int server_id TSRMLS_DC)
```
jmikola: Should return a `bool`.

php_phongo.c
```diff
@@ -451,10 +451,63 @@ mongoc_bulk_operation_t *phongo_bulkwrite_init(zend_bool ordered) { /* {{{ */
 #define PHONGO_WRITECONCERN_ALLOWED   0x01
 #define PHONGO_READPREFERENCE_ALLOWED 0x02
 
-static int process_read_preference(zval *option, zval **zreadPreference TSRMLS_DC)
+static int process_read_concern(zval *option, bson_t *mongoc_opts TSRMLS_DC)
+{
+	if (Z_TYPE_P(option) == IS_OBJECT && instanceof_function(Z_OBJCE_P(option), php_phongo_readconcern_ce TSRMLS_CC)) {
+		const mongoc_read_concern_t *read_concern = phongo_read_concern_from_zval(option TSRMLS_CC);
+
+		mongoc_read_concern_append((mongoc_read_concern_t*)read_concern, mongoc_opts);
```
jmikola: `mongoc_read_concern_append()` returns a success boolean. If false, you should throw an exception saying we couldn't append it properly.

php_phongo.c
```diff
@@ -466,10 +519,14 @@ static int process_read_preference(zval *option, zval **zreadPreference TSRMLS_D
 	return true;
 }
 
-static int process_write_concern(zval *option, zval **zwriteConcern TSRMLS_DC)
+static int process_write_concern(zval *option, bson_t *mongoc_opts, zval **zwriteConcern TSRMLS_DC)
```
jmikola: Should return a `bool`.

php_phongo.c
```diff
@@ -466,10 +519,14 @@ static int process_read_preference(zval *option, zval **zreadPreference TSRMLS_D
...
 {
 	if (Z_TYPE_P(option) == IS_OBJECT && instanceof_function(Z_OBJCE_P(option), php_phongo_writeconcern_ce TSRMLS_CC)) {
+		const mongoc_write_concern_t *write_concern = phongo_write_concern_from_zval(option TSRMLS_CC);
+
 		*zwriteConcern = option;
+
+		mongoc_write_concern_append((mongoc_write_concern_t*) write_concern, mongoc_opts);
```
jmikola: `mongoc_write_concern_append()` returns a success boolean. If false, you should throw an exception saying we couldn't append it properly.

php_phongo.c
```diff
@@ -775,7 +813,25 @@ int phongo_execute_command(mongoc_client_t *client, const char *db, zval *zcomma
...
+		case PHONGO_COMMAND_WRITE:
+			result = mongoc_client_write_command_with_opts(client, db, command->bson, opts, &reply, &error);
+			break;
+		case PHONGO_COMMAND_READ_WRITE:
+			/* We can pass NULL as readPreference, as this argument was added historically, but has no function */
+			result = mongoc_client_read_write_command_with_opts(client, db, command->bson, NULL, opts, &reply, &error);
+			break;
+		default:
+			/* Should never happen, but if it does: exception */
+			phongo_throw_exception(PHONGO_ERROR_LOGIC TSRMLS_CC, "Type '%d' should never have been passed to phongo_execute_command, please file a bug report", type);
```
jmikola: Should we return after throwing? Otherwise we might enter the `(!result)` condition below and attempt to throw again from an unset `bson_error_t`.

src/MongoDB/Manager.c
```diff
@@ -301,7 +301,70 @@ static PHP_METHOD(Manager, executeCommand)
...
+{
+	php_phongo_manager_t     *intern;
+	char                     *db;
+	phongo_zpp_char_len       db_len;
+	zval                     *command;
+	zval                     *options = NULL;
+	DECLARE_RETURN_VALUE_USED
+	SUPPRESS_UNUSED_WARNING(return_value_ptr)
+
+	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "sO|z!", &db, &db_len, &command, php_phongo_command_ce, &options) == FAILURE) {
```
jmikola: Are we inconsistent about using `zend_replace_error_handling()`, or was that intentionally only done to ensure we throw exceptions from constructors? This is a general question outside the scope of this PR.

php_phongo.c
derickr: Done

php_phongo.c
derickr: done

php_phongo.c
derickr: done

php_phongo.c
derickr: done

php_phongo.c
derickr: done

php_phongo.c
derickr: done

src/MongoDB/Manager.c
derickr: I think it was intentionally. There is no other way to signal in Constructors that ctor options were wrong.

php_phongo.c
```diff
@@ -451,10 +451,65 @@ mongoc_bulk_operation_t *phongo_bulkwrite_init(zend_bool ordered) { /* {{{ */
...
 #define PHONGO_READPREFERENCE_ALLOWED 0x02
 
-static int process_read_preference(zval *option, zval **zreadPreference TSRMLS_DC)
+static bool process_read_concern(zval *option, bson_t *mongoc_opts TSRMLS_DC)
+{
+	if (Z_TYPE_P(option) == IS_OBJECT && instanceof_function(Z_OBJCE_P(option), php_phongo_readconcern_ce TSRMLS_CC)) {
+		const mongoc_read_concern_t *read_concern = phongo_read_concern_from_zval(option TSRMLS_CC);
+
+		if (!mongoc_read_concern_append((mongoc_read_concern_t*)read_concern, mongoc_opts)) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error appending \"%s\" option", "ReadPreference");
```
jmikola: Still need to return false here.

php_phongo.c
```diff
@@ -451,10 +451,65 @@ mongoc_bulk_operation_t *phongo_bulkwrite_init(zend_bool ordered) { /* {{{ */
...
+			selected_server_id = mongoc_server_description_id(selected_server);
+			bson_append_int32(opts, "serverId", -1, selected_server_id);
+			mongoc_server_description_destroy(selected_server);
+		} else {
+			/* Check for connection related exceptions */
+			if (!EG(exception)) {
+				phongo_throw_exception_from_bson_error_t(&error TSRMLS_CC);
+			}
+
+			return false;
```
jmikola: I suppose this should be zero to jive with the return type. Alternatively, initialize `selected_server_id` to zero up top and remove this line entirely.

php_phongo.c
```diff
@@ -451,10 +451,65 @@ mongoc_bulk_operation_t *phongo_bulkwrite_init(zend_bool ordered) { /* {{{ */
...
+			PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC,
+			"Expected 'readConcern' option to be 'MongoDB\\Driver\\ReadPreference', %s given",
+			zend_get_type_by_const(Z_TYPE_P(option))
+		);
+		return false;
+	}
+	return true;
+}
+
+static int phongo_do_select_server(mongoc_client_t *client, bson_t *opts, zval *zreadPreference, int server_id TSRMLS_DC)
```
jmikola: Swap `int` return type for `uint32_t`.

src/MongoDB/Manager.c
```diff
@@ -301,7 +301,70 @@ static PHP_METHOD(Manager, executeCommand)
...
```
jmikola: Should you use `a!` for `$options` here? There is no BC concern, so users should always pass an array type.

src/MongoDB/Manager.c
```diff
@@ -510,6 +573,9 @@ ZEND_END_ARG_INFO()
 static zend_function_entry php_phongo_manager_me[] = {
 	PHP_ME(Manager, __construct, ai_Manager___construct, ZEND_ACC_PUBLIC|ZEND_ACC_FINAL)
 	PHP_ME(Manager, executeCommand, ai_Manager_executeCommand, ZEND_ACC_PUBLIC|ZEND_ACC_FINAL)
+	PHP_ME(Manager, executeReadCommand, ai_Manager_executeCommand, ZEND_ACC_PUBLIC|ZEND_ACC_FINAL)
```
jmikola: If the `$options` argument for these methods uses `a!`, you should also create a new arg info just for these as well, which enforces the array type hint.

php_phongo.c
derickr: Done

php_phongo.c
derickr: done

php_phongo.c
derickr: done

src/MongoDB/Manager.c
derickr: Done

src/MongoDB/Manager.c
derickr: Done

tests/manager/manager-executeReadWriteCommand_error-001.phpt
```diff
@@ -0,0 +1,29 @@
...
+
+$manager = new MongoDB\Driver\Manager(STANDALONE);
+
+$command = new MongoDB\Driver\Command([]);
+
+echo throws(function() use ($manager, $command) {
+	$manager->executeReadWriteCommand(
+		DATABASE_NAME, $command,
+		[
+			'readPreference' => new \MongoDB\Driver\ReadPreference(\MongoDB\Driver\ReadPreference::RP_SECONDARY),
```
jmikola: We're not in a namespace so the leading `\` isn't really necessary.

derickr: It's a habit... do you want me to change them (everywhere)?

php_phongo.c
```diff
@@ -448,32 +448,30 @@ mongoc_bulk_operation_t *phongo_bulkwrite_init(zend_bool ordered) { /* {{{ */
...
-#define PHONGO_READPREFERENCE_ALLOWED 0x02
-
 static bool process_read_concern(zval *option, bson_t *mongoc_opts TSRMLS_DC)
 {
 	if (Z_TYPE_P(option) == IS_OBJECT && instanceof_function(Z_OBJCE_P(option), php_phongo_readconcern_ce TSRMLS_CC)) {
 		const mongoc_read_concern_t *read_concern = phongo_read_concern_from_zval(option TSRMLS_CC);
 
 		if (!mongoc_read_concern_append((mongoc_read_concern_t*)read_concern, mongoc_opts)) {
-			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error appending \"%s\" option", "ReadPreference");
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error appending \"%s\" option", "ReadConcern");
```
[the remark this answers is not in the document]
derickr: done

PR #680, opened 2017-11-28
PHPC-980: Sessions API

jmikola:
https://jira.mongodb.org/browse/PHPC-980

I reached a road block supporting options for `phongo_execute_query()` and ended up continuing that work in #707 while refactoring the option parsing for execute methods for [PHPC-1057](https://jira.mongodb.org/browse/PHPC-1057).

This PR is limited just to relevant session commits. All included spec tests should be passing.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -538,7 +538,7 @@ static bool process_read_preference(zval *option, bson_t *mongoc_opts, zval **zr
 	return true;
 }
 
-static bool process_session(zval *option, bson_t *mongoc_opts TSRMLS_DC)
+static bool process_session(zval *option, bson_t *mongoc_opts, zval **zsession, mongoc_client_t *client TSRMLS_DC)
 {
 	const mongoc_client_session_t *client_session;
 
@@ -549,9 +549,18 @@ static bool process_session(zval *option, bson_t *mongoc_opts TSRMLS_DC)
 
 	client_session = Z_SESSION_OBJ_P(option)->client_session;
 
+	if (client != mongoc_client_session_get_client(client_session)) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Cannot use Session started from a different Manager");
+		return false;
+	}
+
 	if (!mongoc_client_session_append(Z_SESSION_OBJ_P(option)->client_session, mongoc_opts, NULL)) {
 		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error appending \"session\" option");
 		return false;
+	}
+
+	if (zsession) {
+		*zsession = option;
 	}
 
 	return true;
@@ -580,7 +589,7 @@ static bool process_write_concern(zval *option, bson_t *mongoc_opts, zval **zwri
 	return true;
 }
 
-static int phongo_execute_parse_options(mongoc_client_t* client, int server_id, zval *driver_options, int type, bson_t *mongoc_opts, zval **zreadPreference, zval **zwriteConcern TSRMLS_DC)
+static int phongo_execute_parse_options(mongoc_client_t* client, int server_id, zval *driver_options, int type, bson_t *mongoc_opts, zval **zreadPreference, zval **zwriteConcern, zval **zsession TSRMLS_DC)
 {
 	if (driver_options && Z_TYPE_P(driver_options) == IS_ARRAY) {
 		HashTable *ht_data = HASH_OF(driver_options);
@@ -604,7 +613,7 @@ static int phongo_execute_parse_options(mongoc_client_t* client, int server_id, 
 					return false;
 				}
 			} else if ((!strcmp(ZSTR_VAL(string_key), "session"))) {
-				if (!process_session(driver_option, mongoc_opts)) {
+				if (!process_session(driver_option, mongoc_opts, zsession, client)) {
 					return false;
 				}
 			} else if ((!strcasecmp(ZSTR_VAL(string_key), "writeConcern")) && (type & PHONGO_COMMAND_WRITE)) {
@@ -640,8 +649,8 @@ static int phongo_execute_parse_options(mongoc_client_t* client, int server_id, 
 				if (!process_read_preference(*driver_option, mongoc_opts, zreadPreference, client, server_id TSRMLS_CC)) {
 					return false;
 				}
-			} else if ((!strcmp(ZSTR_VAL(string_key), "session"))) {
-				if (!process_session(*driver_option, mongoc_opts)) {
+			} else if ((!strcasecmp(ZSTR_VAL(string_key), "session"))) {
+				if (!process_session(*driver_option, mongoc_opts, zsession, client TSRMLS_CC)) {
 					return false;
 				}
 			} else if ((!strcasecmp(string_key, "writeConcern")) && (type & PHONGO_COMMAND_WRITE)) {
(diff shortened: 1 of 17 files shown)


php_phongo.c
```diff
@@ -527,6 +538,34 @@ static bool process_read_preference(zval *option, bson_t *mongoc_opts, zval **zr
 	return true;
 }
 
+static bool process_session(zval *option, bson_t *mongoc_opts, zval **zsession, mongoc_client_t *client TSRMLS_DC)
+{
+	const mongoc_client_session_t *client_session;
+
+	if (Z_TYPE_P(option) != IS_OBJECT || !instanceof_function(Z_OBJCE_P(option), php_phongo_session_ce TSRMLS_CC)) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected \"session\" option to be %s, %s given", ZSTR_VAL(php_phongo_session_ce->name), zend_get_type_by_const(Z_TYPE_P(option)));
```
derickr: This should use the newly introduce macro from your other PR to show the class names.

php_phongo.c
```diff
@@ -527,6 +538,34 @@ static bool process_read_preference(zval *option, bson_t *mongoc_opts, zval **zr
...
+
+	client_session = Z_SESSION_OBJ_P(option)->client_session;
+
+	if (client != mongoc_client_session_get_client(client_session)) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Cannot use Session started from a different Manager");
+		return false;
+	}
+
+	if (!mongoc_client_session_append(Z_SESSION_OBJ_P(option)->client_session, mongoc_opts, NULL)) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error appending \"session\" option");
```
derickr: Is this likely to happen? If so, we should improve our error message to suggest how users can fix it.

php_phongo.c
```diff
@@ -606,6 +649,10 @@ static int phongo_execute_parse_options(mongoc_client_t* client, int server_id,
 				if (!process_read_preference(*driver_option, mongoc_opts, zreadPreference, client, server_id TSRMLS_CC)) {
 					return false;
 				}
+			} else if ((!strcmp(ZSTR_VAL(string_key), "session", zsession, client))) {
```
derickr: strcmp here, but strcasecmp below?

src/MongoDB/Manager.c
```diff
@@ -538,6 +538,42 @@ static PHP_METHOD(Manager, selectServer)
...
+	SUPPRESS_UNUSED_WARNING(return_value_ptr) SUPPRESS_UNUSED_WARNING(return_value_used)
+
+
+	intern = Z_MANAGER_OBJ_P(getThis());
+
+	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "|a!", &options) == FAILURE) {
+		return;
+	}
+
+	if (options && php_array_exists(options, "causalConsistency")) {
```
derickr: Do you want to extract option parsing into its own function like we do for many other options arrays?

src/MongoDB/Session.c
```diff
@@ -0,0 +1,275 @@
...
+{
+	php_phongo_session_t *intern;
+	zval                 *zcluster_time;
+	bson_t                cluster_time = BSON_INITIALIZER;
+	SUPPRESS_UNUSED_WARNING(return_value_ptr) SUPPRESS_UNUSED_WARNING(return_value_used)
+
+
+	intern = Z_SESSION_OBJ_P(getThis());
+
+	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "A", &zcluster_time) == FAILURE) {
```
derickr: I presume we shouldn't need to verify the structure of this cluster time object?

src/MongoDB/Session.c
```diff
@@ -0,0 +1,275 @@
...
+#endif
+	}
+
+	{
+		const bson_t *cluster_time;
+
+		php_phongo_bson_state state = PHONGO_BSON_STATE_INITIALIZER;
+		/* Use native arrays for debugging output */
+		state.map.root_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
+		state.map.document_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
```
derickr: You'd almost want to pull this often done conversion with NATIVE/NATIVE out into its own function.

php_phongo.c
jmikola: Same answer as https://github.com/mongodb/mongo-php-driver/pull/707#discussion_r157181533.

This will be updated to use macros in https://github.com/mongodb/mongo-php-driver/pull/705 after that PR is merged and this is rebased.

php_phongo.c
jmikola: Very unlikely. BSON append for primitive types only fails if the [buffer cannot grow](https://github.com/mongodb/mongo-c-driver/blob/6fa2c35119672c8ce7c0ac9a926fc33ecc3b7ae1/src/libbson/src/bson/bson.c#L325) or the [document length exceeds a 32-bit integer range](https://github.com/mongodb/mongo-c-driver/blob/6fa2c35119672c8ce7c0ac9a926fc33ecc3b7ae1/src/libbson/src/bson/bson.c#L394). Similar to other places where we check for failure (e.g. appending "serverId"), I'm just being a good libmongoc citizen here.

php_phongo.c
jmikola: Knowing that we were not going to allow case-insensitive options with [PHPC-1057](https://jira.mongodb.org/browse/PHPC-1057), I opted not to use `strcasecmp()` when I implemented this. That said, this code no longer exists after the refactoring in #705, since we know use the `php_array` functions to access it.

src/MongoDB/Manager.c
jmikola: We can consider that in the future if the options list grows a bit. I believe I modeled this after the ReadPreference constructor, which also parses a single option ("maxStalenessSeconds") at present.

src/MongoDB/Session.c
jmikola: Good question. libmongoc has a private [`_mongoc_parse_cluster_time()`](https://github.com/mongodb/mongo-c-driver/blob/6fa2c35119672c8ce7c0ac9a926fc33ecc3b7ae1/src/mongoc/mongoc-client-session.c#L146) function used to validate its structure. The driver only cares about the timestamp and doesn't validate the signature (only used server-side AFAIK).

[`mongoc_client_session_advance_cluster_time()`](https://github.com/mongodb/mongo-c-driver/blob/6fa2c35119672c8ce7c0ac9a926fc33ecc3b7ae1/src/mongoc/mongoc-client-session.c#L351) only assigns it if it parses successfully, and the function is void so we don't even get an error if it's malformed. I'm curious if that was intentional. Perhaps @ajdavis can chime in.

On a related note, [the spec](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#gossipping-the-cluster-time) doesn't refer to cluster time as an "opaque value" like we do for session and transaction IDs in retryable writes. Therefore, users can certainly access it and poke around with it. The spec also doesn't say anything about reporting errors if users try to advance an invalid value (see [here](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#advanceclustertime)).

src/MongoDB/Session.c
jmikola: Or perhaps a separate initializer struct like `PHONGO_BSON_STATE_INITIALIZER` but used just for debugging? Either way, that's a good idea. I've opened [PHPC-1074](https://jira.mongodb.org/browse/PHPC-1074) to track it.

jmikola:
Rebased on master and using the new `PHONGO_ZVAL_CLASS_OR_TYPE_NAME()` macro.

src/MongoDB/Session.c
ajdavis: Right, we only care that it has a timestamp, we don't know or care about other fields. I didn't think it worthwhile to implement elaborate error handling. Applications should simply save a cluster time from a session, then set it on another session, without modification. If you make a mistake there's an error log.

PR #685, opened 2017-12-01
PHPC-1015: Add test for DNS Initial Seedlist

derickr:
https://jira.mongodb.org/browse/PHPC-1015

We'll need to decide whether we want to always run this. Starting up a whole
replicaset for one test is a waste of resources really. It also requires that we add a line in our /etc/hosts files (with the IP address of the Vagrant host) for this to work.

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -375,7 +375,11 @@ if test "$MONGODB" != "no"; then
 
   m4_include(src/libmongoc/build/autotools/m4/ax_prototype.m4)
   m4_include(src/libmongoc/build/autotools/CheckCompiler.m4)
+
+  dnl We need to convince the libmongoc M4 file to actually run these checks for us
+  enable_srv=auto
   m4_include(src/libmongoc/build/autotools/FindResSearch.m4)
+
   m4_include(src/libmongoc/build/autotools/WeakSymbols.m4)
   m4_include(src/libmongoc/build/autotools/m4/ax_pthread.m4)
   AX_PTHREAD
--- a/scripts/presets/replicaset-dns.json
+++ b/scripts/presets/replicaset-dns.json
@@ -1,0 +1,65 @@
+{
+	"id": "REPLICASET_DNS",
+	"name": "mongod",
+	"members": [
+		{
+			"procParams": {
+				"dbpath": "/tmp/REPLICASET/27017/",
+				"ipv6": true,
+				"logappend": true,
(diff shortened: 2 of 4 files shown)


config.m4
```diff
@@ -375,7 +375,11 @@ if test "$MONGODB" != "no"; then
 
   m4_include(src/libmongoc/build/autotools/m4/ax_prototype.m4)
   m4_include(src/libmongoc/build/autotools/CheckCompiler.m4)
+
+  dnl We need to convince the libmongoc M4 file to actually run these checks for us
+  enable_srv=auto
```
jmikola: Interesting. Is this necessary because we do not include [`ReadCommandLineArguments.m4`](https://github.com/mongodb/mongo-c-driver/blob/1.9.0-rc0/build/autotools/ReadCommandLineArguments.m4)? AFAIK, this is one of the few times we have to set a variable this way (instead of using AC replacement).

derickr:
> Is this required change documented anywhere? Perhaps it can be added in a block comment within the test?

No, but I can add that.

> Am I correct to assume that the test fails without the /etc/hosts change in place?

Yes:

```
Fatal error: Uncaught MongoDB\Driver\Exception\ConnectionTimeoutException: No suitable servers found (`serverSelectionTryOnce` set): [connection refused calling ismaster on 'localhost.test.build.10gen.cc:27018'] [connection refused calling ismaster on 'localhost.test.build.10gen.cc:27017'] in /home/derick/dev/php/derickr-mongo-php-driver/tests/connect/bug1015.php:7
Stack trace:
#0 /home/derick/dev/php/derickr-mongo-php-driver/tests/connect/bug1015.php(7): MongoDB\Driver\Manager->selectServer(Object(MongoDB\Driver\ReadPreference))
#1 {main}
  thrown in /home/derick/dev/php/derickr-mongo-php-driver/tests/connect/bug1015.php on line 7
```

> I think it's worth having the Vagrant environment committed. We can probably leave this as a manual test for the time being and revisit once we can move over to Evergreen. To that end, a separate tracking ticket associated with the Evergreen epic and also linked to PHPC-1015 would be helpful.

Not sure how easy this is going to be with Evergreen either, but: https://jira.mongodb.org/browse/PHPC-1054

config.m4
derickr: Yup.

PR #690, opened 2017-12-05
PHPC-967: Add support for reading and writing OP_COMPRESSED

derickr:


php_phongo.c
```diff
@@ -1388,6 +1388,20 @@ static bool php_phongo_apply_options_to_uri(mongoc_uri_t *uri, bson_t *options T
...
 			continue;
 		}
+
+		if (!strcasecmp(key, MONGOC_URI_COMPRESSORS)) {
+			if (!BSON_ITER_HOLDS_UTF8(&iter)) {
+				PHONGO_URI_INVALID_TYPE(iter, "string");
+				return false;
+			}
+
+			if (!mongoc_uri_set_compressors(uri, bson_iter_utf8(&iter, NULL))) {
```
jmikola: Can we get some error tests?

You should be able to copy existing tests for `PHONGO_URI_INVALID_TYPE()` errors. For `mongoc_uri_set_compressors()`, I see it only returns false if the value is `NULL` (not possible with this code path) or the string contains invalid UTF-8. There should be some invalid UTF-8 strings in other tests (e.g. `bson-toPHP_error-004.phpt`) that you can re-use.

AFAIK, there's no way to determine if compressors are enabled once the client is constructed, so we probably don't need tests to demonstrate successful application of the option. If you did, though, it looks like "noop" is always a supported compressor type (even if snappy and zlib are disabled).

tests/connect/compression_error-002.phpt
```diff
@@ -0,0 +1,19 @@
...
+
+echo throws(function() {
+	$manager = new MongoDB\Driver\Manager(STANDALONE, [ 'compressors' => "foo\xFEbar"] );
+}, 'MongoDB\Driver\Exception\UnexpectedValueException'), "\n";
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+OK: Got MongoDB\Driver\Exception\UnexpectedValueException
+Detected invalid UTF-8 for fieldname "compressors": foo�bar
```
jmikola: As-is, this can use `EXPECT` instead of `EXPECTF`. If you want to be consistent with other UTF-8 error tests, we tend to use `: %s` at the end instead of repeating the input.

tests/connect/compression_error-002.phpt
```diff
@@ -0,0 +1,19 @@
...
+MongoDB\Driver\Manager: Connecting with invalid compressor values
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS('STANDALONE'); CLEANUP(STANDALONE); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+echo throws(function() {
+	$manager = new MongoDB\Driver\Manager(STANDALONE, [ 'compressors' => "foo\xFEbar"] );
```
jmikola: Four spaces in PHP files instead of tabs.

tests/connect/compression-001.phpt
```diff
@@ -0,0 +1,18 @@
...
+require_once __DIR__ . "/../utils/basic.inc";
+
+ini_set('mongodb.debug', 'stdout');
+$manager = new MongoDB\Driver\Manager(STANDALONE, [ 'compressors' => 'noop'] );
+ini_set('mongodb.debug', null);
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+%A"compressors";s:4:"noop"%A
```
jmikola: This doesn't appear to test anything more than PHPC's own `Created client hash:` debugging, so I think we can just remove this test. Even if libmongoc's URI parsing failed to parse compressors entirely, this would still pass.

tests/connect/compression_error-001.phpt
```diff
@@ -0,0 +1,18 @@
+--TEST--
+MongoDB\Driver\Manager: Connecting with unsupported compressor
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS('STANDALONE'); CLEANUP(STANDALONE); ?>
```
jmikola: `CLEANUP(STANDALONE);` shouldn't be needed here, since we neither do IO nor assert existing data.

tests/connect/compression_error-002.phpt
```diff
@@ -0,0 +1,19 @@
+--TEST--
+MongoDB\Driver\Manager: Connecting with invalid compressor values
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS('STANDALONE'); CLEANUP(STANDALONE); ?>
```
jmikola: Also no need for cleanup.

PR #694, opened 2017-12-08
Fix tests with mongodb 3.6

derickr:


php_phongo.c
```diff
@@ -632,10 +633,12 @@ bool phongo_execute_write(mongoc_client_t *client, const char *namespace, php_ph
 		return false;
 	}
 
+	opts = bson_new();
```
jmikola: Where is this freed? Also, AFAIK this will always allocate memory. Using `BSON_INITIALIZER` and calling `bson_destroy()` later allows us to use a stack-allocated buffer within this function.

php_phongo.c
```diff
@@ -632,10 +633,12 @@ bool phongo_execute_write(mongoc_client_t *client, const char *namespace, php_ph
...
+
 	/* FIXME: Legacy way of specifying the writeConcern option into this function */
 	if (options && Z_TYPE_P(options) == IS_OBJECT && instanceof_function(Z_OBJCE_P(options), php_phongo_writeconcern_ce TSRMLS_CC)) {
 		zwriteConcern = options;
-	} else if (!phongo_execute_parse_options(client, server_id, options, PHONGO_COMMAND_WRITE, NULL, NULL, &zwriteConcern TSRMLS_CC)) {
+	} else if (!phongo_execute_parse_options(client, server_id, options, PHONGO_COMMAND_WRITE, opts, NULL, &zwriteConcern TSRMLS_CC)) {
```
jmikola: With no other changes to this function, `opts` are never used when executing the bulk write.

This isn't a concern for read preference and read concern, as those don't apply here. We also handle write concern explicitly below with a setter.

That said, it looks like we're missing an API in libmongoc to allow us to set a session on the bulk write. /cc @ajdavis

php_phongo.c
```diff
@@ -738,11 +742,13 @@ int phongo_execute_query(mongoc_client_t *client, const char *namespace, zval *z
...
 		mongoc_collection_set_read_concern(collection, query->read_concern);
 	}
+	
+	opts = bson_new();
 
 	/* FIXME: Legacy way of specifying the readPreference option into this function */
 	if (options && Z_TYPE_P(options) == IS_OBJECT && instanceof_function(Z_OBJCE_P(options), php_phongo_readpreference_ce TSRMLS_CC)) {
 		zreadPreference = options;
-	} else if (!phongo_execute_parse_options(client, server_id, options, PHONGO_COMMAND_READ, NULL, &zreadPreference, NULL TSRMLS_CC)) {
+	} else if (!phongo_execute_parse_options(client, server_id, options, PHONGO_COMMAND_READ, opts, &zreadPreference, NULL TSRMLS_CC)) {
```
jmikola: Same issues with `opts` as the bulk write fixes (never freed, and never used after options are parsed).

Unlike bulk write, we do have a way to pass opts in with the find command with `query->opts`. I think it would make more sense to pass that in here.

tests/server/server-executeBulkWrite-004.phpt
```diff
@@ -10,13 +10,6 @@ require_once __DIR__ . "/../utils/basic.inc";
 $manager = new MongoDB\Driver\Manager(REPLICASET);
 $server = $manager->selectServer(new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_SECONDARY));
 
-$bulk = new MongoDB\Driver\BulkWrite();
-$bulk->insert(array('wc' => 0));
-
-$result = $server->executeBulkWrite(NS, $bulk, new MongoDB\Driver\WriteConcern(0));
-var_dump($result->isAcknowledged());
-var_dump($result->getInsertedCount());
```
jmikola: @ajdavis: This test currently fails with a "not master" error from the server. Were there changes in libmongoc 1.9.0 regarding the execution path for unacknowledged writes, whereby we now do get a server response and possible raise an error?

tests/server/server-executeCommand-002.phpt
```diff
@@ -64,6 +65,9 @@ object(stdClass)#%d (%d) {
       }
     }
   }
+  ["cursor"]=>
+  object(stdClass)#16 (0) {
+  }%A
```
jmikola: Is `%A` there to catch an implicit session ID (i.e. `lsid`) and `$clusterTime` (since we gossip it now)?

derickr: yes

PR #700, opened 2017-12-11
PHPC-1027: Introduce classes for deprecated BSON types

derickr:


src/BSON/DBPointer.c
```diff
@@ -0,0 +1,386 @@
...
+
+	php_phongo_dbpointer_init_from_hash(intern, props TSRMLS_CC);
+} /* }}} */
+
+/* {{{ proto string MongoDB\BSON\DBPointer::__toString()
+   Return the DBPointer's namespace string. */
+static PHP_METHOD(DBPointer, __toString)
+{
+	php_phongo_dbpointer_t *intern;
+	char                    *retval;
```
jmikola: Alignment is off by a space here.

src/BSON/DBPointer.c
```diff
@@ -0,0 +1,386 @@
...
+	object_init_ex(return_value, php_phongo_dbpointer_ce);
+
+	intern = Z_DBPOINTER_OBJ_P(return_value);
+	props = Z_ARRVAL_P(array);
+
+	php_phongo_dbpointer_init_from_hash(intern, props TSRMLS_CC);
+} /* }}} */
+
+/* {{{ proto string MongoDB\BSON\DBPointer::__toString()
+   Return the DBPointer's namespace string. */
```
jmikola: "namespace string and ObjectId"

src/BSON/DBPointer.c
```diff
@@ -0,0 +1,386 @@
...
+	retval_len = spprintf(&retval, 0, "[%s/%s]", intern->namespace, intern->oid);
+	PHONGO_RETVAL_STRINGL(retval, retval_len);
+	efree(retval);
+} /* }}} */
+
+/* {{{ proto string MongoDB\BSON\DBPointer::serialize()
+*/
+static PHP_METHOD(DBPointer, serialize)
+{
+	php_phongo_dbpointer_t  *intern;
```
jmikola: Indentation is also off by one space here. Is that an issue for all `php_phongo_dbpointer_t` variables?

tests/bson/bson-dbpointer-jsonserialize-001.phpt
```diff
@@ -0,0 +1,17 @@
+--TEST--
+MongoDB\BSON\Javascript::jsonSerialize() return value (without scope)
```
jmikola: I assume this was copy/pasted prematurely, since `DBPointer::jsonSerialize()` isn't implemented yet (although the other `jsonSerialize()` methods appear to be).

tests/bson/bson-dbpointerinterface-001.phpt
```diff
@@ -0,0 +1,14 @@
+--TEST--
+MongoDB\BSON\JavascriptInterface is implemented by MongoDB\BSON\Javascript
```
jmikola: I don't think we'll need interfaces for any of these deprecated classes.

tests/bson/bson-symbol-compare-001.phpt
```diff
@@ -0,0 +1,17 @@
+--TEST--
+MongoDB\BSON\Symbol comparisons
+--FILE--
+<?php
+
+var_dump(MongoDB\BSON\toPHP(MongoDB\BSON\fromJSON('{ "undefined": {"$symbol": "val1"} }')) == MongoDB\BSON\toPHP(MongoDB\BSON\fromJSON('{ "undefined": {"$symbol": "val1"} }')));
```
jmikola: "undefined" is a curious choice here. Something like "x" or "foo" might be more concise, and not cause any confusion with the actual Undefined type.

Edit: I noticed you used "symbol" in a later test. That certainly works.

tests/bson/bson-symbol-set_state-001.phpt
```diff
@@ -0,0 +1,19 @@
+--TEST--
+MongoDB\BSON\Symbol::__set_state()
+--FILE--
+<?php
+
+var_export(MongoDB\BSON\Symbol::__set_state([
+	'symbol' => "symbolValue",
```
jmikola: Tab alert.

tests/bson/bson-toPHP_error-005.phpt
```diff
@@ -3,9 +3,7 @@ MongoDB\BSON\toPHP(): BSON decoding ignores unsupported BSON types
 --FILE--
 <?php
 $tests = [
-    pack('VCa*xx', 10, 0x06, 'foo'), // undefined
     pack('VCa*xVa*xx12x', 37, 0x0C, 'foo', 11, 'collection'), // DBPointer
```
jmikola: I assume this entire test will ultimately need to be deleted.

tests/bson/bson-symbol-set_state-001.phpt
jmikola: @derickr: I'll see your :-1: and raise you a:

src/BSON/Symbol.c
```diff
@@ -0,0 +1,375 @@
...
+
+ZEND_BEGIN_ARG_INFO_EX(ai_Symbol_unserialize, 0, 0, 1)
+	ZEND_ARG_INFO(0, serialized)
+ZEND_END_ARG_INFO()
+
+ZEND_BEGIN_ARG_INFO_EX(ai_Symbol_void, 0, 0, 0)
+ZEND_END_ARG_INFO()
+
+static zend_function_entry php_phongo_symbol_me[] = {
+	PHP_ME(Symbol, __set_state, ai_Symbol___set_state, ZEND_ACC_PUBLIC|ZEND_ACC_STATIC)
```
jmikola: Does it make sense to implement this method instead of disabling it as we do the constructor?

With this gone, that would leave `unserialize()` as the only way for users to manually create these deprecated objects.  I'm not suggesting we remove `unserialize()`, as folks should be able to round-trip these classes through PHP serialization :)

tests/bson/bson-dbpointer-tostring-001.phpt
```diff
@@ -0,0 +1,14 @@
+--TEST--
+MongoDB\BSON\DBPointer::__toString()
```
jmikola: This implementation appears to be missing. Was it inadvertently removed? I remember seeing a `[%s/%s]` pattern in an earlier PR diff.

Also, should we have a `__toString()` method for Undefined that returns an empty string? That would make it consistent with `(string) null`.

src/BSON/Symbol.c
derickr: Removed the __set_state implementation all together.

src/BSON/DBPointer.c
```diff
@@ -0,0 +1,380 @@
...
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	intern = Z_DBPOINTER_OBJ_P(getThis());
+
+	array_init_size(return_value, 2);
+	ADD_ASSOC_STRINGL(return_value, "$ref", intern->ref, intern->ref_len);
+	ADD_ASSOC_STRING(return_value, "$id", intern->id);
```
jmikola: I just realized that the extended JSON spec calls for: `{"$dbPointer": {"$ref": <namespace [3] as a string>, "$id": ObjectId}}` to avoid confusion with DBRef notation. Can we wrap this in an outer array with a `$dbPointer` key?

PR #701, opened 2017-12-12
Test fixes for libmongoc 1.9.0-dev (post rc1)

jmikola:
https://jira.mongodb.org/browse/PHPC-1037

### DIFF
--- a/tests/bulk/bulkwrite-update_error-002.phpt
+++ b/tests/bulk/bulkwrite-update_error-002.phpt
@@ -28,10 +28,10 @@ echo throws(function() use ($bulk) {
 <?php exit(0); ?>
 --EXPECT--
 OK: Got MongoDB\Driver\Exception\InvalidArgumentException
-update document contains invalid key: empty key
+invalid argument for update: empty key
 
 OK: Got MongoDB\Driver\Exception\InvalidArgumentException
-update document contains invalid key: corrupt BSON
+invalid argument for update: corrupt BSON
 
 OK: Got MongoDB\Driver\Exception\InvalidArgumentException
 Invalid key 'z': update only works with $ operators
--- a/tests/manager/manager-executeCommand-004.phpt
+++ b/tests/manager/manager-executeCommand-004.phpt
@@ -1,5 +1,5 @@
 --TEST--
-MongoDB\Driver\Manager::executeCommand() options
+MongoDB\Driver\Manager::executeCommand() options (MONGOC_CMD_RAW)
 --SKIPIF--
 <?php if (getenv("TRAVIS")) exit("skip This currently tails on Travis because it doesn't run 3.6 yet"); ?>
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
--- a/tests/manager/manager-executeCommand-004.phpt
+++ b/tests/manager/manager-executeCommand-004.phpt
@@ -1,5 +1,5 @@
 --TEST--
-MongoDB\Driver\Manager::executeRawCommand()
+MongoDB\Driver\Manager::executeCommand() options
 --SKIPIF--
 <?php if (getenv("TRAVIS")) exit("skip This currently tails on Travis because it doesn't run 3.6 yet"); ?>
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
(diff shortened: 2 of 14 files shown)


tests/bulk/bulkwrite-update_error-002.phpt
```diff
@@ -28,10 +28,10 @@ echo throws(function() use ($bulk) {
 <?php exit(0); ?>
 --EXPECT--
 OK: Got MongoDB\Driver\Exception\InvalidArgumentException
-update document contains invalid key: empty key
+invalid selector for update: empty key
```
jmikola: Note: this expected output will need to be changed again once https://github.com/mongodb/mongo-c-driver/pull/472 is merged.

tests/manager/manager-executeCommand-004.phpt
```diff
@@ -0,0 +1,48 @@
+--TEST--
+MongoDB\Driver\Manager::executeCommand() options
```
jmikola: This file was renamed from `manager-executeRawCommand-001.phpt`

tests/manager/manager-executeCommand-004.phpt
```diff
@@ -0,0 +1,48 @@
...
+        echo "Write Concern: ", $command->writeConcern->w, "\n";
+    }
+);
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+Read Preference: secondary
+Read Concern: local
```
jmikola: Note: this test currently fails because [CDRIVER-2418](https://jira.mongodb.org/browse/CDRIVER-2418) now ignores the `readConcern` option unless the read flag is set for the command (not done for the raw and write modes). This should be fixed by [CDRIVER-2422](https://jira.mongodb.org/browse/CDRIVER-2422).

tests/apm/monitoring-addSubscriber-001.phpt
```diff
@@ -10,18 +10,18 @@ $m = new MongoDB\Driver\Manager(STANDALONE);
...
-	}
-
-	public function commandSucceeded( \MongoDB\Driver\Monitoring\CommandSucceededEvent $event )
-	{
-	}
-
-	public function commandFailed( \MongoDB\Driver\Monitoring\CommandFailedEvent $event )
-	{
-	}
+    public function commandStarted( \MongoDB\Driver\Monitoring\CommandStartedEvent $event )
```
jmikola: I noticed that some of the newer `executeCommand()` tests were using tabs instead of spaces. A text search turned up many more occurrences, so I've fixed them all in a single commit within this PR.

derickr: I think this is a pointless change. And would rather have seen that in a specific PR, as now it's more difficult to review this as I have to go through all of it.

tests/bulk/bulkwrite-update_error-002.phpt
derickr: Do we care enough about the exact word? I'd be happy if we just just %s for the word "selector"/"argument".

tests/manager/manager-executeCommand-004.phpt
derickr: I had specifically named this to match that PR's changes - where we have Raw, Read, Write, and ReadWrite, so I am not sure if this a good idea.

jmikola: Was this the only occurrence of "Raw" from that PR? I gathered it was more confusing to have a test named in the `class-method` for a method that doesn't exist.

We both know that `executeCommand()` implies `MONGO_CMD_RAW`. Would adding "Raw Command" (or similar) to the test title be a fair compromise?

tests/bulk/bulkwrite-update_error-002.phpt
jmikola: I personally don't care, but without this test we wouldn't have caught the mistake in libmongoc. I don't think libmongoc is going to add tests for all of its error messages, so I'm OK with keeping this around.

tests/manager/manager-executeCommand-004.phpt
jmikola: Just noticed that `manager-executeCommand_error-002.phpt` was added in the same PR and didn't use the "RawCommand" name.

PR #705, opened 2017-12-14
PHPC-1070 and PHPC-1071: Report class name for unexpected object values

jmikola:
https://jira.mongodb.org/browse/PHPC-1070
https://jira.mongodb.org/browse/PHPC-1071

Adds a macro for accessing zval's class or type name, depending on whether it's an object type. This replaces existing ternary statements we had for some BSON exceptions. I've also gone through and updated exceptions where object types are entirely unexpected. Options that expect a BSON document (i.e. array or object) remain unchanged and do not use this macro, as any object type would be accepted.

IMO, this makes the exceptions more helpful and also makes PHPC more consistent with the invalid type exception messages in PHPLIB.

### DIFF
--- a/src/bson-encode.c
+++ b/src/bson-encode.c
@@ -143,24 +143,18 @@ static void php_phongo_bson_append_object(bson_t *bson, php_phongo_bson_flags_t 
 			if (Z_TYPE(obj_data) != IS_ARRAY && !(Z_TYPE(obj_data) == IS_OBJECT && instanceof_function(Z_OBJCE(obj_data), zend_standard_class_def TSRMLS_CC))) {
 				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC,
 					"Expected %s::%s() to return an array or stdClass, %s given",
-					Z_OBJCE_P(object)->name->val,
+					ZSTR_VAL(Z_OBJCE_P(object)->name),
 					BSON_SERIALIZE_FUNC_NAME,
-					(Z_TYPE(obj_data) == IS_OBJECT
-						 ? Z_OBJCE(obj_data)->name->val
-						 : zend_get_type_by_const(Z_TYPE(obj_data))
-					)
+					PHONGO_ZVAL_CLASS_OR_TYPE_NAME(obj_data)
 				);
 				zval_ptr_dtor(&obj_data);
 #else
 			if (Z_TYPE_P(obj_data) != IS_ARRAY && !(Z_TYPE_P(obj_data) == IS_OBJECT && instanceof_function(Z_OBJCE_P(obj_data), zend_standard_class_def TSRMLS_CC))) {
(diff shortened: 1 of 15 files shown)


src/bson-encode.c
```diff
@@ -143,24 +143,18 @@ static void php_phongo_bson_append_object(bson_t *bson, php_phongo_bson_flags_t
...
-					)
+					PHONGO_ZVAL_CLASS_OR_TYPE_NAME(obj_data)
 				);
 				zval_ptr_dtor(&obj_data);
 #else
 			if (Z_TYPE_P(obj_data) != IS_ARRAY && !(Z_TYPE_P(obj_data) == IS_OBJECT && instanceof_function(Z_OBJCE_P(obj_data), zend_standard_class_def TSRMLS_CC))) {
 				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC,
 					"Expected %s::%s() to return an array or stdClass, %s given",
-					Z_OBJCE_P(object)->name,
+					ZSTR_VAL(Z_OBJCE_P(object)->name),
```
jmikola: Technically, this isn't necessary since we're in PHP 5 mode here, but it reduces some variation between the PHP 5 and 7 blocks. The only real difference in these blocks is that `obj_data` is a `zval *` for PHP 5.

src/bson-encode.c
```diff
@@ -430,22 +424,13 @@ void php_phongo_zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *
 #endif
 					phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC,
 						"Expected %s::%s() to return an array or stdClass, %s given",
-#if PHP_VERSION_ID >= 70000
-						Z_OBJCE_P(data)->name->val,
-#else
-						Z_OBJCE_P(data)->name,
-#endif
+						ZSTR_VAL(Z_OBJCE_P(data)->name),
```
jmikola: We could have eliminated the `#if` block much earlier, but I just noticed this.

src/bson-encode.c
```diff
@@ -430,22 +424,13 @@ void php_phongo_zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *
...
 #if PHP_VERSION_ID >= 70000
-						(Z_TYPE(obj_data) == IS_OBJECT
-							 ? Z_OBJCE(obj_data)->name->val
-							 : zend_get_type_by_const(Z_TYPE(obj_data))
+						PHONGO_ZVAL_CLASS_OR_TYPE_NAME(obj_data)
 #else
-						(Z_TYPE_P(obj_data) == IS_OBJECT
-							 ? Z_OBJCE_P(obj_data)->name
-							 : zend_get_type_by_const(Z_TYPE_P(obj_data))
+						PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(obj_data)
```
jmikola: Only difference here is that `obj_data` is a `zval *` for PHP 5.

PR #706, opened 2017-12-15
PHPC-1072: mongoc_client_command_with_opts() reply must always be freed

jmikola:
https://jira.mongodb.org/browse/PHPC-1072

Previously, reply might leak if `phongo_execute_command()` returned due to an error (result=false).


jmikola:
> But please don't add random formatting changes to functional pull requests :-)

I removed the changes to `manager-executeCommand-004.phpt` and `server-executeCommand-006.phpt`.

PR #710, opened 2017-12-19
PHPC-987: Causally consistent reads

jmikola:
https://jira.mongodb.org/browse/PHPC-987

### DIFF
--- a/src/MongoDB/Session.c
+++ b/src/MongoDB/Session.c
@@ -27,6 +27,61 @@
 
 zend_class_entry *php_phongo_session_ce;
 
+static bool php_phongo_session_get_timestamp_parts(zval *obj, uint32_t *timestamp, uint32_t *increment TSRMLS_DC)
+{
+	bool retval = false;
+#if PHP_VERSION_ID >= 70000
+	zval ztimestamp;
+	zval zincrement;
+
+	zend_call_method_with_0_params(obj, NULL, NULL, "getTimestamp", &ztimestamp);
+
+	if (Z_ISUNDEF(ztimestamp) || EG(exception)) {
+		goto cleanup;
+	}
+
+	zend_call_method_with_0_params(obj, NULL, NULL, "getIncrement", &zincrement);
+
+	if (Z_ISUNDEF(zincrement) || EG(exception)) {
+		goto cleanup;
+	}
+
+	*timestamp = Z_LVAL(ztimestamp);
+	*increment = Z_LVAL(zincrement);
+#else
+	zval *ztimestamp = NULL;
+	zval *zincrement = NULL;
+
+	zend_call_method_with_0_params(&obj, NULL, NULL, "getTimestamp", &ztimestamp);
+
+	if (Z_ISUNDEF(ztimestamp) || EG(exception)) {
+		goto cleanup;
+	}
+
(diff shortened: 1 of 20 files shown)


src/MongoDB/Session.c
```diff
@@ -27,6 +27,38 @@
 
 zend_class_entry *php_phongo_session_ce;
 
+static bool php_phongo_session_get_timestamp_parts(zval *obj, uint32_t *timestamp, uint32_t *increment)
+{
+#if PHP_VERSION_ID >= 70000
+	zval ztimestamp;
+	zval zincrement;
+
+	zend_call_method_with_0_params(obj, NULL, NULL, "getTimestamp", &ztimestamp);
```
derickr: I'm uncomfortable with calling methods through the PHP/Zend layer here. Is there no other way to obtain the timestamp?

jmikola:
> Please fix the commit message, there is nothing casual about this (it's causally).

The name of the spec is [Causal Consistency](https://github.com/mongodb/specifications/blob/master/source/causal-consistency/causal-consistency.rst) (also: [wiki](https://en.wikipedia.org/wiki/Causal_consistency)). The reads themselves are said to be causally consistent.

> Fails to compile on PHP 5 with zts. On PHP 7 there is a test failure.

I'll look into the test failures and revise. This was admittedly pushed up for review as I was headed out the door last night (wanted you to see something before this morning), so I can't say I'm surprised.

src/MongoDB/Session.c
jmikola: I can certainly add logic to check for a Timestamp object and access these properties directly without invoking methods, and will do so.

That said, there is no other solution if we're dealing with a non-Timestamp TimestampInterface.

In my testing, if the `getTimestamp()` throws, `getIncrement()` is never even called. This may be the result of logic inside `zend_call_method` that no-ops if a previous call failed or an exception has already been thrown. I still found that a bit odd, as here I am invoking `zend_call_method` twice without logic in-between. If you're more comfortable with it, I can also split up those calls and check for errors between them, instead of a combined check for undefined `ztimestamp` and `zincrement` down below.

derickr: As this goes through zend_execute, I expected that the throw would abort further calls. I think it's better to check for EG(exception) in between. But you're right on the non-Timestamp TimestampInterface. I now believe you shouldn't short-cut it as an exception for "Timestamp" too.

jmikola: The `IS_UNDEF` check was inspired by what I saw in `zend_interfaces.c`:

 * https://github.com/php/php-src/blob/php-7.2.0/Zend/zend_interfaces.c#L160
 * https://github.com/php/php-src/blob/php-7.2.0/Zend/zend_interfaces.c#L191
 * https://github.com/php/php-src/blob/php-7.2.0/Zend/zend_interfaces.c#L388

I actually think the best approach is checking for `(Z_TYPE(retval) == IS_UNDEF || EG(exception))`, as is done in the last example (`zend_user_serialize`).

I personally don't see any harm in optimizing for the common use case (Timestamp instance), and simply forgot to make that change before I last updated this PR. I'll leave it out for now given your feedback, but we can always add it in later if you change your mind.

PR #719, opened 2018-01-02
PHPC-1065: Enable BSON corpus tests for deprecated types

derickr:


tests/bson-corpus/multi-type-valid-001.phpt
```diff
@@ -1,33 +0,0 @@
---TEST--
-Multiple types within the same document: All BSON types
```
jmikola: Why was this removed? AFAIK, the [test still exists](https://github.com/mongodb/specifications/blob/master/source/bson-corpus/tests/multi-type.json) and we have it correctly marked as XFAIL due to our 32-bit integer handling.

derickr: Wasn't removed, just renamed according to the original BSON corpus source. Don't know why GIT didn't detect this.

jmikola: I may be missing something, but I see no evidence of a rename in this commit. The new `multi-type-deprecated-valid-001.phpt` file corresponds to [`multi-type-deprecated.json`](https://github.com/mongodb/specifications/blob/master/source/bson-corpus/tests/multi-type-deprecated.json) and includes very different content.

PR #720, opened 2018-01-02
PHPC-1049: Build warnings on CLang/LLVM

derickr:


src/BSON/ObjectId.c
```diff
@@ -369,10 +369,6 @@ static HashTable *php_phongo_objectid_get_properties_hash(zval *object, bool is_
 
 	PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_debug, intern, props, 1);
 
-	if (!intern->oid) {
-		return props;
-	}
```
jmikola: Why was this removed? All of our `get_properties_hash` functions have such a check to return `props` early if the object is not initialized. Binary has:

```
if (!intern->data) {
	return props;
}
```

For structs that don't have a null pointer, we have an `initialized` property, as with UTCDateTime:

```
if (!intern->initialized) {
	return props;
}
```

derickr: As this if() statement always evaluated to false (as intern->oid was always true), this actually never happened, and it didn't fail any tests...

I hadn't realise we did this for stuff that doesn't have a null pointer though, so will need to change that, but the question is really: how do we test this?

derickr: I've pushed a new commit to put back intern->initialized

src/BSON/ObjectId.c
```diff
@@ -369,7 +369,7 @@ static HashTable *php_phongo_objectid_get_properties_hash(zval *object, bool is_
 
 	PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_debug, intern, props, 1);
 
-	if (!intern->oid) {
+	if (!intern->initialized) {
```
jmikola: Good catch. I forgot that `oid` was a statically allocated char array. That might explain why the warning only complained about ObjectId and not the `char *` fields in other structs.

src/BSON/ObjectId.c
jmikola: > (should really be initialised :uk:)

That's poppycock. See: https://stackoverflow.com/questions/157807/gb-english-or-us-english ([this answer](https://stackoverflow.com/a/157841/162228) in particular)

PR #728, opened 2018-01-04
Upgrade vm to trusty64 and revise install for pip and mo

jmikola:


bin/prep-release.php
```diff
@@ -62,7 +62,7 @@ function get_files() {
         "scripts/centos/ldap/users",
         "scripts/presets/*.{json}",
         "scripts/ssl/*.pem",
-        "scripts/*.{php,sh}",
+        "scripts/*.{json,php,py,sh}",
```
jmikola: Note: `json` contains the mongo orchestration config and should have been included long ago.

scripts/ubuntu/mongo-orchestration.sh
```diff
@@ -23,11 +23,11 @@ dpkg -x mongodb-enterprise-server_3.2.0_amd64.deb 3.2.0
...
 dpkg -x mongodb-enterprise-server_3.6.0_amd64.deb 3.6.0
 
-
 # Python stuff for mongo-orchestration
-apt-get install -y python python-dev python-pip
+apt-get install -y python python-dev
+python get-pip.py
 
-pip install --index-url=https://pypi.python.org/simple/ --upgrade 'git+https://github.com/10gen/mongo-orchestration.git#egg=mongo_orchestration'
+pip install --upgrade mongo-orchestration
```
jmikola: This is the [documented method for installing mongo-orchestration](https://github.com/10gen/mongo-orchestration#installation), provided that pip is up-to-date.

The original command was so verbose because the `python-pip` package on Ubuntu Precise (12.04) is rather ancient.

scripts/ubuntu/mongo-orchestration.sh
```diff
@@ -23,11 +23,11 @@ dpkg -x mongodb-enterprise-server_3.2.0_amd64.deb 3.2.0
 dpkg -x mongodb-enterprise-mongos_3.2.0_amd64.deb 3.2.0
 dpkg -x mongodb-enterprise-server_3.6.0_amd64.deb 3.6.0
 
-
 # Python stuff for mongo-orchestration
-apt-get install -y python python-dev python-pip
+apt-get install -y python python-dev
+python get-pip.py
```
jmikola: This is the install process suggested in the [pip documentation](https://pip.pypa.io/en/stable/installing/).

derickr: Sorry Jeremy, I deleted your comment here by mistake. It said:

> This is the install process suggested in the [pip documentation](https://pip.pypa.io/en/stable/installing/).

scripts/ubuntu/mongo-orchestration.sh
derickr: Shouldn't we upgrade our Precise thing to something newer as well then?

jmikola: New commits updates the VM to trusty64.

PR #730, opened 2018-01-04
PHPC-1050: Command cursor should not invoke getMore at execution

jmikola:
https://jira.mongodb.org/browse/PHPC-1050

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -248,6 +248,7 @@ static void phongo_cursor_init(zval *return_value, mongoc_client_t *client, mong
 	intern->cursor = cursor;
 	intern->server_id = mongoc_cursor_get_hint(cursor);
 	intern->client = client;
+	intern->advanced = false;
 
 	if (readPreference) {
 #if PHP_VERSION_ID >= 70000
@@ -285,6 +286,10 @@ static void phongo_cursor_init_for_query(zval *return_value, mongoc_client_t *cl
 
 	/* namespace has already been validated by phongo_execute_query() */
 	phongo_split_namespace(namespace, &intern->database, &intern->collection);
+
+	/* cursor has already been advanced by phongo_execute_query() calling
+	 * phongo_cursor_advance_and_check_for_error() */
+	intern->advanced = true;
 
 #if PHP_VERSION_ID >= 70000
 	ZVAL_ZVAL(&intern->query, query, 1, 0);
@@ -709,9 +714,9 @@ bool phongo_execute_bulk_write(mongoc_client_t *client, const char *namespace, p
 	return success;
 } /* }}} */
 
-/* Advance the cursor and return whether there is an error. On error, the cursor
- * will be destroyed and an exception will be thrown. */
-static bool phongo_advance_cursor_and_check_for_error(mongoc_cursor_t *cursor TSRMLS_DC)
+/* Advance the cursor and return whether there is an error. On error, false is
+ * returned and an exception is thrown. */
+bool phongo_cursor_advance_and_check_for_error(mongoc_cursor_t *cursor TSRMLS_DC) /* {{{ */
 {
 	const bson_t *doc;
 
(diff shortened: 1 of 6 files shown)


tests/cursor/bug1050-002.phpt
```diff
@@ -0,0 +1,74 @@
...
+$cursor = $manager->executeReadCommand(DATABASE_NAME, $cmd);
+printf("Executing command took %0.6f seconds\n", microtime(true) - $start);
+
+$it = new IteratorIterator($cursor);
+
+printf("Current position is valid: %s\n", $it->valid() ? 'yes' : 'no');
+
+$start = microtime(true);
+$it->next();
+printf("Advancing cursor took %0.6f seconds\n", microtime(true) - $start);
```
jmikola: Unlike the first test, this one omits the initial call to `rewind()` and tests that we preserve existing behavior by not requiring the rewind handler to ever be invoked (only relevant when wrapping a cursor with IteratorIterator).

src/MongoDB/Cursor.c
```diff
@@ -96,7 +96,15 @@ static void php_phongo_cursor_iterator_move_forward(zend_object_iterator *iter T
...
-	cursor->current++;
+
+	/* If the cursor has already advanced, increment its position. Otherwise,
+	 * the first call to mongoc_cursor_next() will be made below and we should
+	 * leave its position at zero. */
+	if (cursor->advanced) {
+		cursor->current++;
+	} else {
+		cursor->advanced = true;
+	}
```
jmikola: Future improvement: refactor `phongo_cursor_advance_and_check_for_error()` so that it can take an optional `bson_t *`. Then `phongo_execute_query()` and `php_phongo_cursor_iterator_rewind()` could pass in `null`, since they only care about error checking and don't need to access the document. Meanwhile, `php_phongo_cursor_iterator_move_forward()` could pass in a `bson_t *` and, on success, use that to populate the current element, as is done with:

```c
php_phongo_bson_to_zval_ex(bson_get_data(doc), doc->len, &cursor->visitor_data);
```

php_phongo.c
```diff
@@ -788,7 +763,8 @@ int phongo_execute_query(mongoc_client_t *client, const char *namespace, zval *z
 		mongoc_cursor_set_max_await_time_ms(cursor, query->max_await_time_ms);
 	}
 
-	if (!phongo_advance_cursor_and_check_for_error(cursor TSRMLS_CC)) {
+	if (!phongo_cursor_advance_and_check_for_error(cursor TSRMLS_CC)) {
+		mongoc_cursor_destroy(cursor);
```
derickr: It should *not* destroy the cursor, as ``phongo_cursor_advance_and_check_for_error`` already does that.

php_phongo.c
```diff
@@ -916,15 +892,33 @@ int phongo_execute_command(mongoc_client_t *client, php_phongo_command_type_t ty
...
-	if (!phongo_advance_cursor_and_check_for_error(cmd_cursor TSRMLS_CC)) {
-		/* If an error is found, the cmd_cursor is destroyed already */
-		return false;
-	}
-
 	phongo_cursor_init_for_command(return_value, client, cmd_cursor, db, zcommand, zreadPreference TSRMLS_CC);
 	return true;
 } /* }}} */
 
+/* Advance the cursor and return whether there is an error. On error, false is
```
derickr: Moving this down here made it impossible to see the changes in the diff, can you please put it back in its original location? That way, I would have spotted that you removed the ``mongoc_cursor_destroy(cursor);`` from it.

src/MongoDB/BulkWrite.c
```diff
@@ -188,7 +188,7 @@ static PHP_METHOD(BulkWrite, __construct)
 		ordered = php_array_fetchc_bool(options, "ordered");
 	}
 
-	intern->bulk = phongo_bulkwrite_init(ordered);
+	intern->bulk = mongoc_bulk_operation_new(ordered);
```
derickr: This is really unrelated to the patch, you just removed ``phongo_bulkwrite_init`` as it only did this one thing. I would suggest you don't do this in this single commit.

src/MongoDB/Cursor.c
```diff
@@ -117,6 +125,16 @@ static void php_phongo_cursor_iterator_rewind(zend_object_iterator *iter TSRMLS_
 	php_phongo_cursor_t        *cursor = cursor_it->cursor;
 	const bson_t               *doc;
 
+	/* If the cursor was never advanced (e.g. command cursor), do so now */
+	if (!cursor->advanced) {
+		cursor->advanced = true;
+
+		if (!phongo_cursor_advance_and_check_for_error(cursor->cursor TSRMLS_CC)) {
+			/* Exception should already have been thrown */
```
derickr: Yes, but shouldn't you destroy the cursor now?

php_phongo.c
jmikola: `phongo_cursor_advance_and_check_for_error()` no longer destroys the cursor. That change was made because we don't want to destroy it from within the iterator handlers.

src/MongoDB/Cursor.c
jmikola: Once we create a `MongoDB\Driver\Cursor` with the `mongoc_cursor_t`, the only place we destroy the libmongoc cursor is from `php_phongo_cursor_free_object()`.

jmikola:
The bulk write refactoring was split out into #731.

PR #734, opened 2018-01-09
PHPC-1092: Don't reset hash on successive get_properties calls

jmikola:
https://jira.mongodb.org/browse/PHPC-1092

This removes a call to `zend_hash_clean()` on successive calls to a BSON object's `get_properties` handler, which can lead to an infinite loop when iterating the object's properties with foreach on PHP 7.x.

For historical reference, the original call to `zend_hash_clean()` was introduced in ad1f3f049f5c101a8b4f89f13e6384bbefb6ef6c for #607 (PHPC-939).

Note: I've only added tests for BSON classes that have a custom `get_properties` handler and utilize the `PHONGO_GET_PROPERTY_HASH_INIT_PROPS()` macro. I do not recall why I originally added  `zend_clean_hash()`, but I assume I was trying to be strict about not inheriting hash state from previous calls.

While testing this, I was surprised to find that iterating the object with `foreach` actually invokes `get_properties` multiple times. If I had to guess, the fact that [`zend_hash_clean()` resets the hash pointer](https://github.com/php/php-src/blob/php-7.2.1/Zend/zend_hash.c#L1333) was causing `foreach` to never advance past the first property.

### DIFF
--- a/php_phongo.h
+++ b/php_phongo.h
@@ -196,7 +196,6 @@ zend_bool phongo_writeconcernerror_init(zval *return_value, bson_t *bson TSRMLS_
         ALLOC_HASHTABLE(props);                                                 \
         zend_hash_init((props), (size), NULL, ZVAL_PTR_DTOR, 0);                \
     } else if ((intern)->properties) {                                          \
-        zend_hash_clean((intern)->properties);                                  \
         (props) = (intern)->properties;                                         \
     } else {                                                                    \
         ALLOC_HASHTABLE(props);                                                 \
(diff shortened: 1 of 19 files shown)


jmikola:
> While testing this, I was surprised to find that iterating the object with `foreach` actually invokes `get_properties` multiple times. If I had to guess, the fact that [`zend_hash_clean()` resets the hash pointer](https://github.com/php/php-src/blob/php-7.2.1/Zend/zend_hash.c#L1333) was causing `foreach` to never advance past the first property.

Indeed, `ZEND_FE_FETCH_R` does invoke the `get_properties` handler on each execution: https://github.com/php/php-src/blob/php-7.2.1/Zend/zend_vm_def.h#L5912

PR #736, opened 2018-01-10
PHPC-888: Support additional TLS libraries for bundled libmongoc

jmikola:
https://jira.mongodb.org/browse/PHPC-888
https://jira.mongodb.org/browse/PHPC-1003


scripts/build/autotools/CheckSSL.m4
```diff
@@ -0,0 +1,240 @@
...
+  if test "$PHP_MONGODB_SSL" = "openssl" -a "$found_openssl" != "yes"; then
+    AC_MSG_ERROR([OpenSSL libraries and development headers could not be found])
+  fi
+])
+
+AS_IF([test "$PHP_MONGODB_SSL" = "libressl" -o "$PHP_MONGODB_SSL" = "auto"],[
+  found_libressl="no"
+
+  dnl TODO: this may require adding libcrypto to the modules list, as pkg-config does not seem to include it (unlike openssl)
+  PKG_CHECK_MODULES([PHP_MONGODB_SSL],[libtls],[
```
jmikola: Note: when testing this on my Ubuntu system, I actually had to change "libtls" to "libressl-tls" (since I'm using [a custom PPA](https://launchpad.net/~strukturag/+archive/ubuntu/libressl)). I also noticed that `pkg-config --libs libressl-tls` returns only `-lressl-tls`, unlike `pkg-config --libs openssl`, which returns `-lssl -lcrypto`. I'm not sure if that's specific to the PPA package I'm using, so I'll have to test with LibreSSL proper.

FWIW, libmongoc also only specifics "libtls", so I'm curious if their LibreSSL builds are linking `-lcrypto` or not.

config.m4
```diff
@@ -83,7 +79,7 @@ AC_DEFUN([PHP_BSON_CLOCK],
   fi
 ])
 
-if test "$MONGODB" != "no"; then
+if test "$PHP_MONGODB" != "no"; then
```
derickr: Heh - this must never have worked then.

scripts/build/autotools/CheckHost.m4
```diff
@@ -0,0 +1,54 @@
...
+os_gnu=no
+
+case "$host" in
+    *-mingw*|*-*-cygwin*)
+        os_win32=yes
+        TARGET_OS=windows
+        ;;
+    *-*-*netbsd*)
+        os_netbsd=yes
+	TARGET_OS=unix
```
derickr: There are some whitespace issues in this file.

scripts/build/autotools/CheckSSL.m4
```diff
@@ -0,0 +1,240 @@
+PHP_ARG_WITH([mongodb-ssl],
+             [whether to enable crypto and TLS],
+             [AS_HELP_STRING([--with-mongodb-ssl=@<:@auto/no/openssl/libressl/darwin@:>@],
+                             [Enable TLS connections and SCRAM-SHA-1 authentication [default=auto]])],
```
derickr: I think all the help options, should be prefixed with "  MongoDB: ", just as I currently do at https://github.com/xdebug/xdebug/blob/master/config.m4#L7

scripts/build/autotools/CheckSSL.m4
```diff
@@ -0,0 +1,240 @@
...
+      PHP_MONGODB_SSL="openssl"
+      found_openssl="yes"
+    elif test "$PHP_MONGODB_SSL" = "openssl"; then
+      AC_MSG_ERROR([OpenSSL development headers could not be found])
+    fi
+
+    AC_CHECK_DECLS([ASN1_STRING_get0_data],
+                   [have_ASN1_STRING_get0_data="yes"],
+                   [have_ASN1_STRING_get0_data="no"],
+                   [[#include <openssl/asn1.h>]])
```
derickr: Why are there two sets of [ ] here?

scripts/build/autotools/CheckSSL.m4
```diff
@@ -0,0 +1,240 @@
...
+             AC_HELP_STRING([--enable-system-ciphers],
+                            [whether to use system crypto profile (deprecated for --enable-mongodb-crypto-system-profile) [default=no]]),
+             [no],
+             [no])
+
+AS_IF([test "$PHP_MONGODB_SSL" = "openssl" -o "$PHP_MONGODB_SSL" = "auto"],[
+  found_openssl="no"
+
+  PKG_CHECK_MODULES([PHP_MONGODB_SSL],[openssl],[
+    dnl PKG_CHECK_MODULES() doesn't check for headers, so confirm they exist (mainly for macOS)
```
derickr: That seems odd. The whole idea behind pkg_config  is that its --includes (or whatever the flag is), shows all the header locations. I don't think we should go check all the headers manually where pkg-config works.

scripts/build/autotools/CheckSSL.m4
derickr: It's possible that your PPA doesn't need the -lcrypto, and just that -lressl-tls contains all of what -lssl and -lcrypt export. I would actually go as far to make sure that we can support your custom PPA, and the normal libressl.

scripts/build/autotools/CheckSSL.m4
```diff
@@ -0,0 +1,240 @@
...
+AS_IF([test "$PHP_MONGODB_SSL" = "auto"],[
+  PHP_MONGODB_SSL="no"
+])
+
+AC_MSG_CHECKING([which TLS library to use])
+AC_MSG_RESULT([$PHP_MONGODB_SSL])
+
+dnl Disable Windows SSL and crypto
+AC_SUBST(MONGOC_ENABLE_SSL_SECURE_CHANNEL, 0)
+AC_SUBST(MONGOC_ENABLE_CRYPTO_CNG, 0)
```
derickr: Is the idea to support these later?

scripts/build/autotools/CheckSSL.m4
```diff
@@ -0,0 +1,240 @@
...
+  CFLAGS="$PHP_MONGODB_SSL_CFLAGS"
+
+  AC_CHECK_DECL([ASN1_STRING_get0_data],
+                [AC_SUBST(MONGOC_HAVE_ASN1_STRING_GET0_DATA, 1)],
+                [AC_SUBST(MONGOC_HAVE_ASN1_STRING_GET0_DATA, 0)],
+                [[#include <openssl/asn1.h>]])
+
+  CFLAGS="$old_CFLAGS"
+else
+  AC_SUBST(MONGOC_HAVE_ASN1_STRING_GET0_DATA, 0)
```
derickr: You've already done the ASN1_STRING_get0_data check earlier. Do you need it twice?

config.m4
jmikola: No sir, it did not :)

scripts/build/autotools/CheckSSL.m4
jmikola: I believe I forgot to remove this if/else block from the file. The problem with this is that `CFLAGS` may change depending on whether we found OpenSSL with `pkg-config` (sets `PHP_MONGODB_SSL_CFLAGS`) or manual header searching and `PHP_CHECK_LIBRARY` (sets `OPENSSL_INCDIR`).

Therefore, this if/else block should be removed and we should only keep the one below here:

```m4
if test "x$have_ASN1_STRING_get0_data" = "xyes"; then
   AC_SUBST(MONGOC_HAVE_ASN1_STRING_GET0_DATA, 1)
else
   AC_SUBST(MONGOC_HAVE_ASN1_STRING_GET0_DATA, 0)
fi
```

scripts/build/autotools/CheckSSL.m4
jmikola: Quoting https://lists.gnu.org/archive/html/bug-gnulib/2011-09/msg00172.html:

> # is a comment character in m4, which reads until end-of-line to end the comment. Which means all the [] after the comment character are treated as part of the comment, instead of as m4 [] quoting.

https://lists.gnu.org/archive/html/bug-gnulib/2011-09/msg00193.html discusses modifying `AC_CHECK_DECLS` to double-quote.

Note that we also used double-quoting in the [original check in `config.m4`](https://github.com/mongodb/mongo-php-driver/blob/e51a184717728e20f6f50280bf7dcd3156900951/config.m4#L309).

scripts/build/autotools/CheckSSL.m4
jmikola: I don't believe so, since Windows builds use `config.w32` and will never use Autotools. We do still need to make the substitutions, though.

scripts/build/autotools/CheckHost.m4
jmikola: It looks like the `TARGET_OS` assignments all used tab indents. Fixing now.

jmikola:
@sibidharan: Since you did such a great job solving the BoringSSL issues, I wonder if you'd be willing to help us test this PR on macOS. I plan to add macOS to our Travis CI matrix, but would still appreciate a fresh set of eyes on it.

This PR introduces a new `configure` option, `--with-mongodb-ssl=[auto/no/openssl/libressl/darwin]`. Specifying `--with-mongodb-ssl=darwin` should allow you to build the driver with its bundled version of libbson and libmongoc _and_ instruct libmongoc to use [Secure Transport](https://developer.apple.com/documentation/security/secure_transport) instead of OpenSSL. This will ultimately remove the need in your BoringSSL workaround to first install libbson and libmongoc as system libraries. Once compiled, you should also be able to verify that the driver is using its bundled libmongoc with Secure Transport via `phpinfo()` output.

sibidharan:
That is indeed such a great news @jmikola

I will test it out on all our peer systems running on couple different versions of MacOS.

jmikola:
@derickr: The last commit adds three macOS environments to the build matrix:

 * `xcode8.3` with `php56` and `--with-mongodb-ssl=auto` (OpenSSL)
 * `xcode9.2` with `php72` and `--with-mongodb-ssl=auto` (OpenSSL)
 * `xcode9.2` with `php72` and `--with-mongodb-ssl=darwin` (Secure Transport)

Note that since we only have a standalone `mongod` running on Travis, this is only testing that the driver builds successfully and can execute non-SSL code paths. We'll definitely want to do a manually test on a Mac later on.

I've no idea when these will actually complete, as [Travis CI status](https://www.traviscistatus.com/) is reporting degraded performance for macOS builds and their backlog appears to consistently be >1000 jobs.

.travis.scripts/before_script.sh
```diff
@@ -0,0 +1,21 @@
...
+    CONFIGURE_OPTS="${CONFIGURE_OPTS} --with-mongodb-ssl=${SSL_VERSION}"
+fi
+
+phpize
+./configure ${CONFIGURE_OPTS}
+make all -j4
+make install
+
+# Use latest run-tests.php, as older versions don't report a non-zero exit code for failures
+wget -O run-tests.php https://raw.githubusercontent.com/php/php-src/master/run-tests.php
```
derickr: Do we still need to do this?

tests/connect/standalone-x509-extract_username-002.phpt
```diff
@@ -2,6 +2,7 @@
 Connect to MongoDB with SSL and X509 auth and username retrieved from cert (stream context)
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS_SSL(['OpenSSL', 'Secure Transport', 'Secure Channel']); ?>
```
derickr: Why can this not be LibreSSL?

scripts/build/autotools/m4/pkg.m4
```diff
@@ -0,0 +1,199 @@
+# pkg.m4 - Macros to locate and utilise pkg-config.            -*- Autoconf -*-
+# serial 1 (pkg-config-0.24)
```
derickr: This file isn't one you wrote yourself, right?

config.m4
```diff
@@ -502,7 +477,7 @@ if test "$MONGODB" != "no"; then
       AC_OUTPUT($srcdir/src/libmongoc/src/zlib-1.2.11/zconf.h)
     fi
   fi
-  if test "$PHP_LIBBSON" == "no"; then
+  if test "$PHP_LIBBSON" = "no"; then
```
derickr: Was this always broken like this before? Or do both = and == work (on bash)?

.travis.yml
```diff
@@ -13,24 +48,16 @@ env:
...
     - REPORT_EXIT_STATUS=1
-    - MONGODB_ORCHESTRATION="http://localhost:8889"
-    - MONGODB_ORCHESTRATION_REPO="git+git://github.com/mongodb/mongo-orchestration@master"
-    - MONGODB_ORCHESTRATION_PRESETS_ROOT=`pwd`
-    - secure: "Dofd9NvXIKEat1PMiia3RQgNsJLaKtzZRf9rM4MsqFaBQPbJmVWPANJkayJPcK+l8pLxHyzzH7WqxML+SlBhBTHdJB++1woXxoVwm+sDdwSye6e+eXc3/zRUxb7zyItBZcvVp2tIJpx5kXJb+CoUQakbfWEQQFk4xrOF0R56+8Y="
-    - secure: "pfGVPucvgcgcefhtzAVeejlpOtn4TnAqJSTM8mJkBl36KdU9P7hMuD3czQ4drWgzZ373/VT5UVhLu/HNsdbW0YBTeqPKJ4YNjqVVLytI8xT7y2Lw9l+I7o93j98TMgAoo8nVRmp/E4D6yutbKK1eddrcmf899R1iJbw8v8d1Ht8="
 
 before_install:
-  - sudo pip install --upgrade cpp-coveralls
-  - sudo pip install ${MONGODB_ORCHESTRATION_REPO}
```
derickr: I can't see where mongo orchestration is installed through in the new code.

.travis.scripts/before_script.sh
jmikola: While I very much remember the problem with `run-tests.php` exiting with zero after failures, I don't quite remember what versions of PHP demonstrated the issue. There is evidence of `exit(1)` going back to [PHP-5.3](https://github.com/php/php-src/blob/PHP-5.3/run-tests.php#L952).

I did come across https://github.com/php/php-src/commit/16efc77608789234c92c1cad906417d06ec3428d, which changed the logic for `exit(1)` in 5.5. Do you recall if this was only an issue up to 5.4 (inclusive)?

If you're able to narrow it down, we can certainly get rid of this.

.travis.yml
jmikola: See ef72444d5239f43792a942cd9938b89bb4fbde1c. These were merely artifacts from when Hannes was using Mongo Orchestration locally to spin up test environments ([PHPC-128](https://jira.mongodb.org/browse/PHPC-128)). I found nothing in our current `.travis.yml` files that actually launched Mongo Orchestration, so this all appears to be dead code.

config.m4
jmikola: Bash supports both `==` and `=`, but `=` is POSIX compliant. You'll note from the new shell scripts that I'm aiming to keep things portable with `#!/bin/sh` and POSIX-friendly syntax.

scripts/build/autotools/m4/pkg.m4
jmikola: Correct. There may be whitespace issues (mixed tabs and spaces) that I can clean up, though. We imported from libmongoc, but they also pulled it from elsewhere.

tests/connect/standalone-x509-extract_username-002.phpt
jmikola: See: https://github.com/mongodb/mongo-c-driver/blob/1.9.2/src/mongoc/mongoc-ssl.c#L78

.travis.scripts/before_script.sh
jmikola: I'll narrow this down in a bit and if 5.5+ reports exit codes, we can remove this.

jmikola:
@derickr: I verified locally with macOS 10.12.5 that I'm able to successfully compile the driver for Secure Transport using `--with-mongodb-ssl=darwin`. `./configure` reports that `MONGODB_SHARED_LIBADD` includes `-framework Security -framework CoreFoundation` and `phpinfo()` reports "Secure Transport" and "Common Crypto". Additionally, I successfully connected to an Atlas cluster with SSL and executed a `buildInfo` command.

.travis.scripts/before_script.sh
jmikola: Confirmed that PHP 5.4 was the last version with this problem. I'll remove this.

PR #738, opened 2018-01-16
 PHPC-1096: Allow building with libmongoc's bundled zlib

jmikola:
https://jira.mongodb.org/browse/PHPC-1096


config.m4
```diff
@@ -292,6 +292,9 @@ if test "$MONGODB" != "no"; then
...
 
+    dnl Generated with: find src/libmongoc/src/zlib-1.2.11 -maxdepth 1 -name '*.c' -print0 | cut -sz -d / -f 5- | sort -z | tr '\000' ' '
```
jmikola: `src/libmongoc/src/zlib-1.2.11` does contain nested `contrib` and `example` directories, but those did not appear relevant. Additionally, libmongoc doesn't include sources in those directories when building with bundled zlib either.

config.m4
```diff
@@ -329,9 +332,21 @@ if test "$MONGODB" != "no"; then
...
     AC_CHECK_TYPE([socklen_t], [AC_SUBST(MONGOC_HAVE_SOCKLEN, 1)], [AC_SUBST(MONGOC_HAVE_SOCKLEN, 0)], [#include <sys/socket.h>])
 
-    AC_SUBST(MONGOC_ENABLE_COMPRESSION_SNAPPY, 0)
-    AC_SUBST(MONGOC_ENABLE_COMPRESSION_ZLIB, 0)
-    AC_SUBST(MONGOC_ENABLE_COMPRESSION, 0)
+    with_snappy=auto
+    with_zlib=auto
+    m4_include(src/libmongoc/build/autotools/m4/pkg.m4)
+    m4_include(src/libmongoc/build/autotools/CheckSnappy.m4)
+    m4_include(src/libmongoc/build/autotools/CheckZlib.m4)
```
jmikola: https://github.com/mongodb/mongo-php-driver/pull/690 originally added this down with our SASL checks, neither of which are relevant if we're using system libraries for libmongoc and libbson. Since I was touching this code anyway, I've moved this up within the condition for configuring the driver with bundled libmongoc.

config.m4
```diff
@@ -434,6 +437,10 @@ if test "$MONGODB" != "no"; then
   if test "$PHP_LIBMONGOC" == "no"; then
     PHP_ADD_INCLUDE([$ext_srcdir/src/libmongoc/src/mongoc/])
     PHP_ADD_BUILD_DIR([$ext_builddir/src/libmongoc/src/mongoc/])
+    if test "x$with_zlib" = "xbundled"; then
+      PHP_ADD_INCLUDE([$ext_srcdir/src/libmongoc/src/zlib-1.2.11/])
+      PHP_ADD_BUILD_DIR([$ext_srcdir/src/libmongoc/src/zlib-1.2.11/])
```
jmikola: For the record, I'm not sure what `PHP_ADD_BUILD_DIR` does but I've added this line for consistency. Reading its definition in `acinclude.m4` (after running `phpize`), it looks like it just appends the the directory to `$BUILD_DIRS`, which is later used by PHP_GEN_BUILD_DIRS (called from our `configure.ac`) to ensure all of those directories exist (`mkdir` is called for them). The directory always exists, so this does seem redundant.

config.m4
```diff
@@ -491,6 +498,9 @@ if test "$MONGODB" != "no"; then
     AC_SUBST(MONGOC_VERSION)
     AC_OUTPUT($srcdir/src/libmongoc/src/mongoc/mongoc-config.h)
     AC_OUTPUT($srcdir/src/libmongoc/src/mongoc/mongoc-version.h)
+    if test "x$with_zlib" = "xbundled"; then
+      AC_OUTPUT($srcdir/src/libmongoc/src/zlib-1.2.11/zconf.h)
```
jmikola: `zconf.h.in` actually doesn't have an substitutions, but we need this line to ensure it's copied to `zconf.h`, which `zlib.h` includes.

config.m4
```diff
@@ -329,9 +332,21 @@ if test "$MONGODB" != "no"; then
 
     AC_CHECK_TYPE([socklen_t], [AC_SUBST(MONGOC_HAVE_SOCKLEN, 1)], [AC_SUBST(MONGOC_HAVE_SOCKLEN, 0)], [#include <sys/socket.h>])
 
-    AC_SUBST(MONGOC_ENABLE_COMPRESSION_SNAPPY, 0)
-    AC_SUBST(MONGOC_ENABLE_COMPRESSION_ZLIB, 0)
-    AC_SUBST(MONGOC_ENABLE_COMPRESSION, 0)
+    with_snappy=auto
+    with_zlib=auto
```
jmikola: Note: I manually tested this PR by changing this line to `with_zlib=bundled` since my system has zlib installed and `configure` picks that up first. Examining the `configure` output, I then checked for `whether to enable bundled zlib... 
yes` and verified that `MONGODB_SHARED_LIBADD` did not include `-lz`.

Since we don't report compression engines anywhere, runtime testing entailed tweaking `tests/connect/compression_error-001.phpt` to check that `zlib` was still accepted when building the driver with bundled zlib (it was).

derickr: I think we should report compression engines then?

config.m4
derickr: These mkdirs make sense if you compile the extension outside of the source directory I believe. For example, you can do things like:

``
git clone ...
cd mongo-php-driver
phpize
mkdir build
cd build
../configure 
make
```

jmikola: If you're calling `phpize` from `mongo-php-driver/`, it's going to create `configure` in the project root. And then calling `../configure` from `mongo-php-driver/build` creates all of these directories under `build/` (essentially recreating much of the project source hierarchy) and the resulting object files will get dumped into those newly created directories. Is that correct?

Seems odd, but I suppose it makes sense.

config.m4
jmikola: Opened [PHPC-1097](https://jira.mongodb.org/browse/PHPC-1097).

config.m4
derickr: Make sure to add it to .gitignore though.

config.m4
derickr: That's how I understand it. I've never seen the need for this, nor used this.

config.m4
jmikola: https://github.com/mongodb/mongo-c-driver/blob/1.9.2/.gitignore#L119

PR #743, opened 2018-01-23
Simplify Matrix

photodude:
dist: trusty is now the default container for travis containers 
os: linux is always true unless otherwise specified in the matrix
only need to specify sudo: false once to run containers

Note: if you need sudo: true for a single job in the matrix you can add it to that single matrix line
Note2: if you need a different dist for a single job in the matrix can add it to that single matrix line
Note3: if you need a different os for a single job in the matrix can add it to that single matrix line
no need for all the duplication

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: php
 
 addons:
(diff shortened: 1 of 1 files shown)


jmikola:
The verbose syntax here was leftover from #736. Thanks for cleaning this up.

photodude:
You are welcome.

PR #749, opened 2018-02-02
[WAIT] PHPC-759: Ensure that the driver can be compiled as a built-in PHP extension

derickr:
https://jira.mongodb.org/browse/PHPC-759

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -82,12 +82,18 @@ AC_DEFUN([PHP_BSON_CLOCK],
 ])
 
 if test "$PHP_MONGODB" != "no"; then
-  AC_MSG_CHECKING([Check for supported PHP versions])
-  PHP_MONGODB_FOUND_VERSION=`${PHP_CONFIG} --version`
-  PHP_MONGODB_FOUND_VERNUM=`echo "${PHP_MONGODB_FOUND_VERSION}" | $AWK 'BEGIN { FS = "."; } { printf "%d", ([$]1 * 100 + [$]2) * 100 + [$]3;}'`
-  AC_MSG_RESULT($PHP_MONGODB_FOUND_VERSION)
-  if test "$PHP_MONGODB_FOUND_VERNUM" -lt "50500"; then
-    AC_MSG_ERROR([not supported. Need a PHP version >= 5.5.0 (found $PHP_MONGODB_FOUND_VERSION)])
+  AC_MSG_CHECKING([for PHP_CONFIG binary])
+  AC_CHECK_PROG(PHP_CONFIG_BINARY, [${PHP_CONFIG}], [yes], [no])
+  AC_MSG_CHECKING([supported PHP versions])
+  if test x"$PHP_CONFIG_BINARY" == x"yes"; then
+    PHP_MONGODB_FOUND_VERSION=`${PHP_CONFIG} --version`
+    PHP_MONGODB_FOUND_VERNUM=`echo "${PHP_MONGODB_FOUND_VERSION}" | $AWK 'BEGIN { FS = "."; } { printf "%d", ([$]1 * 100 + [$]2) * 100 + [$]3;}'`
+    AC_MSG_RESULT($PHP_MONGODB_FOUND_VERSION)
+    if test "$PHP_MONGODB_FOUND_VERNUM" -lt "50500"; then
+      AC_MSG_ERROR([not supported. Need a PHP version >= 5.5.0 (found $PHP_MONGODB_FOUND_VERSION)])
+    fi
+  else
+    AC_MSG_RESULT([undeterminate, continuing, but need a PHP version >= 5.5.0])
   fi
 
   PHP_ARG_ENABLE([developer-flags],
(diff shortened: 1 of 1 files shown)


config.m4
```diff
@@ -82,12 +82,18 @@ AC_DEFUN([PHP_BSON_CLOCK],
...
-  AC_MSG_CHECKING([Check for supported PHP versions])
-  PHP_MONGODB_FOUND_VERSION=`${PHP_CONFIG} --version`
-  PHP_MONGODB_FOUND_VERNUM=`echo "${PHP_MONGODB_FOUND_VERSION}" | $AWK 'BEGIN { FS = "."; } { printf "%d", ([$]1 * 100 + [$]2) * 100 + [$]3;}'`
-  AC_MSG_RESULT($PHP_MONGODB_FOUND_VERSION)
-  if test "$PHP_MONGODB_FOUND_VERNUM" -lt "50500"; then
-    AC_MSG_ERROR([not supported. Need a PHP version >= 5.5.0 (found $PHP_MONGODB_FOUND_VERSION)])
+  AC_MSG_CHECKING([for PHP_CONFIG binary])
+  AC_CHECK_PROG(PHP_CONFIG_BINARY, [${PHP_CONFIG}], [yes], [no])
+  AC_MSG_CHECKING([supported PHP versions])
+  if test x"$PHP_CONFIG_BINARY" == x"yes"; then
```
derickr: The PHP binary doesn't always exist... so we need to check for it.

config.m4
```diff
@@ -309,12 +320,19 @@ if test "$PHP_MONGODB" != "no"; then
...
-    m4_include(scripts/build/autotools/m4/pkg.m4)
+    sinclude(ext/mongodb/scripts/build/autotools/m4/pkg.m4)
+    sinclude(ext/mongodb/scripts/build/autotools/CheckHost.m4)
+    sinclude(ext/mongodb/scripts/build/autotools/CheckSSL.m4)
 
-    m4_include(scripts/build/autotools/CheckHost.m4)
-    m4_include(scripts/build/autotools/CheckSSL.m4)
+    sinclude(scripts/build/autotools/m4/pkg.m4)
+    sinclude(scripts/build/autotools/CheckHost.m4)
+    sinclude(scripts/build/autotools/CheckSSL.m4)
```
derickr: ``sinclude`` works like ``m4_include``, but doesn't bail out if it can't find the file. As this happens during ``./buildconf``, we can't make use of shell magic so this seems the only (ugly) way: we just try them both.

config.m4
```diff
@@ -440,11 +471,21 @@ if test "$PHP_MONGODB" != "no"; then
...
+    sinclude(ext/mongodb/src/libbson/build/autotools/CheckAtomics.m4)
+    sinclude(ext/mongodb/src/libbson/build/autotools/FindDependencies.m4)
+    sinclude(ext/mongodb/src/libbson/build/autotools/m4/ac_compile_check_sizeof.m4)
+    sinclude(ext/mongodb/src/libbson/build/autotools/m4/ac_create_stdint_h.m4)
+    if test -d "${srcdir}/src/libbson/"; then
+      prefixdir=${srcdir}
+    else
+      prefixdir=${srcdir}/ext/mongodb
+    fi
+    AC_CREATE_STDINT_H([$prefixdir/src/libbson/src/bson/bson-stdint.h])
```
derickr: We need to create files in the right directory.

config.m4
```diff
@@ -484,36 +525,50 @@ if test "$PHP_MONGODB" != "no"; then
...
     AC_SUBST(BSON_API_VERSION)
     AC_SUBST(BSON_VERSION)
-    AC_OUTPUT($srcdir/src/libbson/src/bson/bson-config.h)
-    AC_OUTPUT($srcdir/src/libbson/src/bson/bson-version.h)
+    AC_OUTPUT($prefixdir/src/libbson/src/bson/bson-config.h)
+    AC_OUTPUT($prefixdir/src/libbson/src/bson/bson-version.h)
+    srcdir=${backup_srcdir}
   fi
 
+  EXTRA_LIBS="$EXTRA_LIBS $MONGODB_SHARED_LIBADD"
```
derickr: Without this, extra libs like snappy or zlib aren't passed on to PHP.

jmikola: Is this ignored for shared builds?

I see bits of `EXTRA_LIBS` in PHP core but it's oddly absent from any extension `config.m4` files. Given that many of those link extra libraries and can also be built-in, is there a reason for that?

config.m4
jmikola: When was `PHP_CONFIG` introduced? Just curious where you found it missing.

Note that the leading "x" for equality checks is only necessary if an operand is potentially an empty string. In this case, `AC_CHECK_PROG()` will either assign "yes" or "no", so you're save comparing the variable to "yes" directly.

config.m4
```diff
@@ -82,12 +82,18 @@ AC_DEFUN([PHP_BSON_CLOCK],
...
+  AC_MSG_CHECKING([supported PHP versions])
+  if test x"$PHP_CONFIG_BINARY" == x"yes"; then
+    PHP_MONGODB_FOUND_VERSION=`${PHP_CONFIG} --version`
+    PHP_MONGODB_FOUND_VERNUM=`echo "${PHP_MONGODB_FOUND_VERSION}" | $AWK 'BEGIN { FS = "."; } { printf "%d", ([$]1 * 100 + [$]2) * 100 + [$]3;}'`
+    AC_MSG_RESULT($PHP_MONGODB_FOUND_VERSION)
+    if test "$PHP_MONGODB_FOUND_VERNUM" -lt "50500"; then
+      AC_MSG_ERROR([not supported. Need a PHP version >= 5.5.0 (found $PHP_MONGODB_FOUND_VERSION)])
+    fi
+  else
+    AC_MSG_RESULT([undeterminate, continuing, but need a PHP version >= 5.5.0])
```
jmikola: "indeterminate" would be more common to use here.

config.m4
jmikola: How is this different than deciding to call `PHP_ADD_SOURCES()` or `PHP_ADD_SOURCES_X()` based on the value of `$ext_shared`? Does that solution not apply here because `m4_include()` is evaluated earlier to generate the `configure` script, long before we'd be able to check the value of `$ext_shared`?

I tried using `m4_include(PHP_EXT_DIR(mongodb)[scripts/build/autotools/m4/pkg.m4])` but that results in:

```
/usr/bin/m4:config.m4:314: cannot open `""scripts/build/autotools/m4/pkg.m4': No such file or directory
```

Is `PHP_EXT_DIR(mongodb)` only an empty string when building a shared extension? I found it [declared in `phpize.m4`](https://github.com/php/php-src/blob/39c58571350692b4cee3203280546f795335f19e/scripts/phpize.m4#L11) as an empty string. There is a more substantial definition in [`configure.ac`](https://github.com/php/php-src/blob/master/configure.ac#L123), but I assume that's for built-in extensions.

config.m4
```diff
@@ -440,11 +471,21 @@ if test "$PHP_MONGODB" != "no"; then
     fi
   fi
   if test "$PHP_LIBBSON" = "no"; then
-    m4_include(src/libbson/build/autotools/CheckAtomics.m4)
-    m4_include(src/libbson/build/autotools/FindDependencies.m4)
-    m4_include(src/libbson/build/autotools/m4/ac_compile_check_sizeof.m4)
-    m4_include(src/libbson/build/autotools/m4/ac_create_stdint_h.m4)
-    AC_CREATE_STDINT_H([$srcdir/src/libbson/src/bson/bson-stdint.h])
+    sinclude(src/libbson/build/autotools/CheckAtomics.m4)
```
jmikola: Side question: once we `sinclude()` a file, is an include within that file relative to the file's own path? I wonder if we could clean up a lot of this by having a single gateway include that then includes all of these general m4 files.

config.m4
jmikola: Since this is after `PHP_NEW_EXTENSION()` is called, can we use `$ext_srcdir` here and avoid the conditional?

config.m4
```diff
@@ -484,36 +525,50 @@ if test "$PHP_MONGODB" != "no"; then
...
   if test "$PHP_LIBMONGOC" = "no"; then
     backup_srcdir=${srcdir}
-    srcdir=${srcdir}/src/libmongoc/
-    m4_include(src/libmongoc/build/autotools/Versions.m4)
-    srcdir=${backup_srcdir}
+    if test -d "${srcdir}/src/libmongoc/"; then
+      prefixdir=${srcdir}
+    else
+      prefixdir=${srcdir}/ext/mongodb
+    fi
```
jmikola: Same question here regarding `$ext_srcdir`.

config.m4
jmikola: We discussed why `EXTRA_LIBS` is needed. Long-term, let's see if we can use `PHP_ADD_LIBPATH()`.

jmikola:
Superseded by https://github.com/mongodb/mongo-php-driver/pull/830.

PR #750, opened 2018-02-05
PHPC-1109: Support batchSize getMore option for command cursors

kvwalker:
https://jira.mongodb.org/browse/PHPC-1109

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -903,7 +903,7 @@ int phongo_execute_command(mongoc_client_t *client, php_phongo_command_type_t ty
 
 		if (command->max_await_time_ms) {
 			bson_append_bool(&initial_reply, "awaitData", -1, 1);
-			bson_append_int32(&initial_reply, "maxAwaitTimeMS", -1, command->max_await_time_ms);
+			bson_append_int64(&initial_reply, "maxAwaitTimeMS", -1, command->max_await_time_ms);
 			bson_append_bool(&initial_reply, "tailable", -1, 1);
 		}
 
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -907,6 +907,10 @@ int phongo_execute_command(mongoc_client_t *client, php_phongo_command_type_t ty
 			bson_append_bool(&initial_reply, "tailable", -1, 1);
 		}
 
+		if (command->batch_size) {
+			bson_append_int64(&initial_reply, "batchSize", -1, command->batch_size);
+		}
+
 		cmd_cursor = mongoc_cursor_new_from_command_reply(client, &initial_reply, server_id);
 		bson_destroy(&reply);
 	} else {
--- a/src/MongoDB/Command.c
+++ b/src/MongoDB/Command.c
@@ -59,6 +59,9 @@ static bool php_phongo_command_init_max_await_time_ms(php_phongo_command_t *inte
  * (where applicable). */
 static bool php_phongo_command_init(php_phongo_command_t *intern, zval *filter, zval *options TSRMLS_DC) /* {{{ */
 {
+	bson_iter_t iter;
+	bson_iter_t sub_iter;
+
 	intern->bson = bson_new();
 
 	php_phongo_zval_to_bson(filter, PHONGO_BSON_NONE, intern->bson, NULL TSRMLS_CC);
@@ -67,6 +70,14 @@ static bool php_phongo_command_init(php_phongo_command_t *intern, zval *filter, 
 	 * invoke php_phongo_query_free_object to destruct the object. */
 	if (EG(exception)) {
 		return false;
+	}
+
+	if (bson_iter_init(&iter, intern->bson) && bson_iter_find_descendant(&iter, "cursor.batchSize", &sub_iter) && BSON_ITER_HOLDS_INT(&sub_iter)) {
+		int64_t batch_size = bson_iter_as_int64(&sub_iter);
+		
+		if (batch_size >= 0 && batch_size <= UINT32_MAX) {
+			intern->batch_size = (uint32_t) batch_size;
+		}
 	}
 
 	if (!options) {
--- a/tests/command/cursor-batchsize-001.phpt
+++ b/tests/command/cursor-batchsize-001.phpt
@@ -1,0 +1,83 @@
+--TEST--
+MongoDB\Driver\Command non-zero batchSize applies to getMore
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS('STANDALONE'); CLEANUP(STANDALONE); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+class Test implements MongoDB\Driver\Monitoring\CommandSubscriber
+{
+    public function executeCommand()
+    {
+        MongoDB\Driver\Monitoring\addSubscriber($this);
+
+        $manager = new MongoDB\Driver\Manager(STANDALONE);
+
+        $bulkWrite = new MongoDB\Driver\BulkWrite;
+
+        for ($i = 0; $i < 5; $i++) {
+            $bulkWrite->insert(['_id' => $i]);
+        }
+
+        $writeResult = $manager->executeBulkWrite(NS, $bulkWrite);
+        printf("Inserted: %d\n", $writeResult->getInsertedCount());
+
+        $command = new MongoDB\Driver\Command([
+            'aggregate' => COLLECTION_NAME,
+            'pipeline' => [['$match' => new stdClass]],
+            'cursor' => ['batchSize' => 2]
+        ]);
+        $cursor = $manager->executeCommand(DATABASE_NAME, $command);
+
+        $cursor->toArray();
+
+        MongoDB\Driver\Monitoring\removeSubscriber($this);
+    }
+
+    public function commandStarted(MongoDB\Driver\Monitoring\CommandStartedEvent $event)
+    {
+        $command = $event->getCommand();
+
+        if ($event->getCommandName() === 'aggregate') {
+            printf("aggregate command specifies batchSize: %d\n", $command->cursor->batchSize);
+        }
+
+        if ($event->getCommandName() === 'getMore') {
+            printf("getMore command specifies batchSize: %d\n", $command->batchSize);
+        }
+    }
+
+    public function commandSucceeded(MongoDB\Driver\Monitoring\CommandSucceededEvent $event)
+    {
+        $reply = $event->getReply();
+
+        if ($event->getCommandName() === 'aggregate') {
+            printf("aggregate response contains %d document(s)\n", count($reply->cursor->firstBatch));
+        }
+
+        if ($event->getCommandName() === 'getMore') {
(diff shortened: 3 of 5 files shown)


php_phongo.c
```diff
@@ -907,6 +907,10 @@ int phongo_execute_command(mongoc_client_t *client, php_phongo_command_type_t ty
 			bson_append_bool(&initial_reply, "tailable", -1, 1);
 		}
 
+		if (command->batch_size) {
+			bson_append_int32(&initial_reply, "batchSize", -1, command->batch_size);
```
jmikola: On second thought, let's use `bson_append_int64()` here. The corresponding documentation in libmongoc ([`find_with_opts()`](http://mongoc.org/libmongoc/current/mongoc_collection_find_with_opts.html)) says "batchSize" and "maxAwaitTimeMS" are both "non-negative int64". Logic in `Command.c` already ensures they fall between `0` and `UINT32_MAX` (inclusive), so we don't have to worry about that. Assigning as an int64 will also ensure that `UINT32_MAX` will not be interpreted as a negative int64 down the line (an edge case, I know).

I'd also suggest changing the append function for "maxAwaitTimeMS" a few lines up in a second commit.

src/MongoDB/Command.c
```diff
@@ -69,6 +72,19 @@ static bool php_phongo_command_init(php_phongo_command_t *intern, zval *filter,
...
 
+	if (bson_iter_init (&iter, intern->bson) && bson_iter_find_descendant (&iter, "cursor.batchSize", &sub_iter) && BSON_ITER_HOLDS_INT(&sub_iter)) {
+		int64_t batch_size = bson_iter_as_int64(&sub_iter);
+		if (batch_size < 0) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected \"batchSize\" option to be >= 0, %" PRId64 " given", batch_size);
+		}
+
+		if (batch_size > UINT32_MAX) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected \"batchSize\" option to be <= %" PRIu32 ", %" PRId64 " given", UINT32_MAX, batch_size);
+		}
```
jmikola: IIRC, we discussed ignoring out-of-range values instead of throwing an exception, since this is internal behavior hidden from the user. Also, if they happen to use an invalid value, chances are the main command itself will yield an error. I'd suggest:

```c
if (batch_size >= 0 && batch_size <= UINT32_MAX) {
	intern->batch_size = (uint32_t) batch_size;
}
```

Note that if we were going to throw an exception here, it'd be important to return `false` after doing so. This all comes after the `if (EG(exception)) return false;` that trails BSON conversion, and it would therefore be possible for this `php_phongo_command_init()` function to return `true` so long as `php_phongo_command_init_max_await_time_ms()` didn't fail below.

Returning false on the first error will ensure that the calling function can act accordingly (technically not a problem in this case, since this function is the last statement of the Command constructor) and avoid cases where we end up throwing multiple exceptions (in that case, previous exceptions will get masked under the most recent and can only be accessed with [`Exception::getPrevious()`](http://php.net/manual/en/exception.getprevious.php)).

src/MongoDB/Command.c
```diff
@@ -69,6 +72,19 @@ static bool php_phongo_command_init(php_phongo_command_t *intern, zval *filter,
 		return false;
 	}
 
+	if (bson_iter_init (&iter, intern->bson) && bson_iter_find_descendant (&iter, "cursor.batchSize", &sub_iter) && BSON_ITER_HOLDS_INT(&sub_iter)) {
```
jmikola: Spaces between the function name and open parenthesis is part of libmongoc's coding standard, but we don't follow it within the PHP extension. `bson_iter_init(...)` works fine.

jmikola:
I didn't see a test, which might have been an oversight if you just forgot to add a new file. I whipped up the following test file and verified that this works as it should:

```php
--TEST--
MongoDB\Driver\Command batchSize applies to getMore
--SKIPIF--
<?php NEEDS('STANDALONE'); CLEANUP(STANDALONE); ?>
--FILE--
<?php

class Test implements MongoDB\Driver\Monitoring\CommandSubscriber
{
    public function executeCommand()
    {
        $manager = new MongoDB\Driver\Manager(STANDALONE);

        $bulk = new MongoDB\Driver\BulkWrite;
        $bulk->insert(['_id' => 1]);
        $bulk->insert(['_id' => 2]);
        $bulk->insert(['_id' => 3]);
        $bulk->insert(['_id' => 4]);
        $bulk->insert(['_id' => 5]);

        $manager->executeBulkWrite(NS, $bulk);

        MongoDB\Driver\Monitoring\addSubscriber($this);

        $command = new MongoDB\Driver\Command([
            'aggregate' => COLLECTION_NAME,
            'pipeline' => [],
            'cursor' => ['batchSize' => 2],
        ]);

        $cursor = $manager->executeCommand(DATABASE_NAME, $command);

        $cursor->toArray();

        MongoDB\Driver\Monitoring\removeSubscriber($this);
    }

    public function commandStarted(MongoDB\Driver\Monitoring\CommandStartedEvent $event)
    {
        $command = $event->getCommand();

        if ($event->getCommandName() === 'aggregate') {
            printf("aggregate command specifies batchSize: %d\n", $command->cursor->batchSize);
        }

        if ($event->getCommandName() === 'getMore') {
            printf("getMore command specifies batchSize: %d\n", $command->batchSize);
        }
    }

    public function commandSucceeded(MongoDB\Driver\Monitoring\CommandSucceededEvent $event)
    {
        $reply = $event->getReply();

        if ($event->getCommandName() === 'aggregate') {
            printf("aggregate response contains %d document(s)\n", count($reply->cursor->firstBatch));
        }

        if ($event->getCommandName() === 'getMore') {
            printf("getMore response contains %d document(s)\n", count($reply->cursor->nextBatch));
        }
    }

    public function commandFailed(MongoDB\Driver\Monitoring\CommandFailedEvent $event)
    {
    }
}

(new Test)->executeCommand();

?>
===DONE===
<?php exit(0); ?>
--EXPECT--
aggregate command specifies batchSize: 2
aggregate response contains 2 document(s)
getMore command specifies batchSize: 2
getMore response contains 2 document(s)
getMore command specifies batchSize: 2
getMore response contains 1 document(s)
===DONE===
```

I might suggest adding one more test to demonstrate that a `batchSize` of `0` is ignored and is omitted from the `getMore` command. In that case, in the context of this test, the first `getMore` response should contain three documents and there shouldn't be a second `getMore` at all.

php_phongo.c
derickr: On that note, should we make the new ``batch_size`` member of the struct an ``uint64_t`` instead of an ``uint32_t`` below too then?

jmikola: `uint32_t` should cast up to `uint64_t` without issue and we only store these values if they call within the unsigned 32-bit range. I think this is OK to remain as-is.

tests/command/cursor-batchsize-002.phpt
```diff
@@ -0,0 +1,84 @@
...
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+Inserted: 5
+aggregate command specifies batchSize: 0
+aggregate response contains 0 document(s)
+getMore command specifies batchSize: no
+getMore response contains 5 document(s)
```
kvwalker: According to the [manual](https://docs.mongodb.com/manual/reference/method/db.collection.aggregate/index.html#example-aggregate-method-initial-batch-size), a batchSize of 0 means an empty first batch and is useful for quickly returning a cursor or failure message without doing significant server-side work so the aggregate should return 0 documents and the getMore should return the remaining documents (5 in this case).

jmikola: Ah, that's right. I was only thinking about the two `getMore` requests in my last comment.

Side note: this is related to [PHPLIB-312](https://jira.mongodb.org/browse/PHPLIB-312).

tests/command/cursor-batchsize-001.phpt
```diff
@@ -0,0 +1,86 @@
...
+
+class Test implements MongoDB\Driver\Monitoring\CommandSubscriber
+{
+    public function executeCommand()
+    {
+        MongoDB\Driver\Monitoring\addSubscriber($this);
+
+        $manager = new MongoDB\Driver\Manager(STANDALONE);
+
+		$bulkWrite = new MongoDB\Driver\BulkWrite;
```
jmikola: Indentation looks off here. PHP source uses four-space indents (not tabs as in C sources). I realize that's confusing :)

tests/command/cursor-batchsize-001.phpt
```diff
@@ -0,0 +1,86 @@
...
+
+		for ($i = 0; $i < 5; $i++) {
+		    $bulkWrite->insert(array('_id' => $i));
+		}
+
+		$writeResult = $manager->executeBulkWrite(NS, $bulkWrite);
+		printf("Inserted: %d\n", $writeResult->getInsertedCount());
+
+		$command = new MongoDB\Driver\Command(array(
```
jmikola: We can use short array syntax (e.g. `[ ]`) whenever possible.

tests/command/cursor-batchsize-002.phpt
```diff
@@ -0,0 +1,84 @@
+--TEST--
+MongoDB\Driver\Command batchSize of zero ignored
```
jmikola: "batchSize of zero is ignored for getMore"

tests/command/cursor-batchsize-001.phpt
```diff
@@ -0,0 +1,86 @@
+--TEST--
+MongoDB\Driver\Command batchSize applies to getMore
```
jmikola: "non-zero batchSize applies to getMore" (balances out the other test name)

tests/command/cursor-batchsize-002.phpt
```diff
@@ -0,0 +1,84 @@
...
+
+class Test implements MongoDB\Driver\Monitoring\CommandSubscriber
+{
+    public function executeCommand()
+    {
+        MongoDB\Driver\Monitoring\addSubscriber($this);
+
+        $manager = new MongoDB\Driver\Manager(STANDALONE);
+
+		$bulkWrite = new MongoDB\Driver\BulkWrite;
```
jmikola: Indentation here as well.

tests/command/cursor-batchsize-002.phpt
```diff
@@ -0,0 +1,84 @@
+--TEST--
+MongoDB\Driver\Command batchSize of zero ignored
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS_CRYPTO(); ?>
```
jmikola: Looks like copypasta from the causal consistency tests and can be removed.

tests/command/cursor-batchsize-001.phpt
```diff
@@ -0,0 +1,86 @@
...
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS_CRYPTO(); ?>
```
jmikola: Can remove.

PR #751, opened 2018-02-06
Contributing notes for alpha and beta releases

jmikola:


CONTRIBUTING.md
```diff
@@ -235,6 +243,13 @@ Commit this change:
 $ git commit -m "Back to -dev" php_phongo.h
 ```
 
+> **Note:** If this is an alpha or beta release, the version string revert back
+> to its pre-release value without incrementing the patch version. For example,
+> the constants may have indicated version "1.4.0-dev" and stability "devel"
+> before being changed to "1.4.0beta1" and "beta" for a beta release,
+> respectively. After the release, the constants should revert back to
+> "1.4.0-dev" and "devel".
```
derickr: That's not recommended. It should then move to 1.4.0beta2-dev, in order to sort them properly.

jmikola: See: https://github.com/mongodb/mongo-php-driver/pull/752

PR #755, opened 2018-02-13
PHPC-1118: Create clang-format file and Makefile targets

derickr:


src/BSON/Binary.c
```diff
@@ -210,14 +207,14 @@ static PHP_METHOD(Binary, jsonSerialize)
...
 {
-	php_phongo_binary_t      *intern;
+	php_phongo_binary_t* intern;
 #if PHP_VERSION_ID >= 70000
-	zval                      retval;
+	zval retval;
 #else
-	zval                     *retval;
+	zval* retval;
 #endif
```
derickr: Should we make this into a ``#define ZVAL_RETVAL_TYPE`` so we don't have the if/else/endif for these everywhere?

src/MongoDB/BulkWrite.c
```diff
@@ -102,27 +102,27 @@ static bool php_phongo_bulkwrite_opts_append_document(bson_t *opts, const char *
 	return true;
 } /* }}} */
 
-#define PHONGO_BULKWRITE_APPEND_BOOL(opt, value) \
-	if (!BSON_APPEND_BOOL(boptions, (opt), (value))) { \
+#define PHONGO_BULKWRITE_APPEND_BOOL(opt, value)                                                                 \
```
derickr: We can turn this setting one ``Off`` as well, in which case it doesn't touch the ``\``. (Although doing that now, would be a bit annoying).

src/MongoDB/Command.c
```diff
@@ -74,9 +74,9 @@ static bool php_phongo_command_init(php_phongo_command_t *intern, zval *filter,
 
 	if (bson_iter_init(&iter, intern->bson) && bson_iter_find_descendant(&iter, "cursor.batchSize", &sub_iter) && BSON_ITER_HOLDS_INT(&sub_iter)) {
 		int64_t batch_size = bson_iter_as_int64(&sub_iter);
-		
+
 		if (batch_size >= 0 && batch_size <= UINT32_MAX) {
-			intern->batch_size = (uint32_t) batch_size;
+			intern->batch_size = (uint32_t)batch_size;
```
derickr: We can chose to have the space after the cast to be inserted here as well, if we want. I would suggest to do this in a separate PR then, to see whether we want to.

src/MongoDB/CursorId.c
```diff
@@ -88,33 +91,32 @@ static phongo_create_object_retval php_phongo_cursorid_create_object(zend_class_
...
+static HashTable* php_phongo_cursorid_get_debug_info(zval* object, int* is_temp TSRMLS_DC) /* {{{ */
 {
-	php_phongo_cursorid_t  *intern;
+	php_phongo_cursorid_t* intern;
 #if PHP_VERSION_ID >= 70000
-	zval                    retval;
+	zval retval;
 #else
-	zval                    retval = zval_used_for_init;
+	zval retval = zval_used_for_init;
```
derickr: Should we macroify this too? Just like I suggested for ``zval*`` vs ``zval`` for the return value?

src/MongoDB/Exception/WriteException.c
```diff
@@ -41,7 +41,7 @@ static PHP_METHOD(WriteException, getWriteResult)
 #if PHP_VERSION_ID >= 70000
 	writeresult = zend_read_property(php_phongo_writeexception_ce, getThis(), ZEND_STRL("writeResult"), 0, &rv TSRMLS_CC);
 #else
-	writeresult = zend_read_property(php_phongo_writeexception_ce, getThis(), ZEND_STRL("writeResult"), 0 TSRMLS_CC);
+	writeresult                  = zend_read_property(php_phongo_writeexception_ce, getThis(), ZEND_STRL("writeResult"), 0 TSRMLS_CC);
```
derickr: I don't understand why it made this change.

src/BSON/Regex.c
```diff
@@ -72,20 +72,20 @@ static bool php_phongo_regex_init(php_phongo_regex_t *intern, const char *patter
...
-static bool php_phongo_regex_init_from_hash(php_phongo_regex_t *intern, HashTable *props TSRMLS_DC) /* {{{ */
+static bool php_phongo_regex_init_from_hash(php_phongo_regex_t* intern, HashTable* props TSRMLS_DC) /* {{{ */
 {
 #if PHP_VERSION_ID >= 70000
 	zval *pattern, *flags;
 
-	if ((pattern = zend_hash_str_find(props, "pattern", sizeof("pattern")-1)) && Z_TYPE_P(pattern) == IS_STRING &&
-	    (flags = zend_hash_str_find(props, "flags", sizeof("flags")-1)) && Z_TYPE_P(flags) == IS_STRING) {
+	if ((pattern = zend_hash_str_find(props, "pattern", sizeof("pattern") - 1)) && Z_TYPE_P(pattern) == IS_STRING &&
+		(flags = zend_hash_str_find(props, "flags", sizeof("flags") - 1)) && Z_TYPE_P(flags) == IS_STRING) {
```
derickr: I'm fixing this with a newline before the next line.

src/BSON/UTCDateTime.c
```diff
@@ -105,37 +105,37 @@ static bool php_phongo_utcdatetime_init_from_hash(php_phongo_utcdatetime_t *inte
...
 	int64_t sec, usec;
 
 	/* The following assignments use the same logic as date_format() in php_date.c */
 	sec = datetime_obj->time->sse;
 #if PHP_VERSION_ID >= 70200
-	usec = (int64_t) floor(datetime_obj->time->us);
+	usec = (int64_t)floor(datetime_obj->time->us);
 #else
-	usec = (int64_t) floor(datetime_obj->time->f * 1000000 + 0.5);
+	usec                  = (int64_t)floor(datetime_obj->time->f * 1000000 + 0.5);
```
derickr: Don't get this change either.

src/MongoDB/Query.c
```diff
@@ -268,33 +268,33 @@ static bool php_phongo_query_init(php_phongo_query_t *intern, zval *filter, zval
 	}
 
 	PHONGO_QUERY_OPT_BOOL("allowPartialResults", options, "allowPartialResults")
-		else PHONGO_QUERY_OPT_BOOL("allowPartialResults", options, "partial");
+	else PHONGO_QUERY_OPT_BOOL("allowPartialResults", options, "partial");
```
derickr: This is a silly one. I'm putting the ``else...`` on the same line to avoid this nonsense.

src/MongoDB/Server.c
```diff
@@ -528,35 +520,37 @@ ZEND_BEGIN_ARG_INFO_EX(ai_Server_void, 0, 0, 0)
...
 /* }}} */
 
 /* {{{ MongoDB\Driver\Server object handlers */
 static zend_object_handlers php_phongo_handler_server;
 
-static int php_phongo_server_compare_objects(zval *o1, zval *o2 TSRMLS_DC) /* {{{ */
+static int php_phongo_server_compare_objects(zval* o1, zval* o2 TSRMLS_DC) /* {{{ */
 {
-	php_phongo_server_t         *intern1, *intern2;
+	php_phongo_server_t *        intern1, *intern2;
```
derickr: Need to split this up into multiple lines.

src/bson-encode.c
```diff
@@ -447,13 +448,13 @@ void php_phongo_zval_to_bson(zval *data, php_phongo_bson_flags_t flags, bson_t *
...
-						BSON_SERIALIZE_FUNC_NAME,
+										   "Expected %s::%s() to return an array or stdClass, %s given",
+										   ZSTR_VAL(Z_OBJCE_P(data)->name),
+										   BSON_SERIALIZE_FUNC_NAME,
 #if PHP_VERSION_ID >= 70000
-						PHONGO_ZVAL_CLASS_OR_TYPE_NAME(obj_data)
+										   PHONGO_ZVAL_CLASS_OR_TYPE_NAME(obj_data)
 #else
-						PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(obj_data)
+										   PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(obj_data)
```
derickr: Fixing this as well by putting the first argument on the next line already.

src/MongoDB/Query.c
derickr: Except that that doesn't work...

php_phongo.c
```diff
@@ -2228,60 +2259,60 @@ static mongoc_client_t *php_phongo_make_mongo_client(const mongoc_uri_t *uri TSR
...
+		"Creating Manager, phongo-%s[%s] - mongoc-%s(%s), libbson-%s(%s), php-%s",
 		PHP_MONGODB_VERSION,
 		PHP_MONGODB_STABILITY,
 		MONGOC_VERSION_S,
 		mongoc_version,
 		BSON_VERSION_S,
 		bson_version,
-		PHP_VERSION
-	);
+		PHP_VERSION);
```
jmikola: I noticed that libmongoc would format this by indenting all subsequent lines at the next column after `MONGOC_DEBUG(`. I assume that won't work well because we use tabs instead of spaces for indents (it'd require mixing both for a leading indent).

Would you prefer `);` on its own line as in the original, or is this something we have no control over (like multi-line `if` conditions)?

php_phongo.c
```diff
@@ -2228,60 +2259,60 @@ static mongoc_client_t *php_phongo_make_mongo_client(const mongoc_uri_t *uri TSR
...
 } /* }}} */
 
-static void php_phongo_persist_client(const char *hash, size_t hash_len, mongoc_client_t *client TSRMLS_DC)
+static void php_phongo_persist_client(const char* hash, size_t hash_len, mongoc_client_t* client TSRMLS_DC)
 {
-	php_phongo_pclient_t *pclient = (php_phongo_pclient_t *) pecalloc(1, sizeof(php_phongo_pclient_t), 1);
+	php_phongo_pclient_t* pclient = (php_phongo_pclient_t*)pecalloc(1, sizeof(php_phongo_pclient_t), 1);
 
-	pclient->pid = (int) getpid();
+	pclient->pid    = (int)getpid();
```
jmikola: Do you see a benefit in leaving whitespace after a cast? I think it does help the function or identifier stand out more.

php_phongo.c
```diff
@@ -2518,12 +2551,13 @@ ZEND_INI_MH(OnUpdateDebug)
...
-		|| strcasecmp("no", ZSTR_VAL(new_value)) == 0
-		|| strcasecmp("false", ZSTR_VAL(new_value)) == 0
-	   ) {
+	if (!new_value ||
+		(new_value && !ZSTR_VAL(new_value)[0]) ||
+		strcasecmp("0", ZSTR_VAL(new_value)) == 0 ||
+		strcasecmp("off", ZSTR_VAL(new_value)) == 0 ||
+		strcasecmp("no", ZSTR_VAL(new_value)) == 0 ||
+		strcasecmp("false", ZSTR_VAL(new_value)) == 0) {
+
```
jmikola: I noticed several other multi-line conditionals earlier in the PR that didn't include a blank line after `) {`. Were you going to go back and add them since you added a `clang-format` rule to preserve them?

Is there no way to have `clang-format` inject the newline on its own?

php_phongo.c
```diff
@@ -1125,24 +1121,26 @@ void php_phongo_read_preference_prep_tagsets(zval *tagSets TSRMLS_DC) /* {{{ */
...
-		HashPosition   pos;
-		zval         **tagSet;
+		HashPosition pos;
+		zval** tagSet;
 
 		for (zend_hash_internal_pointer_reset_ex(ht_data, &pos);
-		     zend_hash_get_current_data_ex(ht_data, (void **) &tagSet, &pos) == SUCCESS;
-		     zend_hash_move_forward_ex(ht_data, &pos)) {
+			 zend_hash_get_current_data_ex(ht_data, (void**)&tagSet, &pos) == SUCCESS;
+			 zend_hash_move_forward_ex(ht_data, &pos)) {
```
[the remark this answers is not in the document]
jmikola: Is a newline preserved after the `for`'s `) {` (like multi-line `if` statements)? If so, I think it'd benefit here.

php_phongo.c
```diff
@@ -1302,19 +1333,19 @@ static bool php_phongo_apply_options_to_uri(mongoc_uri_t *uri, bson_t *options T
...
-		    !strcasecmp(key, MONGOC_URI_W) ||
-		    !strcasecmp(key, MONGOC_URI_WTIMEOUTMS)) {
+			!strcasecmp(key, MONGOC_URI_MAXSTALENESSSECONDS) ||
+			!strcasecmp(key, MONGOC_URI_READCONCERNLEVEL) ||
+			!strcasecmp(key, MONGOC_URI_READPREFERENCE) ||
+			!strcasecmp(key, MONGOC_URI_READPREFERENCETAGS) ||
+			!strcasecmp(key, MONGOC_URI_SAFE) ||
+			!strcasecmp(key, MONGOC_URI_SLAVEOK) ||
+			!strcasecmp(key, MONGOC_URI_W) ||
+			!strcasecmp(key, MONGOC_URI_WTIMEOUTMS)) {
```
jmikola: We can add a newline after this multi-line `if`.

php_phongo.c
```diff
@@ -1714,9 +1743,9 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t *uri, bson_t *option
 	}
 
 	if (!bson_iter_init_find_case(&iter, options, MONGOC_URI_JOURNAL) &&
-	    !bson_iter_init_find_case(&iter, options, MONGOC_URI_SAFE) &&
-	    !bson_iter_init_find_case(&iter, options, MONGOC_URI_W) &&
-	    !bson_iter_init_find_case(&iter, options, MONGOC_URI_WTIMEOUTMS)) {
+		!bson_iter_init_find_case(&iter, options, MONGOC_URI_SAFE) &&
+		!bson_iter_init_find_case(&iter, options, MONGOC_URI_W) &&
+		!bson_iter_init_find_case(&iter, options, MONGOC_URI_WTIMEOUTMS)) {
```
jmikola: Newline below this line.

php_phongo.h
```diff
@@ -35,32 +35,32 @@ extern zend_module_entry mongodb_module_entry;
...
-#else
-# ifdef ZTS
-#  define MONGODB_G(v) TSRMG(mongodb_globals_id, zend_mongodb_globals *, v)
-#  define mglo mongodb_globals_id
-# else
-#  define MONGODB_G(v) (mongodb_globals.v)
-#  define mglo mongodb_globals
-# endif
 #endif
+#else /* PHP_VERSION_ID >= 70000 */
```
jmikola: Would it make more sense for this to display the logic for entering the `else` block (e.g `PHP_VERSION_ID < 70000`), or is that uncommon?

Based on what I see below with `ZTS`, I assume it's easier to just repeat the original `#if` condition for matching purposes.

php_phongo_classes.h
```diff
@@ -20,289 +20,317 @@
...
 
-static inline php_phongo_bulkwrite_t* php_bulkwrite_fetch_object(zend_object *obj) {
-    return (php_phongo_bulkwrite_t *)((char *)obj - XtOffsetOf(php_phongo_bulkwrite_t, std));
+static inline php_phongo_bulkwrite_t* php_bulkwrite_fetch_object(zend_object* obj)
+{
+	return (php_phongo_bulkwrite_t*)((char*)obj - XtOffsetOf(php_phongo_bulkwrite_t, std));
 }
-static inline php_phongo_command_t* php_command_fetch_object(zend_object *obj) {
-    return (php_phongo_command_t *)((char *)obj - XtOffsetOf(php_phongo_command_t, std));
+static inline php_phongo_command_t* php_command_fetch_object(zend_object* obj)
```
jmikola: Unrelated to formatting changes, I wonder if these functions should have line breaks between them. Are they currently considered members of the same "paragraph"?

scripts/clang-format.sh
```diff
@@ -0,0 +1,13 @@
...
+if test x"$1" = x; then 
+	FILES=`git ls-files | grep '\.[ch]$'`
+fi
+if test x"$1" = xchanged; then 
+	FILES=`git diff --name-only | grep '\.[ch]$'`
+fi
+
+for i in $FILES; do
+	clang-format-7 $i > /tmp/tmp-clang.c
+	mv /tmp/tmp-clang.c $i
```
jmikola: `clang-format` supports a `-i` option for in place updates. This shouldn't be necessary.

See: https://clang.llvm.org/docs/ClangFormat.html

Makefile.frag
```diff
@@ -64,6 +64,12 @@ test-bootstrap:
 	vagrant ssh mo -c 'sudo mongo-orchestration -f mongo-orchestration-config.json -b 192.168.112.10 --enable-majority-read-concern start'
 	php scripts/start-servers.php
 
+format:
+	$(top_srcdir)/scripts/clang-format.sh
+
+format-changed:
+	$(top_srcdir)/scripts/clang-format.sh changed
```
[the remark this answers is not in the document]
jmikola: Ah, I see the `changed` is actually meant to only target files in the `git` diff, which is separate from `make`.

scripts/clang-format.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+if test x"$1" = x; then 
+	FILES=`git ls-files | grep '\.[ch]$'`
+fi
+if test x"$1" = xchanged; then 
+	FILES=`git diff --name-only | grep '\.[ch]$'`
```
jmikola: If the intention of `changed` is to affect files relevant to a new PR, wouldn't you also want to target untracked files?

Alternatively, will this collect files that you've staged for commit with `git add` but not yet committed?

src/BSON/Binary.c
```diff
@@ -50,29 +50,29 @@ static bool php_phongo_binary_init(php_phongo_binary_t *intern, const char *data
...
+		(type = zend_hash_str_find(props, "type", sizeof("type") - 1)) && Z_TYPE_P(type) == IS_LONG) {
 		return php_phongo_binary_init(intern, Z_STRVAL_P(data), Z_STRLEN_P(data), Z_LVAL_P(type) TSRMLS_CC);
 	}
 #else
 	zval **data, **type;
 
-	if (zend_hash_find(props, "data", sizeof("data"), (void**) &data) == SUCCESS && Z_TYPE_PP(data) == IS_STRING &&
-	    zend_hash_find(props, "type", sizeof("type"), (void**) &type) == SUCCESS && Z_TYPE_PP(type) == IS_LONG) {
+	if (zend_hash_find(props, "data", sizeof("data"), (void**)&data) == SUCCESS && Z_TYPE_PP(data) == IS_STRING &&
+		zend_hash_find(props, "type", sizeof("type"), (void**)&type) == SUCCESS && Z_TYPE_PP(type) == IS_LONG) {
```
jmikola: Newline below?

src/BSON/Binary.c
```diff
@@ -50,29 +50,29 @@ static bool php_phongo_binary_init(php_phongo_binary_t *intern, const char *data
...
-static bool php_phongo_binary_init_from_hash(php_phongo_binary_t *intern, HashTable *props TSRMLS_DC) /* {{{ */
+static bool php_phongo_binary_init_from_hash(php_phongo_binary_t* intern, HashTable* props TSRMLS_DC) /* {{{ */
 {
 #if PHP_VERSION_ID >= 70000
 	zval *data, *type;
 
-	if ((data = zend_hash_str_find(props, "data", sizeof("data")-1)) && Z_TYPE_P(data) == IS_STRING &&
-	    (type = zend_hash_str_find(props, "type", sizeof("type")-1)) && Z_TYPE_P(type) == IS_LONG) {
+	if ((data = zend_hash_str_find(props, "data", sizeof("data") - 1)) && Z_TYPE_P(data) == IS_STRING &&
+		(type = zend_hash_str_find(props, "type", sizeof("type") - 1)) && Z_TYPE_P(type) == IS_LONG) {
```
jmikola: Newline below?

src/BSON/Binary.c
jmikola: Works for me. You can add that to `phongo_compat.h`. In that case, I'd suggest ordering that commit before the one that changes formatting with the first `clang-format` invocation.

scripts/clang-format.sh
```diff
@@ -0,0 +1,13 @@
...
+fi
+
+for i in $FILES; do
+	clang-format-7 $i > /tmp/tmp-clang.c
```
jmikola: Why `clang-format-7` instead of just `clang-format`? On my local Ubuntu 16.04 system, `clang-format --version` reports `3.8.0`. Are there really that many versions between?

src/BSON/DBPointer.c
```diff
@@ -57,21 +57,20 @@ static bool php_phongo_dbpointer_init(php_phongo_dbpointer_t *intern, const char
...
-static bool php_phongo_dbpointer_init_from_hash(php_phongo_dbpointer_t *intern, HashTable *props TSRMLS_DC) /* {{{ */
+static bool php_phongo_dbpointer_init_from_hash(php_phongo_dbpointer_t* intern, HashTable* props TSRMLS_DC) /* {{{ */
 {
 #if PHP_VERSION_ID >= 70000
 	zval *ref, *id;
 
-	if ((ref = zend_hash_str_find(props, "ref", sizeof("ref")-1)) && Z_TYPE_P(ref) == IS_STRING &&
-	    (id = zend_hash_str_find(props, "id", sizeof("id")-1)) && Z_TYPE_P(id) == IS_STRING) {
+	if ((ref = zend_hash_str_find(props, "ref", sizeof("ref") - 1)) && Z_TYPE_P(ref) == IS_STRING &&
+		(id = zend_hash_str_find(props, "id", sizeof("id") - 1)) && Z_TYPE_P(id) == IS_STRING) {
```
jmikola: Newline below?

src/BSON/DBPointer.c
```diff
@@ -57,21 +57,20 @@ static bool php_phongo_dbpointer_init(php_phongo_dbpointer_t *intern, const char
...
 		return php_phongo_dbpointer_init(intern, Z_STRVAL_P(ref), Z_STRLEN_P(ref), Z_STRVAL_P(id), Z_STRLEN_P(id) TSRMLS_CC);
 	}
 #else
 	zval **ref, **id;
 
-	if (zend_hash_find(props, "ref", sizeof("ref"), (void**) &ref) == SUCCESS && Z_TYPE_PP(ref) == IS_STRING &&
-	    zend_hash_find(props, "id", sizeof("id"), (void**) &id) == SUCCESS && Z_TYPE_PP(id) == IS_STRING) {
-
+	if (zend_hash_find(props, "ref", sizeof("ref"), (void**)&ref) == SUCCESS && Z_TYPE_PP(ref) == IS_STRING &&
+		zend_hash_find(props, "id", sizeof("id"), (void**)&id) == SUCCESS && Z_TYPE_PP(id) == IS_STRING) {
```
jmikola: Newline below?

src/BSON/DBPointer.c
```diff
@@ -103,19 +102,19 @@ static PHP_METHOD(DBPointer, __toString)
...
 static PHP_METHOD(DBPointer, jsonSerialize)
 {
-	php_phongo_dbpointer_t *intern;
+	php_phongo_dbpointer_t* intern;
 #if PHP_VERSION_ID >= 70000
-	zval                    zdb_pointer, zoid;
+	zval zdb_pointer, zoid;
 #else
-	zval                   *zdb_pointer, *zoid;
+	zval *zdb_pointer, *zoid;
```
jmikola: Note: If you come up with a `ZVAL_RETVAL_TYPE` compat macro, you're going to need to break these into their own lines.

src/BSON/UTCDateTime.c
jmikola: This is very strange. It's one character off from the `intern->` assignments below.

src/BSON/functions.c
```diff
@@ -24,19 +24,21 @@
...
 {
-	zval   *data;
-	bson_t *bson;
+	zval*   data;
+	bson_t* bson;
 
-	SUPPRESS_UNUSED_WARNING(return_value_ptr) SUPPRESS_UNUSED_WARNING(this_ptr) SUPPRESS_UNUSED_WARNING(return_value_used) /* We don't use these */
+	SUPPRESS_UNUSED_WARNING(return_value_ptr)
+	SUPPRESS_UNUSED_WARNING(this_ptr)
+	SUPPRESS_UNUSED_WARNING(return_value_used) /* We don't use these */
```
jmikola: If you have a moment, can you check if these are actually still needed? I seem to recall some `PHP_FUNCTION` and `PHP_METHOD` blocks no using them.

src/MongoDB/BulkWrite.c
jmikola: I'm find with leaving it on. This looks ugly in GitHub because it uses 8-space tabs, but I imagine it'd be an improvement in our editors.

src/MongoDB/BulkWrite.c
```diff
@@ -168,12 +168,13 @@ static bool php_phongo_bulkwrite_delete_apply_options(bson_t *boptions, zval *zo
...
-	zend_error_handling      error_handling;
-	zval                    *options = NULL;
-	zend_bool                ordered = 1;
-	SUPPRESS_UNUSED_WARNING(return_value_ptr) SUPPRESS_UNUSED_WARNING(return_value) SUPPRESS_UNUSED_WARNING(return_value_used)
-
+	php_phongo_bulkwrite_t* intern;
+	zend_error_handling     error_handling;
+	zval*                   options = NULL;
+	zend_bool               ordered = 1;
+	SUPPRESS_UNUSED_WARNING(return_value_ptr)
```
jmikola: I think these would be best put in their own paragraph.

I'm not sure if it impacts `clang-format` at all, as it probably considers them function calls.

src/MongoDB/Command.c
jmikola: Ah, I commented on that earlier. Space after the cast is preferred for me.

Would it not make sense to do it in the same PR, since it's our existing convention and would limit the diff here?

src/MongoDB/CursorId.c
jmikola: What would you use for PHP 7, `= { 0 }`?

src/MongoDB/Exception/WriteException.c
jmikola: Odd, but I gather it's related to the odd indent we saw earlier.

src/MongoDB/Manager.c
```diff
@@ -206,19 +211,22 @@ static void php_phongo_manager_prep_uri_options(zval *options TSRMLS_DC) /* {{{
...
+		HashPosition pos;
+		zval**       option;
 
-		for (zend_hash_internal_pointer_reset_ex(ht_data, &pos);
-		     zend_hash_get_current_data_ex(ht_data, (void **) &option, &pos) == SUCCESS;
-		     zend_hash_move_forward_ex(ht_data, &pos)) {
-			char  *string_key = NULL;
-			uint   string_key_len = 0;
-			ulong  num_key = 0;
+		for (
```
jmikola: Did `clang-format` add the line break here (akin to `MONGO_DEBUG(` in `php_phongo.c`), or did you?

src/contrib/php_array_api.h
```diff
@@ -26,13 +26,13 @@
 #include "zend_list.h"
 
 #ifdef ZEND_ENGINE_3
-# define PAA_LENGTH_ADJ(l) (l)
-# define PAA_SYM_EXISTS zend_symtable_str_exists
-# define PAA_SYM_DEL    zend_symtable_str_del
+#define PAA_LENGTH_ADJ(l) (l)
```
jmikola: Can we exclude files in `contrib/`, for the same reason we don't want to process things in `src/libmongoc`, etc.?

src/BSON/Binary.c
jmikola: Oh, I was hoping the changes introduced by `clang-format` were all isolated in a single commit (separate from manual changes like you adding line breaks). That would allow you to throw out that commit and regenerate as needed. If not, don't worry about the ordering.

src/MongoDB/Query.c
jmikola: What happens? It injects a line break anyway? I don't really mind, as `else` still stands out from `PHONGO` and shouldn't hurt readability too much.

Is the trailing semicolon necessary? All of these macros are just `if` blocks, so I assume not. Perhaps we can delete it so that the lines starting with `PHONGO` and `else` both end consistently.

src/BSON/Binary.c
derickr: I did end up fixing this btw, including right order.

src/MongoDB/Query.c
derickr: It indeed injects a line break anyway. clang-format really likes things that look like statements end with a statement separator (``;``), so IMO, we should have one there.

src/MongoDB/BulkWrite.c
```diff
@@ -451,26 +454,21 @@ static phongo_create_object_retval php_phongo_bulkwrite_create_object(zend_class
...
+static HashTable* php_phongo_bulkwrite_get_debug_info(zval* object, int* is_temp TSRMLS_DC) /* {{{ */
 {
-#if PHP_VERSION_ID >= 70000
-	zval                      retval;
-#else
-	zval                      retval = zval_used_for_init;
-#endif
-	php_phongo_bulkwrite_t   *intern = NULL;
-
+	zval                    retval = ZVAL_STATIC_INIT;
```
jmikola: I suppose this is a bit of a WTF that we don't use `ZVAL_RETVAL_TYPE` here, but that can always be helped later by renaming `retval` to something else. This isn't a `retval` in the normal sense of a `PHP_METHOD`.

Nothing to change here, just making a note.

src/MongoDB/Manager.c
```diff
@@ -773,30 +783,28 @@ static phongo_create_object_retval php_phongo_manager_create_object(zend_class_e
...
 {
-	php_phongo_manager_t         *intern;
-	mongoc_server_description_t **sds;
+	php_phongo_manager_t*         intern;
+	mongoc_server_description_t** sds;
 	size_t                        i, n = 0;
+	zval                          retval = ZVAL_STATIC_INIT;
 #if PHP_VERSION_ID >= 70000
-	zval                          retval, cluster;
+	zval cluster;
```
jmikola: Can we rewrite this as `ZVAL_RETVAL_TYPE cluster`? That removes the `NULL` assignment for PHP 5.x, but I don't think we need to worry about that since we'll pass it to `MAKE_STD_ZVAL()` anyway.

PR #764, opened 2018-02-21
PHPC-1121: Added Appveyor config / PHPC-1114: Add template.rc file

derickr:


.appveyor.yml
```diff
@@ -0,0 +1,82 @@
+image: Visual Studio 2015
+version: '{branch}.{build}'
+
+cache:
+  - c:\build-cache -> .appveyor.yml
```
jmikola: Given that you've offloaded most of the work to scripts in `.appveyor/`, shouldn't you also invalidate if any of those change, too? It looks like Appveyor would support this, per [Multiple build cache dependencies](https://help.appveyor.com/discussions/questions/1788-multiple-build-cache-dependencies#comment_38254037).

    - c:\build-cache -> .appveyor.yml, appveyor\*.cmd

I'm thinking ahead, but this might avoid needing to [manually invalidate the cache](https://www.appveyor.com/docs/build-cache/#cleaning-up-cache).

.appveyor/build_task.cmd
```diff
@@ -0,0 +1,73 @@
...
+	if "%PHP_SDK_ARCH%"=="x86" set ARCH_IN_FILENAME=
+	if "%PHP_SDK_ARCH%"=="x64" set ARCH_IN_FILENAME=-x86_64
+
+	cd /d C:\projects\php-src
+
+	cmd /c buildconf.bat --force
+
+	if %errorlevel% neq 0 exit /b 3
+
+	cmd /c configure.bat --disable-all --with-mp=auto --enable-cli --%ZTS_STATE%-zts --enable-json --with-openssl --enable-mongodb=shared --enable-object-out-dir=%PHP_BUILD_OBJ_DIR% --with-config-file-scan-dir=%APPVEYOR_BUILD_FOLDER%\build\modules.d --with-prefix=%APPVEYOR_BUILD_FOLDER%\build --with-php-build=%DEPS_DIR%
```
jmikola: Other than the "mongodb" replacements here, was this basically copied from another project?

phongo_version.h
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef PHONGO_VERSION_H
+#define PHONGO_VERSION_H
+
+#define PHP_MONGODB_VERSION "1.5.0-dev"
+#define PHP_MONGODB_STABILITY "devel"
+#define PHP_MONGODB_VERSION_DESC 1,5,0,0
```
jmikola: Instructions for bumping the version in `CONTRIBUTING.md` should be updated in the same commit (assuming that still refers to `php_phongo.h`).

template.rc
```diff
@@ -0,0 +1,57 @@
...
+    BLOCK "StringFileInfo"
+    BEGIN
+        BLOCK "040904b0"
+        BEGIN
+            VALUE "Comments", THANKS_GUYS "\0"
+            VALUE "CompanyName", "MongoDB, Inc.\0"
+            VALUE "FileDescription", FILE_DESCRIPTION "\0"
+            VALUE "FileVersion", PHP_MONGODB_VERSION
+            VALUE "InternalName", FILE_NAME "\0"
+            VALUE "LegalCopyright", "Copyright � 2014-2018 MongoDB, Inc.\0"
```
jmikola: Why do I see a Unicode replacement character here?

I'd also suggest using "present" as the upper bound, per [DRIVERS-329](https://jira.mongodb.org/browse/DRIVERS-329). This could be rewritten as:

> Copyright 2014-present MongoDB, Inc.

template.rc
```diff
@@ -0,0 +1,57 @@
...
+ FILESUBTYPE VFT2_UNKNOWN
+BEGIN
+    BLOCK "StringFileInfo"
+    BEGIN
+        BLOCK "040904b0"
+        BEGIN
+            VALUE "Comments", THANKS_GUYS "\0"
+            VALUE "CompanyName", "MongoDB, Inc.\0"
+            VALUE "FileDescription", FILE_DESCRIPTION "\0"
+            VALUE "FileVersion", PHP_MONGODB_VERSION
```
jmikola: Is this being pulled directly from `phongo_version.h`? If so, both it and `THANKS_GUYS` are declared with `#define`. Why do you add a `"\0"` to `THANKS_GUYS` but not here?

.appveyor.yml
```diff
@@ -0,0 +1,82 @@
...
+      ARCHITECTURE: x86
+      ZTS_STATE: disable
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      PHP_BUILD_CRT: vc14
+
+install:
+  - .appveyor\install.cmd
+
+build_script:
+  - .appveyor\build.cmd
```
jmikola: This currently runs now tests and is only verifying that we can build the DLL, correct?

.appveyor/install.cmd
```diff
@@ -0,0 +1,64 @@
...
+
+	if not exist "%PHP_BUILD_OBJ_DIR%" (
+		echo Creating %PHP_BUILD_OBJ_DIR%
+		mkdir "%PHP_BUILD_OBJ_DIR%"
+	)
+
+	if not exist "%PHP_BUILD_CACHE_SDK_DIR%" (
+		echo Cloning remote SDK repository
+		rem git clone -q --depth=1 --branch %SDK_BRANCH% %SDK_REMOTE% "%PHP_BUILD_CACHE_SDK_DIR%" 2>&1
+		git clone --branch %SDK_BRANCH% %SDK_REMOTE% "%PHP_BUILD_CACHE_SDK_DIR%" 2>&1
```
jmikola: Why does the commented command above have `-q --depth=1`. You use those arguments for cloning PHP below, so why not for the SDK as well?

.appveyor/install.cmd
```diff
@@ -0,0 +1,64 @@
...
+	) else (
+		echo Fetching remote SDK repository
+		git --git-dir="%PHP_BUILD_CACHE_SDK_DIR%\.git" --work-tree="%PHP_BUILD_CACHE_SDK_DIR%" fetch --prune origin 2>&1
+		echo Checkout SDK repository branch
+		git --git-dir="%PHP_BUILD_CACHE_SDK_DIR%\.git" --work-tree="%PHP_BUILD_CACHE_SDK_DIR%" checkout --force %SDK_BRANCH%
+	)
+
+	if "%PHP_REL%"=="master" (
+		echo git clone -q --depth=1 --branch=%PHP_REL% https://github.com/php/php-src C:\projects\php-src
```
jmikola: Just pointing out that these `echo` lines are inconsistent with those for the SDK repository. Would it be preferable to write `"Cloning remote PHP repository (%PHP_REL%)"`?

.appveyor/install.cmd
```diff
@@ -0,0 +1,64 @@
...
+		git clone -q --depth=1 --branch=%PHP_REL% https://github.com/php/php-src C:\projects\php-src
+	) else (
+		echo git clone -q --depth=1 --branch=PHP-%PHP_REL% https://github.com/php/php-src C:\projects\php-src
+		git clone -q --depth=1 --branch=PHP-%PHP_REL% https://github.com/php/php-src C:\projects\php-src
+	)
```
jmikola: Any reason we can't cache the `php-src` repository as well?

.appveyor.yml
```diff
@@ -0,0 +1,82 @@
+image: Visual Studio 2015
+version: '{branch}.{build}'
+
+cache:
+  - c:\build-cache -> .appveyor.yml
+  - c:\build-cache\sdk -> .appveyor.yml
```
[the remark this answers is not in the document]
derickr: Yes, I think so. But it was in the original example that Anatol had for me.

.appveyor.yml
```diff
@@ -0,0 +1,82 @@
+image: Visual Studio 2015
+version: '{branch}.{build}'
```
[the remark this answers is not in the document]
derickr: Yes — it creates the "build name" on the right hand side at: https://ci.appveyor.com/project/derickr/mongo-php-driver/history

.appveyor.yml
```diff
@@ -0,0 +1,82 @@
...
+    - PHP_REL: 7.1
+      ARCHITECTURE: x86
+      ZTS_STATE: disable
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      PHP_BUILD_CRT: vc14
+    - PHP_REL: 7.0
+      ARCHITECTURE: x86
+      ZTS_STATE: disable
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      PHP_BUILD_CRT: vc14
```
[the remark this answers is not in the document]
derickr: No real reason, except that we'd be creating binaries for outdated PHP versions, and building binaries takes a looong time on AppVeyor already.

.appveyor.yml
derickr: I've never had the cache create a problem for me with Xdebug, so didn't really think much about this.

.appveyor/build_task.cmd
derickr: Yes, copied from Xdebug, which was copied from something else.

We're still building php.exe too. I think that's fine for now, as "it works", and messing with this just takes too long for it to be worthwhile.

phongo_version.h
derickr: OK, will fix that (thought I had).

template.rc
derickr: The file encoding is win-1252 (see line 12), I guess GitHub doesn't get that.

.appveyor.yml
derickr: It does run tests that don't need a mongod server, but right now I'm not making the build fail due to failing tests. We need to follow up on that.

template.rc
derickr: Don't know. I copied this from Xdebug where it worked for the past 10 years :-)

.appveyor/install.cmd
derickr: Snippet I got from Anatol. I guess the ``rem`` line can go.

.appveyor/install.cmd
derickr: The cache is limited in space, you can't just chuck everything in it.

.appveyor/install.cmd
derickr: Maybe. I don't care enough to go through another hours long iteration for this.

.appveyor.yml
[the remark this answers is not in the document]
jmikola: I think it's worth adding that if we're to rely on this instead of manually creating test builds for PHP 5.5 and 5.6 before releasing.

If build times are very long, I'd also be amenable to disabling Appveyor for PRs and simply having it monitor branches on the mongodb repo.

.appveyor/build_task.cmd
jmikola: Ok. IIRC, @weltling was talking about in #php.pecl:

```
<weltling> Derick, btw. perhaps you could check the phpize based approach https://github.com/krakjoe/apcu/blob/master/appveyor.yml 
<weltling> you can spare some time per single job by not building all the php, still need to fetch deps in your case
<weltling> see the build_script: section in the yml file
<weltling> for the tests - also of course php is needed, but in that case it fetches the precompiled version, too
<weltling> so except the extension itself, in most case nothing needs to be compiled
<weltling> with some exts i was helping out, the build time dropped reasonably
<weltling> here it fetches https://github.com/krakjoe/apcu/blob/master/appveyor.yml#L99
<weltling> basically, there it fetches the dev package, generates a job file and runs it, all packed into yml
<weltling> if there are no external deps, it's a way easier and faster approach, with deps might need some more lines
```

If you don't want to handle this in this PR, I would suggest creating a PHPC ticket so we can track this for later. Feel free to copy/paste that chat log into the ticket for future reference, too.

.appveyor/install.cmd
jmikola: Noted. I was expecting this wouldn't be massive, since you're only using a depth of 1.

In any event, I suppose we can revisit this when optimizing later to not recompile `php.exe` each time.

.appveyor/install.cmd
jmikola: I'd suggest adding `-q --depth=1` to this `git clone`, though. I don't see any reason you need the full commit history, and `-q` will make the output a bit quieter. Since progress won't be reported to stderr, we can probably remove the redirection with `2>&1` as well.

CONTRIBUTING.md
```diff
@@ -222,25 +225,27 @@ $ git tag -a -m "Release X.Y.Z" X.Y.Z
...
 #define PHP_MONGODB_STABILITY "stable"
+#define PHP_MONGODB_VERSION_DESC 1,1,8,1
 ```
 
 The above would be changed to:
 
 ```
 #define PHP_MONGODB_VERSION "1.1.9-dev"
 #define PHP_MONGODB_STABILITY "devel"
+#define PHP_MONGODB_VERSION_DESC 1,1,9,0
```
jmikola: How do we increment this if going from 1.4.0-beta1 back to 1.4.0-dev and later to 1.4.0-rc1? I assume the value might transition from the following states:

 * `1,4,0,0` (original 1.4.0-dev)
 * `1,4,0,1` (first beta)
 * `1,4,0,2` (reverted back to dev)
 * `1,4,0,3` (first RC)

If so, can we summarize this in the "**Note:** If this is an alpha, beta, or RC release..." paragraph below?

template.rc
jmikola: Ok. I think we should still use "present" as the date's upper bound, though. That will ensure this never goes out of date.

.appveyor/build_task.cmd
```diff
@@ -0,0 +1,73 @@
...
+	if %errorlevel% neq 0 exit /b 3
+
+	nmake /NOLOGO
+	if %errorlevel% neq 0 exit /b 3
+
+	nmake install
+
+	if %errorlevel% neq 0 exit /b 3
+
+	mkdir c:\tests_tmp
```
jmikola: I missed that the test suite was being executed here.

I think this really belongs in its own `test_script.cmd` file, which can be run from the `test_script` phase in `appveyor.yml`. https://www.appveyor.com/docs/appveyor-yml/ has an example of this:

```yml
# to run your custom scripts instead of automatic tests
test_script:
  - echo This is my custom test script
```

.appveyor/install.cmd
```diff
@@ -0,0 +1,63 @@
...
+
+	if not exist "%PHP_BUILD_OBJ_DIR%" (
+		echo Creating %PHP_BUILD_OBJ_DIR%
+		mkdir "%PHP_BUILD_OBJ_DIR%"
+	)
+
+	if not exist "%PHP_BUILD_CACHE_SDK_DIR%" (
+		echo Cloning remote SDK repository
+		git clone --branch %SDK_BRANCH% %SDK_REMOTE% "%PHP_BUILD_CACHE_SDK_DIR%" 2>&1
```
jmikola: I'd suggest adding `-q --depth=1` to this `git clone`, though. I don't see any reason you need the full commit history, and `-q` will make the output a bit quieter. Since progress won't be reported to stderr, we can probably remove the redirection with `2>&1` as well.

.appveyor/build_task.cmd
jmikola: If this poses problems with making the build fail (as you mentioned in https://github.com/mongodb/mongo-php-driver/pull/764#discussion_r170301548), which can't be resolved by [Allow Failing Jobs](https://www.appveyor.com/docs/build-configuration/#allow-failing-jobs), then we can leave this as-is. Please make a PHPC ticket to revisit this later, though.

derickr: The problem was more that in order to run the tests in a separate script, you'd have to store the output of the build tasks, and basically restore things for every task. Not worth the effort, as this works just fine™. I wouldn't even want to add a PHPC ticket for this.

.appveyor.yml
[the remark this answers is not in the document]
derickr: I tried that, but no luck:
https://ci.appveyor.com/project/derickr/mongo-php-driver/build/master.18/job/yc9jeqbrawjkuh97#L51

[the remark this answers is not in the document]
derickr: The new SDK only supports 2015 and 2017: https://github.com/Microsoft/php-sdk-binary-tools#requirements

phongo_version.h
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright 2014-2018 MongoDB, Inc.
```
jmikola: Since this file was just created, the date should start at 2018. This should read "2018-present" per [DRIVERS-329](https://jira.mongodb.org/browse/DRIVERS-329).

template.rc
```diff
@@ -0,0 +1,57 @@
...
+    BLOCK "StringFileInfo"
+    BEGIN
+        BLOCK "040904b0"
+        BEGIN
+            VALUE "Comments", THANKS_GUYS "\0"
+            VALUE "CompanyName", "MongoDB, Inc.\0"
+            VALUE "FileDescription", FILE_DESCRIPTION "\0"
+            VALUE "FileVersion", PHP_MONGODB_VERSION "\0"
+            VALUE "InternalName", FILE_NAME "\0"
+            VALUE "LegalCopyright", "Copyright � 2014-2018 MongoDB, Inc.\0"
```
jmikola: This should read "2018-present" per [DRIVERS-329](https://jira.mongodb.org/browse/DRIVERS-329).

My [last comment](https://github.com/mongodb/mongo-php-driver/pull/764/commits/baeeb395be2a4d9bcc4875b8d4e0193fc6bd7051#r170291984) on this appears to have been hidden by recent diffs.

PR #766, opened 2018-02-22
PHPC-1118: Add detection of the correct clang-format binary

derickr:


scripts/clang-format.sh
```diff
@@ -12,6 +12,29 @@ if test x"$1" = xchanged; then
...
+which clang-format
+if [ $? = 0 ]; then
+	VERSION=`clang-format -version | cut -d " " -f 3 | cut -c 1-3`
+	if [ x"$VERSION" = "x6.0" ]; then
+		CLANG_FORMAT=`which clang-format`
+	fi
+fi
+
+if [ x"$CLANG_FORMAT" = "x" ]; then
+	echo "Couldn't find the right clang-format (needs version 6.0)"
```
jmikola: Due to unrelated issues with the https://apt.llvm.org/ returning a "hash sum mismatch" error for the 6.0 packages only (unfortunate coincidence, I know), I'm only able to install `clang-format-7` on my Ubuntu 16.04 machine.

If I modify this script to avoid the error and accept version "7.0", I can successfully run `make format` just fine. Given that, can we tweak this to require >= 6.0 or perhaps major versions 6 and 7 if you're concerned about things breaking in a forthcoming 8.0 release?

derickr: Supporting 6.0 and 7 works for me. But 8 can break things again...

jmikola: I suggest we just deal with that when the time comes. Any error will be hard to ignore when running `make format`.

derickr: Do you want to make a PR against my branch for the "7.0" check?

jmikola: https://github.com/derickr/mongo-php-driver/pull/2

PR #768, opened 2018-02-26
PHPC-1090: Clean up switch statements for exception class selection

kvwalker:
https://jira.mongodb.org/browse/PHPC-1090

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -144,57 +144,16 @@ zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_d
 			return php_phongo_connectiontimeoutexception_ce;
 		case MONGOC_ERROR_CLIENT_AUTHENTICATE:
 			return php_phongo_authenticationexception_ce;
-
 		case MONGOC_ERROR_COMMAND_INVALID_ARG:
 			return php_phongo_invalidargumentexception_ce;
-
-		case MONGOC_ERROR_STREAM_INVALID_TYPE:
-		case MONGOC_ERROR_STREAM_INVALID_STATE:
-		case MONGOC_ERROR_STREAM_NAME_RESOLUTION:
-		case MONGOC_ERROR_STREAM_CONNECT:
-		case MONGOC_ERROR_STREAM_NOT_ESTABLISHED:
-			return php_phongo_connectionexception_ce;
-		case MONGOC_ERROR_CLIENT_NOT_READY:
-		case MONGOC_ERROR_CLIENT_TOO_BIG:
-		case MONGOC_ERROR_CLIENT_TOO_SMALL:
-		case MONGOC_ERROR_CLIENT_GETNONCE:
-		case MONGOC_ERROR_CLIENT_NO_ACCEPTABLE_PEER:
-		case MONGOC_ERROR_CLIENT_IN_EXHAUST:
-		case MONGOC_ERROR_PROTOCOL_INVALID_REPLY:
-		case MONGOC_ERROR_PROTOCOL_BAD_WIRE_VERSION:
-		case MONGOC_ERROR_CURSOR_INVALID_CURSOR:
-		/*case MONGOC_ERROR_PROTOCOL_ERROR:*/
-		case MONGOC_ERROR_BSON_INVALID:
-		case MONGOC_ERROR_MATCHER_INVALID:
-		case MONGOC_ERROR_NAMESPACE_INVALID:
-		case MONGOC_ERROR_GRIDFS_INVALID_FILENAME:
-			return php_phongo_runtimeexception_ce;
-		case MONGOC_ERROR_QUERY_FAILURE:
-		case MONGOC_ERROR_QUERY_COMMAND_NOT_FOUND:
-		case MONGOC_ERROR_QUERY_NOT_TAILABLE:
-		case MONGOC_ERROR_COLLECTION_INSERT_FAILED:
-			return php_phongo_serverexception_ce;
 	}
 	switch (domain) {
+		case MONGOC_ERROR_COMMAND:
+			// TODO: return php_phongo_commandexception_ce after PHPC-1089 is merged
 		case MONGOC_ERROR_SERVER:
 			return php_phongo_serverexception_ce;
-		case MONGOC_ERROR_CLIENT:
 		case MONGOC_ERROR_STREAM:
-		case MONGOC_ERROR_PROTOCOL:
-		case MONGOC_ERROR_CURSOR:
-		case MONGOC_ERROR_QUERY:
-		case MONGOC_ERROR_INSERT:
-		case MONGOC_ERROR_SASL:
-		case MONGOC_ERROR_BSON:
-		case MONGOC_ERROR_MATCHER:
-		case MONGOC_ERROR_NAMESPACE:
-		case MONGOC_ERROR_COMMAND:
-		case MONGOC_ERROR_COLLECTION:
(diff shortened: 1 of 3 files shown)


kvwalker:
I removed the MONGOC_ERROR cases that are not used in libmongoc, according to [Jesse's comment](https://jira.mongodb.org/browse/HELP-5525?focusedCommentId=1770461&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1770461) on a HELP ticket.

tests/standalone/write-error-001.phpt
```diff
@@ -19,11 +19,11 @@ $wtimeout = 1000;
...
 throws(function() use($bulk, $writeConcern, $manager) {
     $result = $manager->executeBulkWrite(NS, $bulk, $writeConcern);
-}, "MongoDB\Driver\Exception\ConnectionException");
+}, "MongoDB\Driver\Exception\BulkWriteException");
 
 ?>
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
-OK: Got MongoDB\Driver\Exception\ConnectionException
```
jmikola: Can you add the exception message to this test? Should entail changing `throws(...)` to `echo throws(...), "\n";`. I expect the error is due to using `w` >= 1 on a standalone.

tests/standalone/write-error-001.phpt
```diff
@@ -17,13 +17,14 @@ $hannes_id = $bulk->insert($hannes);
...
     $result = $manager->executeBulkWrite(NS, $bulk, $writeConcern);
-}, "MongoDB\Driver\Exception\BulkWriteException");
+}, "MongoDB\Driver\Exception\BulkWriteException"),  "\n";
 
 ?>
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
 OK: Got MongoDB\Driver\Exception\BulkWriteException
+cannot use 'w' > 1 when a host is not replicated
```
jmikola: This error message dates back to https://github.com/mongodb/mongo/commit/13226efcd6638b707e0f4a64564d655cea9bc9c7#diff-750135e45674cd2fe8dc31115771a990R64 (server 2.6.0) and hasn't changed since, so we should be fine without a pattern here. `EXPECTF` and be changed to `EXPECT`.

php_phongo.c
```diff
@@ -144,38 +144,15 @@ zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_d
...
 			return php_phongo_authenticationexception_ce;
-
 		case MONGOC_ERROR_COMMAND_INVALID_ARG:
 			return php_phongo_invalidargumentexception_ce;
-
-		case MONGOC_ERROR_STREAM_INVALID_TYPE:
-		case MONGOC_ERROR_STREAM_INVALID_STATE:
 		case MONGOC_ERROR_STREAM_NAME_RESOLUTION:
 		case MONGOC_ERROR_STREAM_CONNECT:
-		case MONGOC_ERROR_STREAM_NOT_ESTABLISHED:
```
jmikola: Apologies if I asked previously, but are the three `MONGOC_ERROR_STREAM` codes being removed here not longer used in libmongoc?

Does it make sense to remove all of these and instead add a case for the `MONGOC_ERROR_STREAM` domain below, where we return `php_phongo_connectionexception_ce`?

php_phongo.c
```diff
@@ -144,38 +144,15 @@ zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_d
...
-			return php_phongo_runtimeexception_ce;
-		case MONGOC_ERROR_QUERY_FAILURE:
-		case MONGOC_ERROR_QUERY_COMMAND_NOT_FOUND:
-		case MONGOC_ERROR_QUERY_NOT_TAILABLE:
-		case MONGOC_ERROR_COLLECTION_INSERT_FAILED:
-			return php_phongo_serverexception_ce;
 	}
 	switch (domain) {
+		case MONGOC_ERROR_COMMAND:
+			// TODO: return php_phongo_commandexception_ce after PHPC-1089 is merged
```
jmikola: Since #771 is also open, does one of these PRs depend on the other? Based on this comment, it looks like you might want to get this PR merged and then rebase #771 afterwards.

php_phongo.c
```diff
@@ -144,38 +144,15 @@ zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_d
...
-		case MONGOC_ERROR_BSON_INVALID:
-		case MONGOC_ERROR_MATCHER_INVALID:
-		case MONGOC_ERROR_NAMESPACE_INVALID:
-		case MONGOC_ERROR_GRIDFS_INVALID_FILENAME:
-			return php_phongo_runtimeexception_ce;
-		case MONGOC_ERROR_QUERY_FAILURE:
-		case MONGOC_ERROR_QUERY_COMMAND_NOT_FOUND:
-		case MONGOC_ERROR_QUERY_NOT_TAILABLE:
-		case MONGOC_ERROR_COLLECTION_INSERT_FAILED:
-			return php_phongo_serverexception_ce;
```
jmikola: Based on the domain cases below, these four error codes will now be reported as `php_phongo_runtimeexception_ce`.

If I understand [libmongoc's Error Reporting](http://mongoc.org/libmongoc/current/errors.html) correctly, the `MONGOC_ERROR_QUERY` domain is no longer used in error API v2. By extension, none of the `MONGOC_ERROR_QUERY_*` codes are used. 

As for `MONGOC_ERROR_COLLECTION_INSERT_FAILED`, I believe that's only used if we would execute an empty insert (legacy `OP_INSERT` or newer `insert` command). Since PHPC only uses the bulk write API, we'd never hit that and only reach the error in [`mongoc_bulk_operation_execute()`](https://github.com/mongodb/mongo-c-driver/blob/1.9.3/src/mongoc/mongoc-bulk-operation.c#L728) instead.

Given that, I think these cases are safe to remove.

php_phongo.c
kvwalker: Yes, the removed codes are no longer used in libmongoc, according to [Jesse's comment on a help ticket](https://jira.mongodb.org/browse/HELP-5525?focusedCommentId=1770461&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1770461) .

php_phongo.c
kvwalker: Yep, that sounds like a good plan.

php_phongo.c
```diff
@@ -157,6 +154,7 @@ zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_d
 			return php_phongo_serverexception_ce;
 		case MONGOC_ERROR_CLIENT:
 		case MONGOC_ERROR_STREAM:
+			return php_phongo_connectionexception_ce;
```
jmikola: This may change behavior for the `MONGOC_ERROR_CLIENT` domain, since its case falls through.

Rather than move `case MONGOC_ERROR_CLIENT:` below this `return`, I'd propose you delete it and also remove all of the cases between this `return` and `default:`. This way, we only have cases for non-default return values.

jmikola: Additionally, can you remove the following lines down below?

```c
/* FIXME: We don't have the Exceptions mocked yet.. */
#if 0
	return phongo_ce_mongo_connection_exception;
#endif
```

`phongo_ce_mongo_connection_exception` does not exist and that's simply dead code.

PR #770, opened 2018-02-27
PHPC-1132, PHPC-1133: Remove HHVM checks and clean up includes

jmikola:
https://jira.mongodb.org/browse/PHPC-1132
https://jira.mongodb.org/browse/PHPC-1133

There was some overlap with the commit for removing HHVM checks, so I've bundled these tickets into one PR.

### DIFF
--- a/tests/manager/manager-getreadconcern-001.phpt
+++ b/tests/manager/manager-getreadconcern-001.phpt
@@ -1,20 +1,17 @@
 --TEST--
 MongoDB\Driver\Manager::getReadConcern()
---SKIPIF--
-<?php require __DIR__ . "/../utils/basic-skipif.inc"?>
 --FILE--
 <?php
-require_once __DIR__ . "/../utils/basic.inc";
 
 $tests = [
-    [STANDALONE, []],
-    [STANDALONE . '/?readconcernlevel=local', []],
-    [STANDALONE . '/?readconcernlevel=majority', []],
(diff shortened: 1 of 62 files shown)


tests/manager/manager-getreadconcern-001.phpt
```diff
@@ -1,20 +1,17 @@
...
-    [STANDALONE, ['readconcernlevel' => 'not-yet-supported']],
-    [STANDALONE . '/?readconcernlevel=local', ['readconcernlevel' => 'majority']],
+    [null, []],
+    ['mongodb://127.0.0.1/?readconcernlevel=local', []],
+    ['mongodb://127.0.0.1/?readconcernlevel=majority', []],
+    ['mongodb://127.0.0.1/?readconcernlevel=not-yet-supported', []],
+    [null, ['readconcernlevel' => 'local']],
+    [null, ['readconcernlevel' => 'majority']],
+    [null, ['readconcernlevel' => 'not-yet-supported']],
+    ['mongodb://127.0.0.1/?readconcernlevel=local', ['readconcernlevel' => 'majority']],
```
derickr: Do these tests not connect at all?

jmikola: They do not. Since they don't call an execute method nor `selectServer()`, SDAM is never initialized.

I considered using "example.com" here, but decided to keep the same default that is effectively used when we pass `null` for the cases above.

PR #771, opened 2018-02-27
PHPC-1089: Introduce CommandException class

kvwalker:
https://jira.mongodb.org/browse/PHPC-1089


src/MongoDB/Exception/CommandException.c
```diff
@@ -0,0 +1,82 @@
...
+{
+	zend_class_entry ce;
+
+	INIT_NS_CLASS_ENTRY(ce, "MongoDB\\Driver\\Exception", "CommandException", php_phongo_commandexception_me);
+#if PHP_VERSION_ID >= 70000
+	php_phongo_commandexception_ce = zend_register_internal_class_ex(&ce, php_phongo_serverexception_ce);
+#else
+	php_phongo_commandexception_ce = zend_register_internal_class_ex(&ce, php_phongo_serverexception_ce, NULL TSRMLS_CC);
+#endif
+	php_phongo_commandexception_ce->ce_flags |= ZEND_ACC_EXPLICIT_ABSTRACT_CLASS;
```
jmikola: This looks like WriteException copypasta. While WriteException is abstract, CommandException probably shouldn't be.

src/MongoDB/Exception/CommandException.c
```diff
@@ -0,0 +1,82 @@
...
+#include <php.h>
+
+#include "phongo_compat.h"
+#include "php_phongo.h"
+
+zend_class_entry* php_phongo_commandexception_ce;
+
+/* {{{ proto document MongoDB\Driver\Exception\CommandException::getResultDocument()
+   Returns the result document from the failed command. */
+static PHP_METHOD(CommandException, getResultDocument)
```
jmikola: In the interest of improving test coverage, I think it'd be a good idea to create `tests/exception/commandexception-getresultdocument-001.phpt` for this.

The test won't need to use servers, so you can do without a `--SKIPIF--` and the typical "basic" include. I'd suggest creating a new instance of the exception (message/code is irrelevant, so no constructor arguments), using [reflection](http://php.net/reflection) to assign an arbitrary document to the `resultDocument` property, and then verifying that this method returns the same document. For the comparison, I'd suggest starting with `var_dump($resultDocument === $exception->getResultDocument())` and asserting that it outputs "bool(true)". Based on the implementation, I think an assigned object (assuming stdClass is used as the document) would be the same. `===` should certainly work with an array document.

If this works out, you can then copy the test for WriteException (either extend it locally in the test or use BulkWriteException), too.

jmikola:
I'll hold off on reviewing this further until #768 is merged and this gets rebased.

src/MongoDB/Exception/CommandException.c
```diff
@@ -0,0 +1,81 @@
...
+	RETURN_ZVAL(resultdocument, 1, 0);
+} /* }}} */
+
+/* {{{ MongoDB\Driver\Exception\CommandException function entries */
+ZEND_BEGIN_ARG_INFO_EX(ai_CommandException_void, 0, 0, 0)
+ZEND_END_ARG_INFO()
+
+static zend_function_entry php_phongo_commandexception_me[] = {
+	PHP_ME(CommandException, getResultDocument, ai_CommandException_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+		PHP_FE_END
```
derickr: Did clang-format format it like this? It should have put PHP_FE_END in the same column as PHP_ME. I'll have a look.

tests/exception/commandexception-getresultdocument-001.phpt
```diff
@@ -0,0 +1,22 @@
+--TEST--
+MongoDB\Driver\Exception\CommandException result document
```
jmikola: This can be: `MongoDB\Driver\Exception\CommandException::getResultDocument()`

tests/exception/bulkwriteexception-getresultdocument-001.phpt
```diff
@@ -0,0 +1,22 @@
+--TEST--
+MongoDB\Driver\Exception\BulkWriteException write result
```
jmikola: This can be: `MongoDB\Driver\Exception\BulkWriteException::getWriteResult()`

Also, the filename should be `bulkwriteexception-getwriteresult-001.phpt`.

src/MongoDB/Exception/CommandException.c
jmikola: Looks like clang-format's fault. See: https://github.com/mongodb/mongo-php-driver/blob/17e0a4b12edcc11d438881128a54998134898f70/src/MongoDB/Manager.c#L732

PR #773, opened 2018-02-28
PHPC-355: Add test for ConnectionTimeoutException

kvwalker:
https://jira.mongodb.org/browse/PHPC-355

### DIFF
--- a/tests/manager/manager-executeCommand_error-003.phpt
+++ b/tests/manager/manager-executeCommand_error-003.phpt
@@ -1,0 +1,21 @@
+--TEST--
+MongoDB\Driver\Manager::executeCommand() connection error
+--FILE--
+<?php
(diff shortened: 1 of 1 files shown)


tests/manager/connect-error-001.phpt
```diff
@@ -0,0 +1,22 @@
+--TEST--
+MongoDB\Driver\Manager - no connection (command)
```
jmikola: I'd suggest naming the file `manager-executeCommand_error-003.phpt` (there is a gap between `002` and `004`, so we can fill it. And then "MongoDB\Driver\Manager::executeCommand() connection error" as a test name.

Down the line, we can make copies of this test for other execute methods on Manager and Server. The naming convention becomes much more practical in that case.

PR #776, opened 2018-03-07
Fixed PHPC-922: Improve specificity in generic corrupt BSON exception message

derickr:


tests/bson/bug0531-002.phpt
```diff
@@ -0,0 +1,21 @@
+--TEST--
+PHPC-531: Segfault due to double free by corrupt BSON visitor (nested document)
```
jmikola: It's a bit odd for this to add a new PHPC-531 test; however, `bson-toPHP_error-004.phpt` is the only other place where we test for "Detected unknown BSON type" outside of BSON spec tests. 

Ideally, this would be a new test case in that file (using `pack()`), or a newly created `bson-toPHP_error-005.phpt` file (there is a gap between 4 and 6). The latter would be fairly low effort, but I won't argue if you want to leave this as-is.

tests/bson/bson-toPHP_error-006.phpt
```diff
@@ -0,0 +1,36 @@
+--TEST--
+MongoDB\BSON\toPHP(): Field path values with bson_iter_visit_all() failures
```
jmikola: Any reason this is `006` if `005` doesn't exist?

tests/bson/bson-fromPHP_error-008.phpt
```diff
@@ -0,0 +1,47 @@
+--TEST--
+BSON\fromPHP(): PHP documents with circular references
```
jmikola: `MongoDB\BSON\fromPHP()`

tests/bson/bson-fromPHP_error-008.phpt
```diff
@@ -0,0 +1,47 @@
...
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/tools.php';
+
+echo "\nTesting packed array with circular reference\n";
+
+echo throws(function() {
+    $document = ['x' => 1, 'y' => [1, 2, 3]];
```
jmikola: This looks redundant, since `bson-fromPHP_error-004.phpt` already tests `$document = ['x' => 1, 'y' => []];` and asserts that the field path is "y.0". All this is testing is that `$document['y'][] = &$document['y'];` on the subsequent line appends the fourth element in a packed array instead of the first.

tests/bson/bson-fromPHP_error-008.phpt
```diff
@@ -0,0 +1,47 @@
...
+    fromPHP($document);
+}, 'MongoDB\Driver\Exception\UnexpectedValueException'), "\n";
+
+echo "\nTesting object with circular reference\n";
+
+echo throws(function() {
+    $document = (object) ['x' => 1, 'y' => (object) []];
+    $document->y->z = &$document->y;
+    fromPHP($document);
+}, 'MongoDB\Driver\Exception\UnexpectedValueException'), "\n";
```
jmikola: This test case is identical to the one in `bson-fromPHP_error-004.phpt` and should be removed.

tests/bson/bson-fromPHP_error-008.phpt
```diff
@@ -0,0 +1,47 @@
...
+
+echo "\nTesting associative array with circular reference\n";
+
+echo throws(function() {
+    $document = ['x' => 1, 'y' => []];
+    $document['y'][0]['z'] = &$document['y'];
+    fromPHP($document);
+}, 'MongoDB\Driver\Exception\UnexpectedValueException'), "\n";
```
jmikola: This appears to be the only new test, which is only checking that ".0." is incorporated into the exception message's field path. I think this should be be added to the `004` test as an additional case since it's so closely related to the other cases. The test case description should then be: "Testing associative array in packed array with circular reference" or something to that effect.

Noting the historic argument of not modifying existing test cases, I'd argue that spreading closely related tests out across multiple files only complicates maintainability. This would also only be a test case addition to an existing file (not altering existing tests in any way).

tests/bson/bson-fromPHP_error-005.phpt
```diff
@@ -7,7 +7,7 @@ require_once __DIR__ . '/../utils/tools.php';
 
 class MyRecursiveSerializable implements MongoDB\BSON\Serializable
 {
-    public $x = 1;
+    public $child = 1;
```
jmikola: Based on the message, I assume this change was necessary to demonstrate that BSON serialization never even reaches this property. That would have been difficult before since both properties shared the same name ("x").

tests/bson/bson-toPHP_error-004.phpt
```diff
@@ -32,15 +32,15 @@ foreach ($tests as $bson) {
 ?>
 ===DONE===
 <?php exit(0); ?>
---EXPECTF--
+--EXPECT--
```
jmikola: Good catch.

src/bson.c
```diff
@@ -52,15 +52,162 @@ static void php_phongo_bson_visit_corrupt(const bson_iter_t* iter ARG_UNUSED, vo
...
+	php_phongo_field_path_write_item_at_current_level(field_path, element);
+
+	field_path->current_level++;
+
+	return true;
+}
+
+bool php_phongo_field_path_pop(php_phongo_field_path* field_path)
+{
+	field_path->current_level--;
```
jmikola: Would it make sense for this to also check if `field_path->elements[field_path->current_level]` is null before decrementing? Currently, it looks like this might leave an element beyond the post-pop current element allocated. That runs contrary to logic you have in `php_phongo_field_path_ensure_allocation()` to ensure that pointers beyond the current level are set to null.

src/bson.c
```diff
@@ -52,15 +52,162 @@ static void php_phongo_bson_visit_corrupt(const bson_iter_t* iter ARG_UNUSED, vo
...
+	if (field_path->elements[field_path->current_level] != NULL) {
+		efree(field_path->elements[field_path->current_level]);
+		field_path->elements[field_path->current_level] = NULL;
+	}
+
+	return true;
+}
+
+static bool php_phongo_bson_state_is_initialized(php_phongo_bson_state* state)
```
jmikola: Any reason not to use an `inline` hint here? AFAIK, it's used very sparingly (BulkWrite.c and php_array_api.h).

src/bson.c
```diff
@@ -52,15 +52,162 @@ static void php_phongo_bson_visit_corrupt(const bson_iter_t* iter ARG_UNUSED, vo
...
+
+void php_phongo_bson_state_ctor(php_phongo_bson_state* state)
+{
+	state->field_path = php_phongo_field_path_alloc();
+}
+
+void php_phongo_bson_state_copy_ctor(php_phongo_bson_state* dst, php_phongo_bson_state* src)
+{
+	dst->map = src->map;
+	src->field_path->ref_count++;
```
jmikola: Wouldn't it be prudent to check that `src->field_path` is allocated before doing this?

src/bson.c
```diff
@@ -666,23 +870,30 @@ static bool php_phongo_bson_visit_document(const bson_iter_t* iter ARG_UNUSED, c
...
 static bool php_phongo_bson_visit_array(const bson_iter_t* iter ARG_UNUSED, const char* key, const bson_t* v_array, void* data) /* {{{ */
 {
-	zval*       retval = PHONGO_BSON_STATE_ZCHILD(data);
-	bson_iter_t child;
+	zval*                  retval = PHONGO_BSON_STATE_ZCHILD(data);
+	bson_iter_t            child;
+	php_phongo_bson_state* parent_state = (php_phongo_bson_state*) data;
 	TSRMLS_FETCH();
 
+	php_phongo_field_path_push(((php_phongo_bson_state*) data)->field_path, key);
```
jmikola: `((php_phongo_bson_state*) data)` can just be `parent_state`, as you did in `php_phongo_bson_visit_document()`.

src/bson.c
jmikola: In the interest of just ensuring unused pointers are set to `NULL`, let's do the follwowing before decrementing the current level, too:

```c
if (field_path->elements[field_path->current_level] != NULL) {
	field_path->elements[field_path->current_level] = NULL;
}
```

src/bson.c
```diff
@@ -41,50 +41,201 @@
...
+{
+	size_t i;
+
+	if (field_path->elements) {
+		for (i = 0; i < field_path->allocated_levels; i++) {
+			if (!field_path->elements[i]) {
+				continue;
+			}
+			field_path->elements[i] = NULL;
+		}
```
jmikola: I think we can do without this `for` loop, since we're going to free the entire `elements` array anyway. This may be a leftover from when you were actually freeing each element string.

src/bson.c
```diff
@@ -41,50 +41,201 @@
...
+{
+	if (level >= field_path->allocated_levels) {
+		size_t i;
+
+		field_path->allocated_levels = field_path->current_level + PHONGO_FIELD_PATH_EXPANSION;
+		field_path->elements         = erealloc(field_path->elements, sizeof(char**) * field_path->allocated_levels);
+
+		for (i = level; i < field_path->allocated_levels; i++) {
+			field_path->elements[i] = NULL;
+		}
```
jmikola: I was going to suggest using `memset()` here to remove another loop, but https://stackoverflow.com/a/2141373/162228 suggests why this is preferable. That said, I doubt PHP is running on any platforms where a NULL pointer is not zero bytes (famous last words).

src/MongoDB/Cursor.c
```diff
@@ -107,7 +107,9 @@ static void php_phongo_cursor_iterator_move_forward(zend_object_iterator* iter T
 	}
 
 	if (mongoc_cursor_next(cursor->cursor, &doc)) {
+		php_phongo_bson_state_ctor(&cursor->visitor_data);
 		php_phongo_bson_to_zval_ex(bson_get_data(doc), doc->len, &cursor->visitor_data);
+		php_phongo_bson_state_dtor(&cursor->visitor_data);
```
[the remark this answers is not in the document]
jmikola: Were you going to look into this before merging? If not, can you create another ticket in the epic to go back and look into it later?

PR #777, opened 2018-03-12
Update Exception expectations for manager-set-uri-options-002

kvwalker:
https://jira.mongodb.org/browse/PHPC-332

### DIFF
--- a/tests/manager/manager-set-uri-options-002.phpt
+++ b/tests/manager/manager-set-uri-options-002.phpt
@@ -26,10 +26,10 @@ $manager = new MongoDB\Driver\Manager(STANDALONE_SSL, $options, array("context" 
 
 $bulk = new MongoDB\Driver\BulkWrite;
 $bulk->insert(array("my" => "value"));
-throws(function() use ($manager, $bulk) {
+echo throws(function() use ($manager, $bulk) {
     $inserted = $manager->executeBulkWrite(NS, $bulk)->getInsertedCount();
     printf("Inserted incorrectly: %d\n", $inserted);
-}, "Exception");
+}, "MongoDB\Driver\Exception\ConnectionTimeoutException"), "\n";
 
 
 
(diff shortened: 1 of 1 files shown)


tests/manager/manager-set-uri-options-002.phpt
```diff
@@ -46,6 +46,6 @@ printf("Inserted: %d\n", $inserted);
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
-OK: Got Exception
+OK: Got MongoDB\Driver\Exception\ConnectionTimeoutException
```
jmikola: At a glance, I'm not sure why failure to specify `ssl=true` when connecting to a server requiring SSL manifests itself as a ConnectionTimeoutException instead of a generic ConnectionException.

It may be helpful to add the exception message here (adding `echo` before `throws()`, followed by `, "\n";` afterwards). [PHPC-332](https://jira.mongodb.org/browse/PHPC-332) is quite old, so I don't recall the original context of the issue.

tests/manager/manager-set-uri-options-002.phpt
```diff
@@ -47,5 +47,6 @@ printf("Inserted: %d\n", $inserted);
 <?php exit(0); ?>
 --EXPECTF--
 OK: Got MongoDB\Driver\Exception\ConnectionTimeoutException
+No suitable servers found (`serverSelectionTryOnce` set): [connection closed calling ismaster on '%s']
```
jmikola: Ah, I didn't realize and/or forgot that server selection failures are ConnectionTimeoutExceptions. That's clear from the logic in `phongo_exception_from_mongoc_domain()`, though.

I will take this moment to update the docs on http://php.net/manual/en/class.mongodb-driver-exception-connectiontimeoutexception.php.

jmikola: Updated: http://svn.php.net/viewvc?view=revision&revision=344498

PR #781, opened 2018-03-15
PHPC-578: Expose result document for failed commands

kvwalker:
https://jira.mongodb.org/browse/PHPC-578

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -853,7 +853,25 @@ int phongo_execute_command(mongoc_client_t* client, php_phongo_command_type_t ty
 			return false;
 	}
 	if (!result) {
-		phongo_throw_exception_from_bson_error_t(&error TSRMLS_CC);
+		if (error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN) {
+#if PHP_VERSION_ID >= 70000
+			zval zv;
+#else
+			zval* zv;
+#endif
+
+			zend_throw_exception(php_phongo_commandexception_ce, error.message, error.code TSRMLS_CC);
+			php_phongo_bson_to_zval(bson_get_data(&reply), reply.len, &zv);
+
+#if PHP_VERSION_ID >= 70000
+			phongo_add_exception_prop(ZEND_STRL("resultDocument"), &zv);
+#else
+			phongo_add_exception_prop(ZEND_STRL("resultDocument"), zv TSRMLS_CC);
+#endif
+			zval_ptr_dtor(&zv);
+		} else {
+			phongo_throw_exception_from_bson_error_t(&error TSRMLS_CC);
+		}
 		bson_destroy(&reply);
 		bson_destroy(&opts);
 		return false;
--- a/tests/manager/manager-executeWriteCommand_error-002.phpt
+++ b/tests/manager/manager-executeWriteCommand_error-002.phpt
@@ -1,0 +1,64 @@
+--TEST--
+MongoDB\Driver\Manager::executeWriteCommand() throws CommandException for invalid writeConcern
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS('REPLICASET'); CLEANUP(REPLICASET); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(REPLICASET);
+
+$command = new MongoDB\Driver\Command([
+    'findAndModify' => COLLECTION_NAME,
+    'query' => ['_id' => 'foo'],
+    'update' => ['foo' => ['bar']],
+    'upsert' => true,
+    'new' => true,
+]);
+
+try {
+    $manager->executeWriteCommand(DATABASE_NAME, $command, ['writeConcern' => new MongoDB\Driver\WriteConcern("undefined")]);
+} catch (MongoDB\Driver\Exception\CommandException $e) {
+    printf("%s(%d): %s\n", get_class($e), $e->getCode(), $e->getMessage());
+    var_dump($e->getResultDocument());
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+MongoDB\Driver\Exception\CommandException(79): Write Concern error: No write concern mode named 'undefined' found in replica set configuration
+object(stdClass)#%d (%d) {
+  ["lastErrorObject"]=>
+  object(stdClass)#%d (3) {
+    ["n"]=>
+    int(1)
+    ["updatedExisting"]=>
+    bool(false)
+    ["upserted"]=>
+    string(3) "foo"
+  }
+  ["value"]=>
+  object(stdClass)#%d (2) {
+    ["_id"]=>
+    string(3) "foo"
+    ["foo"]=>
+    array(1) {
+      [0]=>
+      string(3) "bar"
+    }
+  }
+  ["writeConcernError"]=>
+  object(stdClass)#%d (3) {
+    ["code"]=>
+    int(79)
+    ["codeName"]=>
+    string(23) "UnknownReplWriteConcern"
+    ["errmsg"]=>
+    string(74) "No write concern mode named 'undefined' found in replica set configuration"
+  }
+  ["ok"]=>
+  float(1)%A
+}
+===DONE===
--- a/tests/manager/manager-executeWriteCommand_error-003.phpt
+++ b/tests/manager/manager-executeWriteCommand_error-003.phpt
@@ -1,0 +1,41 @@
+--TEST--
+MongoDB\Driver\Manager::executeWriteCommand() throws CommandException for unsupported update operator
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS('STANDALONE'); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(STANDALONE);
+
+$command = new MongoDB\Driver\Command([
+    'findAndModify' => COLLECTION_NAME,
+    'query' => ['_id' => 'foo'],
+    'upsert' => true,
+    'new' => true,
+]);
+
+try {
+    $manager->executeWriteCommand(DATABASE_NAME, $command);
+} catch (MongoDB\Driver\Exception\CommandException $e) {
+    printf("%s(%d): %s\n", get_class($e), $e->getCode(), $e->getMessage());
+    var_dump($e->getResultDocument());
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+MongoDB\Driver\Exception\CommandException(9): Either an update or remove=true must be specified
+object(stdClass)#4 (4) {
+  ["ok"]=>
+  float(0)
+  ["errmsg"]=>
+  string(49) "Either an update or remove=true must be specified"
+  ["code"]=>
+  int(9)
+  ["codeName"]=>
+  string(13) "FailedToParse"
+}
+===DONE===


php_phongo.c
```diff
@@ -665,9 +665,11 @@ bool phongo_execute_bulk_write(mongoc_client_t* client, const char* namespace, p
 		if ((error.domain == MONGOC_ERROR_COMMAND && error.code != MONGOC_ERROR_COMMAND_INVALID_ARG) ||
 			error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN) {
 			phongo_throw_exception(PHONGO_ERROR_WRITE_FAILED TSRMLS_CC, "%s", error.message);
+			phongo_add_exception_prop(ZEND_STRL("resultDocument"), return_value TSRMLS_CC);
```
derickr: I don't think this is quite right, as we are already adding the ``writeResult`` to this exception in the next line (669). The issue is a bit sparse, and talks about a new getter instead of a property too. In any case, I would have expected that line 672 would use ``writeResult`` instead of ``resultDocument``.

jmikola: `phongo_execute_command()` basically needs a copy of this `if (!success)` conditional.

I'd start by only using a CommandException for `error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN`. Otherwise, we can fall through to `phongo_throw_exception_from_bson_error_t()`, as this area of the code does.

This means if  `phongo_execute_command()` is throwing an instance of ServerException, it's actually a CommandException.

php_phongo.c
```diff
@@ -843,7 +845,12 @@ int phongo_execute_command(mongoc_client_t* client, php_phongo_command_type_t ty
 			return false;
 	}
 	if (!result) {
-		phongo_throw_exception_from_bson_error_t(&error TSRMLS_CC);
+		if (error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN) {
+			phongo_add_exception_prop(ZEND_STRL("resultDocument"), (zval*) &reply TSRMLS_CC);
+			phongo_throw_exception(PHONGO_ERROR_COMMAND_FAILED TSRMLS_CC, "%s", error.message);
```
jmikola: `manager-executeWriteCommand_error-001.phpt` currently exists to test invalid options. Can you create new 002 and 003 tests to check for a CommandException and dump the result document?

You can consult `manager-executeBulkWrite_error-001.phpt` and `manager-executeBulkWrite_error-003.phpt` for prior art.

I'd suggest using `findAndModify` for both new tests. The `MONGOC_ERROR_WRITE_CONCERN` test can rely on a replica set and specify an outlandish value for `w`. For `MONGOC_ERROR_SERVER`, you can just use an unsupported update operator in the `findAndModify` command, which should trigger a server error.

jmikola: Using `phongo_throw_exception()` here is problematic, as it always sets the exception code to 0. This actually causes quite a bit of the test suite to be skipped because [`CLEANUP()`](https://github.com/mongodb/mongo-php-driver/blob/0a745ee8dee43ca7952afea6217678302e13b5bc/tests/utils/tools.php#L239) can no longer catch specific server-side exception codes when attempting to drop a nonexistent collection.

Now, we also can't use `phongo_throw_exception_from_bson_error_t()` here because all that knows is that the `bson_error_t`'s domain is either `MONGOC_ERROR_SERVER` or `MONGOC_ERROR_WRITE_CONCERN`, which would return a ServerException or RuntimeException, respectively.

Since we know the exact exception class we'd like to use here, and we have the error message and code (from `bson_error_t`), let's call `zend_throw_exception()` directly:

```c
zend_throw_exception(php_phongo_commandexception_ce, error.message, error.code TSRMLS_CC);
```

Note: `phongo_exception_from_mongoc_domain()` should probably be fixed to return a ServerException for the `MONGOC_ERROR_WRITE_CONCERN` domain, since libmongoc [documents](http://mongoc.org/libmongoc/current/errors.html) that its error code always originates from the server. I've opened [PHPC-1143](https://jira.mongodb.org/browse/PHPC-1143) to track that.

php_phongo.c
```diff
@@ -850,7 +852,12 @@ int phongo_execute_command(mongoc_client_t* client, php_phongo_command_type_t ty
 			return false;
 	}
 	if (!result) {
-		phongo_throw_exception_from_bson_error_t(&error TSRMLS_CC);
+		if (error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN) {
+			phongo_add_exception_prop(ZEND_STRL("resultDocument"), (zval*) &reply TSRMLS_CC);
```
jmikola: If you look at the definition for `phongo_add_exception_prop()` in `phongo_compat.c`, you'll notice that it is a no-op unless `EG(exception)` is non-zero (i.e. an exception has been thrown). This function should be called after we throw the exception on the next line. That explains why the `resultDocument` property was null in the tests.

Independent of that, we can't simply cast a `bson_t *` to a `zval *` (although it'd be great if BSON conversation was that easy). We'll need to use `php_phongo_bson_to_zval()` or `php_phongo_bson_to_zval_ex()` to convert the BSON document to a PHP zval. Since we're not supporting custom type maps here, and this isn't debug output where we'd convert everything to arrays, I'll suggest we use `php_phongo_bson_to_zval()` and its default type map. For prior art, see: https://github.com/mongodb/mongo-php-driver/blob/0a745ee8dee43ca7952afea6217678302e13b5bc/src/MongoDB/Command.c#L187-L192

To make life simpler, you should be able to reuse the `return_value` variable instead of conditionally declaring a zval for PHP 5.x and 7+. This is similar to what we do in `phongo_execute_bulk_write()`, where `return_value` (always a `zval *`, regardless of PHP version) is initialized as a WriteResult object and later assigned to an exception property.

Therefore, after using `return_value` as the destination for BSON conversion, it should then we suitable to use with `phongo_add_exception_prop()` directly.

tests/manager/manager-executeWriteCommand_error-003.phpt
```diff
@@ -0,0 +1,36 @@
...
+    'findAndModify' => COLLECTION_NAME,
+    'query' => ['_id' => 'foo'],
+    'upsert' => true,
+    'new' => true,
+]);
+
+try {
+    $manager->executeWriteCommand(DATABASE_NAME, $command);
+} catch (MongoDB\Driver\Exception\CommandException $e) {
+	printf("CommandException: %s\n", $e->getMessage());
```
jmikola: This indentation looks off. Make sure all test files use 4-space indents.

tests/manager/manager-executeWriteCommand_error-003.phpt
```diff
@@ -0,0 +1,36 @@
...
+
+    echo "\n===> ResultDocument\n";
```
jmikola: I think we can remove this. Simply `var_dump()`-ing the property after the `printf()` above should be fine, as that's all we're doing in this test.

The tests that use `printWriteResult()` are some of the oldest, so they often have funny formatting and makeshift titles for output.

tests/manager/manager-executeWriteCommand_error-003.phpt
```diff
@@ -0,0 +1,36 @@
+--TEST--
+MongoDB\Driver\Manager::executeWriteCommand() throws CommandException for unsupported update operator
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php NEEDS('STANDALONE'); CLEANUP(STANDALONE); ?>
```
jmikola: This shouldn't need `CLEANUP()`, since the write never completes successfully and no data will be left behind in the collection.

tests/manager/manager-executeWriteCommand_error-002.phpt
```diff
@@ -0,0 +1,39 @@
...
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(REPLICASET);
+
+$document = ['foo' => ['bar']];
+
+$command = new MongoDB\Driver\Command([
+    'findAndModify' => COLLECTION_NAME,
+    'query' => ['_id' => 'foo'],
+    'update' => $document,
```
jmikola: This can just be a literal `['foo' => ['bar']]` if we can remove the `$document` variable above.

tests/manager/manager-executeWriteCommand_error-002.phpt
```diff
@@ -0,0 +1,39 @@
...
+    'upsert' => true,
+    'new' => true,
+]);
+
+try {
+    $manager->executeWriteCommand(DATABASE_NAME, $command, ['writeConcern' => new MongoDB\Driver\WriteConcern("minority")]);
```
jmikola: Since this is an intentionally undefined mode, I'll suggest using `'undefined'` here. This is similar enough to the actual "majority" mode that it might confuse someone that comes across the test later.

tests/manager/manager-executeWriteCommand_error-002.phpt
```diff
@@ -0,0 +1,39 @@
...
+} catch (MongoDB\Driver\Exception\CommandException $e) {
+	printf("CommandException: %s\n", $e->getMessage());
```
jmikola: I'd suggest adding the exception code to this output, so we can test that `phongo_execute_command()` successfully sets that when throwing the exception.

I typically use `printf("%s(%d): %s\n", get_class($e), $e->getCode(), $e->getMessage())` in my own scripts, but you can certainly hard-code the short class name here.

tests/manager/manager-executeWriteCommand_error-002.phpt
```diff
@@ -0,0 +1,33 @@
...
+    'upsert' => true,
+    'new' => true,
+]);
+
+try {
+    $manager->executeWriteCommand(DATABASE_NAME, $command, ['writeConcern' => new MongoDB\Driver\WriteConcern("undefined")]);
+} catch (MongoDB\Driver\Exception\CommandException $e) {
+    printf("%s(%d): %s\n", get_class($e), $e->getCode(), $e->getMessage());
+    assert($e->getMessage() === "Write Concern error: " . $e->getResultDocument()->writeConcernError->errmsg);
+    assert($e->getCode() === $e->getResultDocument()->writeConcernError->code);
```
jmikola: I use "assert" loosely when referring to the PHPC tests; sorry if that came across as a suggestion for using [`assert()`](http://php.net/assert). If you read further down the page, you'll see that `assert()` behavior can vary wildly by INI settings, which makes it less than ideal for testing purposes. This is why `phpt` files (in PHP core and our extension) mostly rely on dumping and comparing output.

Rather than change `assert()` to `var_dump()` and add "bool(true)" to the expected output, let's just `var_dump($e->getResultDocument())` here as you do in the next test.

Since the `printf()` includes the code and message and `var_dump()` covers the result document itself, these comparisons are redundant. I don't think I was thinking about `var_dump($e->getResultDocument())` when I suggested assertions earlier today :)

tests/manager/manager-executeWriteCommand_error-003.phpt
```diff
@@ -0,0 +1,43 @@
...
+    'new' => true,
+]);
+
+try {
+    $manager->executeWriteCommand(DATABASE_NAME, $command);
+} catch (MongoDB\Driver\Exception\CommandException $e) {
+    printf("%s(%d): %s\n", get_class($e), $e->getCode(), $e->getMessage());
+    var_dump($e->getResultDocument());
+    assert($e->getMessage() === $e->getResultDocument()->errmsg);
+    assert($e->getCode() === $e->getResultDocument()->code);
```
jmikola: Likewise, I think both of these `assert()` calls can be removed.

php_phongo.c
```diff
@@ -850,7 +850,13 @@ int phongo_execute_command(mongoc_client_t* client, php_phongo_command_type_t ty
 			return false;
 	}
 	if (!result) {
-		phongo_throw_exception_from_bson_error_t(&error TSRMLS_CC);
+		if (error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN) {
+			zend_throw_exception(php_phongo_commandexception_ce, error.message, error.code TSRMLS_CC);
+			php_phongo_bson_to_zval(bson_get_data(&reply), reply.len, return_value);
```
jmikola: The third argument to `php_phongo_bson_to_zval()` is a `zval *` or `zval **` for PHP 7 and 5, respectively. See: https://github.com/mongodb/mongo-php-driver/blob/5bcfdcec81a3802f8f7c8e4c06ecdceb39ceb70f/src/bson.c#L986

```c
#if PHP_VERSION_ID >= 70000
	php_phongo_bson_to_zval(bson_get_data(&reply), reply.len, return_value);
#else
	php_phongo_bson_to_zval(bson_get_data(&reply), reply.len, &return_value);
#endif
```

That may do the trick.

jmikola: Ah, this actually leaks because `return_value` is already allocated, and `php_phongo_bson_to_zval_ex()` tries to allocate it again for PHP 5.x: https://github.com/mongodb/mongo-php-driver/blob/5bcfdcec81a3802f8f7c8e4c06ecdceb39ceb70f/src/bson.c#L1028

Solution would be to go back to the Command class' debug info handler. Declare either a `zval zv` or `zval *zv` for PHP 7 and 5, respectively, and then we can just call `**php_phongo_bson_to_zval(bson_get_data(&reply), reply.len, &zv)` for all versions.

php_phongo.c
```diff
@@ -852,8 +852,15 @@ int phongo_execute_command(mongoc_client_t* client, php_phongo_command_type_t ty
 	if (!result) {
 		if (error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN) {
 			zend_throw_exception(php_phongo_commandexception_ce, error.message, error.code TSRMLS_CC);
-			php_phongo_bson_to_zval(bson_get_data(&reply), reply.len, return_value);
-			phongo_add_exception_prop(ZEND_STRL("resultDocument"), return_value TSRMLS_CC);
+#if PHP_VERSION_ID >= 70000
+			zval zv;
```
jmikola: Declarations need to appear first in a scope. You may need to break this up as:

```c
#if PHP_VERSION_ID >= 70000
zval zv;
#else
zval *zv;
#endif

zend_throw_exception(php_phongo_commandexception_ce, error.message, error.code TSRMLS_CC);
php_phongo_bson_to_zval(bson_get_data(&reply), reply.len, &zv);

#if PHP_VERSION_ID >= 70000
phongo_add_exception_prop(ZEND_STRL("resultDocument"), &zv);
#else
phongo_add_exception_prop(ZEND_STRL("resultDocument"), zv TSRMLS_CC);
#endif
```

I omitted `TSRMLS_CC` for PHP 7 since it's defined as nothing (regardless of whether PHP 7 is compiled with thread safety). This is only need for PHP 5, so we can omit it in any PHP 7-specific code.

tests/manager/manager-executeWriteCommand_error-002.phpt
```diff
@@ -21,13 +21,73 @@ try {
...
+    var_dump($e->getResultDocument());
 }
 
 ?>
 ===DONE===
 <?php exit(0); ?>
---EXPECT--
+--EXPECTF--
 MongoDB\Driver\Exception\CommandException(79): Write Concern error: No write concern mode named 'undefined' found in replica set configuration
+object(stdClass)#13 (6) {
```
jmikola: The object identifiers should always use `%d` in `EXPECTF` output, since these can change. The property counter ("(6)" in this case) can usually be left as-is, but I'd also suggest replacing it with `(%d)` so that we can use a wildcard for some extra fields here.

tests/manager/manager-executeWriteCommand_error-002.phpt
```diff
@@ -21,13 +21,73 @@ try {
...
+    ["code"]=>
+    int(79)
+    ["codeName"]=>
+    string(23) "UnknownReplWriteConcern"
+    ["errmsg"]=>
+    string(74) "No write concern mode named 'undefined' found in replica set configuration"
+  }
+  ["ok"]=>
+  float(1)
+  ["operationTime"]=>
```
jmikola: Since `operationTime` and `$clusterTime` are 3.6+ fields, let's replace these with a wildcard pattern as is done in some other tests.

```
  ["ok"]=>
  float(1)%A
}
```

php_phongo.c
```diff
@@ -850,7 +850,24 @@ int phongo_execute_command(mongoc_client_t* client, php_phongo_command_type_t ty
...
+#endif
+
+			zend_throw_exception(php_phongo_commandexception_ce, error.message, error.code TSRMLS_CC);
+			php_phongo_bson_to_zval(bson_get_data(&reply), reply.len, &zv);
+
+#if PHP_VERSION_ID >= 70000
+			phongo_add_exception_prop(ZEND_STRL("resultDocument"), &zv);
+#else
+			phongo_add_exception_prop(ZEND_STRL("resultDocument"), zv TSRMLS_CC);
+#endif
```
jmikola: Testing this locally, I found that it was still leaking. A debug build of PHP typically reports simple leaks on its own, which is how I caught it. Running the tests with Valgrind confirmed the same.

I don't expect that Homebrew uses debug builds of PHP, so no worries there.

Ultimately, I believe this is due to PHP adding a reference when it updates the property of an object. When `phongo_add_exception_prop()` uses [`zend_update_property()`](http://lxr.jpauli.tech/source/xref/php-7.2/Zend/zend_API.c#3953). Unless the object is some internal class that is also using a custom handler, this will call [`zend_std_write_property()`](http://lxr.jpauli.tech/source/xref/php-7.2/Zend/zend_object_handlers.c#683). In that function, you might notice that the zval gets a reference added to it.

I came up with https://github.com/kvwalker/mongo-php-driver/pull/1 while debugging this, although in hingsight it is possible that a simple `zval_ptr_dtor(&zv)` before the end of this scope might have fixed the leak. If you determine it does, feel free to do that and ignore my PR.

As for why this differs from `phongo_execute_bulk_write()` and it's assignment of `BulkWriteException::$writeResult`, I believe there are a few reasons:

 * There, we always initialize `return_value` as a WriteResult object, since it's either going to be used as the actual return value or stuffed into an exception. The only time we don't initialize it is if we know there isn't an error and the user is ignoring the return value.
 * When `phongo_add_exception_prop()` assigns `return_value` to the exception property, it gets a reference added. At that point, I expect its refcount is 2; however, PHP likely decrements the refcount on `return_value` when returning from `executeBulkWrite()` back to userland when an exception is thrown since it knows that userland has no way to access the function's return value.
 * In the event of an exception, we're left with a `return_value` with refcount=1 assigned to the exception object. It is ultimately freed along with the exception itself.

php_phongo.c
```diff
@@ -673,8 +673,9 @@ bool phongo_execute_bulk_write(mongoc_client_t* client, const char* namespace, p
 
 	/* The Write failed */
 	if (!success) {
-		if (error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN) {
-			zend_throw_exception(php_phongo_bulkwriteexception_ce, error.message, error.code TSRMLS_CC);
+		if ((error.domain == MONGOC_ERROR_COMMAND && error.code != MONGOC_ERROR_COMMAND_INVALID_ARG) ||
+			error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN) {
+			phongo_throw_exception(PHONGO_ERROR_WRITE_FAILED TSRMLS_CC, "%s", error.message);
```
jmikola: The changes to these three lines look to have been introduced by https://github.com/mongodb/mongo-php-driver/pull/781/commits/373d2f26d1249c81caa67d84e7ba40a8de87dd2e. Should these be removed?

IIRC, @derickr caught this in one of his earlier reviews, when the PR was originally changing `phongo_execute_bulk_write()` instead of `phongo_execute_command()`.

As-is, this reverts changes from https://github.com/mongodb/mongo-php-driver/commit/110d13e4b3532354d128101b1bc701a3fadf6889 and https://github.com/mongodb/mongo-php-driver/commit/8e0e836913a6c826246585c43d64e6a8f82d444f.

jmikola: On a semi-related note: [PHPC-1086](https://jira.mongodb.org/browse/PHPC-1086).

PR #784, opened 2018-03-21
PHPC-1090: phongo_exception_from_mongoc_domain fixes

jmikola:
https://jira.mongodb.org/browse/PHPC-1090

The first commit addresses an issue I noticed with:

 * https://github.com/mongodb/mongo-php-driver/commit/1ad6d75b5c47da1095d7647896f66f673019a4b7#diff-c06c6e1c9374aecabcf544157f9d0c26R151
 * https://github.com/mongodb/mongo-php-driver/commit/395719ad47527fd239aef02cd73528521730dbac#diff-c06c6e1c9374aecabcf544157f9d0c26R152

The second commit addresses a concern that was in the description for [PHPC-1090](https://jira.mongodb.org/browse/PHPC-1090), which I didn't pick up on while reviewing #768.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -146,8 +146,6 @@ zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_d
 			return php_phongo_invalidargumentexception_ce;
 	}
 	switch (domain) {
-		case MONGOC_ERROR_COMMAND:
-			return php_phongo_commandexception_ce;
 		case MONGOC_ERROR_SERVER:
 			if (code == 50) {
 				return php_phongo_executiontimeoutexception_ce;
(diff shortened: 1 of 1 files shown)


php_phongo.c
```diff
@@ -136,28 +136,35 @@ zend_class_entry* phongo_exception_from_phongo_domain(php_phongo_error_domain_t
...
+
+	if (domain == MONGOC_ERROR_STREAM) {
+		if (code == MONGOC_ERROR_STREAM_SOCKET) {
+			return php_phongo_connectiontimeoutexception_ce;
+		}
+
+		return php_phongo_connectionexception_ce;
+	}
+
+	return php_phongo_runtimeexception_ce;
```
jmikola: I opted to remove the `switch` entirely since we are frequently matching domains and codes together. Also, a single `default:` would not have been sufficient, since some of the domains would have needed to `break;` out of the switch entirely and require this final `return`.

kvwalker: This looks good, but it seems to me that the codes are unique so is it not redundant to also check the domain for the cases where there is only one outcome, such as `domain == MONGOC_ERROR_COMMAND && code == MONGOC_ERROR_COMMAND_INVALID_ARG`?

jmikola: Since some codes may come from the server, it's theoretically possible that a server-side code might match one of libmongoc's constants. In that case, I wanted to be defensive so that we're only considering codes for the expected domain.

PR #785, opened 2018-03-22
PHPC-1144: Use server-side error code for BulkWriteException

kvwalker:
https://jira.mongodb.org/browse/PHPC-1144

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -673,8 +673,7 @@ bool phongo_execute_bulk_write(mongoc_client_t* client, const char* namespace, p
 
 	/* The Write failed */
 	if (!success) {
-		if ((error.domain == MONGOC_ERROR_COMMAND && error.code != MONGOC_ERROR_COMMAND_INVALID_ARG) ||
-			error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN) {
+		if (error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN) {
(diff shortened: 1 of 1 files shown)


php_phongo.c
```diff
@@ -671,7 +671,7 @@ bool phongo_execute_bulk_write(mongoc_client_t* client, const char* namespace, p
 	if (!success) {
 		if ((error.domain == MONGOC_ERROR_COMMAND && error.code != MONGOC_ERROR_COMMAND_INVALID_ARG) ||
 			error.domain == MONGOC_ERROR_SERVER || error.domain == MONGOC_ERROR_WRITE_CONCERN) {
-			phongo_throw_exception(PHONGO_ERROR_WRITE_FAILED TSRMLS_CC, "%s", error.message);
+			zend_throw_exception(php_phongo_bulkwriteexception_ce, error.message, error.code TSRMLS_CC);
```
jmikola: Independent of this, I think `(error.domain == MONGOC_ERROR_COMMAND && error.code != MONGOC_ERROR_COMMAND_INVALID_ARG)`  in the above condition is probably redundant now that we're using libmongoc error API v2. I think I missed this in https://github.com/mongodb/mongo-php-driver/commit/1ad6d75b5c47da1095d7647896f66f673019a4b7#diff-c06c6e1c9374aecabcf544157f9d0c26L705.

If we leave that in place, I suppose there's a small chance we might use a BulkWriteException for a MONGOC_ERROR_PROTOCOL_BAD_WIRE_VERSION code under the MONGOC_ERROR_COMMAND domain.

Could you address it in a second commit associated with PHPC-1090?

PR #793, opened 2018-04-04
PHPC-314: Implement type map syntax for documents within field paths

derickr:
First step of implementing the fieldPath syntax. It doesn't do wild cards yet, or typemaps for scalar types.

### DIFF
--- a/src/bson-encode.c
+++ b/src/bson-encode.c
@@ -365,6 +365,7 @@ try_again:
 				}
 
 				bson_append_array_begin(bson, key, key_len, &child);
+				php_phongo_field_path_write_type_at_current_level(field_path, PHONGO_FIELD_PATH_ITEM_ARRAY);
 				field_path->current_level++;
 				php_phongo_zval_to_bson_internal(entry, field_path, flags, &child, NULL TSRMLS_CC);
 				field_path->current_level--;
@@ -391,6 +392,7 @@ try_again:
 				ZEND_HASH_INC_APPLY_COUNT(tmp_ht);
 			}
 
+			php_phongo_field_path_write_type_at_current_level(field_path, PHONGO_FIELD_PATH_ITEM_DOCUMENT);
 			field_path->current_level++;
 			php_phongo_bson_append_object(bson, field_path, flags, key, key_len, entry TSRMLS_CC);
 			field_path->current_level--;
--- a/src/bson-encode.c
+++ b/src/bson-encode.c
@@ -366,9 +366,9 @@ try_again:
 
 				bson_append_array_begin(bson, key, key_len, &child);
 				php_phongo_field_path_write_type_at_current_level(field_path, PHONGO_FIELD_PATH_ITEM_ARRAY);
-				field_path->current_level++;
+				field_path->size++;
 				php_phongo_zval_to_bson_internal(entry, field_path, flags, &child, NULL TSRMLS_CC);
-				field_path->current_level--;
+				field_path->size--;
 				bson_append_array_end(bson, &child);
 
 				if (tmp_ht && ZEND_HASH_APPLY_PROTECTION(tmp_ht)) {
@@ -393,9 +393,9 @@ try_again:
 			}
 
 			php_phongo_field_path_write_type_at_current_level(field_path, PHONGO_FIELD_PATH_ITEM_DOCUMENT);
-			field_path->current_level++;
+			field_path->size++;
 			php_phongo_bson_append_object(bson, field_path, flags, key, key_len, entry TSRMLS_CC);
-			field_path->current_level--;
+			field_path->size--;
 
 			if (tmp_ht && ZEND_HASH_APPLY_PROTECTION(tmp_ht)) {
 				ZEND_HASH_DEC_APPLY_COUNT(tmp_ht);
@@ -650,7 +650,7 @@ cleanup:
  * will be used. */
 void php_phongo_zval_to_bson(zval* data, php_phongo_bson_flags_t flags, bson_t* bson, bson_t** bson_out TSRMLS_DC) /* {{{ */
 {
-	php_phongo_field_path* field_path = php_phongo_field_path_alloc();
+	php_phongo_field_path* field_path = php_phongo_field_path_alloc(false);
 
 	php_phongo_zval_to_bson_internal(data, field_path, flags, bson, bson_out TSRMLS_CC);
 
--- a/src/bson.c
+++ b/src/bson.c
@@ -29,6 +29,8 @@
 #include "phongo_compat.h"
 #include "php_array_api.h"
 
+#define DEBUG 0
+
 #undef MONGOC_LOG_DOMAIN
 #define MONGOC_LOG_DOMAIN "PHONGO-BSON"
 
@@ -62,7 +64,7 @@ char* php_phongo_field_path_as_string(php_phongo_field_path* field_path)
 		return estrdup("");
 	}
 
-	for (i = 0; i <= field_path->current_level; i++) {
+	for (i = 0; i <= field_path->size; i++) {
 		if (!field_path->elements[i]) {
 			continue;
 		}
@@ -72,7 +74,7 @@ char* php_phongo_field_path_as_string(php_phongo_field_path* field_path)
 	path = emalloc(length);
 	ptr  = path;
 
-	for (i = 0; i <= field_path->current_level; i++) {
+	for (i = 0; i <= field_path->size; i++) {
 		if (!field_path->elements[i]) {
 			continue;
 		}
@@ -86,16 +88,24 @@ char* php_phongo_field_path_as_string(php_phongo_field_path* field_path)
 	return path;
 }
 
-php_phongo_field_path* php_phongo_field_path_alloc(void)
+php_phongo_field_path* php_phongo_field_path_alloc(bool free_elements)
 {
 	php_phongo_field_path* tmp = ecalloc(1, sizeof(php_phongo_field_path));
 	tmp->ref_count             = 1;
+	tmp->free_elements         = free_elements;
 
 	return tmp;
 }
 
 void php_phongo_field_path_free(php_phongo_field_path* field_path)
 {
+	if (field_path->free_elements) {
+		size_t i;
+
+		for (i = 0; i < field_path->size; i++) {
+			efree(field_path->elements[i]);
+		}
+	}
 	if (field_path->elements) {
 		efree(field_path->elements);
 	}
@@ -107,14 +117,14 @@ void php_phongo_field_path_free(php_phongo_field_path* field_path)
 
 static void php_phongo_field_path_ensure_allocation(php_phongo_field_path* field_path, size_t level)
 {
-	if (level >= field_path->allocated_levels) {
+	if (level >= field_path->allocated) {
 		size_t i;
 
-		field_path->allocated_levels = field_path->current_level + PHONGO_FIELD_PATH_EXPANSION;
-		field_path->elements         = erealloc(field_path->elements, sizeof(char**) * field_path->allocated_levels);
-		field_path->element_types    = erealloc(field_path->element_types, sizeof(php_phongo_bson_field_path_item_types*) * field_path->allocated_levels);
+		field_path->allocated     = field_path->size + PHONGO_FIELD_PATH_EXPANSION;
+		field_path->elements      = erealloc(field_path->elements, sizeof(char**) * field_path->allocated);
+		field_path->element_types = erealloc(field_path->element_types, sizeof(php_phongo_bson_field_path_item_types*) * field_path->allocated);
 
-		for (i = level; i < field_path->allocated_levels; i++) {
+		for (i = level; i < field_path->allocated; i++) {
 			field_path->elements[i]      = NULL;
 			field_path->element_types[i] = PHONGO_FIELD_PATH_ITEM_NONE;
 		}
@@ -123,16 +133,20 @@ static void php_phongo_field_path_ensure_allocation(php_phongo_field_path* field
 
 void php_phongo_field_path_write_item_at_current_level(php_phongo_field_path* field_path, const char* element)
 {
-	php_phongo_field_path_ensure_allocation(field_path, field_path->current_level);
+	php_phongo_field_path_ensure_allocation(field_path, field_path->size);
 
-	field_path->elements[field_path->current_level] = element;
+	if (field_path->free_elements) {
+		field_path->elements[field_path->size] = estrdup(element);
+	} else {
+		field_path->elements[field_path->size] = (char*) element;
+	}
 }
 
 void php_phongo_field_path_write_type_at_current_level(php_phongo_field_path* field_path, php_phongo_bson_field_path_item_types element_type)
 {
-	php_phongo_field_path_ensure_allocation(field_path, field_path->current_level);
+	php_phongo_field_path_ensure_allocation(field_path, field_path->size);
 
-	field_path->element_types[field_path->current_level] = element_type;
+	field_path->element_types[field_path->size] = element_type;
 }
 
 bool php_phongo_field_path_push(php_phongo_field_path* field_path, const char* element, php_phongo_bson_field_path_item_types element_type)
@@ -140,20 +154,20 @@ bool php_phongo_field_path_push(php_phongo_field_path* field_path, const char* e
 	php_phongo_field_path_write_item_at_current_level(field_path, element);
 	php_phongo_field_path_write_type_at_current_level(field_path, element_type);
 
-	field_path->current_level++;
+	field_path->size++;
 
 	return true;
 }
 
 bool php_phongo_field_path_pop(php_phongo_field_path* field_path)
 {
-	field_path->elements[field_path->current_level]      = NULL;
-	field_path->element_types[field_path->current_level] = PHONGO_FIELD_PATH_ITEM_NONE;
+	field_path->elements[field_path->size]      = NULL;
+	field_path->element_types[field_path->size] = PHONGO_FIELD_PATH_ITEM_NONE;
 
-	field_path->current_level--;
+	field_path->size--;
 
-	field_path->elements[field_path->current_level]      = NULL;
-	field_path->element_types[field_path->current_level] = PHONGO_FIELD_PATH_ITEM_NONE;
+	field_path->elements[field_path->size]      = NULL;
+	field_path->element_types[field_path->size] = PHONGO_FIELD_PATH_ITEM_NONE;
 
 	return true;
 }
@@ -165,7 +179,7 @@ inline static bool php_phongo_bson_state_is_initialized(php_phongo_bson_state* s
 
 void php_phongo_bson_state_ctor(php_phongo_bson_state* state)
 {
-	state->field_path = php_phongo_field_path_alloc();
+	state->field_path = php_phongo_field_path_alloc(false);
 }
 
 void php_phongo_bson_state_copy_ctor(php_phongo_bson_state* dst, php_phongo_bson_state* src)
@@ -782,6 +796,55 @@ static const bson_visitor_t php_bson_visitors = {
 	{ NULL }
 };
 
+static bool map_element_matches_field_path(php_phongo_field_path_map_element* map_element, php_phongo_field_path* current)
+{
+	size_t i;
+
+	if (map_element->entry->size != current->size) {
+		return false;
+	}
+	for (i = 0; i < current->size; i++) {
+		if (strcmp(map_element->entry->elements[i], current->elements[i]) != 0) {
+			return false;
+		}
+	}
+	return true;
+}
+
+static php_phongo_field_path_map_element* map_find_field_path_entry(php_phongo_bson_state* state)
+{
+	size_t i;
+
+	/* Loop over all field path mappings, and for each, try to see whether it matches the current path */
+	for (i = 0; i < state->map.field_paths.size; i++) {
+		if (map_element_matches_field_path(state->map.field_paths.map[i], state->field_path)) {
+			return state->map.field_paths.map[i];
+		}
+	}
+	return NULL;
+}
+
+static void php_phongo_handle_field_path_entry_for_compound_type(php_phongo_bson_state* state, php_phongo_bson_typemap_types* type, zend_class_entry** ce)
+{
+	php_phongo_field_path_map_element* entry = map_find_field_path_entry(state);
+
+	if (entry) {
+		switch (entry->node_type) {
+			case PHONGO_TYPEMAP_NATIVE_ARRAY:
+			case PHONGO_TYPEMAP_NATIVE_OBJECT:
+				*type = entry->node_type;
+				break;
+			case PHONGO_TYPEMAP_CLASS:
+				*type = entry->node_type;
+				*ce   = entry->node_ce;
+				break;
+			default:
+				/* Do nothing - pacify compiler */
+				break;
+		}
+	}
+}
+
 static bool php_phongo_bson_visit_document(const bson_iter_t* iter ARG_UNUSED, const char* key, const bson_t* v_document, void* data) /* {{{ */
 {
 	zval*                  retval = PHONGO_BSON_STATE_ZCHILD(data);
@@ -804,6 +867,10 @@ static bool php_phongo_bson_visit_document(const bson_iter_t* iter ARG_UNUSED, c
 #endif
 
 		if (!bson_iter_visit_all(&child, &php_bson_visitors, &state) && !child.err_off) {
+			/* Check for entries in the fieldPath type map key, and use them to
+			 * override the default ones for this type */
+			php_phongo_handle_field_path_entry_for_compound_type(&state, &state.map.document_type, &state.map.document);
+
 			/* If php_phongo_bson_visit_binary() finds an ODM class, it should
 			 * supersede a default type map and named document class. */
 			if (state.odm && state.map.document_type == PHONGO_TYPEMAP_NONE) {
@@ -917,6 +984,9 @@ static bool php_phongo_bson_visit_array(const bson_iter_t* iter ARG_UNUSED, cons
 #endif
 
 		if (!bson_iter_visit_all(&child, &php_bson_visitors, &state) && !child.err_off) {
+			/* Check for entries in the fieldPath type map key, and use them to
+			 * override the default ones for this type */
+			php_phongo_handle_field_path_entry_for_compound_type(&state, &state.map.array_type, &state.map.array);
 
 			switch (state.map.array_type) {
 				case PHONGO_TYPEMAP_CLASS: {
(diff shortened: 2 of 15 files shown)


src/bson-encode.c
```diff
@@ -648,7 +650,7 @@ static void php_phongo_zval_to_bson_internal(zval* data, php_phongo_field_path*
  * will be used. */
 void php_phongo_zval_to_bson(zval* data, php_phongo_bson_flags_t flags, bson_t* bson, bson_t** bson_out TSRMLS_DC) /* {{{ */
 {
-	php_phongo_field_path* field_path = php_phongo_field_path_alloc();
+	php_phongo_field_path* field_path = php_phongo_field_path_alloc(false);
```
jmikola: Is `php_phongo_field_path* field_path` something that was introduced when you improved BSON exception messages to report the entire field path?

src/bson-encode.c
```diff
@@ -365,9 +365,10 @@ static void php_phongo_bson_append(bson_t* bson, php_phongo_field_path* field_pa
 				}
 
 				bson_append_array_begin(bson, key, key_len, &child);
-				field_path->current_level++;
+				php_phongo_field_path_write_type_at_current_level(field_path, PHONGO_FIELD_PATH_ITEM_ARRAY);
+				field_path->size++;
 				php_phongo_zval_to_bson_internal(entry, field_path, flags, &child, NULL TSRMLS_CC);
-				field_path->current_level--;
+				field_path->size--;
```
jmikola: Would it make sense to have macros or inline functions for incrementing and decrementing `size`?

Does `php_phongo_field_path_write_type_at_current_level()` ensure that there is always sufficient buffer space to increment `size`?

src/bson.c
```diff
@@ -86,59 +88,86 @@ char* php_phongo_field_path_as_string(php_phongo_field_path* field_path)
...
+	if (field_path->element_types) {
+		efree(field_path->element_types);
+	}
 	efree(field_path);
 }
 
 static void php_phongo_field_path_ensure_allocation(php_phongo_field_path* field_path, size_t level)
 {
-	if (level >= field_path->allocated_levels) {
+	if (level >= field_path->allocated) {
```
jmikola: For readability, would `allocated_size` make more sense?

src/bson.c
```diff
@@ -86,59 +88,86 @@ char* php_phongo_field_path_as_string(php_phongo_field_path* field_path)
...
 		size_t i;
 
-		field_path->allocated_levels = field_path->current_level + PHONGO_FIELD_PATH_EXPANSION;
-		field_path->elements         = erealloc(field_path->elements, sizeof(char**) * field_path->allocated_levels);
+		field_path->allocated     = field_path->size + PHONGO_FIELD_PATH_EXPANSION;
```
jmikola: Any reason we don't want to grow the allocation by a power-of-two here?

src/bson.c
```diff
@@ -86,59 +88,86 @@ char* php_phongo_field_path_as_string(php_phongo_field_path* field_path)
...
+		field_path->elements      = erealloc(field_path->elements, sizeof(char**) * field_path->allocated);
```
jmikola: I realize that `sizeof(char *)` is equal to `sizeof(char **)`, but if `elements` is declared as a `char**` (array of strings) then shouldn't we use `sizeof(char *)` here?

src/bson.c
```diff
@@ -86,59 +88,86 @@ char* php_phongo_field_path_as_string(php_phongo_field_path* field_path)
...
+		field_path->element_types = erealloc(field_path->element_types, sizeof(php_phongo_bson_field_path_item_types*) * field_path->allocated);
```
jmikola: Similarly, shouldn't this be `sizeof(php_phongo_bson_field_path_item_types)` as well, since `element_types` points to an array of enums?

src/bson.c
```diff
@@ -86,59 +88,86 @@ char* php_phongo_field_path_as_string(php_phongo_field_path* field_path)
...
 		}
 	}
 }
 
 void php_phongo_field_path_write_item_at_current_level(php_phongo_field_path* field_path, const char* element)
 {
-	php_phongo_field_path_ensure_allocation(field_path, field_path->current_level);
+	php_phongo_field_path_ensure_allocation(field_path, field_path->size);
+
+	if (field_path->free_elements) {
```
jmikola: If `free_elements` isn't just used for free logic, I might suggest calling it `copy_elements` instead. That makes it clearer that this function is doing to `estrdup()` and likewise makes sense when you check it to free each element later on.

As-is, I would expect the caller to do `estrdup()` on its own.

src/bson-encode.c
jmikola: Yes it does: https://github.com/mongodb/mongo-php-driver/pull/793/files#diff-f7d29a680148f52d6601f59ed787f577R147

src/bson.c
```diff
@@ -767,14 +796,66 @@ static const bson_visitor_t php_bson_visitors = {
 	{ NULL }
 };
 
+static bool map_element_matches_field_path(php_phongo_field_path_map_element* map_element, php_phongo_field_path* current)
```
jmikola: I realize these are static, but shouldn't we be prefixing with `php_phongo` anyway for good measure?

src/bson.c
```diff
@@ -767,14 +796,66 @@ static const bson_visitor_t php_bson_visitors = {
...
+{
+	size_t i;
+
+	if (map_element->entry->size != current->size) {
+		return false;
+	}
+	for (i = 0; i < current->size; i++) {
+		if (strcmp(map_element->entry->elements[i], "$") == 0) {
+			continue;
```
jmikola: I imagine this is responsible for my earlier questions about wildcards have less precedence than an exact field name, despite appearing earlier in the list of field paths.

I do think it's more intuitive to follow first-match-wins logic here. If we match a wildcard or an exact field in this `for` iteration, return true immediately. If nothing matches, we can return `false` after iteration has succeeded.

I realize that the current implementation can return early. If you're concerned with overhead, we can always benchmark and optimize down the line.

src/bson.c
```diff
@@ -767,14 +796,66 @@ static const bson_visitor_t php_bson_visitors = {
...
+	return true;
+}
+
+static php_phongo_field_path_map_element* map_find_field_path_entry(php_phongo_bson_state* state)
+{
+	size_t i;
+
+	/* Loop over all field path mappings, and for each, try to see whether it matches the current path */
+	for (i = 0; i < state->map.field_paths.size; i++) {
+		if (map_element_matches_field_path(state->map.field_paths.map[i], state->field_path)) {
```
jmikola: `map_element_matches_field_path()` may be a good candidate for an `inline` hint, where supported.

src/bson.c
```diff
@@ -767,14 +796,66 @@ static const bson_visitor_t php_bson_visitors = {
...
+			case PHONGO_TYPEMAP_NATIVE_OBJECT:
+				*type = entry->node_type;
+				break;
+			case PHONGO_TYPEMAP_CLASS:
+				*type = entry->node_type;
+				*ce   = entry->node_ce;
+				break;
+			default:
+				/* Do nothing - pacify compiler */
+				break;
```
jmikola: Is `php_phongo_bson_state state` always initialized with `PHONGO_BSON_STATE_INITIALIZER` (and layered with the default type map) before `php_phongo_handle_field_path_entry_for_compound_type()` is called at each level? I just want to make sure there is no concern with this "do nothing" behavior inadvertently leaving a previously applied field path applied to the state.

Would it make sense to revert to the default type map before calling this function?

src/bson.c
```diff
@@ -1201,6 +1289,195 @@ static bool php_phongo_bson_state_parse_type(zval* options, const char* name, ph
 	return retval;
 } /* }}} */
 
+static void field_path_map_element_set_info(php_phongo_field_path_map_element* element, php_phongo_bson_typemap_types type, zend_class_entry* ce TSRMLS_DC)
```
jmikola: `TSRMLS_DC` shouldn't be necessary here.

src/bson.c
```diff
@@ -1201,6 +1289,195 @@ static bool php_phongo_bson_state_parse_type(zval* options, const char* name, ph
...
+	element->node_ce   = ce;
+}
+
+static void map_add_field_path_element(php_phongo_bson_typemap* map, php_phongo_field_path_map_element* element)
+{
+	/* Make sure we have allocated enough */
+	if (map->field_paths.allocated < map->field_paths.size + 1) {
+		map->field_paths.allocated += PHONGO_FIELD_PATH_EXPANSION;
+		map->field_paths.map = erealloc(map->field_paths.map, sizeof(php_phongo_field_path_map_element) * map->field_paths.allocated);
+	}
```
jmikola: Should we have an "ensure" function for this, or is this really the only place we need to increase allocation?

src/bson.c
```diff
@@ -1201,6 +1289,195 @@ static bool php_phongo_bson_state_parse_type(zval* options, const char* name, ph
...
+
+	map->field_paths.map[map->field_paths.size] = element;
+	map->field_paths.size++;
+}
+
+static php_phongo_field_path_map_element* field_path_map_element_alloc(void)
+{
+	php_phongo_field_path_map_element* tmp = ecalloc(1, sizeof(php_phongo_field_path_map_element));
+
+	tmp->entry = php_phongo_field_path_alloc(true);
```
jmikola: In this case, the element names should always be `estrdup()`-ed?

src/bson.c
```diff
@@ -1201,6 +1289,195 @@ static bool php_phongo_bson_state_parse_type(zval* options, const char* name, ph
...
+
+	return tmp;
+}
+
+bool php_phongo_bson_state_add_field_path(php_phongo_bson_typemap* map, char* field_path_original, php_phongo_bson_typemap_types type, zend_class_entry* ce TSRMLS_DC)
```
jmikola: If `field_path_map_element_set_info()` doesn't need `TSRMLS_DC`, does this?

src/bson.c
```diff
@@ -1201,6 +1289,195 @@ static bool php_phongo_bson_state_parse_type(zval* options, const char* name, ph
...
+	char*                              saveptr;
+	char*                              field_path_str = estrdup(field_path_original);
+	const char*                        element;
+	php_phongo_field_path_map_element* field_path_map_element;
+
+	field_path_map_element = field_path_map_element_alloc();
+
+	/* Loop over all the segments. A segment is delimited by a "." */
+	element = php_strtok_r(field_path_str, ".", &saveptr);
+	while (element) {
```
jmikola: Is there a case where we don't enter the `while` loop on the first iteration?

It may be good to have error tests for field paths such as `"..."`, `".foo"`, and `""` so we can test how that is handled.

src/bson.c
```diff
@@ -1201,6 +1289,195 @@ static bool php_phongo_bson_state_parse_type(zval* options, const char* name, ph
...
+			if (!string_key) {
+				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "The 'fieldPaths' element is not an associative array");
+				return false;
+			}
+
+			if (!php_phongo_bson_state_parse_type(fieldpaths, ZSTR_VAL(string_key), &map_type, &map_ce TSRMLS_CC)) {
+				return false;
+			}
+
+			if (!php_phongo_bson_state_add_field_path(map, ZSTR_VAL(string_key), map_type, map_ce TSRMLS_CC)) {
```
jmikola: `php_phongo_bson_state_add_field_path()` always returns true. In the event a failure case is introduced, I suppose we'll want to ensure it throws an exception. Correct?

Is that why you're preemptively passing in `TSRMLS_CC` here (consistency with `php_phongo_bson_state_parse_type()`)?

src/bson-encode.c
derickr: Yeah, it does. I can make it a macro, but as we're only doing this twice, I didn't see much use of it. Happy to do so if you would prefer it.

src/bson-encode.c
derickr: Yes.

src/bson.c
derickr: Yes, I think so. I'll change that.

src/bson.c
derickr: Good catch, and yes. Will fix that.

src/bson.c
derickr: No, but as this is the amount of nested levels, I think we hardly even hit the first size of 8. Happy to change that if you'd like, but IMO, it is not going to be much of a difference.

src/bson.c
derickr: Probably - will have a look and update if needed.

src/bson.c
derickr: I agree, and I had already updated this locally. I had changed it to "owns_elements" though — let me know which one you think is better.

src/bson.c
derickr: This is the one that I was pondering about a little, I think I would lean to "no". It makes it much clearer which ones are statics, and we also benefit from a shorter name, and much easier auto-completion in the editors.

src/bson.c
derickr: It currently implements the "first-match-wins" logic. If we find a wild card, we instantly skip to the next *level*, and not to the next element in the field path that we're trying to match against.

derickr: I did actually run some benchmarks, as you can see in ``typemap-008.phpt``. The ``25000`` is there for a reason.

src/bson.c
derickr: I'll change that. But it would surprise me if compilers didn't already do that on their own.

src/bson.c
derickr: Indeed - will remove it.

src/bson.c
derickr: Yes, it is. We create a new state for each nested level, which is assigned with ``php_phongo_bson_state state = PHONGO_BSON_STATE_INITIALIZER;`` in ``php_phongo_bson_visit_document`` and ``php_phongo_bson_visit_array``: https://github.com/mongodb/mongo-php-driver/pull/793/files#diff-f7d29a680148f52d6601f59ed787f577R861

src/bson.c
derickr: It's the only place, and we always add to the end. Which is not the case when we build the *current* conversion field path for error messages and mapping, as there we leave nested sections as often as we go into them. I'll add a comment.

src/bson.c
derickr: Yes, and they will be through https://github.com/mongodb/mongo-php-driver/pull/793/files#diff-f7d29a680148f52d6601f59ed787f577R139  if ``true`` is used.

src/bson.c
derickr: Doesn't look like it, will remove it.

src/bson.c
derickr: I don't think we won't ever enter, but I will add test cases. I know that ``strtok`` handles ``...`` as ``.`` for example.

src/bson.c
derickr: I'd wish to say that I added ``TSRMLS_CC`` because of that, but that's not true :-) But yes, perhaps we should leave it as long as the strtok loop can create failure notifications with the three cases that you suggested.

tests/bson/typemap-002.phpt
```diff
@@ -8,14 +8,6 @@ MongoDB\Driver\Cursor::setTypeMap(): Setting using type "object"
...
 require_once __DIR__ . "/../utils/basic.inc";
 
-class MyArrayObject extends ArrayObject implements MongoDB\BSON\Unserializable
-{
-    function bsonUnserialize(array $data)
-    {
-        parent::__construct($data);
-    }
-}
-
```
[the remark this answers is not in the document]
derickr: Yes.

tests/bson/typemap-003.phpt
```diff
@@ -0,0 +1,81 @@
...
+
+$document1 = [
+    '_id' => 1,
+    'array' => [1, 2, 3],
+    'object' => ['string' => ['sleutels', 'keys'] ]
+];
+$document2 = [
+    '_id' => 2,
+    'array' => [4, 5, 6],
+    'object' => ['associative' => ['elementen', 'elements' ]]
```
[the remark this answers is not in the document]
derickr: This test was originally to see if we could set the typemap twice without leaking or memory issues.

But I added a functional bit to test that a bit better, and might ended up having just a test for PROPS_AS_ARRAY. Because of the memory issue/doubble setting, I would like to keep it though.

tests/bson/typemap-003.phpt
```diff
@@ -0,0 +1,81 @@
...
+$cursor->setTypeMap($typemap1);
+$cursor->setTypeMap($typemap2);
+    
+$documents = $cursor->toArray();
+
+var_dump($documents[0] instanceof stdClass);
+var_dump(is_array($documents[0]->array));
+var_dump($documents[0]->object instanceof MyArrayObject);
+var_dump(is_array($documents[0]->object['string']));
+var_dump(is_array($documents[0]->object->string));
```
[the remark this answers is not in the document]
derickr: See above.

tests/bson/typemap-004.phpt
```diff
@@ -0,0 +1,82 @@
...
+$manager->executeBulkWrite(NS, $bulk);
+
+function fetch($manager, $typemap = []) {
+    $cursor = $manager->executeQuery(NS, new MongoDB\Driver\Query([]));
+    if ($typemap) {
+        $cursor->setTypeMap($typemap);
+    }
+
+    $documents = $cursor->toArray();
+    return $documents;
```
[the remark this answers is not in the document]
derickr: Sure, will fix that. I probably had a debug statement in between at some point.... which might still be useful in case an IDE is used though?

tests/bson/typemap-005.phpt
```diff
@@ -0,0 +1,103 @@
...
+    $documents = $cursor->toArray();
+    return $documents;
+}
+
+
+echo "Default\n";
+$documents = fetch($manager);
+var_dump($documents[0] instanceof stdClass);
+var_dump(is_array($documents[0]->array0));
+var_dump(is_object($documents[0]->array1));
```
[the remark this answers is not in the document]
derickr: Perhaps, but I'm just being thorough.

tests/bson/typemap-005.phpt
```diff
@@ -0,0 +1,103 @@
...
+
+echo "\nSetting 'array1.1' path to 'MyArrayObject'\n";
+$documents = fetch($manager, ["fieldPaths" => [
+    'array1.1'  => "MyArrayObject",
+]]);
+var_dump($documents[0] instanceof stdClass);
+var_dump(is_object($documents[0]->array1));
+var_dump($documents[0]->array1 instanceof stdClass);
+$a = ((array) $documents[0]->array1);
+var_dump($a[1] instanceof MyArrayObject);
```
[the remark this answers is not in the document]
derickr: Didn't work in < PHP 7.2 - it's what I had at first: https://travis-ci.org/mongodb/mongo-php-driver/jobs/362202660#L1326

tests/bson/typemap-006.phpt
```diff
@@ -0,0 +1,141 @@
+--TEST--
+MongoDB\Driver\Cursor::setTypeMap(): Setting fieldPath typemaps for compound types with wild card keys
```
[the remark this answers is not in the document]
derickr: Or it can be two ;-) Will fix it.

tests/bson/typemap-006.phpt
```diff
@@ -0,0 +1,141 @@
...
+
+echo "\nSetting 'array.$' to 'MyWildCardArrayObject' and 'array.1' path to 'MyArrayObject'\n";
+$documents = fetch($manager, ["fieldPaths" => [
+    'array.$' => "MyWildCardArrayObject",
+    'array.1' => "MyArrayObject"
+]]);
+var_dump($documents[0] instanceof stdClass);
+var_dump(is_array($documents[0]->array));
+var_dump($documents[0]->array[0] instanceof MyWildCardArrayObject);
+var_dump($documents[0]->array[1] instanceof MyArrayObject);
```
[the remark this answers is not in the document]
derickr: And it should... I'm not sure why this test shouldn't fail. I'll re-check.

tests/bson/typemap-006.phpt
```diff
@@ -0,0 +1,141 @@
...
+
+echo "\nSetting 'object.$' to 'MyWildCardArrayObject' and 'object.one' path to 'MyArrayObject'\n";
+$documents = fetch($manager, ["fieldPaths" => [
+    'object.$' => "MyWildCardArrayObject",
+    'object.one' => "MyArrayObject"
+]]);
+var_dump($documents[0] instanceof stdClass);
+var_dump(is_object($documents[0]->object));
+var_dump($documents[0]->object->one instanceof MyArrayObject);
+var_dump($documents[0]->object->two instanceof MyWildCardArrayObject);
```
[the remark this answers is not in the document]
derickr: Will check this too.

tests/bson/typemap-007.phpt
```diff
@@ -0,0 +1,133 @@
...
+
+class MyArrayObject extends ArrayObject implements MongoDB\BSON\Unserializable
+{
+    function bsonUnserialize(array $data)
+    {
+        parent::__construct($data, ArrayObject::ARRAY_AS_PROPS);
+    }
+}
+
+class MyWildCardArrayObject extends MyArrayObject {};
```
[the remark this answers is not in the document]
derickr: PHP is case insensitive.

tests/bson/typemap-007.phpt
```diff
@@ -0,0 +1,133 @@
+--TEST--
+MongoDB\Driver\Cursor::setTypeMap(): Setting fieldPath typemaps for compound types with wild card keys (nested)
```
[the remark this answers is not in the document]
derickr: They're all testing the typemap functionality, so I prefer them all in the same place.

tests/bson/typemap-007.phpt
```diff
@@ -0,0 +1,133 @@
...
+var_dump($documents[0]->object->parent1->child1 instanceof MyWildCardArrayObject);
+var_dump($documents[0]->object->parent1->child2 instanceof MyWildCardArrayObject);
+var_dump($documents[0]->object->parent2 instanceof stdClass);
+var_dump($documents[0]->object->parent2->child1 instanceof MyArrayObject);
+var_dump(is_array($documents[0]->object->parent2->child2));
+
+echo "\nSetting 'object.parent1.$' path to 'MyWildCardArrayObject' and 'object.$.$' to 'MyArrayObject'\n";
+$documents = fetch($manager, ["fieldPaths" => [
+    'object.parent1.$'    => "MyWildCardArrayObject",
+    'object.$.$'          => "MyArrayObject",
```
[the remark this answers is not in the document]
derickr: Happy to add one more.

tests/bson/typemap-008.phpt
```diff
@@ -0,0 +1,122 @@
+--TEST--
+MongoDB\Driver\Cursor::setTypeMap(): fieldPath typemaps without server
```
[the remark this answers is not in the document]
derickr: I really want all the typemap stuff in one place.

tests/bson/typemap-008.phpt
```diff
@@ -0,0 +1,122 @@
...
+        'parent2' => [
+            'child1' => [  7,  8,  9 ],
+            'child2' => [ 10, 11, 12 ],
+        ],
+    ],
+] );
+
+function fetch($bson, $typeMap = [])
+{
+    for ($i = 0; $i < 25000; $i++) {
```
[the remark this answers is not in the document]
derickr: Testing for memory leaks, and so that we can use this as a benchmarking script. It's a good thing that I added this test (with 50000) in the first place, as it *did* find a memory leak bug.

tests/bson/typemap-008.phpt
```diff
@@ -0,0 +1,122 @@
...
+        $documents = [ \MongoDB\BSON\toPHP($bson, $typeMap) ];
+    }
+    return $documents;
+}
+
+
+echo "\nSetting 'object.$.child1' path to 'MyWildCardArrayObject'\n";
+$documents = fetch($bson, ["fieldPaths" => [
+    'object.$.child1' => "MyWildCardArrayObject"
```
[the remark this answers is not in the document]
derickr: That was on purpose.

tests/cursor/cursor-setTypeMap_error-003.phpt
```diff
@@ -0,0 +1,67 @@
+--TEST--
+Cursor::setTypeMap(): fieldPaths must be an array, with single key/string elements
```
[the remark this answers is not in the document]
derickr: They did not: https://github.com/mongodb/mongo-php-driver/blob/master/tests/cursor/cursor-setTypeMap_error-001.phpt#L2 — and it tests the same thing as the error-001 and error-002 tests, so it should be the same here — and not as ``toPHP_error``.

tests/bson/typemap-003.phpt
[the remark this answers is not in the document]
jmikola: I'm still not sure how `ARRAY_AS_PROPS` relates to the double setting of a type map, but no objections if you'd like to keep this in place.

tests/bson/typemap-004.phpt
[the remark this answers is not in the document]
jmikola: I don't follow the relation to an IDE. If you're referring to an IDE indicating the type information for a `$documents` variable on these two lines, I don't think that's useful.

src/bson.c
jmikola: Aye, let's revisit this if you decide to add exceptions for tokenization errors.

src/bson.c
derickr: Sigh, strtok_r is pretty useless, as it can't handle empty elements (like in ``foo..bar``), so I've had to write this, which also means that we needed to keep some ``TSRMLS_[DC]C`` things around.

src/bson.c
derickr: Which I've now just done :-).

tests/bson/typemap-003.phpt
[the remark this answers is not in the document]
derickr: It doesn't — ``ARRAY_AS_PROPS`` was just added to make the asserts a little easier to write.

tests/bson/typemap-004.phpt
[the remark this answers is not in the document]
derickr: No, if there is no variable, an IDE can't inspect its contents, as it can't see the direct return values of a function.

tests/bson/typemap-006.phpt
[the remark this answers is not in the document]
derickr: Hah!::

```
class MyWildcardArrayObject extends MyArrayObject {};
```

So of course it matches :-) I will get rid of the extends, and repeat the ``MyArrayObject`` implementation.

Luckily the implementation was fine, just not the test.

tests/bson/typemap-006.phpt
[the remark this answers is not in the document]
derickr: Same as above.

tests/bson/typemap-007.phpt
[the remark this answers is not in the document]
derickr: [Narrator: They were changed to MyWildcardArrayObject anyway]

tests/bson/typemap-007.phpt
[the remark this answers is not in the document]
derickr: Done, added this.

tests/cursor/cursor-setTypeMap_error-003.phpt
```diff
@@ -0,0 +1,67 @@
...
+foreach ($fieldPaths as $fieldPath) {
+    $typeMap = ['fieldPaths' => $fieldPath];
+
+    printf("Test typeMap: %s\n", json_encode($typeMap));
+
+    echo throws(function() use ($cursor, $typeMap) {
+        $cursor->setTypeMap($typeMap);
+    }, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
+
+    echo "\n";
```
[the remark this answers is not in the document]
derickr: Done

src/bson.c
```diff
@@ -1201,6 +1289,230 @@ static bool php_phongo_bson_state_parse_type(zval* options, const char* name, ph
...
+	php_phongo_field_path_map_element* field_path_map_element;
+
+	if (field_path_original[0] == '.') {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "A 'fieldPaths' key may not start with a '.'");
+		return false;
+	}
+
+	if (field_path_original[strlen(field_path_original) - 1] == '.') {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "A 'fieldPaths' key may not end with a '.'");
+		return false;
```
jmikola: Assuming both of these cases are handled by the "empty segment" error below, I'd propose allowing that to catch this. IMO, `".foo"`, `"foo..bar"`, and `"foo."` all contain empty segments.

tests/cursor/cursor-setTypeMap_error-003.phpt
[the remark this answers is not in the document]
jmikola: > and it tests the same thing as the error-001 and error-002 tests, so it should be the same here — and not as toPHP_error

Not sure I follow. My suggestion was that once we finalize these tests for `Cursor::setTypeMap()` we copy them for `toPHP()` to ensure that function passes as well.

Separate from this, can we have all of the current tests organized under `cursor/cursor-setTypeMap*`? Any other tests that get copied can then go under `bson/bson-toPHP*`

tests/bson/typemap-007.phpt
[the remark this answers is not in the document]
jmikola: Aye, I realize class name casing makes no different. This is mainly for aesthetics :)

tests/bson/typemap-008.phpt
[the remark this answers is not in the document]
jmikola: Ah, just realized that this still has `MongoDB\Driver\Cursor::setTypeMap()` in the name. Anything not using a live server and instead testing `toPHP()` should note `MongoDB\BSON\toPHP()` in its title.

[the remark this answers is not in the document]
derickr: Done

src/bson-encode.c
jmikola: I'm OK with keeping this as-is if it's only two uses.

src/bson.c
jmikola: `owns_elements` works for me.

src/bson.c
derickr: I had that first, but thought it better we have more precise error messages. This one is also faster to check as nothing has been allocated yet.

tests/bson/bson-toPHP-007.phpt
```diff
@@ -0,0 +1,122 @@
+--TEST--
+MongoDB\BSON\toPHP(): fieldPath typemaps without server
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
```
jmikola: Server-less tests shouldn't import this (see other BSON tests), as it ultimately requires `PHONGO-SERVERS.json` to exist. The extension check in `basic-skipif.inc` isn't very helpful anyway.

PR #795, opened 2018-04-10
PHPC-1156: Prefer Secure Transport over OpenSSL with --with-mongodb-ssl on Darwin

derickr:
https://jira.mongodb.org/browse/PHPC-1156

@kvwalker — as I don't have a Mac, could you test this patch by running ``./configure --with-mongodb-ssl`` and see if it picks Secure Transport please? Before the patch it should (try to) pick OpenSSL.

### DIFF
--- a/scripts/build/autotools/CheckSSL.m4
+++ b/scripts/build/autotools/CheckSSL.m4
@@ -12,6 +12,20 @@ PHP_ARG_WITH([openssl-dir],
              [auto],
              [no])
 
+AS_IF([test "$PHP_MONGODB_SSL" = "darwin" -o \( "$PHP_MONGODB_SSL" = "auto" -a "$os_darwin" = "yes" \)],[
+  if test "$os_darwin" = "no"; then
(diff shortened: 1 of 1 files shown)


jmikola:
Should just require running `./configure` as `--with-mongodb-ssl` already defaults to `auto`.

kvwalker:
I think this works! Before the change, my Mac shows 
`Build configuration:`
  `CFLAGS                                           : -g -O2`
   ...
  `MONGODB_SHARED_LIBADD                            : -lsasl2  -lssl -lcrypto -lz     -lz`

After the change, it shows
`Build configuration:`
  `CFLAGS                                           : -g -O2`
   ...
  `MONGODB_SHARED_LIBADD                            : -lsasl2 -framework Security -framework CoreFoundation      -lz`

jmikola:
@derickr: For context, @venyii is @alcaeus' co-worker who previously experienced some issues with OpenSSL on macOS.

PR #803, opened 2018-04-12
PHPC-1152, PHPC-1161: Emulate implicit sessions for command cursors

jmikola:
https://jira.mongodb.org/browse/PHPC-1152
https://jira.mongodb.org/browse/PHPC-1161

Tests for this patch depend on #811, although I've included a temporary commit with the patch (sans the full patch and tests in #811) to give this PR's own tests a chance at passing.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -749,7 +749,7 @@ bool phongo_cursor_advance_and_check_for_error(mongoc_cursor_t *cursor TSRMLS_DC
 	return true;
 } /* }}} */
 
-int phongo_execute_query(mongoc_client_t *client, const char *namespace, zval *zquery, zval *options, uint32_t server_id, zval *return_value, int return_value_used TSRMLS_DC) /* {{{ */
+bool phongo_execute_query(mongoc_client_t *client, const char *namespace, zval *zquery, zval *options, uint32_t server_id, zval *return_value, int return_value_used TSRMLS_DC) /* {{{ */
 {
 	const php_phongo_query_t *query;
 	mongoc_cursor_t *cursor;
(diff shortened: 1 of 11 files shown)


php_phongo.c
```diff
@@ -837,38 +859,53 @@ int phongo_execute_command(mongoc_client_t *client, php_phongo_command_type_t ty
 	mongoc_cursor_t *cmd_cursor;
 	zval *zreadPreference = NULL;
 	zval *zsession = NULL;
-	int result;
+	bool result = false;
+	bool free_reply = false;
+	bool free_zsession = false;
```
derickr: Before you merge this into master, you should run ``clang-format`` on this.

php_phongo.c
```diff
@@ -837,38 +859,53 @@ int phongo_execute_command(mongoc_client_t *client, php_phongo_command_type_t ty
...
 		/* Exception should already have been thrown */
-		bson_destroy(&opts);
-		return false;
+		goto cleanup;
+	}
+
+	/* If an explicit session was not provided, attempt to create an implicit
+	 * client session (ignoring any errors).
+	 *
+	 * TODO: Ensure that this doesn't conflict with any commands users may run
```
derickr: How are you going to "ensure" this? Right now, it just looks like a comment that we're going to forget about :-)

php_phongo.c
jmikola: This is in the v1.4 branch, which isn't formatted. I intend to run it through clang-format after (or during) the merge to master. Does that work?

php_phongo.c
jmikola: Indeed. This is something I wanted to check manually but running through a few cases in [Exceptions to sending the session ID to the server on all commands](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#exceptions-to-sending-the-session-id-to-the-server-on-all-commands). In particular, I'd like to test:

 * `isMaster`
 * `executeWriteCommand()` with an unacknowledged write concern

I'm also curious about [`killCursors`](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#when-sending-a-killcursors-command). The spec says that an LSID is optional, but testing in [PHPC-1152](https://jira.mongodb.org/browse/PHPC-1152) demonstrated that `killCursors` wasn't always using the LSID of the command from which the cursor originated. I want to confirm that the exceptions for mismatched LSIDs only applied to `getMore` and that `killCursors` truly doesn't care. If that's not the case, I think the spec language may need some corrections.

jmikola: @kvwalker answered the `killCursors` question in https://github.com/mongodb/mongo-php-library/pull/528#discussion_r182174656.

jmikola: `isMaster` isn't a problem.

Unacknowledged write concerns is a legitimate issue, and may also affect libmongoc. I've opened [PHPC-1163](https://jira.mongodb.org/browse/PHPC-1163) to track that.

PR #816, opened 2018-04-24
PHPC-1137/PHPC-1138 Deprecate maxScan and snapshot options

derickr:
http://jira.mongodb.org/browse/PHPC-1137
http://jira.mongodb.org/browse/PHPC-1138

### DIFF
--- a/src/MongoDB/Query.c
+++ b/src/MongoDB/Query.c
@@ -100,13 +100,19 @@ static bool php_phongo_query_opts_append_document(bson_t* opts, const char* opts
 /* Note: handling of integer options will depend on SIZEOF_ZEND_LONG and we
 	 * are not converting strings to 64-bit integers for 32-bit platforms. */
 
-#define PHONGO_QUERY_OPT_INT64(opt, zarr, key)                                                                       \
-	if ((zarr) && php_array_existsc((zarr), (key))) {                                                                \
-		if (!BSON_APPEND_INT64(intern->opts, (opt), php_array_fetchc_long((zarr), (key)))) {                         \
-			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error appending \"%s\" option", (opt)); \
-			return false;                                                                                            \
-		}                                                                                                            \
+#define PHONGO_QUERY_OPT_INT64_EX(opt, zarr, key, deprecated)                                                                             \
+	if ((zarr) && php_array_existsc((zarr), (key))) {                                                                                     \
+		if ((deprecated)) {                                                                                                               \
+			php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"%s\" option is deprecated and will be removed in a future release", key); \
+		}                                                                                                                                 \
+		if (!BSON_APPEND_INT64(intern->opts, (opt), php_array_fetchc_long((zarr), (key)))) {                                              \
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error appending \"%s\" option", (opt));                      \
+			return false;                                                                                                                 \
+		}                                                                                                                                 \
 	}
+
+#define PHONGO_QUERY_OPT_INT64(opt, zarr, key) PHONGO_QUERY_OPT_INT64_EX((opt), (zarr), (key), 0)
+#define PHONGO_QUERY_OPT_INT64_DEPRECATED(opt, zarr, key) PHONGO_QUERY_OPT_INT64_EX((opt), (zarr), (key), 1)
 
 #define PHONGO_QUERY_OPT_STRING(opt, zarr, key)                                                   \
 	if ((zarr) && php_array_existsc((zarr), (key))) {                                             \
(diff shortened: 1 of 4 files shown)


src/MongoDB/Query.c
```diff
@@ -90,6 +90,15 @@ static bool php_phongo_query_opts_append_document(bson_t* opts, const char* opts
 		}                                                                                                            \
 	}
 
+#define PHONGO_QUERY_OPT_BOOL_DEPRECATED(opt, zarr, key)                                                                    \
```
jmikola: Rather than duplicate these macros in their entirety, what do you think of:

```c
#define PHONGO_QUERY_OPT_BOOL_EX(opt, zarr, key, deprecated)
	if ((zarr) && php_array_existsc((zarr), (key))) {
		if (deprecated) {
			php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"%s\" option is deprecated and will be removed in a future release", key);
		}
		if (!BSON_APPEND_BOOL(intern->opts, (opt), php_array_fetchc_bool((zarr), (key)))) {
			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error appending \"%s\" option", (opt));
			return false;
		}
	}

#define PHONGO_QUERY_OPT_BOOL(opt, zarr, key) PHONGO_QUERY_OPT_BOOL_EX((opt), (zarr), (key), 0)
#define PHONGO_QUERY_OPT_BOOL_DEPRECATED(opt, zarr, key) PHONGO_QUERY_OPT_BOOL_EX((opt), (zarr), (key), 1)
```

This would mitigate the risk of the macros getting out of sync.

src/MongoDB/Query.c
```diff
@@ -90,6 +90,15 @@ static bool php_phongo_query_opts_append_document(bson_t* opts, const char* opts
...
+	if ((zarr) && php_array_existsc((zarr), (key))) {                                                                       \
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The '%s' option is deprecated and will be removed in a future release", key); \
```
jmikola: Nit: `\"%s\"` for consistency with the "Error appending" message (I did so in my suggestion above)

src/MongoDB/Query.c
```diff
@@ -98,7 +107,7 @@ static bool php_phongo_query_opts_append_document(bson_t* opts, const char* opts
 	}
 
 /* Note: handling of integer options will depend on SIZEOF_ZEND_LONG and we
-	 * are not converting strings to 64-bit integers for 32-bit platforms. */
+ * are not converting strings to 64-bit integers for 32-bit platforms. */
```
jmikola: Was this introduced by clang-format and fixed manually?

tests/query/query-ctor-002.phpt
```diff
@@ -49,6 +49,9 @@ var_dump(new MongoDB\Driver\Query(
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
+Deprecated: MongoDB\Driver\Query::__construct(): The 'maxScan' option is deprecated and will be removed in a future release in %squery-ctor-002.php on line %d
```
jmikola: I believe most of our other tests used `in %s on line %d`. I think we can do the same here and avoid necessary changes if this ever gets renamed.

src/MongoDB/Query.c
```diff
@@ -277,8 +295,8 @@ static bool php_phongo_query_init(php_phongo_query_t* intern, zval* filter, zval
 	PHONGO_QUERY_OPT_BOOL("exhaust", options, "exhaust");
 	PHONGO_QUERY_OPT_DOCUMENT("max", options, "max")
 	else PHONGO_QUERY_OPT_DOCUMENT("max", modifiers, "$max");
-	PHONGO_QUERY_OPT_INT64("maxScan", options, "maxScan")
-	else PHONGO_QUERY_OPT_INT64("maxScan", modifiers, "$maxScan");
+	PHONGO_QUERY_OPT_INT64_DEPRECATED("maxScan", options, "maxScan")
+	else PHONGO_QUERY_OPT_INT64_DEPRECATED("maxScan", modifiers, "$maxScan");
```
jmikola: AFAIK, `modifiers` is deprecated ([DRIVERS-347](https://jira.mongodb.org/browse/DRIVERS-347)). Perhaps we should simply log a notice when _that_ option is used. If so, that should be its own PHPC ticket. I think we can still do it for 1.5.0; however, I'd then suggest we raise that notice once for `modifiers` and forego additional notices for sub-options like `$maxScan`.

Thoughts?

src/MongoDB/Query.c
derickr: No, clang had ignored it because it was a tab - so I fixed it by using a space.

src/MongoDB/Query.c
derickr: I copied this from the PHPLIB change - thought it would be better to have it consistent!

src/MongoDB/Query.c
derickr: Done

src/MongoDB/Query.c
derickr: But I now see PHPLIB uses " too, so fixed.

tests/query/query-ctor-002.phpt
derickr: Fixed

src/MongoDB/Query.c
derickr: It should still throw a notice for ``$maxScan`` and ``snapshot``, as they're still deprecated options on their own. Otherwise people change the modifiers to options (and ``$maxScan`` to ``maxScan``), and they would *then* get a notice. Better to see them all in one go.

PR #840, opened 2018-05-18
PHPC-1076: Expose result document for failed commands via CommandFailedEvent

jmikola:
https://jira.mongodb.org/browse/PHPC-1076

Supersedes #818. This includes the bump to libmongoc 1.10-dev and test suite changes for connection strings and skip functions.

### DIFF
--- a/tests/apm/monitoring-commandFailed-003.phpt
+++ b/tests/apm/monitoring-commandFailed-003.phpt
@@ -48,7 +48,7 @@ try {
 --EXPECTF--
 started: findAndModify
 failed: findAndModify
-object(stdClass)#%d (4) {
+object(stdClass)#%d (%d) {
   ["ok"]=>
   float(0)
   ["errmsg"]=>
@@ -56,5 +56,5 @@ object(stdClass)#%d (4) {
   ["code"]=>
   int(9)
   ["codeName"]=>
-  string(13) "FailedToParse"
+  string(13) "FailedToParse"%A
 }


tests/apm/monitoring-commandFailed-003.phpt
```diff
@@ -0,0 +1,60 @@
...
+object(stdClass)#%d (4) {
+  ["ok"]=>
+  float(0)
+  ["errmsg"]=>
+  string(49) "Either an update or remove=true must be specified"
+  ["code"]=>
+  int(9)
+  ["codeName"]=>
+  string(13) "FailedToParse"
+}
```
kvwalker: This test fails because `var_dump($event->getReply())` returns an `object` that includes `operationTime`, `$clusterTime`, and `signature`.

This can be fixed by adding
```
["operationTime"]=>
  object(MongoDB\BSON\Timestamp)#%d (2) {
    ["increment"]=>
    string(1) "1"
    ["timestamp"]=>
    string(10) "%d"
  }
  ["$clusterTime"]=>
  object(stdClass)#%d (2) {
    ["clusterTime"]=>
    object(MongoDB\BSON\Timestamp)#%d (2) {
      ["increment"]=>
      string(1) "1"
      ["timestamp"]=>
      string(10) "%d"
    }
    ["signature"]=>
    object(stdClass)#%d (2) {
      ["hash"]=>
      object(MongoDB\BSON\Binary)#%d (2) {
        ["data"]=>
        string(20) "%s"
        ["type"]=>
        int(0)
      }
      ["keyId"]=>
      int(0)
    }
```

jmikola: Thanks for catching that. It didn't come up with `overview.php` because I was already using `%A` there for 3.2 compatibility.

jmikola:
Note: Changes in https://github.com/mongodb/mongo-php-driver/pull/841 should address Windows builds.

PR #854, opened 2018-06-19
PHPC-1083: Change "available" RP constant to use libmongoc's constant instead of hard coding it

derickr:


php_phongo.c
```diff
@@ -982,11 +982,11 @@ bool phongo_execute_command(mongoc_client_t* client, php_phongo_command_type_t t
 			goto cleanup;
 		}
 
-		cmd_cursor = mongoc_cursor_new_from_command_reply(client, &initial_reply, server_id);
+		cmd_cursor = mongoc_cursor_new_from_command_reply_with_opts(client, &initial_reply, &opts);
 	} else {
 		bson_t* wrapped_reply = create_wrapped_command_envelope(db, &reply);
 
-		cmd_cursor = mongoc_cursor_new_from_command_reply(client, wrapped_reply, server_id);
+		cmd_cursor = mongoc_cursor_new_from_command_reply_with_opts(client, wrapped_reply, &opts);
```
jmikola: Does this change belong in the same commit?

derickr: It's now split out. Some more work had to be done here — see email to the driver lists.

php_phongo.c
```diff
@@ -957,36 +957,36 @@ bool phongo_execute_command(mongoc_client_t* client, php_phongo_command_type_t t
...
-		if (zsession && !mongoc_client_session_append(Z_SESSION_OBJ_P(zsession)->client_session, &initial_reply, &error)) {
+		if (zsession && !mongoc_client_session_append(Z_SESSION_OBJ_P(zsession)->client_session, &opts, &error)) {
 			phongo_throw_exception_from_bson_error_t(&error TSRMLS_CC);
 			bson_destroy(&initial_reply);
 			result = false;
 			goto cleanup;
 		}
 
-		cmd_cursor = mongoc_cursor_new_from_command_reply(client, &initial_reply, server_id);
+		cmd_cursor = mongoc_cursor_new_from_command_reply_with_opts(client, &initial_reply, &opts);
```
jmikola: Since `server_id` is no longer being passed as an explicit parameter, I think it should be added to `opts`. Otherwise, we may run the risk of not pinning subsequent getMores to the same server.

See: https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/doc/mongoc_cursor_new_from_command_reply_with_opts.rst

derickr: There is no need, as that is already done at https://github.com/mongodb/mongo-php-driver/pull/854/files/0c9caa69c3d9d95cad53337b5c6e03c4d1d923ca#diff-c06c6e1c9374aecabcf544157f9d0c26L899

php_phongo.c
jmikola: Likely also want to set `serverId` on `opts` here as well.

php_phongo.c
```diff
@@ -957,36 +957,36 @@ bool phongo_execute_command(mongoc_client_t* client, php_phongo_command_type_t t
...
 		bson_t       initial_reply = BSON_INITIALIZER;
 		bson_error_t error         = { 0 };
 
 		bson_copy_to(&reply, &initial_reply);
 
 		if (command->max_await_time_ms) {
-			bson_append_bool(&initial_reply, "awaitData", -1, 1);
-			bson_append_int64(&initial_reply, "maxAwaitTimeMS", -1, command->max_await_time_ms);
-			bson_append_bool(&initial_reply, "tailable", -1, 1);
+			bson_append_bool(&opts, "awaitData", -1, 1);
```
jmikola: At this point `opts` is the original document we used to build options for `mongoc_client_command_with_opts()` (and related functions). I think we should start with a new `opts` document for `mongoc_cursor_new_from_command_reply_with_opts()` (e.g. `cursor_opts`), as not all options are shared.

This re-use may be why server-pinning was working (related to me asking to set `serverId` on `opts` below), but that's not terribly clear.

php_phongo.c
```diff
@@ -957,36 +957,36 @@ bool phongo_execute_command(mongoc_client_t* client, php_phongo_command_type_t t
...
+			bson_append_bool(&opts, "tailable", -1, 1);
 		}
 
 		if (command->batch_size) {
-			bson_append_int64(&initial_reply, "batchSize", -1, command->batch_size);
+			bson_append_int64(&opts, "batchSize", -1, command->batch_size);
 		}
 
-		if (zsession && !mongoc_client_session_append(Z_SESSION_OBJ_P(zsession)->client_session, &initial_reply, &error)) {
+		if (zsession && !mongoc_client_session_append(Z_SESSION_OBJ_P(zsession)->client_session, &opts, &error)) {
```
jmikola: Another reason to split `opts` from a new `cursor_opts` document: this is currently appending explicit sessions twice to the same document.

php_phongo.c
jmikola: This should also use a `cursor_opts`:

```c
bson_t cursor_opts = BSON_INITIALIZER;
bson_t* wrapped_reply = create_wrapped_command_envelope(db, &reply);


bson_append_int32(&cursor_opts, "serverId", -1, server_id);
cmd_cursor = mongoc_cursor_new_from_command_reply_with_opts(client, wrapped_reply, &opts);
bson_destroy(&cursor_opts);
```

derickr: updated

PR #864, opened 2018-06-27
Update .travis.yml

719media:
php 7.0 in there twice. I know, I know, this is a super awesome pr

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ matrix:
     - php: 7.0
     - php: 7.1
     - php: 7.2
-    - php: 7.0
       env:
         - SERVER_VERSION=3.0.15
     - php: 7.0


derickr:
This is on purpose, the ``php: 7.0`` for this one is specifically for the running of the tests with ``SERVER_VERSION=3.0.15``.

719media:
But isn’t it already included for this version just a few lines up?

719media:
Oops! Thanks

.travis.yml
```diff
@@ -20,7 +20,6 @@ matrix:
     - php: 7.0
     - php: 7.1
     - php: 7.2
-    - php: 7.0
```
jmikola: > But isn’t it already included for this version just a few lines up?

@719media: If I'm reading this diff correctly, it combines two separate jobs (PHP 7.2 with server 3.6.2 and PHP 7.0 with server 3.0.15) with one (PHP 7.2 with server 3.0.15).

For no real reason in particular, we run tests with all older server versions under PHP 7.0.

PR #867, opened 2018-06-28
PHPC-1221: Accept "yes" and no value for non-binary PHP_ARG_WITH() options

derickr:
The checks for ICU were already as expected, but both SASL and SSL needed improvements.

I also added ``configure: checking whether {x} is available`` for the SSL checks so that it is more clear which section checks for what in configure output (the dual checks for OpenSSL and LibreSSL confused me at first).

### DIFF
--- a/scripts/autotools/libmongoc/CheckSSL.m4
+++ b/scripts/autotools/libmongoc/CheckSSL.m4
@@ -12,7 +12,15 @@ PHP_ARG_WITH([openssl-dir],
              [auto],
              [no])
 
+dnl PHP_ARG_WITH without a value assigns "yes". Treat it like "auto" but required.
+AS_IF([test "$PHP_MONGODB_SSL" = "yes"],[
+  crypto_required="yes"
+  PHP_MONGODB_SSL="auto"
+])
+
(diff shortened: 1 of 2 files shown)


scripts/autotools/libmongoc/CheckSSL.m4
```diff
@@ -144,6 +153,9 @@ AS_IF([test "$PHP_MONGODB_SSL" = "libressl" -o "$PHP_MONGODB_SSL" = "auto"],[
 ])
 
 AS_IF([test "$PHP_MONGODB_SSL" = "auto"],[
+  if test "$crypto_required" = "yes"; then
```
jmikola: Since `$crypto_required` is sometimes unassigned, this is actually a place where you want to do `"x$crypto_required" = "xyes"`.

scripts/autotools/libmongoc/CheckSSL.m4
```diff
@@ -12,7 +12,14 @@ PHP_ARG_WITH([openssl-dir],
              [auto],
              [no])
 
+AS_IF([test "$PHP_MONGODB_SSL" = "yes"],[
```
jmikola: Would it make sense to add a comment explaining why we handle "yes" even though it's not a documented option. Something like:

> PHP_ARG_WITH will substitute "yes" in the absence of a value. Treat it like "auto" but required.

Likewise for CheckSasl.m4 (above the first place you reference "yes" as a value).

PR #870, opened 2018-06-29
PHPC-1222: Implement accessor for error labels on exception classes

derickr:
Before we can merge this, we need to branch 1.5 from 1.6 of course. Just wanted your opinion on it. The second commit in this PR we can safely drop. I did the work so I can audit, but am now unsure whether it's wise to get rid of that older API that doesn't take a reply_t as well.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -181,9 +181,49 @@ void phongo_throw_exception(php_phongo_error_domain_t domain TSRMLS_DC, const ch
 	va_end(args);
 }
 
-void phongo_throw_exception_from_bson_error_t(bson_error_t* error TSRMLS_DC)
+static void phongo_exception_add_error_labels(bson_t* reply TSRMLS_DC)
 {
-	zend_throw_exception(phongo_exception_from_mongoc_domain(error->domain, error->code), error->message, error->code TSRMLS_CC);
+	bson_iter_t iter;
+
+	if (!reply) {
+		return;
+	}
+
+	if (bson_iter_init_find(&iter, reply, "errorLabels")) {
+		bson_iter_t error_labels;
+#if PHP_VERSION_ID >= 70000
+		zval labels;
+
+		array_init(&labels);
+#else
+		zval* labels = NULL;
+
+		ALLOC_INIT_ZVAL(labels);
+		array_init(labels);
+#endif
+
+		bson_iter_recurse(&iter, &error_labels);
+		while (bson_iter_next(&error_labels)) {
+			if (BSON_ITER_HOLDS_UTF8(&error_labels)) {
+				const char*    error_label;
+				uint32_t       error_label_len;
+
+				error_label = bson_iter_utf8(&error_labels, &error_label_len);
+#if PHP_VERSION_ID >= 70000
+				ADD_NEXT_INDEX_STRINGL(&labels, error_label, error_label_len);
+#else
+				ADD_NEXT_INDEX_STRINGL(labels, error_label, error_label_len);
+#endif
+			}
+		}
+
+#if PHP_VERSION_ID >= 70000
+		phongo_add_exception_prop(ZEND_STRL("errorLabels"), &labels);
+#else
+		phongo_add_exception_prop(ZEND_STRL("errorLabels"), labels TSRMLS_CC);
+#endif
+		zval_ptr_dtor(&labels);
+	}
 }
 
 void phongo_throw_exception_from_bson_error_and_reply_t(bson_error_t* error, bson_t* reply TSRMLS_DC)
@@ -192,7 +232,7 @@ void phongo_throw_exception_from_bson_error_and_reply_t(bson_error_t* error, bso
 	 * may use CommandException and report the result document for the
 	 * failed command. For BC, ExceededTimeLimit errors will continue to use
 	 * ExcecutionTimeoutException and omit the result document. */
-	if ((error->domain == MONGOC_ERROR_SERVER && error->code != PHONGO_SERVER_ERROR_EXCEEDED_TIME_LIMIT) || error->domain == MONGOC_ERROR_WRITE_CONCERN) {
+	if (reply && ((error->domain == MONGOC_ERROR_SERVER && error->code != PHONGO_SERVER_ERROR_EXCEEDED_TIME_LIMIT) || error->domain == MONGOC_ERROR_WRITE_CONCERN)) {
 #if PHP_VERSION_ID >= 70000
 		zval zv;
 #else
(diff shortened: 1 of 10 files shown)


src/MongoDB/Exception/RuntimeException.c
```diff
@@ -23,12 +23,90 @@
...
+#include "php_array_api.h"
 
 zend_class_entry* php_phongo_runtimeexception_ce;
 
+static bool php_phongo_has_string_array_element(zval* labels, char* label TSRMLS_DC)
+{
+	HashTable* ht_data;
+
+	if (Z_TYPE_P(labels) != IS_ARRAY) {
+		return false;
```
jmikola: I pondered whether a LogicException or something similar would work here, but it's probably best to keep `hasErrorLabel()` as simple as possible. The edge case I was thinking of was if someone uses reflection to re-assign our private property.

src/MongoDB/Exception/RuntimeException.c
```diff
@@ -23,12 +23,90 @@
...
+	ht_data = HASH_OF(labels);
+
+#if PHP_VERSION_ID >= 70000
+	{
+		zval* z_label;
+
+		ZEND_HASH_FOREACH_VAL(ht_data, z_label)
+		{
+			if (Z_TYPE_P(z_label) == IS_STRING) {
+				if (strcmp(Z_STRVAL_P(z_label), label) == 0) {
```
jmikola: I'd be in favor of combining these into a single `if` condition. That won't make the line too long (we have much worse offenders).

php_phongo.h
```diff
@@ -110,7 +110,6 @@ void              phongo_throw_exception(php_phongo_error_domain_t domain TSRMLS
 #endif /* PHP_WIN32 */
 #endif /* PHP_VERSION_ID < 70000 */
 	;
-void phongo_throw_exception_from_bson_error_t(bson_error_t* error TSRMLS_DC);
```
jmikola: > whether it's wise to get rid of that older API that doesn't take a reply_t as well.

I think it'd be best to keep `phongo_throw_exception_from_bson_error_t()` and simply have it call `phongo_throw_exception_from_bson_error_and_reply_t()` with a `NULL` reply. I imagine it'd be a quick compiler optimization and it's also cut down on many of the code replacements in this PR.

php_phongo.c
```diff
@@ -181,9 +181,43 @@ void phongo_throw_exception(php_phongo_error_domain_t domain TSRMLS_DC, const ch
...
+	bson_iter_t iter;
+
+	if (!reply) {
+		return;
+	}
+
+	if (bson_iter_init_find(&iter, reply, "errorLabels")) {
+		bson_iter_t error_labels;
+#if PHP_VERSION_ID >= 70000
+		zval* labels = ecalloc(sizeof(zval), 1);
```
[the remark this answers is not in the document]
derickr: FWIW, I took the same approach as https://github.com/derickr/mongo-php-driver/blob/6e1e2643fb7b6202f72a6e6684ac60cb4c07e07c/php_phongo.c#L1216-L1230

php_phongo.c
```diff
@@ -209,8 +243,9 @@ void phongo_throw_exception_from_bson_error_and_reply_t(bson_error_t* error, bso
 #endif
 		zval_ptr_dtor(&zv);
 	} else {
-		phongo_throw_exception_from_bson_error_t(error TSRMLS_CC);
+		zend_throw_exception(phongo_exception_from_mongoc_domain(error->domain, error->code), error->message, error->code TSRMLS_CC);
```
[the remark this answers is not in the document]
derickr: Yes, because ``phongo_throw_exception_from_bson_error_t`` is now a wrapper for ``phongo_throw_exception_from_bson_error_and_reply_t``.

src/MongoDB/Exception/RuntimeException.c
derickr: Done

src/MongoDB/Exception/RuntimeException.c
derickr: Added

tests/session/transaction-integration-002.phpt
```diff
@@ -61,9 +61,7 @@ try {
     ] );
     $manager->executeCommand(DATABASE_NAME, $cmd, ['session' => $sessionB]);
 } catch (MongoDB\Driver\Exception\CommandException $e) {
-    $rd = $e->getResultDocument();
-
-    echo (isset($rd->errorLabels) && in_array("TransientTransactionError", $rd->errorLabels)) ?
+    echo $e->hasErrorLabel('TransientTransactionError') ?
```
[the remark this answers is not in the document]
derickr: I had written these, but forgotten to ``git add`` them.

src/MongoDB/Exception/RuntimeException.c
jmikola: What was added? I wasn't suggesting actually throwing, just explaining my thought process. I do think it's preferable to just return false here.

src/MongoDB/Exception/RuntimeException.c
```diff
@@ -43,6 +119,8 @@ void php_phongo_runtimeexception_init_ce(INIT_FUNC_ARGS) /* {{{ */
 	php_phongo_runtimeexception_ce = zend_register_internal_class_ex(&ce, spl_ce_RuntimeException, NULL TSRMLS_CC);
 #endif
 	zend_class_implements(php_phongo_runtimeexception_ce TSRMLS_CC, 1, php_phongo_exception_ce);
+
+	zend_declare_property_null(php_phongo_runtimeexception_ce, ZEND_STRL("errorLabels"), ZEND_ACC_PROTECTED TSRMLS_CC);
```
jmikola: Is there a particular reason this can't be private? `hasErrorLabel()` is `final` on this class and should be the only method that accesses it.

I realize `WriteException::$writeResult` is also protected (and only accessed by `getWriteResult()`. Perhaps `protected` is needed so that we can update the property on any class? If so, I wonder if that's simply a problem with `phongo_add_exception_prop()` not using the base class entry when it calls `zend_update_property()`.

src/MongoDB/Exception/RuntimeException.c
derickr: Added "added" to the wrong place I think. Want to say I added the test cases, perhaps? /me shrugs

src/MongoDB/Exception/RuntimeException.c
derickr: | I realize WriteException::$writeResult is also protected (and only accessed by getWriteResult(). Perhaps protected is needed so that we can update the property on any class?

Yes, that's the case:
```
001+ Fatal error: Uncaught ReflectionException: Property errorLabels does not exist in /home/derick/dev/php/derickr-mongo-php-driver/tests/exception/bulkwriteexception-haserrorlabel-001.php
```

|  If so, I wonder if that's simply a problem with phongo_add_exception_prop() not using the base class entry when it calls zend_update_property().

It doesn't matter when I change it to use the base class (php_phongo_runtimeexception_ce). So, I think we're stuck with this.

tests/session/transaction-integration-002.phpt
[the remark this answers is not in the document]
jmikola: If I understand correctly, this test was only using `executeCommand()` so that you could inspect the error label by calling `CommandException::getResultDocument()`. I think this should be rewritten to execute a bulk write and then call `hasErrorLabel` on the BulkWriteException.

That may not be possible in this PR, as `phongo_execute_bulk_write()` doesn't appear to be considering the reply document when throwing its exception. If so, please link a new ticket to PHPC-1222 to populate error labels for bulk writes and make a note about updating this test.

php_phongo.c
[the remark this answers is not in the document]
derickr: No, we can't, as ``phongo_throw_exception_from_bson_error_t`` is now a wrapper for ``phongo_throw_exception_from_bson_error_and_reply_t``:

```
void phongo_throw_exception_from_bson_error_t(bson_error_t* error TSRMLS_DC)
{
    phongo_throw_exception_from_bson_error_and_reply_t(error, NULL TSRMLS_CC);
}
```

PR #871, opened 2018-07-02
PHPC-1184: Add alternative topologies to Travis CI

derickr:


tests/apm/monitoring-commandFailed-001.phpt
```diff
@@ -41,17 +41,23 @@ $subscriber = new MySubscriber;
...
+    $primary->executeCommand(DATABASE_NAME, $command);
+} catch (Exception $e) {
+    /* Swallow */
+}
 ?>
 --EXPECT--
-started: drop
-failed: drop
+started: geoNear
+failed: geoNear
```
jmikola: `geoNear` is deprecated ([PM-939](https://jira.mongodb.org/browse/PM-939)). I think a more robust solution would just be to issue an `aggregate` with an unsupported pipeline operator (e.g. `[ { "$unsupported": 1 } ]`) or an `update` with a similarly unsupported modifier.

tests/utils/tools.php
```diff
@@ -95,6 +95,25 @@ function get_primary_server($uri)
...
 
+/**
+ * Runs a command and returns whether an exception was thrown or not
+ *
+ * @param string $uri Connection string
+ * @param Command $command
+ * @return bool
+ * @throws RuntimeException
+ */
+function command_works($uri, $command)
```
jmikola: Any reason not to type-hint `$command` here?

Alternatively, allow it to take any `array|object` and then wrap it in a Command as needed. This is what PHPLIB's [`DatabaseCommand`](https://github.com/mongodb/mongo-php-library/blob/1.4.1/src/Operation/DatabaseCommand.php#L81) helper does. It would save you from needing to import and construct `MongoDB\Driver\Command` from `skip_if_sleep_command_unavailable()`, and anything else that may use this in the future.

tests/standalone/connectiontimeoutexception-001.phpt
```diff
@@ -5,6 +5,7 @@ ConnectionTimeoutException: exceeding sockettimeoutms
 <?php skip_if_not_live(); ?>
 <?php skip_if_not_clean(); ?>
 <?php skip_if_test_commands_disabled(); ?>
+<?php skip_if_sleep_command_unavailable(); ?>
```
jmikola: Since we already have a skip for disabled test commands on the line above, what is the case where [`sleep`](https://mongodb-documentation.readthedocs.io/en/latest/reference/command/sleep.html) isn't supported? Unfortunately, it looks like the command is no longer in our own manual, so I've linked to one of the Chinese translations (cloned from an older version).

tests/apm/overview.phpt
```diff
@@ -96,9 +102,9 @@ object(MongoDB\Driver\Monitoring\CommandFailedEvent)#%d (%d) {
     ["ok"]=>
     float(0)
     ["errmsg"]=>
-    string(12) "ns not found"
+    string(13) "can't find ns"
     ["code"]=>
-    int(26)%A
+    int(8)%A
```
jmikola: The fact that this still reports error code 8 ("UnknownError") instead of 26 ("NamespaceNotFound") is more evidence that `geoNear` is not being actively maintained and we should probably avoid using it for any testing.

c.f. https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err

scripts/presets/travis/replica_sets/replicaset.json
```diff
@@ -12,8 +12,7 @@
                 "nssize": 1,
                 "port": 3000,
                 "bind_ip_all": true,
-                "smallfiles": true,
-                "setParameter": {"enableTestCommands": 1}
+                "smallfiles": true
```
jmikola: Why are you removing `enableTestCommand`?

I'm curious if this is the reason you needed to introduce `skip_if_sleep_command_unavailable()`. Perhaps there's a bug in `skip_if_test_commands_disabled()`?

tests/standalone/connectiontimeoutexception-001.phpt
jmikola: Interesting. I wonder if that's related to fail points on mongos not being supported. Depending on the outcome of that ticket, it may make sense to simply skip this test for mongos and leave `/* See: SERVER-36066 */` as a comment between the skip function and `?>`.

derickr: The work for skipping this properly (based on the sleep function existing) is already done, so I don't see a reason to remove that again.

.travis.yml
```diff
@@ -20,25 +21,40 @@ matrix:
...
 
 before_install:
   - pip install "mongo-orchestration>=0.6.7,<1.0" --user `whoami`
   - .travis.scripts/before_install.sh
-  - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-${SERVER_VERSION}.tgz
-  - tar zxf mongodb-linux-x86_64-${SERVER_VERSION}.tgz
-  - export PATH=${PWD}/mongodb-linux-x86_64-${SERVER_VERSION}/bin/:${PATH}
+  - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1404-${SERVER_VERSION}.tgz
+  - tar zxf mongodb-linux-x86_64-ubuntu1404-${SERVER_VERSION}.tgz
+  - export PATH=${PWD}/mongodb-linux-x86_64-ubuntu1404-${SERVER_VERSION}/bin:${PATH}
```
jmikola: Was there a problem with the standard Linux builds? Is it that they ship without SSL support?

I'm just curious if we might run into issues with the build team no longer supplying 14.04 builds for newer releases down the line (assuming Travis doesn't upgrade to 16.04 sooner).

.travis.yml
```diff
@@ -47,12 +63,14 @@ before_install:
...
   - mongo-orchestration start
-  - pushd ${MO_PATH} && ${TRAVIS_BUILD_DIR}/.travis.scripts/mo.sh configurations/servers/clean.json start && popd
+  - .travis.scripts/setup_mo.sh
   - ulimit -a
   - ulimit -c unlimited || true
   - .travis.scripts/before_script.sh
 
 script:
+  - export MONGODB_URI=`cat /tmp/uri.txt`
+  - echo $MONGODB_URI
```
jmikola: `setup_mo.sh` already prints the URI (prefixed with "MongoDB Test URI: "). Why bother doing so again?

scripts/presets/travis/replica_sets/replicaset.json
```diff
@@ -0,0 +1,66 @@
...
+            "procParams": {
+                "dbpath": "/tmp/REPLICASET/3000/",
+                "ipv6": true,
+                "logappend": true,
+                "logpath": "/tmp/REPLICASET/3000/mongod.log",
+                "journal": true,
+                "nssize": 1,
+                "port": 3000,
+                "bind_ip_all": true,
+                "smallfiles": true
```
jmikola: Per https://github.com/10gen/mongo-orchestration/pull/247#issuecomment-403333778, I think we can remove `nssize`, `smallfiles`, and `noprealloc` from all of these configs.

Only the `STANDALONE_OLD` case under the 3.0.x server version will use MMAPv1 (where `noprealloc` might still apply); however, I'm not convinced it's worth optimizing for that case (especially since it's not even a replica set). Having these servers use `/tmp` as their DB path should be enough optimization, since I would expect that to be a RAM disk.

I had previously removed `noprealloc` from our non-Travis configs in https://github.com/mongodb/mongo-php-driver/commit/2ef765acaff37872e26a767ec1db5da9d579df6a but stupidly added the others in https://github.com/mongodb/mongo-php-driver/commit/9a83bde5824deca68be8cda688133f3f235f06a1. Perhaps we can remove all three options across the board (for these Travis configs and the non-Travis configs).

scripts/presets/travis/replica_sets/replicaset.json
```diff
@@ -0,0 +1,66 @@
+{
+    "id": "REPLICASET",
+    "name": "mongod",
+    "members": [
+        {
+            "procParams": {
+                "dbpath": "/tmp/REPLICASET/3000/",
+                "ipv6": true,
+                "logappend": true,
```
jmikola: Is this option important given that the Travis build boxes are ephemeral?

scripts/presets/travis/replica_sets/replicaset.json
```diff
@@ -0,0 +1,66 @@
...
+                "logpath": "/tmp/REPLICASET/3000/mongod.log",
+                "journal": true,
```
jmikola: I expect this was copied over from our non-Travis configs. It looks like [`storage.journal.enabled`](https://docs.mongodb.com/manual/reference/configuration-options/#storage.journal.enabled) defaults to true only on 64-bit platforms, so it's probably sensible to keep this set explicitly in case.

scripts/presets/travis/sharded_clusters/cluster.json
```diff
@@ -0,0 +1,89 @@
+{
+    "configsvrs": [ {
+		"members" : [
```
jmikola: These JSON files appear to be using a mix of four-space and tab indents. I'd suggest standardizing on four-space indents, as I did for our non-Travis configs in https://github.com/mongodb/mongo-php-driver/commit/8b19fcbc152af9fa9c2e0e67b34e4b348f2ff2d9.

tests/cursor/bug1050-001.phpt
```diff
@@ -57,7 +57,7 @@ if (isset($document)) {
 <?php exit(0); ?>
 --EXPECTF--
 Executing command took 0.%d seconds
-Rewinding cursor took 1.%d seconds
+Rewinding cursor took 0.%r(4|5)%r%d seconds
```
jmikola: Good catch. :+1:

tests/connect/bug1045.phpt
```diff
@@ -12,7 +12,7 @@ PHPC-1045: Segfault if username is not provided for SCRAM-SHA-1 authMechanism
 require_once __DIR__ . "/../utils/basic.inc";
 
 // URI does not support auth, but that is not necessary for the test
-$m = new MongoDB\Driver\Manager(URI, ['authMechanism' => 'SCRAM-SHA-1', 'ssl' => false]);
+$m = new MongoDB\Driver\Manager(URI, ['authMechanism' => 'SCRAM-SHA-1']);
```
jmikola: Is the "URI does not support auth" comment above still relevant, given that we may be testing against a server using authentication?

Perhaps this should be changed to: "URI may or may not support auth, but that..."

tests/standalone/connectiontimeoutexception-001.phpt
jmikola: If I understand `skip_if_sleep_command_unavailable()` correctly, it's going to introduce a one-second delay if the server _does_ support the command. If this is also the only place that function would be used, I think the skip is preferable.

I think the best improvement might actually be modifying `skip_if_test_commands_disabled()` to also skip if the primary server is a `mongos` node, since it doesn't seem to support any test commands (despite reporting otherwise when we query its configuration).

That will also prove more useful down the line if we inevitably add more tests that want to set fail points.

I'm actually curious how the following tests pass on mongos:

 *  [`tests/standalone/executiontimeoutexception-001.phpt`](https://github.com/mongodb/mongo-php-driver/blob/1.5.1/tests/standalone/executiontimeoutexception-001.phpt)
 *  [`tests/standalone/executiontimeoutexception-002.phpt`](https://github.com/mongodb/mongo-php-driver/blob/1.5.1/tests/standalone/executiontimeoutexception-002.phpt)

The other two tests that set fail points are currently disabled:

 * [`tests/cursor/cursor-getmore-005.phpt`](https://github.com/mongodb/mongo-php-driver/blob/1.5.1/tests/cursor/cursor-getmore-005.phpt)
 * [`tests/cursor/cursor-getmore-006.phpt`](https://github.com/mongodb/mongo-php-driver/blob/1.5.1/tests/cursor/cursor-getmore-006.phpt)

.travis.yml
derickr: Yes, the standard ones ship without SSL support :-/ And yes, we will run into issues when we stop supporting 14.04 and Travis doesn't upgrade :-/

.travis.yml
derickr: Originally, I added the echo here because for some reason the environment variable didn't propagate between steps (i.e., ``before_script`` and ``script``). I then later added the ``export`` to fix this, and the ``echo`` stayed for testing. I am inclined to keep it.

scripts/presets/travis/replica_sets/replicaset.json
derickr: It was useful for Travis in debug mode.

scripts/presets/travis/replica_sets/replicaset.json
derickr: I've made a ticket for it, to do this as a low priority thing later: https://jira.mongodb.org/browse/PHPC-1240

scripts/presets/travis/sharded_clusters/cluster.json
derickr: It was only this one... → fixed.

tests/connect/bug1045.phpt
derickr: Updated

tests/standalone/connectiontimeoutexception-001.phpt
derickr: I can't explain why those two tests actually work, but I checked, and they indeed fail with ``operation exceeded time limit``, even when talking to ``mongos``. I can't reproduce this effect in the shell though.

derickr: And also, setting the failpoint on ``mongos`` succeeds:

```
mongos> db.runCommand( { configureFailPoint: 'maxTimeAlwaysTimeOut', mode: 'alwaysOn' } );
{
	"ok" : 1,
	"operationTime" : Timestamp(1531733041, 1),
	"$clusterTime" : {
		"clusterTime" : Timestamp(1531733041, 1),
		"signature" : {
			"hash" : BinData(0,"AAAAAAAAAAAAAAAAAAAAAAAAAAA="),
			"keyId" : NumberLong(0)
		}
	}
}
```

jmikola: Ah, so it is a case of this specific test command not working on mongos. Works for me, then.

PR #875, opened 2018-07-14
Fix build against OpenSSL 1.1

DaveRandom:
OpenSSL 1.1 [deprecates](https://www.openssl.org/docs/man1.1.0/ssl/SSL_library_init.html) the `SSL_library_init()` function in favour of `OPENSSL_init_ssl()`.

While backward-compatibility macros are provided so that programs will generally work, `AC_CHECK_LIB` does not find the function and configure time because it does not include the header with the macro definition.

This patch will let the configure script search for the new function if the old one is not found.

Specifically, this fixes building against [the openssl package for debian stretch](https://packages.debian.org/stretch/openssl).

*N.B. while this patch definitely fixes the build, I have not yet been able to run the test suite against the resulting binary, I will be able to do so next week, it may be best if this patch is not merged before I can confirm it works properly.*

### DIFF
--- a/scripts/autotools/libmongoc/CheckSSL.m4
+++ b/scripts/autotools/libmongoc/CheckSSL.m4
@@ -82,11 +82,24 @@ AS_IF([test "$PHP_MONGODB_SSL" = "openssl" -o "$PHP_MONGODB_SSL" = "auto"],[
                       [have_crypto_lib="yes"],
                       [have_crypto_lib="no"],
                       [$OPENSSL_LIBDIR_LDFLAG])
+
+    have_ssl_lib="no"
+
+    dnl OpenSSL < 1.1.0
     PHP_CHECK_LIBRARY([ssl],
                       [SSL_library_init],
                       [have_ssl_lib="yes"],
-                      [have_ssl_lib="no"],
+                      [],
                       [$OPENSSL_LIBDIR_LDFLAG -lcrypto])
+
+    dnl OpenSSL >= 1.1.0
+    if test "$have_ssl_lib" = "no"; then
+      PHP_CHECK_LIBRARY([ssl],
+                        [OPENSSL_init_ssl],
+                        [have_ssl_lib="yes"],
+                        [],
+                        [$OPENSSL_LIBDIR_LDFLAG -lcrypto])
+    fi
 
     if test "$have_ssl_lib" = "yes" -a "$have_crypto_lib" = "yes"; then
       PHP_ADD_LIBRARY([ssl],,[MONGODB_SHARED_LIBADD])


DaveRandom:
I spent more or less the whole day overthinking this, and I do remember finding that was a problem at one point, and then later finding that it wasn't, I can't remember specifically why/how, I will find it again.

DaveRandom:
I remember: The build system is using [`dh_auto_configure`](https://manpages.debian.org/stretch/debhelper/dh_auto_configure.1.en.html) which passes (amongst many other things determined by magic) `--prefix=/usr --libdir=\${prefix}/lib/x86_64-linux-gnu`. The end result of this is that `$PHP_LIBDIR` ends up defined as `lib/x86_64-linux-gnu`, so when I pass `--with-openssl-dir=auto` the `/usr` entry in `$PHP_OPENSSL_DIR` results in everything being found. (n.b. I am building standalone as a shared module after `phpize`)

I followed it through and I definitely understood it at one point and it did make sense, unfortunately I've now forgotten it again. This is well outside my comfort zone in many ways tbh, however if you want more info I can point you in the direction of people who actually know what they are talking about.

scripts/autotools/libmongoc/CheckSSL.m4
```diff
@@ -82,10 +82,21 @@ AS_IF([test "$PHP_MONGODB_SSL" = "openssl" -o "$PHP_MONGODB_SSL" = "auto"],[
...
+    dnl OpenSSL < 1.1.0
     PHP_CHECK_LIBRARY([ssl],
                       [SSL_library_init],
                       [have_ssl_lib="yes"],
-                      [have_ssl_lib="no"],
+                      [],
+                      [$OPENSSL_LIBDIR_LDFLAG -lcrypto])
+
+    dnl OpenSSL >= 1.1.0
+    PHP_CHECK_LIBRARY([ssl],
```
jmikola: Rather than always perform `PHP_CHECK_LIBRARY()` twice, I think we can improve this by wrapping the second attempt for OpenSSL >= 1.1.0 in a conditional:

```
AC_MSG_NOTICE([checking whether OpenSSL < 1.1.0 is available])
PHP_CHECK_LIBRARY([ssl],
                  [SSL_library_init],
                  [have_ssl_lib="yes"],
                  [have_ssl_lib="no"],
                  [$OPENSSL_LIBDIR_LDFLAG -lcrypto])

if test "$have_ssl_lib" = "no"; then
  AC_MSG_NOTICE([checking whether OpenSSL >= 1.1.0 is available])
  PHP_CHECK_LIBRARY([ssl],
                    [OPENSSL_init_ssl],
                    [have_ssl_lib="yes"],
                    [have_ssl_lib="no"],
                    [$OPENSSL_LIBDIR_LDFLAG -lcrypto])
fi
```

DaveRandom: Good call, updated

derickr:
I'm closing this in favour of #879, which incorporates your commit, and @jmikola's (reversed) suggestion. Thanks for the patch and explanation.

scripts/autotools/libmongoc/CheckSSL.m4
derickr: Sorry, I missed this commit while I was working on this :-/

DaveRandom:
👍

PR #876, opened 2018-07-16
Don't check for specific error string in test

DaveRandom:
Test checks for a specific error string that may vary between systems,

This patch brings the test into line with several other tests that check for timeout exceptions, e.g. https://github.com/mongodb/mongo-php-driver/blob/master/tests/standalone/bug0655.phpt

### DIFF
--- a/tests/manager/manager-executeCommand_error-003.phpt
+++ b/tests/manager/manager-executeCommand_error-003.phpt
@@ -17,5 +17,5 @@ echo throws(function() use($manager, $command) {
 <?php exit(0); ?>
 --EXPECTF--
 OK: Got MongoDB\Driver\Exception\ConnectionTimeoutException
-No suitable servers found (`serverSelectionTryOnce` set): [connection refused calling ismaster on '%s']
+No suitable servers found (`serverSelectionTryOnce` set): %s
 ===DONE===


derickr:
Hey Chris, I'm happy to merge this. Did you actually run into a problem where this test failed due to the message?

DaveRandom:
I did, the test was failing on a containerized debian stretch because it was producing "connection failed" rather than "connection refused".

So far it is the only test that I have encountered problems with - I've not yet managed to run the test suite with a server available, but I'm working on that today.

tests/manager/manager-executeCommand_error-003.phpt
```diff
@@ -17,5 +17,5 @@ echo throws(function() use($manager, $command) {
 <?php exit(0); ?>
 --EXPECTF--
 OK: Got MongoDB\Driver\Exception\ConnectionTimeoutException
-No suitable servers found (`serverSelectionTryOnce` set): [connection refused calling ismaster on '%s']
+No suitable servers found (`serverSelectionTryOnce` set): %s
```
jmikola: AFAIK, ["connection refused"](https://github.com/mongodb/mongo-c-driver/blob/1.11.0/src/libmongoc/src/mongoc/mongoc-async.c#L145) comes from libmongoc encountering a `POLLHUP` error on the socket during monitoring.

I found no record of libmongoc assigning a "connection failed" string via `bson_set_error()` in its topology scanner or async components, although there are some similar strings (e.g. ["connection closed"](https://github.com/mongodb/mongo-c-driver/blob/1.11.0/src/libmongoc/src/mongoc/mongoc-async.c#L148), ["connection timeout"](https://github.com/mongodb/mongo-c-driver/blob/1.11.0/src/libmongoc/src/mongoc/mongoc-async.c#L178)). It's possible that "connection failed" comes from some other source outside of libmongoc (perhaps a function that converts a socket error to a string).

In any event, I don't see the harm in relaxing this error message, as we're mainly testing for a ConnectionTimeoutException and server selection failure.

derickr:
I've merged this, but also into v1.5. GitHub doesn't seem to pick that up, so closing this manually. Thanks for the patch! (I'll test with the OpenSSL one (#875) tomorrow)

PR #879, opened 2018-07-17
 PHPC-1241: OpenSSL 1.1 not found if pkg-config is not available 

derickr:
I've taking @DaveRandom 's patch from #875 and added the conditional check for OpenSSL < 1.1 from @jmikola's comment on that. I did switch the order of the checking, to check for >= 1.1 before checking for < 1.1. I've tested this locally with 1. pkg-config available, 2. pkg-config unavailable and OpenSSL 1.0.2, and 3. pkg-config unavailable and OpenSSL 1.1.

### DIFF
--- a/scripts/autotools/libmongoc/CheckSSL.m4
+++ b/scripts/autotools/libmongoc/CheckSSL.m4
@@ -83,22 +83,21 @@ AS_IF([test "$PHP_MONGODB_SSL" = "openssl" -o "$PHP_MONGODB_SSL" = "auto"],[
                       [have_crypto_lib="no"],
                       [$OPENSSL_LIBDIR_LDFLAG])
 
-    have_ssl_lib="no"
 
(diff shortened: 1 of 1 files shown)


scripts/autotools/libmongoc/CheckSSL.m4
```diff
@@ -82,12 +82,24 @@ AS_IF([test "$PHP_MONGODB_SSL" = "openssl" -o "$PHP_MONGODB_SSL" = "auto"],[
                       [have_crypto_lib="yes"],
                       [have_crypto_lib="no"],
                       [$OPENSSL_LIBDIR_LDFLAG])
+
+
+    AC_MSG_NOTICE([checking whether OpenSSL >= 1.1.0 is available])
```
jmikola: Great idea reversing the order to prefer >= 1.1.0.

PR #880, opened 2018-07-17
PHPC-1229: Populate error labels for bulk write exceptions

derickr:
https://jira.mongodb.org/browse/PHPC-1229

### DIFF
--- a/tests/session/transaction-integration-003.phpt
+++ b/tests/session/transaction-integration-003.phpt
@@ -1,0 +1,58 @@
+--TEST--
+MongoDB\Driver\Session::startTransaction() Transient Error Test
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_clean(DATABASE_NAME, COLLECTION_NAME); ?>
+<?php skip_if_not_replica_set(); ?>
+<?php skip_if_server_version('<', '4.0'); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(URI);
+
+/* Create collections as that can't be (automatically) done in a transaction */
+$cmd = new \MongoDB\Driver\Command([
+    'create' => COLLECTION_NAME,
+]);
+$manager->executeCommand(DATABASE_NAME, $cmd);
+
+/* Insert Data */
+$bw = new \MongoDB\Driver\BulkWrite();
+$bw->insert( [ '_id' => 0, 'msg' => 'Initial Value' ] );
+$manager->executeBulkWrite(NS, $bw);
+
+/* First 'thread', try to update document, but don't close transaction */
+$sessionA = $manager->startSession();
+$sessionA->startTransaction( [
+    'readConcern' => new \MongoDB\Driver\ReadConcern( "snapshot" ),
+    'writeConcern' => new \MongoDB\Driver\WriteConcern( \MongoDB\Driver\WriteConcern::MAJORITY )
+] );
+
+$bw = new \MongoDB\Driver\BulkWrite();
+$bw->update( [ '_id' => 0 ], [ '$set' => [ 'msg' => 'Update from session A' ] ] );
+$manager->executeBulkWrite(NS, $bw, ['session' => $sessionA]);
+ 
+
+/* Second 'thread', try to update the same document, should trigger exception. In handler, commit
+ * first settion, verify result, and redo this transaction. */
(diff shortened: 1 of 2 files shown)


tests/session/transaction-integration-003.phpt
```diff
@@ -0,0 +1,58 @@
...
+
+try {
+    $bw = new \MongoDB\Driver\BulkWrite();
+    $bw->update( [ '_id' => 0 ], [ '$set' => [ 'msg' => 'Update from session B' ] ] );
+    $manager->executeBulkWrite(NS, $bw, ['session' => $sessionB]);
+} catch (MongoDB\Driver\Exception\BulkWriteException $e) {
+    echo $e->hasErrorLabel('TransientTransactionError') ?
+        "found a TransientTransactionError" : "did NOT get a TransientTransactionError", "\n";
+}
+?>
```
jmikola: This test would do well to explicitly commit/abort its transactions. Without doing so, the transaction may left open on the server for some time and and cause delays in subsequent tests using the same collection. You may be able to experience that by attempting to run this test multiple times.

See https://github.com/mongodb/mongo-php-library/pull/564/files#r202395078 for more context.

If we have other tests that are leaving transactions open, you may want to update those as well in a separate commit.

derickr: Updated, and added a commit to do it for -002 too (The only one where there could be a dangling transaction)

tests/session/transaction-integration-002.phpt
```diff
@@ -64,6 +64,9 @@ try {
     echo $e->hasErrorLabel('TransientTransactionError') ?
         "found a TransientTransactionError" : "did NOT get a TransientTransactionError", "\n";
 }
+
+$sessionA->abortTransaction();
+$sessionB->abortTransaction();
```
jmikola: See: https://github.com/mongodb/mongo-php-library/pull/564#discussion_r203069307

After some more testing, it looks like this isn't actually necessary. In [endSession changes](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#endsession-changes) the spec requires that an in-progress transaction be aborted when the session ends. This may only be a problem in https://github.com/mongodb/mongo-php-library/pull/564#discussion_r202395078 because of a libmongoc bug.

Since neither of us could reproduce the blocking I saw in PHPLIB in these PHPC tests, I'd suggest removing these explicit aborts. Even if libmongoc has a bug, it looks like it does abort the transaction before the process terminates.

tests/session/transaction-integration-003.phpt
```diff
@@ -0,0 +1,61 @@
...
+    $bw = new \MongoDB\Driver\BulkWrite();
+    $bw->update( [ '_id' => 0 ], [ '$set' => [ 'msg' => 'Update from session B' ] ] );
+    $manager->executeBulkWrite(NS, $bw, ['session' => $sessionB]);
+} catch (MongoDB\Driver\Exception\BulkWriteException $e) {
+    echo $e->hasErrorLabel('TransientTransactionError') ?
+        "found a TransientTransactionError" : "did NOT get a TransientTransactionError", "\n";
+}
+
+$sessionA->abortTransaction();
+$sessionB->abortTransaction();
```
jmikola: Per https://github.com/mongodb/mongo-php-driver/pull/880/files#r203070704, these can be removed. Sorry for the false alarm.

tests/session/transaction-integration-003.phpt
```diff
@@ -0,0 +1,58 @@
...
+<?php skip_if_server_version('<', '4.0'); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(URI);
+
+/* Create collections as that can't be (automatically) done in a transaction */
+$cmd = new \MongoDB\Driver\Command([
+    'create' => COLLECTION_NAME,
```
jmikola: I believe this should use a majority write concern. Per [Transactions and Locks](https://docs.mongodb.com/manual/core/transactions/#transactions-and-locks):

> When creating or dropping a collection immediately before starting a transaction, if the collection is accessed within the transaction, issue the create or drop operation with write concern "majority" to ensure that the transaction can acquire the required locks.

Would you mind updating that for all transaction tests? Feel free to make a separate ticket if you like (akin to [PHPLIB-377](https://jira.mongodb.org/browse/PHPLIB-377)), but I'm fine without.

jmikola: https://jira.mongodb.org/browse/PHPC-1244

PR #882, opened 2018-07-17
PHPC-1231: Add Session::isInTransaction to indicate whether a transaction is active

derickr:


src/MongoDB/Session.c
```diff
@@ -390,6 +390,24 @@ static PHP_METHOD(Session, endSession)
 	mongoc_client_session_destroy(intern->client_session);
 } /* }}} */
 
+/* {{{ proto void MongoDB\Driver\Session::isInTransaction(void)
+   Returns whether a multi-document transaction is in progress */
+static PHP_METHOD(Session, isInTransaction)
```
jmikola: This method name looks strange to me, but I think it's preferable to the more verbose option of `isTransactionInProgress()` or the like.

src/MongoDB/Session.c
```diff
@@ -390,6 +390,24 @@ static PHP_METHOD(Session, endSession)
...
+	SUPPRESS_UNUSED_WARNING(return_value_ptr)
+	SUPPRESS_UNUSED_WARNING(return_value_used)
+
+	intern = Z_SESSION_OBJ_P(getThis());
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	RETURN_BOOL(!!mongoc_client_session_in_transaction(intern->client_session));
```
jmikola: Is the quick bool cast with `!!` necessary? [The function](https://github.com/mongodb/mongo-c-driver/blob/ba4a1261b31be4aa103a601f63553191c1a65b04/src/libmongoc/src/mongoc/mongoc-client-session.c#L772) returns a `bool` and [`_mongoc_client_session_in_txn()`](https://github.com/mongodb/mongo-c-driver/blob/ba4a1261b31be4aa103a601f63553191c1a65b04/src/libmongoc/src/mongoc/mongoc-client-session.c#L909) uses `true` and `false` return values.

I would imagine there are a few other places where we trust libmongoc's `bool` return values.

src/MongoDB/Session.c
```diff
@@ -408,15 +426,16 @@ ZEND_END_ARG_INFO()
 
 static zend_function_entry php_phongo_session_me[] = {
 	/* clang-format off */
+	PHP_ME(Session, abortTransaction, ai_Session_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
 	PHP_ME(Session, advanceClusterTime, ai_Session_advanceClusterTime, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
 	PHP_ME(Session, advanceOperationTime, ai_Session_advanceOperationTime, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(Session, commitTransaction, ai_Session_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(Session, endSession, ai_Session_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(Session, isInTransaction, ai_Session_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
```
jmikola: Since it looks like you're alphabetizing these, `isInTransaction` should go after the `get` methods.

tests/session/session-isInTransaction-001.phpt
```diff
@@ -0,0 +1,36 @@
+--TEST--
+MongoDB\Driver\Session::isInTransaction()
```
jmikola: I think this test would do well to cover both the "started" and "in progress" transaction states. At the moment, I believe it only covers "started", which is before the first command is sent. See [`test_in_transaction()`](https://github.com/mongodb/mongo-c-driver/commit/8868a3bbe79fe9ca826b64fcb7df17c19fbb5bb7#diff-c96a6ebf9cd16bc38cf93c1d65dc69e8R112) for a good example. Ideally, we should just port that to a PHP test.

src/MongoDB/Session.c
derickr: Change it to ``inTransaction`` ? Don't like that much either...

src/MongoDB/Session.c
derickr: Removed the ``!!``.

src/MongoDB/Session.c
derickr: Yeah, I had alphabetised them in the first commit, ... and then put the new one in the wrong place :-)

tests/session/session-isInTransaction-001.phpt
derickr: Our test was basically that test already, but I added two sections that actually have an operation in the transaction too.

src/MongoDB/Session.c
jmikola: I didn't want to suggest `inTransaction` because it's not consistent with prefixing boolean-returning methods with "is". Totally OK with sticking with `isInTransaction()`.

tests/session/session-isInTransaction-001.phpt
```diff
@@ -0,0 +1,74 @@
...
+var_dump($session->isInTransaction());
+$session->abortTransaction();
+var_dump($session->isInTransaction());
+
+/* Empty transaction, and committed empty transaction */
+var_dump($session->isInTransaction());
+$session->startTransaction();
+var_dump($session->isInTransaction());
+$session->commitTransaction();
+var_dump($session->isInTransaction());
```
jmikola: Given that transactions are started lazily during the first command, how does this interact when the only operation is an abort or commit? I assume the driver is still sending something in this case and the server just registers it as a no-op?

Perhaps a question for @ShaneHarvey.

PR #886, opened 2018-07-19
Clean up SKIPIFs for session tests

jmikola:
Several things here:

 * All session tests require crypto to generate logical session IDs.
 * `skip_if_not_replica_set()` was not sufficient, as sessions also require 3.6+ (or 4.0+ for transactions). This happened to pass because Travis didn't have a job using a 3.0 replica set.
 * `skip_if_server_version()` should follow a test for a live server.
 * `skip_if_not_clean()` defaults to the namespace constants, so we need only provide arguments when custom collection names are used. This skip should also come last (after testing for a live server).


jmikola:
Rebased on `v1.5`. Changes for `session-isInTransaction-001.phpt` and `transaction-integration-003.phpt` can be made after merging up to master.

derickr:
The changes in here will likely conflict with the ones for fixing the MongoDB 3.0 replication PR from #887 — so this will need another rebase after that one is merged.

PR #887, opened 2018-07-19
 PHPC-1249: Fix tests running against MongoDB 3.0 replicaset 

derickr:


derickr:
You were a bit too eager approving this, Travis hadn't completed :-þ

jmikola:
Looks like that was caused by a malformed URI, which is skipping most tests.

I'll note that `tests/session/session_error-001.phpt` only failed because we've yet to remove the `new Manager(URI)` bit in https://github.com/mongodb/mongo-php-driver/pull/886/files#diff-d25ed7ef0c41dba8d45f80a9aeb26cb1R7. That highlights the fact that we should have a test that doesn't  talk to the server but does attempt to successfully parse the URI. May be worth adding that in its own ticket.

jmikola:
Created [PHPC-1248](https://jira.mongodb.org/browse/PHPC-1248) to track addition of a URI-validation test.

derickr:
That wasn't really the case though, as I had forgotten to add the MO config file. Without this session_error test failing, I would not have noticed it actually having been broken.

tests/replicaset/manager-getservers-001.phpt
```diff
@@ -116,7 +116,7 @@ array(3) {
     int(%d)
   }
 }
-%s:3000 - primary: 1, secondary: 0, arbiter: 0
-%s:3001 - primary: 0, secondary: 1, arbiter: 0
-%s:3002 - primary: 0, secondary: 0, arbiter: 1
+%s:%d0 - primary: 1, secondary: 0, arbiter: 0
+%s:%d1 - primary: 0, secondary: 1, arbiter: 0
+%s:%d2 - primary: 0, secondary: 0, arbiter: 1
```
jmikola: Missed these.

PR #892, opened 2018-07-23
PHPC-1244: Use majority WC to drop/create collections for transaction tests

derickr:


tests/session/transaction-integration-001.phpt
```diff
@@ -20,12 +20,20 @@ $manager = new MongoDB\Driver\Manager(URI);
 $cmd = new \MongoDB\Driver\Command([
     'create' => $EMPLOYEES_COL
 ]);
-$manager->executeCommand(DATABASE_NAME, $cmd);
+$manager->executeCommand(
+    DATABASE_NAME,
+    $cmd,
```
jmikola: If we're splitting `executeCommand()` up into multiple lines, I think we can easily inline the command itself, too.

Suggest replacing `$cmd` with `new \MongoDB\Driver\Command([ 'create' => $EMPLOYEES_COL ])`.

PR #893, opened 2018-07-23
PHPC-1239: Passing empty driversOptions array to Manager constructor overrides SSL options in URI

derickr:


php_phongo.c
```diff
@@ -1934,9 +1945,10 @@ static inline char* php_phongo_fetch_ssl_opt_string(zval* zoptions, const char*
 	return value;
 }
 
-static mongoc_ssl_opt_t* php_phongo_make_ssl_opt(zval* zoptions TSRMLS_DC)
+static mongoc_ssl_opt_t* php_phongo_make_ssl_opt(mongoc_uri_t* uri, zval* zoptions TSRMLS_DC)
 {
 	mongoc_ssl_opt_t* ssl_opt;
+	bool              any_ssl_option_set = false;
```
derickr: We still need any_ssl_option_set here, to see whether any changes were actually made. And if not, we need to deallocate the new ssl_opt and return NULL, so that we don't inadvertently end up setting an empty SSL opt struct on the client (that would turn SSL on).

Technically, that possibly means we don't have to do the ``mongoc_client_get_ssl`` check at https://github.com/mongodb/mongo-php-driver/pull/893/files#diff-c06c6e1c9374aecabcf544157f9d0c26R2479 any more, but I left that for now.

php_phongo.c
```diff
@@ -1921,6 +1921,17 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t* uri, bson_t* option
 } /* }}} */
 
 #ifdef MONGOC_ENABLE_SSL
+
+/* This is copied from mongoc-ssl.c */
+static void php_phongo_mongoc_ssl_opts_from_uri(mongoc_ssl_opt_t* ssl_opt, mongoc_uri_t* uri)
+{
+	ssl_opt->pem_file               = mongoc_uri_get_option_as_utf8(uri, MONGOC_URI_SSLCLIENTCERTIFICATEKEYFILE, NULL);
+	ssl_opt->pem_pwd                = mongoc_uri_get_option_as_utf8(uri, MONGOC_URI_SSLCLIENTCERTIFICATEKEYPASSWORD, NULL);
+	ssl_opt->ca_file                = mongoc_uri_get_option_as_utf8(uri, MONGOC_URI_SSLCERTIFICATEAUTHORITYFILE, NULL);
```
jmikola: [`mongoc_uri_get_option_as_utf8()`](http://mongoc.org/libmongoc/current/mongoc_uri_get_option_as_utf8.html) returns a `const char *` that is only valid for the lifetime of the URI structure.

I think these need to be run through `estrdup()` as we do in [`php_phongo_fetch_ssl_opt_string()`](https://github.com/mongodb/mongo-php-driver/blob/1.5.1/php_phongo.c#L1924) (if the string from `php_array.h` is not already duplicated). Ideally, we'd use `estrndup()` here but we don't have the string length (and real values likely shouldn't include null bytes).

Before you fix this, are you able to observe a double-free with this current implementation when `php_phongo_free_ssl_opt()` is called and runs `str_efree` on these three string fields?

php_phongo.c
```diff
@@ -1921,6 +1921,17 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t* uri, bson_t* option
...
+	ssl_opt->weak_cert_validation   = mongoc_uri_get_option_as_bool(uri, MONGOC_URI_SSLALLOWINVALIDCERTIFICATES, false);
+	ssl_opt->allow_invalid_hostname = mongoc_uri_get_option_as_bool(uri, MONGOC_URI_SSLALLOWINVALIDHOSTNAMES, false);
```
jmikola: Ideas for new test cases:

 * Check that any single SSL-related URI option enables SSL for the client
 * Confirm that the `$driverOptions` equivalent overrides the corresponding URI option
 * Confirm that a separate `$driverOptions` key leaves an SSL-related URI option in place

php_phongo.c
```diff
@@ -1963,6 +1975,14 @@ static mongoc_ssl_opt_t* php_phongo_make_ssl_opt(zval* zoptions TSRMLS_DC)
 
 	ssl_opt = ecalloc(1, sizeof(mongoc_ssl_opt_t));
 
+	/* If SSL options are set in the URL, we need to read them and set them on
+	 * the options struct so we can merge potential options from passed in
+	 * driverOptions (zoptions)*/
```
jmikola: Add a space between "(zoptions)" and "*".

php_phongo.c
jmikola: Another alternative would be declaring `mongoc_ssl_opt_t empty_ssl_opt = {0}` up top and later checking if `memcmp(ssl_opt, &empty_ssl_ssl_opt, sizeof(mongoc_ssl_opt_t))` returns zero later to check if any modifications have been made.

tests/manager/manager-ctor-ssl-002.phpt
```diff
@@ -0,0 +1,17 @@
+--TEST--
+PHPC-1239: Passing SSL driverOptions overrides SSL options from URI
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_live(); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(URI, array(), array());
```
jmikola: We can use `[]` instead of `array()` here. Down the line, I'd like to go back and clean up all `array()` usages.

tests/manager/manager-ctor-ssl-002.phpt
```diff
@@ -0,0 +1,17 @@
+--TEST--
+PHPC-1239: Passing SSL driverOptions overrides SSL options from URI
```
jmikola: If I understand correctly, this test is only significant when we're connecting over SSL, since MO will add SSL parameters in the URI string. I realize my previous PR review suggested creating this test, but I just want to sanity check as my mind has wandered since I last reviewed :)

php_phongo.c
```diff
@@ -1963,6 +1975,14 @@ static mongoc_ssl_opt_t* php_phongo_make_ssl_opt(zval* zoptions TSRMLS_DC)
...
+	if (mongoc_uri_get_ssl(uri)) {
```
jmikola: In https://github.com/mongodb/mongo-php-driver/pull/893/files#r204696800, you said this may no longer be necessary; however, I think without it you'd always end up assigning `any_ssl_option_set` to true.

As-is, there's still an edge case where `any_ssl_option_set` is true and we still end up returning an unmodified `mongoc_ssl_opt_t` struct. `mongoc_uri_get_ssl()` may return true even if only the `ssl=true` URI option is set. In that case, `php_phongo_mongoc_ssl_opts_from_uri()` isn't really copying anything and simply assigning defaults and we return an empty struct. Assigning the empty struct is a pointless copy and will redundantly enable SSL on the client.

I suppose a more correct approach would be passing `&any_ssl_option_set` into `php_phongo_mongoc_ssl_opts_from_uri()` so that it can be set only if actual values are copied (i.e. a non-null `const char *` or a true value for one of the two boolean options). _Then_, you could remove `mongoc_uri_get_ssl()` and guarantee that this function returns `NULL` if the struct is truly never modified (i.e. non-zero).

php_phongo.c
jmikola: Note that you'll probably want to create a local `const char *` to re-use for each of these three options:

 * Declare `const char * tmp`
 * `tmp = mongoc_uri_get_option_as_utf8(...)`
 * If `tmp` isn't null, `strdup()` and assign to the struct

php_phongo.c
```diff
@@ -1921,6 +1921,17 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t* uri, bson_t* option
 } /* }}} */
 
 #ifdef MONGOC_ENABLE_SSL
+
+/* This is copied from mongoc-ssl.c */
```
jmikola: Looking at [`_mongoc_ssl_opts_from_uri()`](https://github.com/mongodb/mongo-c-driver/blob/1.12.0/src/libmongoc/src/mongoc/mongoc-ssl.c#L97), I wonder if libmongoc also has a bug where it's copying these `const char *` into the `mongoc_ssl_opt_t` and later allowing them to be freed. I think the same fix I propose in https://github.com/mongodb/mongo-php-driver/pull/893/files#r204735780 to duplicate them first is warranted. If so, I'd suggest creating a CDRIVER ticket for that and opening a PR.

php_phongo.c
derickr: Yeah, that's an alternative, but not one I like. :-þ

php_phongo.c
derickr: I meant for: https://github.com/mongodb/mongo-php-driver/pull/893/files#diff-c06c6e1c9374aecabcf544157f9d0c26R2479

tests/manager/manager-ctor-ssl-002.phpt
derickr: It's relevant only for SSL, but that doesn't mean we can't spend the 0.0005 seconds to run it on the other topologies...

tests/manager/manager-ctor-ssl-002.phpt
derickr: We can fix that then then. It's ``array()`` for now here, just as in the original test.

php_phongo.c
jmikola: At the end of manager initialization, the `mongoc_ssl_opt_t` returned by `php_phongo_make_ssl_opt()` [will be freed](https://github.com/mongodb/mongo-php-driver/blob/1.5.1/php_phongo.c#L2462) via [`php_phongo_free_ssl_opt()`](https://github.com/mongodb/mongo-php-driver/blob/1.5.1/php_phongo.c#L2009). That will involve calling `str_efree()` on every non-NULL string on the struct before we `efree()` the struct pointer itself.

Currently, some of those strings may end up being `const char *` that point to memory within the `mongoc_uri_t`. That data will have been freed when [`phongo_manager_init()` calls [`mongoc_uri_destroy()`](https://github.com/mongodb/mongo-php-driver/blob/1.5.1/php_phongo.c#L2458) immediately before the `php_phongo_free_ssl_opt()` call. This can cause the double-free.

When `mongoc_client_set_ssl_opts()` talks about duplicating strings, it's referring to copying the `mongoc_ssl_opt_t` into the client/pool. In addition to copying the struct, it duplicates any non-NULL strings to avoid referring back to any memory from the original `mongoc_ssl_opt_t` that was copied. In libmongoc's case, they [use a stack-allocated struct](https://github.com/mongodb/mongo-c-driver/blob/1.12.0/src/libmongoc/src/mongoc/mongoc-client.c#L1030) to quickly collect strings from the URI and ferry them over to a client/pool. This stack-allocated `mongoc_ssl_opt_t` only exists for a subset of the corresponding `mongoc_uri_t` lifetime, so they don't have to worry about freeing pointers like `mongoc_ssl_opt_t.pem_pwd` at all.

php_phongo.c
jmikola: The original description in [PHPC-1239](https://jira.mongodb.org/browse/PHPC-1239) mentioned a URI with `ssl=true&sslallowinvalidcertificates=true`. Where did `sslallowinvalidcertificates=true` come from?

Assuming our MO environment still depends on self-signed certificates, I suppose we could rely on these undocumented URI options instead of passing the equivalent in `$driverOptions`. And the test could set an unrelated key in `$driverOptions` just to demonstrate that `sslallowinvalidcertificates=true` URI option is still respected.

We may not be able to use this approach for everything, but perhaps it can test one case, which would be better than none.

php_phongo.c
jmikola: Assuming `php_phongo_mongoc_ssl_opts_from_uri()` will need to be updated for https://github.com/mongodb/mongo-php-driver/pull/893#discussion_r204735780, can we also pass in `&any_ssl_option_set` so it's only set when a struct field is actually assigned? That would still remove this conditional.

Alternatively, I suppose the code from `php_phongo_mongoc_ssl_opts_from_uri()` can simply be inlined in this function.

jmikola:
Superseded by https://github.com/mongodb/mongo-php-driver/pull/897.

php_phongo.c
jmikola: Disregard. This wasn't a bug in libmongoc as their struct was stack-allocated.

PR #894, opened 2018-07-23
Remove unused autotools macro AC_COMPILE_CHECK_SIZEOF

petk:
Hello,

The `AC_COMPILE_CHECK_SIZEOF` is from the [GNU Autoconf Archive](https://www.gnu.org/software/autoconf-archive/) and was renamed to `AX_COMPILE_CHECK_SIZEOF`. However this macro is not used in current MongoDB PHP extension nor in the mongo-c-driver dependency anymore.


derickr:
I've merged this after rebasing. Thanks again, and I'm closing this out.

PR #895, opened 2018-07-23
Replace obsolete AC_TRY_LINK with AC_LINK_IFELSE

petk:
The `AC_TRY_LINK` macro has been made obsolete in Autoconf 2.50 in 2001. It should be replaced with `AC_LINK_IFELSE` instead.

Systems should by now well support this, since PHP 5.3 required Autoconf 2.13+ (released in 1999) or newer, PHP 5.4 required Autoconf 2.59+ (released in 2003), and since PHP 7.2, Autoconf 2.64+ (released in 2008) is required and for PHP 7.2 phpize the Autoconf 2.59+ is required.

Refs:
- https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/AC_005fACT_005fIFELSE-vs-AC_005fTRY_005fACT.html
- https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Obsolete-Macros.html

### DIFF
--- a/scripts/autotools/libmongoc/CheckResolv.m4
+++ b/scripts/autotools/libmongoc/CheckResolv.m4
@@ -7,20 +7,20 @@ old_LIBS="$LIBS"
 LIBS="$LIBS -lresolv"
 
 dnl Thread-safe DNS query function for _mongoc_client_get_srv.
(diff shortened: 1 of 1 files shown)


derickr:
I've merged this after rebasing. Thanks again, and I'm closing this out.

PR #896, opened 2018-07-24
Fix AC_CHECK_FUNC macro call syntax for rand_r

petk:
The [AC_CHECK_FUNC](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Generic-Functions.html) macro has only the following form:

```m4
AC_CHECK_FUNC(function, [action-if-found], [action-if-not-found])
````
In most cases adding more arguments to macro call don't cause errors, but it also doesn't have any functionality either, so spotting such typos is not obvious at first glance.

The second argument of the macro `[]`, can be also left out.


derickr:
Thank for the patch, I've merged it.

PR #897, opened 2018-07-25
PHPC-1239: Passing SSL driversOptions to Manager constructor overrides SSL options in URI

derickr:
https://jira.mongodb.org/browse/PHPC-1239

Starting over, as there were too many unrelated and wrong comments on the PR, and I lost track.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1921,6 +1921,29 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t* uri, bson_t* option
 } /* }}} */
 
 #ifdef MONGOC_ENABLE_SSL
+
+static void php_phongo_mongoc_ssl_opts_from_uri(mongoc_ssl_opt_t* ssl_opt, mongoc_uri_t* uri, bool* any_ssl_option_set)
+{
+	const char* pem_file = mongoc_uri_get_option_as_utf8(uri, MONGOC_URI_SSLCLIENTCERTIFICATEKEYFILE, NULL);
+	const char* pem_pwd  = mongoc_uri_get_option_as_utf8(uri, MONGOC_URI_SSLCLIENTCERTIFICATEKEYPASSWORD, NULL);
+	const char* ca_file  = mongoc_uri_get_option_as_utf8(uri, MONGOC_URI_SSLCERTIFICATEAUTHORITYFILE, NULL);
+
+	ssl_opt->pem_file               = pem_file ? estrdup(pem_file) : NULL;
+	ssl_opt->pem_pwd                = pem_pwd ? estrdup(pem_pwd) : NULL;
+	ssl_opt->ca_file                = ca_file ? estrdup(ca_file) : NULL;
+	ssl_opt->weak_cert_validation   = mongoc_uri_get_option_as_bool(uri, MONGOC_URI_SSLALLOWINVALIDCERTIFICATES, false);
+	ssl_opt->allow_invalid_hostname = mongoc_uri_get_option_as_bool(uri, MONGOC_URI_SSLALLOWINVALIDHOSTNAMES, false);
+
+	/* Boolean options default to false, so we cannot consider them for
+	 * any_ssl_option_set. This isn't actually a problem as libmongoc will
+	 * already have assigned them when creating the client, enabling SSL, and
+	 * assigning SSL options. Therefore, we only need to check for non-defaults
+	 * (i.e. non-NULL strings, true booleans). */
+	if (pem_file || pem_pwd || ca_file || ssl_opt->weak_cert_validation || ssl_opt->allow_invalid_hostname) {
+		*any_ssl_option_set = true;
+	}
+}
+
 static inline char* php_phongo_fetch_ssl_opt_string(zval* zoptions, const char* key, int key_len)
 {
 	int       plen;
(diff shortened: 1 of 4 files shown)


php_phongo.c
```diff
@@ -1963,44 +1975,69 @@ static mongoc_ssl_opt_t* php_phongo_make_ssl_opt(zval* zoptions TSRMLS_DC)
 
 	ssl_opt = ecalloc(1, sizeof(mongoc_ssl_opt_t));
 
+	/* If SSL options are set in the URL, we need to read them and set them on
+	 * the options struct so we can merge potential options from passed in
+	 * driverOptions (zoptions) */
+	if (mongoc_uri_get_ssl(uri)) {
+		php_phongo_mongoc_ssl_opts_from_uri(ssl_opt, uri);
+		any_ssl_option_set = true;
```
jmikola: My comments in https://github.com/mongodb/mongo-php-driver/pull/893#discussion_r204740634 still apply here. This may set `any_ssl_option_set` to true even if `php_phongo_mongoc_ssl_opts_from_uri()` doesn't actually do anything, because `mongoc_uri_get_ssl()` will return true when only `ssl=true` is set in the URI (with none of the other SSL-related options).

My suggestion was to either pass in the address of `any_ssl_option_set` to have `php_phongo_mongoc_ssl_opts_from_uri()` set it accordingly (as you do later in this very function for each option) or to just inline `php_phongo_mongoc_ssl_opts_from_uri()` here.

php_phongo.c
```diff
@@ -1963,44 +1975,69 @@ static mongoc_ssl_opt_t* php_phongo_make_ssl_opt(zval* zoptions TSRMLS_DC)
...
 		ssl_opt->weak_cert_validation = php_array_fetchc_bool(zoptions, "weak_cert_validation");
+		any_ssl_option_set            = true;
 	} else if (php_array_existsc(zoptions, "allow_self_signed")) {
 		ssl_opt->weak_cert_validation = php_array_fetchc_bool(zoptions, "allow_self_signed");
+		any_ssl_option_set            = true;
 	}
 
 	if (php_array_existsc(zoptions, "pem_file")) {
-		ssl_opt->pem_file = php_phongo_fetch_ssl_opt_string(zoptions, ZEND_STRL("pem_file"));
+		ssl_opt->pem_file  = php_phongo_fetch_ssl_opt_string(zoptions, ZEND_STRL("pem_file"));
```
jmikola: I think there is a risk of leaking memory for any strings that `php_phongo_mongoc_ssl_opts_from_uri()` previously copied with `estrdup()`. Before assigning any strings on the `mongoc_ssl_opt_t` struct after the call to `php_phongo_mongoc_ssl_opts_from_uri()`, you'll need to check if the string is non-null and then `efree()` it.

I might suggest just creating a temporary macro before the first time you need to do this and then undefining it before the end of this function. We do that elsewhere for temporary macros in BulkWrite.c and Query.c.

Before fixing this, it'd be wise to add a test that demonstrates the memory leak. It actually shouldn't require a live server, as we're just building the libmongoc client. You should use `skip_if_not_libmongoc_ssl()`, though.

derickr: Iv'e updated the code, and added a test file (``tests/manager/manager-set-uri-options-003.phpt``)

php_phongo.c
derickr: I've done that now for ``pem_file || pem_pwd || ca_file`` as these are ``null`` if not set, but it's hard to say whether ``weak_cert_validation`` or ``allow_invalid_hostname`` are like that — as they're booleans defaulting to ``false``.

php_phongo.c
```diff
@@ -1921,6 +1921,25 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t* uri, bson_t* option
 } /* }}} */
 
 #ifdef MONGOC_ENABLE_SSL
+
+/* This is copied from mongoc-ssl.c */
```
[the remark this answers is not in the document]
derickr: Removed.

php_phongo.c
```diff
@@ -1921,6 +1921,25 @@ static bool php_phongo_apply_wc_options_to_uri(mongoc_uri_t* uri, bson_t* option
...
+	const char* pem_pwd  = mongoc_uri_get_option_as_utf8(uri, MONGOC_URI_SSLCLIENTCERTIFICATEKEYPASSWORD, NULL);
+	const char* ca_file  = mongoc_uri_get_option_as_utf8(uri, MONGOC_URI_SSLCERTIFICATEAUTHORITYFILE, NULL);
+
+	ssl_opt->pem_file               = pem_file ? estrdup(pem_file) : NULL;
+	ssl_opt->pem_pwd                = pem_pwd ? estrdup(pem_pwd) : NULL;
+	ssl_opt->ca_file                = ca_file ? estrdup(ca_file) : NULL;
+	ssl_opt->weak_cert_validation   = mongoc_uri_get_option_as_bool(uri, MONGOC_URI_SSLALLOWINVALIDCERTIFICATES, false);
+	ssl_opt->allow_invalid_hostname = mongoc_uri_get_option_as_bool(uri, MONGOC_URI_SSLALLOWINVALIDHOSTNAMES, false);
+
+	if (pem_file || pem_pwd || ca_file) {
```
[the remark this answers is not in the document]
derickr: I've added `` || ssl_opt->weak_cert_validation || ssl_opt->allow_invalid_hostname`` and the comment you wrote.

PR #898, opened 2018-07-26
PHPC-1060: Atlas connectivity tests

derickr:


.travis.yml
```diff
@@ -54,6 +58,7 @@ matrix:
         - SERVER_VERSION=3.6.5
 
 before_install:
+  - openssl aes-256-cbc -K $encrypted_b354efda2943_key -iv $encrypted_b354efda2943_iv -in .travis.scripts/atlas-uris.txt.enc -out .travis.scripts/atlas-uris.txt -d
```
derickr: Couldn't use just encrypted env-vars, as they were "too much data" — encrypted files are suggested at https://docs.travis-ci.com/user/encrypting-files/#automated-encryption

tests/atlas.phpt
```diff
@@ -0,0 +1,40 @@
+--TEST--
+Atlas Connectivity Tests
+--SKIPIF--
+<?php
+if (!file_exists('.travis.scripts/atlas-uris.txt')) { echo "skip Atlas URIs not found\n"; }
```
derickr: We need to test for this, as in forks, this file can't be unencrypted (i.e., for Pull Requests)

tests/atlas.phpt
```diff
@@ -0,0 +1,40 @@
...
+if ($_ENV['TESTS'] !== 'tests/atlas.phpt') { echo "skip Atlas tests not wanted\n"; }
```
jmikola: Do we still need a CI matrix entry that overrides `TESTS` to just run this Atlas test?

jmikola: To elaborate, normally the test suite is run with `make test` and `TESTS` is not explicitly set. Does this test only get run when running `make test TESTS=tests/atlas.phpt`?

Is there ever a case where `$_ENV['TESTS']` would not be set and this could raise a warning?

derickr: Technically, this could warn. But skip-if code never throws a warning, as ``display_errors`` is disabled for SKIPIFs: https://github.com/php/php-src/blob/master/run-tests.php#L1626

.travis.yml
```diff
@@ -16,6 +16,10 @@ env:
 
 matrix:
   include:
+    - php: 7.2
+      env:
+        - TEST_PHP_ARGS="-q -s output.txt -x --show-diff"
+        - TESTS=tests/atlas.phpt
```
jmikola: Was this added after https://github.com/mongodb/mongo-php-driver/pull/898/files#r205480285, or did I just miss this during the initial review.

derickr: You missed it :-)

derickr:
Merged after rebase.

PR #901, opened 2018-07-27
PHPC-1227: Add test for disallowed per-op read/write concerns in transactions

derickr:
This is just a test - not sure which more to add. Documentation still to follow.

### DIFF
--- a/tests/session/transaction-integration_error-001.phpt
+++ b/tests/session/transaction-integration_error-001.phpt
@@ -1,0 +1,70 @@
+--TEST--
+MongoDB\Driver\Session: Setting per-op readConcern or writeConcern in transaction
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_libmongoc_crypto(); ?>
+<?php skip_if_not_replica_set(); ?>
+<?php skip_if_server_version('<', '4.0'); ?>
+<?php skip_if_not_clean(); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(URI);
+
+/* Create collections as that can't be (automatically) done in a transaction */
+$manager->executeCommand(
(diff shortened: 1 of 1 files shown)


tests/session/transaction-integration_error-001.phpt
```diff
@@ -0,0 +1,70 @@
+--TEST--
+MongoDB\Driver\Session: Setting per-op readConcern or writeConcern in transaction
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_libmongoc_crypto(); ?>
+<?php skip_if_not_replica_set(); ?>
+<?php skip_if_server_version('<', '4.0'); ?>
+<?php skip_if_not_clean(DATABASE_NAME, COLLECTION_NAME); ?>
```
jmikola: `skip_if_not_clean()`. No need to use arguments unless you're customizing the defaults (e.g. custom collection).

tests/session/transaction-integration_error-001.phpt
```diff
@@ -0,0 +1,70 @@
...
+            'writeConcern' => new \MongoDB\Driver\WriteConcern(
+                \MongoDB\Driver\WriteConcern::MAJORITY ) 
+        ]
+    );
+}, "MongoDB\Driver\Exception\InvalidArgumentException"), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
```
jmikola: `--EXPECT--`

tests/session/transaction-integration_error-001.phpt
derickr: Fixed.

tests/session/transaction-integration_error-001.phpt
derickr: I think we should be okay with ``EXPECTF`` — it doesn't make any difference, and makes things easier if tests need to be modified for changes in version — there are plenty of situations where we had to do that.

I'll fix it for this one.

PR #902, opened 2018-07-30
PHPC-1227: Add more tests for disallowed per-op read/write concerns in transactions

derickr:
Added tests for ``executeReadCommand``, ``executeWriteCommand`` and ``executeReadWriteCommand`` too.

### DIFF
--- a/tests/session/transaction-integration_error-004.phpt
+++ b/tests/session/transaction-integration_error-004.phpt
@@ -1,5 +1,5 @@
 --TEST--
-MongoDB\Driver\Session: Setting per-op readConcern or writeConcern in transaction (executeCommand)
+MongoDB\Driver\Session: Setting per-op readConcern or writeConcern in transaction (executeReadWriteCommand)
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
 <?php skip_if_not_libmongoc_crypto(); ?>
@@ -30,10 +30,10 @@ $session->startTransaction( [
 
 echo throws(function() use ($manager, $session) {
     $cmd = new \MongoDB\Driver\Command( [
-        'update' => COLLECTION_NAME,
-        'updates' => [ [ 'q' => [ 'employee' => 3 ], 'u' => [ '$set' => [ 'status' => 'Inactive' ] ] ] ]
+        'count' => COLLECTION_NAME,
+        'query' => [ 'q' => [ 'employee' => 3 ] ]
     ] );
-    $manager->executeCommand(
+    $manager->executeReadWriteCommand(
         DATABASE_NAME,
         $cmd,
         [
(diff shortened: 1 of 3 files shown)


tests/session/transaction-integration_error-004.phpt
```diff
@@ -0,0 +1,70 @@
...
+
+$session->startTransaction( [
+    'readConcern' => new \MongoDB\Driver\ReadConcern( "snapshot" ),
+    'writeConcern' => new \MongoDB\Driver\WriteConcern( \MongoDB\Driver\WriteConcern::MAJORITY )
+] );
+
+echo throws(function() use ($manager, $session) {
+    $cmd = new \MongoDB\Driver\Command( [
+        'update' => COLLECTION_NAME,
+        'updates' => [ [ 'q' => [ 'employee' => 3 ], 'u' => [ '$set' => [ 'status' => 'Inactive' ] ] ] ]
```
jmikola: I'm not sure this is the best command, as [`update`](https://docs.mongodb.com/manual/reference/command/update/) doesn't actually support a `readConcern` option. Based on the list of supported commands in [the spec](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#read-concern) it looks like `readConcern` is not an option for any commands that write (e.g. `aggregate` supports `readConcern` but not when `$out` is used). In that case, you'd do better to simply execute a proper read command here.

I realize this doesn't matter from an integration standpoint, as the driver will throw long before we send this command to the server, but it would make for a more correct example in the test if anyone reads this later.

derickr: I think you added this to the wrong section, as we're really just checking a ``writeConcern`` here. The comment should have belonged to the previous [``throws`` as there we use a ``readConcern``](https://github.com/mongodb/mongo-php-driver/pull/902/files#diff-01a050f3d755cc3e7644ad65afc22ee2R41), which I've updated.

jmikola: > I think you added this to the wrong section

Indeed. This was intended for the previous section.

jmikola:
> I couldn't actually find a command that takes both readConcern and writeConcern. Not even findAndModify.

I don't think there are such commands. See discussion in [SPEC-1138](https://jira.mongodb.org/browse/SPEC-1138), where I posed that question and asked that we update the RC/WC spec accordingly.

The server documentation doesn't come out and say that "no command accepts both a RC and WC concurrently", but we have the following to support that idea:

 * [Linearizable Reads via `findAndModify`](https://docs.mongodb.com/manual/tutorial/perform-findAndModify-linearizable-reads/) demonstrates how to enforce a particular read concern without using a `readConcern` option.
 * [Override Default Read Concern](https://docs.mongodb.com/manual/reference/command/aggregate/#override-default-read-concern) mentions that `aggregate` with `$out` doesn't support a "majority" read concern (only "local"). It doesn't talk about other levels, but I believe that's just because this documentation hasn't been updated since the first two levels were introduced.

PR #903, opened 2018-07-30
PHPC-1250: Best practices for SKIPIF and improve Testing docs

jmikola:
https://jira.mongodb.org/browse/PHPC-1250

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,38 @@ make clean > /dev/null && make all > /dev/null && make install
 
 ## Testing
 
+The extension's test use the PHPT format from PHP internals. This format is
+documented in the following links:
+
+ * [Introduction to PHPT Files](https://qa.php.net/write-test.php)
+ * [PHPT - Test File Layout](https://qa.php.net/phpt_details.php)
+
+Generally, most tests will be based on the following template:
+
+```
+--TEST--
+Description of API or JIRA issue being tested
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php /* One or more skip functions */ ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+// Test code
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -23,6 +23,74 @@ make clean > /dev/null && make all > /dev/null && make install
...
+should be logically ordered:
+
+ * Any PHP environment requirements should be checked first. For example, if a
+   test requires a 64-bit architecture, start by checking `PHP_INT_SIZE` before
+   anything else.
+ * Any extension build requirements (e.g. `skip_if_not_libmongoc_crypto()`) or
+   test environment requirements (e.g. `skip_if_auth()`) should then be checked.
+   These functions only examine local information, such as `phpinfo()` output or
+   the structure of the `URI` constant, and do not interact with a remote
+   MongoDB server.
```
derickr: How do we remember in which category each ``skip_if`` falls in though? Would it make sense to add this information to the function names?

Alternatively, could we automate the reordering of these functions through a script perhaps? I want to get rid of the mental overhead of having to think about this :-)

jmikola: > How do we remember in which category each ``skip_if`` falls in though? Would it make sense to add this information to the function names?

I don't think prefixing these functions (e.g. `0_skip_if_not...`) is worth it. I'd also be hesitant to indicate the level in a doc block, as that seems prone to getting out-of-date if the implementation changed for some reason.

Frankly, these are all mostly one-line functions in `skipif.php` or `tools.php` so I think it's best to just take a look at the code until this becomes familiar. This is definitely just a best practice and not a technical requirement. If you get things out of order and an error is thrown (e.g. trying to check a server version before ensuring you have a live server), I still catch the error/exception and emit a "skip" line.

> Alternatively, could we automate the reordering of these functions through a script perhaps? I want to get rid of the mental overhead of having to think about this :-)

If you'd like to spend time writing such a script that scans `phpt` file for best practices, feel free. I'd propose you also add a feature to catch other code smells like using `EXPECTF` without format specifiers (e.g. `%s`).

Otherwise, I'm happy to just continue pointing out bad ordering of `skip_if` functions in code reviews as I see them.

PR #905, opened 2018-08-06
PHPC-1184: Added sharded cluster with replicaset

derickr:


.travis.yml
```diff
@@ -58,7 +61,7 @@ matrix:
         - SERVER_VERSION=3.6.5
 
 before_install:
-  - openssl aes-256-cbc -K $encrypted_b354efda2943_key -iv $encrypted_b354efda2943_iv -in .travis.scripts/atlas-uris.txt.enc -out .travis.scripts/atlas-uris.txt -d
+  - openssl aes-256-cbc -K $encrypted_b354efda2943_key -iv $encrypted_b354efda2943_iv -in .travis.scripts/atlas-uris.txt.enc -out .travis.scripts/atlas-uris.txt -d || true
```
jmikola: Does this belong in the commit for adding a sharded replica set? Looks like a fix for the automated Atlas testing ticket.

scripts/presets/travis/sharded_clusters/cluster_replset.json
```diff
@@ -0,0 +1,117 @@
...
+                    "logappend": true,
+                    "nssize": 1,
+                    "port": 4492,
+                    "bind_ip_all": true,
+                    "smallfiles": true
+                }
+            }
+        ]
+    } ],
+    "id": "cluster_2",
```
jmikola: Would "cluster_rs" make more sense, given the IDs you use below? Either way, I assume this name isn't important as we don't reference it in URIs.

.travis.yml
jmikola: Does this belong in the commit for adding a sharded replica set? Looks like a fix for the automated Atlas testing ticket.

tests/atlas.phpt
```diff
@@ -3,6 +3,7 @@ Atlas Connectivity Tests
 --SKIPIF--
 <?php
 if (!file_exists('.travis.scripts/atlas-uris.txt')) { echo "skip Atlas URIs not found\n"; }
+if (filesize('.travis.scripts/atlas-uris.txt') < 10) { echo "skip Atlas URI file empty\n"; }
```
jmikola: This likely also belongs in a separate commit. I assume there was some error where decrypting failed and left a tiny file? If so, is there a better way to catch that in the Travis shell scripts instead of this curious conditional?

.travis.yml
derickr: Without it, PR builds on Travis would always fail as openssl bailed out and aborted the test run. It should go into its own commit, before the PHPC-1184 sharded-rs cluster. I'll break them up.

scripts/presets/travis/sharded_clusters/cluster_replset.json
derickr: Doesn't matter, but I'll change it.

tests/atlas.phpt
derickr: Doing that in the Travis shell scripts was messy, and complicated, (and not working). As we're only doing this once, I felt that a simple solution in the PHP skipif was more warranted.

PR #907, opened 2018-08-08
PHPC-1177: Reimplement parse_url tests

derickr:


tests/server/server-construct-001.phpt
```diff
@@ -10,7 +8,13 @@ parse_url() tests must be reimplemented (PHPC-1177)
 <?php
 require_once __DIR__ . "/../utils/basic.inc";
 
-$parsed = parse_url(STANDALONE);
+$parsed = parse_url(URI);
```
jmikola: This is redundant in the case of a replica set. Ternary assignment would help here:

```php
$uri = is_replica_set(URI) ? get_primary_server(URI)->getInfo()['me'] : URI;
$parsed = parse_url($uri);
```

scripts/presets/travis/standalone/standalone-x509.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "mongod",
+    "id" : "STANDALONE_X509",
+    "authSource": "\$external",
```
jmikola: "$" is not a special character in JSON. There should be no need to escape here.

derickr: But it was definitely necessary. It's likely to do with the way how the ``mo.sh`` tool handles sending things to Mongo Orchestration.

tests/utils/skipif.php
```diff
@@ -97,7 +97,7 @@ function skip_if_not_auth_mechanism($authMechanism)
 {
     $uriAuthMechanism = get_uri_option(URI, 'authMechanism');
 
-    if ($uriAuthMechanism === null) {
+    if ($uriAuthMechanism === null && $authMechanism !== null) {
```
jmikola: This shouldn't be necessary, as `is_auth()` already returns `true` if `authMechanism` is set to something. Better to just use `skip_if_not_auth()`.

Also, if we were to keep this behavior, I'd suggest defaulting the argument to `null` and changing the doc block accordingly to mention that we check for "anything" if no argument is provided.

tests/server/server-executeBulkWrite-005.phpt
```diff
@@ -24,7 +24,10 @@ foreach ($writeConcerns as $wc) {
     var_dump($result->getInsertedCount());
 }
 
-$command = new MongoDB\Driver\Command(array('drop' => COLLECTION_NAME));
+$command = new MongoDB\Driver\Command([
+    'delete' => COLLECTION_NAME,
+    'deletes' => [ [ 'q' => (object) [], 'limit' => 0 ] ]
+]);
```
jmikola: Is this to avoid issue with not having permissions to `drop` (althouga h writes are allowed)?

Rather than hard-code delete commands like this, I think it'd be clearer to use `executeBulkWrite()`.

tests/utils/skipif.php
derickr: This is necessary to make sure there **is** authentication, and it is *not* X509. ``skip_if_not_auth()`` returns true for either "standard", where ``authMechanism`` isn't set to anything, or ``MONGO-X509`` auth.

It's used at https://github.com/mongodb/mongo-php-driver/pull/907/files#diff-15f3c9a5beac9f38bb34f9eef5d058f3R6 to make sure we're just dealing with normal simple auth.

tests/server/server-executeBulkWrite-005.phpt
derickr: "althouga h" ← **:-þ**

Yes, that is the reason. I'll change it to use ``executeBulkWrite()``.

PR #910, opened 2018-08-14
[WAIT] PHPC-1262: Re-enable parse_url tests for X509 credentials 

derickr:


tests/connect/standalone-x509-error-001.phpt
```diff
@@ -1,11 +1,10 @@
 --TEST--
 X509 connection should not reuse previous stream after an auth failure
---XFAIL--
-parse_url() tests must be reimplemented (PHPC-1177)
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
 <?php skip_if_not_libmongoc_ssl(); ?>
 <?php skip_if_not_ssl(); ?>
+<?php skip_if_not_standalone(); ?>
```
jmikola: Is this really necessary given that we have `skip_if_not_auth_mechanism('MONGODB-X509')` on the next line? If the test only depends on X509 it shouldn't care about the topology (or need to assume that our only X509 environment also happens to be a standalone).

jmikola:
Is this still in progress given the test failures in https://travis-ci.org/mongodb/mongo-php-driver/jobs/415904721?

PR #913, opened 2018-08-17
 PHPC-1179: Reimplement tests that start servers with Mongo Orchestration 

derickr:
These tests could not have worked since we moved from a Wire Protocol getMore OP to the OP_COMMAND style of calling ``getMore``. Beats me how we didn't manage to detect this.

The second commit isn't strictly necessary, but it updates the code to be able to throw a ``ConnectionException`` if getMore fails.

Previously, the ``getMore`` tests use a failPoint for the getMore OP, where the driver would then (through ``libmongoc``) throw a ``ConnectionException``.  With the shift to the command-based failPoint, ``libmongoc``/the driver now converted that to a ``ServerException``. This second commit restores that behaviour. Alternatively, we can just update the test to check for a ``ServerException`` instead.

### DIFF
--- a/tests/cursor/cursor-getmore-005.phpt
+++ b/tests/cursor/cursor-getmore-005.phpt
@@ -1,15 +1,20 @@
 --TEST--
 MongoDB\Driver\Cursor query result iteration with getmore failure
---XFAIL--
-START() tests must be reimplemented (PHPC-1179)
 --SKIPIF--
 <?php require __DIR__ . "/" ."../utils/basic-skipif.inc"; ?>
-<?php START("THROWAWAY", ["version" => "30-release"]); CLEANUP(THROWAWAY); ?>
+<?php
+/* This test spins up its own mongod instance, so only run this in the most default "standalone, no
+ * auth" configurations. This way, we can test on multiple server versions, but not waste resources
+ * on f.e. Travis. */
+?>
+<?php skip_if_not_standalone(); ?>
+<?php skip_if_auth(); ?>
 --FILE--
 <?php
 require_once __DIR__ . "/../utils/basic.inc";
 
-$manager = new MongoDB\Driver\Manager(THROWAWAY);
+$uri = createTemporaryMongoInstance();
+$manager = new MongoDB\Driver\Manager($uri);
 
 $bulkWrite = new MongoDB\Driver\BulkWrite;
 
@@ -30,14 +35,13 @@ throws(function() use ($cursor) {
         printf("%d => {_id: %d}\n", $i, $document->_id);
     }
 }, "MongoDB\Driver\Exception\ConnectionException");
-
 ?>
 ===DONE===
-<?php DELETE("THROWAWAY"); ?>
+<?php destroyTemporaryMongoInstance(); ?>
 <?php exit(0); ?>
 --CLEAN--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
-<?php DELETE("THROWAWAY"); ?>
+<?php destroyTemporaryMongoInstance(); ?>
 --EXPECT--
 Inserted: 5
 0 => {_id: 0}
--- a/tests/cursor/cursor-getmore-005.phpt
+++ b/tests/cursor/cursor-getmore-005.phpt
@@ -7,7 +7,9 @@ MongoDB\Driver\Cursor query result iteration with getmore failure
  * auth" configurations. This way, we can test on multiple server versions, but not waste resources
  * on f.e. Travis. */
 ?>
+<?php skip_if_not_live(); ?>
 <?php skip_if_not_standalone(); ?>
+<?php skip_if_no_getmore_failpoint(); ?>
 <?php skip_if_auth(); ?>
 --FILE--
 <?php
--- a/tests/utils/skipif.php
+++ b/tests/utils/skipif.php
@@ -226,3 +226,15 @@ function skip_if_not_clean($databaseName = DATABASE_NAME, $collectionName = COLL
         exit("skip Could not drop '$databaseName.$collectionName': " . $e->getMessage());
     }
 }
+
+function skip_if_no_getmore_failpoint()
+{
+    $serverVersion = get_server_version(URI);
+
+    if (
+        version_compare($serverVersion, '3.2', '>=') &&
+        version_compare($serverVersion, '4.0', '<')
+    ) {
+        exit("skip Server version '$serverVersion' does not support a getMore failpoint'");
+    }
+}
--- a/tests/utils/tools.php
+++ b/tests/utils/tools.php
@@ -148,6 +148,21 @@ function get_server_storage_engine($uri)
 }
 
 /**
+ * Helper to return the version of a specific server.
+ *
+ * @param MongoDB\Driver\Server $server
+ * @return string
+ * @throws RuntimeException
+ */
+function get_server_version_from_server(MongoDB\Driver\Server $server)
+{
+    $command = new Command(['buildInfo' => 1]);
+    $cursor = $server->executeCommand('admin', $command);
+
+    return current($cursor->toArray())->version;
+}
+
+/**
  * Returns the version of the primary server.
  *
  * @param string $uri Connection string
@@ -157,10 +172,7 @@ function get_server_storage_engine($uri)
 function get_server_version($uri)
 {
     $server = get_primary_server($uri);
-    $command = new Command(['buildInfo' => 1]);
-    $cursor = $server->executeCommand('admin', $command);
-
-    return current($cursor->toArray())->version;
+    return get_server_version_from_server($server);
 }
 
 /**
@@ -441,11 +453,19 @@ function loadFixtures(\MongoDB\Driver\Manager $manager, $dbname = DATABASE_NAME,
     }
 }
 
-function START($id, array $options = array()) {
-    /* starting/stopping servers only works using the Vagrant setup */
-    PREDICTABLE();
-
-    $options += array("name" => "mongod", "id" => $id);
+function createTemporaryMongoInstance(array $options = [])
+{
+    $id = 'mo_' . COLLECTION_NAME;
+    $options += [
+        "name" => "mongod",
+        "id" => $id,
+        'procParams' => [
+            'dbpath' => "/tmp/phongo/{$id}",
+            'logpath' => "/tmp/phongo/{$id}.log",
+            'ipv6' => true,
+            'setParameter' => [ 'enableTestCommands' => 1 ],
+        ],
+    ];
     $opts = array(
         "http" => array(
             "timeout" => 60,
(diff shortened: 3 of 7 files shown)


tests/utils/tools.php
```diff
@@ -686,6 +688,8 @@ function fromJSON($var) {
 
 /* Note: this fail point may terminate the mongod process, so you may want to
  * use this in conjunction with a throwaway server. */
-function failGetMore(MongoDB\Driver\Manager $manager) {
-    return configureFailPoint($manager, "failReceivedGetmore", "alwaysOn");
+function failGetMore(MongoDB\Driver\Manager $manager)
+{
+    configureFailPoint($manager, 'failReceivedGetmore', 'alwaysOn');
+    configureFailPoint($manager, 'failCommand', 'alwaysOn', [ 'errorCode' => 237, 'failCommands' => ['getMore'] ]);
```
derickr: We use ``237`` here, as that's the same original code that MongoD would throw if a cursor had already gone by the time we call ``getMore``. This allows us to make things consistent with the getMore OP behaviour from previous mongod versions. An ``errorCode`` is required here for the failPoint to work.

php_phongo.h
```diff
@@ -97,6 +97,12 @@ typedef enum {
  * or command should select ExecutionTimeoutException. */
 #define PHONGO_SERVER_ERROR_EXCEEDED_TIME_LIMIT 50
 
+/* These constants are used for determining if a server error is due to a
+ * broken cursor. If so, we select a ConnectionException as this is what we did
+ * when getMore wasn't a command yet. */
```
jmikola: If I understand correctly, a ConnectionException was previously thrown because we experienced an opcode failure instead of a mere command failure (with `ok: 0`). I understand the desire for consistency in error reporting, but are we lying to users here?

I would think that a ConnectionException implies that the socket itself is discarded. It's possible that libmongoc may very well have done that for the old getmore failure (I don't know offhand so it bears investigation), but I don't think it would for a failed command.

Both of these errors seem very much like ServerExceptions. Although both ConnectionException and ServerException extend RuntimeException (and may report error labels), users can now rest assured that these error codes come from the server, which makes ServerException more sensible.

With the older getmore opcode failures, there was no server-side error code, so ConnectionException (or a base RuntimeException) would have made sense.

tests/cursor/cursor-getmore-005.phpt
```diff
@@ -1,15 +1,22 @@
...
 MongoDB\Driver\Cursor query result iteration with getmore failure
---XFAIL--
-START() tests must be reimplemented (PHPC-1179)
 --SKIPIF--
 <?php require __DIR__ . "/" ."../utils/basic-skipif.inc"; ?>
-<?php START("THROWAWAY", ["version" => "30-release"]); CLEANUP(THROWAWAY); ?>
+<?php
+/* This test spins up its own mongod instance, so only run this in the most default "standalone, no
+ * auth" configurations. This way, we can test on multiple server versions, but not waste resources
+ * on f.e. Travis. */
```
jmikola: Random thought, but is this the same approach you have for testing Atlas connections?

I'm curious if we should simply have a separate matrix entry that essentially implies "the test will use its own environment", so that these "spin up our own MO environment" and Atlas tests run in an isolated matrix entry.

tests/utils/skipif.php
```diff
@@ -226,3 +226,15 @@ function skip_if_not_clean($databaseName = DATABASE_NAME, $collectionName = COLL
         exit("skip Could not drop '$databaseName.$collectionName': " . $e->getMessage());
     }
 }
+
+function skip_if_no_getmore_failpoint()
```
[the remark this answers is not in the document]
jmikola: Why is 4.0 the exclusive upper bound for this failpoint? Was it removed from the server's 4.0+ branches?

tests/utils/tools.php
```diff
@@ -448,24 +456,24 @@ function START($id, array $options = array()) {
...
-    $json = file_get_contents(getMOUri() . "/servers/$id", false, $ctx);
+    $json = file_get_contents(getMongoOrchestrationURI() . "/servers/$id", false, $ctx);
     $result = json_decode($json, true);
 
     /* Failed -- or was already started */
     if (!isset($result["mongodb_uri"])) {
-        DELETE($id);
-        define($id, false);
+        destroyTemporaryMongoInstance($id);
+        die("Could not start temporary server instance\n");
```
jmikola: Can we throw an exception instead? The error handlers in `basic-skipif.inc` will ensure that any uncaught exception gets properly turned into an `exit` statement if run during a SKIPIF. If called as part of a normal test body, I also think an exception is more consistent with what we do elsewhere (avoiding utility functions that simply abort the PHP process on their own).

tests/utils/tools.php
```diff
@@ -630,31 +633,31 @@ function configureFailPoint(MongoDB\Driver\Manager $manager, $failPoint, $mode,
...
-    $result = $manager->executeCommand("admin", $cmd);
-    $arr = current($result->toArray());
-    if (empty($arr->ok)) {
-        var_dump($result);
-        throw new RuntimeException("Failpoint failed");
+    try {
+        $cmd = new MongoDB\Driver\Command($doc);
+        $result = $manager->executeCommand("admin", $cmd);
+    } catch (Exception $e) {
+        /* don't care if the fail point can't be set, it's MongoD version
```
jmikola: If you're now skipping tests when you know a failpoint isn't available, wouldn't you still want to keep the original behavior here (`var_dump()` and throw on unexpected failure to set a failpoint)?

tests/utils/tools.php
```diff
@@ -630,31 +633,31 @@ function configureFailPoint(MongoDB\Driver\Manager $manager, $failPoint, $mode,
...
 }
 
 function failMaxTimeMS(MongoDB\Driver\Manager $manager) {
     return configureFailPoint($manager, "maxTimeAlwaysTimeOut", array("times" => 1));
 }
 
-function getMOUri() {
-    if (!($HOST = getenv("MONGODB_ORCHESTRATION_HOST"))) {
-        $HOST = "192.168.112.10";
+function getMongoOrchestrationURI()
```
jmikola: I reckon we can do away with this method entirely and simply define the appropriate constants in `basic.inc` as we do for others. Consider:

```php
define('MONGODB_ORCHESTRATION_HOST', getenv('MONGODB_ORCHESTRATION_HOST') ?: 'localhost');
define('MONGODB_ORCHESTRATION_PORT', getenv('MONGODB_ORCHESTRATION_HOST') ?: '8889');
define('MONGODB_ORCHESTRATION_URI', sprintf('http://%s:%s/v1', MONGODB_ORCHESTRATION_HOST, MONGODB_ORCHESTRATION_PORT));
```

If the individual host and port values aren't needed as their own constant, you could also use local variables with the same `?:` expression purely to ensure `MONGODB_ORCHESTRATION_URI` gets set properly.

tests/utils/tools.php
jmikola: This seems like something worth adding to a block comment for this line or method.

php_phongo.h
jmikola: But this comes at a cost of inconsistency with regard to error code reporting. We tell users that that any server-originating error codes are used in exceptions within the ServerException hierarchy.

Am I correct in assuming that the error codes for killed cursors using OP_GETMORE differ from the command's error code? I believe so, as `MONGOC_ERROR_CURSOR_INVALID_CURSOR` was simply a sequential value in `mongoc_error_code_t` and that is what is used in `_mongoc_rpc_check_ok()` when inspecting an OP_REPLY. When the getMore command used, libmongoc now reports the error differently and uses the new server-originating error codes, which also changes how users can detect and differentiate each error.

IMO, the error reporting inconsistency is due to the server's API. Given that, I think the variation of exceptions is excusable based on the wire protocol we're using, and using ServerException is actually an improvement when it comes to distinguishing the errors via their code.

tests/utils/tools.php
jmikola: Is it a programming error to call `configureFailPoint()` without first checking that the fail point is supported? I thought that's the reason we have the various skip functions and special behavior like reporting that failpoints are never supported on mongos.

php_phongo.h
derickr: I've removed this now, and split the two test cases in a pre-3.6 and post-3.6 version.

tests/utils/tools.php
derickr: It's complicated. Because we need to set two failPoints for getMore, which one is available depends on the exact mongod version. We can either ignore the exception, or, do lots of more complex checking in ``failGetMore``. I've decided that this now "good enough" for what we need.

tests/utils/tools.php
```diff
@@ -632,31 +647,12 @@ function configureFailPoint(MongoDB\Driver\Manager $manager, $failPoint, $mode,
 
     $cmd = new MongoDB\Driver\Command($doc);
     $result = $manager->executeCommand("admin", $cmd);
```
jmikola: There's no need to assign `$result` anymore if we're not going to inspect the `ok` field. Granted, I'm not even sure the old code would have ever hit the exception, since `executeCommand()` may throw an exception on its own.

Side note: since `configureFailPoint()` is now a void function, let's also remove the `return` statement in `failMaxTimeMS()` and anything else that calls this function (if there even is anything else). That will be consistent with what you did in `failGetMore()`.

tests/utils/tools.php
```diff
@@ -632,31 +647,12 @@ function configureFailPoint(MongoDB\Driver\Manager $manager, $failPoint, $mode,
...
-        throw new RuntimeException("Failpoint failed");
-    }
-    return true;
 }
 
 function failMaxTimeMS(MongoDB\Driver\Manager $manager) {
     return configureFailPoint($manager, "maxTimeAlwaysTimeOut", array("times" => 1));
 }
 
-function getMOUri() {
```
jmikola: Removing this function likely breaks `scripts/start-servers.php` for the Vagrant environment, which was relying on `tools.php`. I realized that while looking at places that still called `getMOPresetBase()`, as I was going to suggest you remove that function as well.

I was going to suggest leaving this here and returning `MONGO_ORCHESTRATION_URI`, but `start-servers.php` doesn't ever include `basic.inc`, which sets that constant. Perhaps we should leave this function in place in this PR and simply create a new ticket to delete both of them later _and_ refactor `start-servers.php`.

tests/utils/tools.php
derickr: Updated

tests/utils/tools.php
derickr: I have soft-deprecated "start-servers.php", and don't want to spend any time on this. There is already a ticket for getting rid of the Vagrant stuff: https://jira.mongodb.org/browse/PHPC-1242 — I've made a note there to remove this function too.

PR #915, opened 2018-08-20
PHPC-1209: is_replica_set() should return false when connected to RS primary in standalone mode

derickr:


tests/utils/tools.php
```diff
@@ -199,7 +199,15 @@ function is_mongos($uri)
  */
 function is_replica_set($uri)
 {
-    return get_primary_server($uri)->getType() === Server::TYPE_RS_PRIMARY;
+    if (get_primary_server($uri)->getType() !== Server::TYPE_RS_PRIMARY) {
+        return false;
+    }
+
+    if (get_uri_option($uri, 'replicaSet') === NULL) {
```
jmikola: I think this would be more technically correct if we also make an allowance for a comma in the `$uri` string. My understanding of the connection string and SDAM specs is:

 * A comma-delimited list of hosts implies either a replica set or sharded cluster (to be determined after discovery)
 * A `replicaSet` URI option removes any ambiguity and implies a replica set connection
 * A URI string with only a single host (i.e. no commas) and no `replicaSet` option is either a standalone connection to a mongod or a connection to a sharded cluster through only a single mongos

Given that, I'd suggest adding `&& strpos($uri, ',') === false` to the condition and perhaps a comment that explains the logic.

derickr: I don't think we should do that unless we have a testing environment that does that specific case. And, I don't think we should spend time on adding another one.

derickr:
Not sure why GitHub didn't close this, but it's merged.

Issue #916, opened 2018-08-20
Fixes for openssl without pkg-config were not merge to 1.5 branch


DaveRandom:
The fixes that came out of #875 were not merged down to the 1.5 branch and are not present in 1.5.2. Specifically the last 3 commits shown [here](https://github.com/mongodb/mongo-php-driver/commits/master/scripts/autotools/libmongoc/CheckSSL.m4) are not present [here](https://github.com/mongodb/mongo-php-driver/commits/v1.5/scripts/autotools/libmongoc/CheckSSL.m4).

DaveRandom:
OK no worries, if I just take a branch based on 1.5 and cherry-pick the commits and PR that is that what you need?

derickr:
Hi,

I think that should work — I don't expect any conflicts. It'd be great if you could use ``PHPC-1241: OpenSSL 1.1 not found if pkg-config is not available`` as commit message.

cheers,
Derick

PR #917, opened 2018-08-21
PHPC-1241: OpenSSL 1.1 not found if pkg-config is not available

DaveRandom:
As per #916


derickr:
Thanks! I've just merged this.

PR #918, opened 2018-08-21
Fix tests for 7.3

DaveRandom:
- Look for `(object) array()` instead of `stdClass::__set_state()` from `var_export()` output (https://github.com/php/php-src/commit/e4e9cd835550990a6b8df7c61d59b6cc0da9b5b2). This is implemented by duplicating the test with appropriate SKIPIF.
- Use strings instead of integers as for scalar bad args, 7.3 uses `int` instead of `integer` in error output (https://github.com/php/php-src/commit/ce1d69a1f6dcf15d43029301059c25e5bc09a577). Using strings instead means the tests pass on all supported PHP versions.


derickr:
Thanks, but this is a duplicate of #914, which is waiting for code review.

PR #919, opened 2018-08-23
PHPC-1225: Add test to catch unsupported server wire protocol versions

derickr:
I'm also converting this error to a ConnectionException instead of a RuntimeException

### DIFF
--- a/tests/manager/manager-ctor-wireversion.phpt
+++ b/tests/manager/manager-ctor-wireversion.phpt
@@ -1,0 +1,20 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): wire version support
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new \MongoDB\Driver\Manager(URI);
+
+$command = new MongoDB\Driver\Command(['ping' => 1]);
+try {
+    $manager->executeCommand("test", $command);
+} catch (\MongoDB\Driver\Exception\ConnectionException $e) {
+    if ($e->getCode() == 15) { // Bad Wire Version
+        echo "Bad wire version detected: ", $e->getMessage(), "\n";
+    }
+}
+?>
+===DONE===
+--EXPECT--
+===DONE===
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -166,6 +166,10 @@ zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_d
 		return php_phongo_serverexception_ce;
 	}
 
+	if (domain == MONGOC_ERROR_PROTOCOL && code == MONGOC_ERROR_PROTOCOL_BAD_WIRE_VERSION) {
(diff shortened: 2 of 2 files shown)


tests/manager/manager-ctor-wireversion.phpt
```diff
@@ -0,0 +1,22 @@
...
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new \MongoDB\Driver\Manager(URI);
+
+$command = new MongoDB\Driver\Command(['ping' => 1]);
+try {
+    $manager->executeCommand("test", $command);
+} catch (\MongoDB\Driver\Exception\RuntimeException $e) {
+    if ($e->getCode() == 15) { // Bad Wire Version
+        echo "Bad wire version detected: ", $e->getMessage(), "\n";
```
jmikola: [PHPC-1225](https://jira.mongodb.org/browse/PHPC-1225) purposefully suggests not skipping the test because we don't have a live server, which explains why the `SKIPIF` above is quite empty. If I understand correctly, since we only echo for a wire protocol error this test will still pass when no server is listening on `URI` (since the error code will differ). If so, LGTM.

tests/manager/manager-ctor-wireversion.phpt
```diff
@@ -0,0 +1,22 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): wire version support
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
```
jmikola: As-is, this only skips when the extension isn't loaded. I think it makes sense to remove `SKIPIF` entirely here, as we do for BSON tests and the like, which obviously require the extension.

The other code in `basic-skipif` (e.g. setting up exception handlers) isn't relevant because we don't actually call any utility functions.

tests/manager/manager-ctor-wireversion.phpt
```diff
@@ -0,0 +1,22 @@
...
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new \MongoDB\Driver\Manager(URI);
+
+$command = new MongoDB\Driver\Command(['ping' => 1]);
+try {
+    $manager->executeCommand("test", $command);
+} catch (\MongoDB\Driver\Exception\RuntimeException $e) {
```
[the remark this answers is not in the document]
derickr: Fixed.

tests/manager/manager-ctor-wireversion.phpt
derickr: Correct:

```
derick@singlemalt:~/dev/php/derickr-mongo-php-driver $ echo $MONGODB_URI
derick@singlemalt:~/dev/php/derickr-mongo-php-driver $ make test TESTS=tests/manager/manager-ctor-wireversion.phpt | grep PASS
PASS MongoDB\Driver\Manager::__construct(): wire version support [tests/manager/manager-ctor-wireversion.phpt] 
```

PR #923, opened 2018-09-04
PHPC-1219: Relax requirement for replica set tagsets assertions

derickr:


derickr:
Yes. Sorta, there are two more tests where we have them, but they're both ``getTags()`` tests.

PR #927, opened 2018-09-19
Fix port via parse_url() when default URI is used

jmikola:
The default URI is "mongodb://127.0.0.1/" and does not include a port. Since parse_url() will not yield a "port" index in that case we should fill in the default (i.e. 27017).

This is related to dd5e9e9b734059d756d1c8089bb94b8eb09f9161 (https://github.com/mongodb/mongo-php-driver/pull/907).


jmikola:
@derickr: If this looks good, please merge into v1.5 on your own (while tending to #926) and then up to master.

For some context, I discovered this while running a single node replica set on my local host and relying on the `URI` constant in `basic.inc` instead of a `MONGODB_URI` environment variable. Should be easy to repro if you like.

PR #928, opened 2018-09-20
PHPC-1272: phongo_execute_query() should not alter state of Query opts

derickr:
https://jira.mongodb.org/browse/PHPC-1272


php_phongo.c
```diff
@@ -777,30 +778,36 @@ bool phongo_execute_query(mongoc_client_t* client, const char* namespace, zval*
 
 	query = Z_QUERY_OBJ_P(zquery);
 
+	bson_copy_to(&opts, query->opts);
```
jmikola: The order of these arguments should be reversed (src -> to).

Issue #754, opened 2018-02-10
Allow to extend `MongoDB\Driver\Cursor`


abellion:
### Description

I would like to be able to set some metadata to the retrieved documents when using `find()`, `findOne()` or `aggregate()`.

To do so, a solution is to wrap the `MongoDB\Driver\Cursor` inside an `IteratorIterator` and set the metadata to the documents when the `current()` method is called.

The downside of this approach is that the returned iterator is not an instance of the `MongoDB\Driver\Cursor` class so it may break some type hints and it doesn't have its methods.

### Proposal

If the `MongoDB\Driver\Cursor` were an instance of `Iterator`, it would be easy to extend it and take advantage of its `current()` method while not loosing type hinting and the other methods the Cursor class provides.

jmikola:
Having `MongoDB\Driver\Cursor` implement [`Iterator`](http://php.net/manual/en/class.iterator.php) still would not allow it to be extended, as the the class is `final`. We're not keen on removing `final` from the core classes due to (a) issues we had with extending core classes in the legacy `ext-mongo` extension and (b) promoting composition over inheritance. As an internal class, we already need to provide the [iterator handlers](http://www.phpinternalsbook.com/classes_objects/iterators.html), which [`IteratorIterator`](php.net/iteratoriterator`IteratorIterator) is already capable of exposing as PHP methods accessible directly from userland.

If I understand correctly, the real need here is having the extension provide an interface implemented by both `MongoDB\Driver\Cursor` and available for a userland `IteratorIterator` sub-class to facilitate polymorphism in any application code expecting a cursor (assuming it type hints with that interface instead of `MongoDB\Driver\Cursor`).

Interfaces for driver classes was something we considered early on in [PHPC-378](https://jira.mongodb.org/browse/PHPC-378) but there was never a pressing need for it. IIRC, I believe I also spoke to @alcaeus about whether future versions of Doctrine ODM might need such an interface, as its current API does have a class that composes a cursor, but we were unsure. Presently, I expect users may be using [`Traversable`](http://php.net/traversable) to pass cursors and iterators interchangeably, since that's the common interface; however, that doesn't help if the code also expects to be able to call methods like [`MongoDB\Driver\Cursor::isDead()`](http://php.net/manual/en/mongodb-driver-cursor.isdead.php).

[...]

PR #932, opened 2018-10-05
PHPC-1123: Create interface for Cursor class

derickr:


tests/cursor/cursorinterface-001.phpt
```diff
@@ -0,0 +1,26 @@
+--TEST--
+MongoDB\Driver\CursorInterface is implemented by MongoDB\Driver\Cursor
```
jmikola: This can technically use `class_implements()` and `in_array()` without requiring a live database server (using only class name strings), but it's fine as is.

src/MongoDB/CursorInterface.c
```diff
@@ -0,0 +1,63 @@
...
+	/* clang-format on */
+};
+/* }}} */
+
+void php_phongo_cursor_interface_init_ce(INIT_FUNC_ARGS) /* {{{ */
+{
+	zend_class_entry ce;
+
+	INIT_NS_CLASS_ENTRY(ce, "MongoDB\\Driver", "CursorInterface", php_phongo_cursor_interface_me);
+	php_phongo_cursor_interface_ce = zend_register_internal_interface(&ce TSRMLS_CC);
```
jmikola: In order for this to fully stand in for Cursor, I think it needs to extend Traversable. This was discussed in my original comment that resulted in PHPC-1123: https://github.com/mongodb/mongo-php-driver/issues/754#issuecomment-365683894

With that change, I think we can then allow Cursor to implement Traversable through CursorInterface, instead of implementing both directly.

tests/cursor/cursorinterface-001.phpt
jmikola: If we'd like to test that CursorInterface extends Traversable, we can use this approach to check the interfaces directly (since we can't create concrete objects to do so).

src/MongoDB/CursorInterface.c
```diff
@@ -0,0 +1,63 @@
...
+ZEND_END_ARG_INFO()
+
+ZEND_BEGIN_ARG_INFO_EX(ai_CursorInterface_void, 0, 0, 0)
+ZEND_END_ARG_INFO()
+
+static zend_function_entry php_phongo_cursor_interface_me[] = {
+	/* clang-format off */
+	ZEND_ABSTRACT_ME(CursorInterface, getId, ai_CursorInterface_void)
+	ZEND_ABSTRACT_ME(CursorInterface, getServer, ai_CursorInterface_void)
+	ZEND_ABSTRACT_ME(CursorInterface, isDead, ai_CursorInterface_void)
```
jmikola: Before we codify this with an interface, what do you think about deprecating `isDead()` and introducing a new method in the next minor release (with this interface)? If so, this was justify a separate ticket to be done before this PR is finalized/merged.

This is based on [`mongoc_cursor_is_alive()`](http://mongoc.org/libmongoc/current/mongoc_cursor_is_alive.html), which libmongoc has superseded with [`mongoc_cursor_more()`](http://mongoc.org/libmongoc/current/mongoc_cursor_more.html). I think the documentation of `more()` is easier to understand for users:

> This function shall indicate if there is potentially more data to be read from the cursor. This is only useful with tailable cursors. Use mongoc_cursor_next() for regular cursors.

Since the underlying logic in libmongoc is based entirely on there _potentially_ being more results (based on the cursor's batch position and whether we have still have a cursor ID), a rename may better serve users who might otherwise assume that "alive" or "dead" are absolute.

src/MongoDB/CursorInterface.c
derickr: Done. In fact, if we add ``zend_class_implements`` to CursorInterface, we *cannot* have it in Cursor:

> ``E_COMPILE_ERROR: Class MongoDB\Driver\CursorInterface cannot implement previously implemented interface Traversable``

tests/cursor/cursorinterface-001.phpt
derickr: Done that in two new tests.

src/MongoDB/CursorInterface.c
derickr: I've not done that yet though.

src/MongoDB/CursorInterface.c
jmikola: Interesting. Is this something that only applies to PHP internals, or would this affect userland?

https://3v4l.org/OhPT2 suggests that it's just an internal restriction. And I wouldn't expect there to be any any issue with a class implementing both Iterator and IteratorAggregate (in userland).

src/MongoDB/CursorInterface.c
jmikola: Due to the nature of a cursor iteration, I don't think we can do better than _potentially_. The cases where a cursor _may_ have more data are as follows:

 * We know there are additional documents in the current batch (buffered on the client side). Iterating will fetch a document from our local buffer.
 * There are no additional documents in the current batch, but we have a cursor ID. Iterating will request the server for more documents, at which point we will either:
   *  Receive a batch containing more documents and the cursor ID will be kept open. This is common to iterating through a large result set.
   * Receive a batch containing more documents and the cursor ID will be zero, which signifies that we've received the final batch from the server (i.e. cursor is closed server-side)
   * Receive an empty batch from the server, but the cursor ID is kept open. This tells us we should consult the server for more documents later (common in tailable cursors).
   * Receive an empty batch from the server and the cursor ID will be zero. AFAIK, this is the only case where we are assured that there are not _potentially_ more documents (i.e. `more()` returns false).

jmikola: Those bullets were in response to your earlier comment:

> What we should have is a method that informs the user whether there is more data available in the cursor, and not just potentially.

I was trying to reason out why I don't think it's possible to provide such a method; although I don't know if you were advocating we introduce one or just wishing it were possible to do so. 

While brain-dumping the above points into [PHPC-1279](https://jira.mongodb.org/browse/PHPC-1279), I gave this some more thought and reconsidered `isDead()`. It seems like the "dead" case is simpler to understand (and probably document) than the "potentially more or alive" case. If you agree or don't feel strongly I'd be LGTM on leaving `isDead()` on the interface and simply improving its documentation down the line.

PR #983, opened 2019-05-30
PHPC-1387: Consider error document for cursor exceptions

jmikola:
https://jira.mongodb.org/browse/PHPC-1387

This allows executeQuery() to throw a CommandException, which exposes the error document.

In 1.6+, it will also ensure that error labels are set on any RuntimeException thrown during cursor iteration, which is relevant for transactions.


jmikola:
FYI: AppVeyor build failures are to be expected. These were fixed in the 1.6 branch by [PHPC-1329](https://jira.mongodb.org/browse/PHPC-1329).

PR #984, opened 2019-05-31
PHPC-1359: Prohibit startTransaction() on sharded clusters

jmikola:
https://jira.mongodb.org/browse/PHPC-1359

In versions of the driver that do not support mongos pinning, we should prohibit starting a transaction on a sharded cluster (even if the server version might actually support transactions).


jmikola:
> Do you also need to check maxWireVersion < 7?

I spoke to @kevinAlbs about this and he said it looks like libmongoc does not implement that check. He also cited the following from the Transaction spec's [startTransaction](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#starttransaction) section:

> startTransaction SHOULD report an error if the driver can detect that transactions are not supported by the deployment. A deployment does not support transactions when the deployment does not support sessions, or maxWireVersion < 7, or the maxWireVersion < 8 and the topology type is Sharded

Since transactions cannot be used without sessions, and libmongoc _does_ report an error if sessions are unsupported, we need only worry about starting a transaction on a 3.6 server. In that case, the driver can rely on a 3.6 server to report an error for an unrecognized command field. Testing this locally, I can produce the following server error message attempting to start a transaction with an insert command:

> BSON field 'insert.startTransaction' is an unknown field

I will open a ticket to propose that libmongoc add the check to better comply with the spec, as I don't think we have good justification for ignoring that SHOULD; however, I think the change in this PR (and [CDRIVER-3067](https://jira.mongodb.org/browse/CDRIVER-3067)) is more independent and more important, since it addresses a scenario where a driver might successfully start a transaction on a mongos server, but then fail to pin the session and invite unexpected behavior/errors at runtime.

jmikola:
Created [CDRIVER-3161](https://jira.mongodb.org/browse/CDRIVER-3161) to track the libmongoc issue, and opened [PHPC-1391](https://jira.mongodb.org/browse/PHPC-1391) to depend on it; however, I don't anticipate making a PHP-specific fix for that issue.

PR #985, opened 2019-05-31
PHPC-1378: Omit CFLAGS/LDFLAGS in client metadata to avoid platform truncation

jmikola:
https://jira.mongodb.org/browse/PHPC-1378

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -283,15 +283,12 @@ if test "$PHP_MONGODB" != "no"; then
     AC_SUBST(MONGOC_ENABLE_SHM_COUNTERS, 0)
     AC_SUBST(MONGOC_TRACE, 1)
 
-    dnl Capture user-set environment variables for metadata handshake
-    dnl We can do this after the above scripts as they only modify STD_CFLAGS
-    AS_VAR_COPY(MONGOC_USER_SET_CFLAGS, [CFLAGS])
-    AS_VAR_COPY(MONGOC_USER_SET_LDFLAGS, [LDFLAGS])
-    AS_VAR_COPY(MONGOC_CC, [CC])
-
-    AC_SUBST(MONGOC_USER_SET_CFLAGS)
-    AC_SUBST(MONGOC_USER_SET_LDFLAGS)
-    AC_SUBST(MONGOC_CC)
+    dnl Assignments for metadata handshake. Leave CFLAGS/LDFLAGS empty as they
+    dnl would likely cause platform info (PHP version) to be truncated. We can
+    dnl consider restoring CFLAGS/LDFLAGS once CDRIVER-3134 is resolved.
+    AC_SUBST(MONGOC_CC, [$CC])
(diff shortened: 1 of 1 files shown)


config.m4
```diff
@@ -283,15 +283,12 @@ if test "$PHP_MONGODB" != "no"; then
     AC_SUBST(MONGOC_ENABLE_SHM_COUNTERS, 0)
     AC_SUBST(MONGOC_TRACE, 1)
 
-    dnl Capture user-set environment variables for metadata handshake
-    dnl We can do this after the above scripts as they only modify STD_CFLAGS
-    AS_VAR_COPY(MONGOC_USER_SET_CFLAGS, [CFLAGS])
-    AS_VAR_COPY(MONGOC_USER_SET_LDFLAGS, [LDFLAGS])
-    AS_VAR_COPY(MONGOC_CC, [CC])
```
jmikola: I don't recall why [`AS_VAR_COPY`](https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Polymorphic-Variables.html) was used here, unless it had something to do with resolving variables within CFLAGS and LDFLAGS; however, I can't make the connection to the polymorphic variable example provided in the Autoconf manual.

In any event, I don't think it was necessary for `$CC` since that should simply be the compiler name.

config.m4
```diff
@@ -283,15 +283,12 @@ if test "$PHP_MONGODB" != "no"; then
...
-
-    AC_SUBST(MONGOC_USER_SET_CFLAGS)
-    AC_SUBST(MONGOC_USER_SET_LDFLAGS)
-    AC_SUBST(MONGOC_CC)
+    dnl Assignments for metadata handshake. Leave CFLAGS/LDFLAGS empty as they
+    dnl would likely cause platform info (PHP version) to be truncated. We can
+    dnl consider restoring CFLAGS/LDFLAGS once CDRIVER-3134 is resolved.
+    AC_SUBST(MONGOC_CC, [$CC])
+    AC_SUBST(MONGOC_USER_SET_CFLAGS, [])
+    AC_SUBST(MONGOC_USER_SET_LDFLAGS, [])
```
jmikola: I manually verified that this produces the same output in `configure` as the original pairing of `AS_VAR_COPY` and `AC_SUBST`:

```
MONGOC_CC=$CC
```

Predictably, there is no assignment for CFLAGS/LDFLAGS in `configure` with the empty `AC_SUBST` macros; however, `MONGOC_USER_SET_LDFLAGS` and `MONGOC_USER_SET_CFLAGS` are still included in `ac_subst_vars` which ensures that they at least get replaced when processing `mongoc-config.h.in`. `@MONGOC_USER_SET_LDFLAGS@` and `@MONGOC_USER_SET_CFLAGS@` are both replaced with empty strings during configuration.

rcsanchez97: That makes sense.  Your change is certainly more straightforward in terms of what is happening.

PR #1006, opened 2019-08-09
 PHPC-1382: Allow applications to set maxTimeMS for commitTransaction

alcaeus:
https://jira.mongodb.org/browse/PHPC-1382

I chose to duplicate the checks in place for the `maxAwaitTimeMS` option on cursors and check for negative and overflowing values, along with a sanity check that the user actually gave us an int for the option.

@jmikola docs are still missing for this PR. While I have my PHP VCS account set up, I'd rather go through the process of writing/publishing docs on php.net with you before adding code there.

### DIFF
--- a/src/MongoDB/Session.c
+++ b/src/MongoDB/Session.c
@@ -247,21 +247,50 @@ static PHP_METHOD(Session, getOperationTime)
 } /* }}} */
 
 /* Creates a opts structure from an array optionally containing an RP, RC,
- * and/or WC object. Returns NULL if no options were found, or there was an
- * invalid option. If there was an invalid option or structure, an exception
- * will be thrown too. */
+ * WC object, and/or maxCommitTimeMS int. Returns NULL if no options were found,
+ * or there was an invalid option. If there was an invalid option or structure,
+ * an exception will be thrown too. */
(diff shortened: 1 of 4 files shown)


src/MongoDB/Session.c
```diff
@@ -247,21 +247,64 @@ static PHP_METHOD(Session, getOperationTime)
...
+ * WC object, and/or maxCommitTimeMS int. Returns NULL if no options were found,
+ * or there was an invalid option. If there was an invalid option or structure,
+ * an exception will be thrown too. */
 mongoc_transaction_opt_t* php_mongodb_session_parse_transaction_options(zval* options TSRMLS_DC)
 {
 	mongoc_transaction_opt_t* opts = NULL;
 
+	if (php_array_existsc(options, "maxCommitTimeMS")) {
+		int64_t max_commit_time_ms;
+		zval* max_commit_time_ms_zval = php_array_fetchc(options, "maxCommitTimeMS");
```
jmikola: PHPLIB certainly does aggressive type checking on _its_ scalar options, but PHPC typically does not. For "maxAwaitTimeMS" in Cursor.c and various other scalar options, I think we tend to just fetch as longs. For example:

```
int64_t max_await_time_ms = php_array_fetchc_long(options, "maxAwaitTimeMS");
```

I'd swap over for consistency until such point we decide to change behavior across the board.

PR #1007, opened 2019-08-09
Test against PHP 7.4 on travis-ci

alcaeus:
Until we can get PHP 7.4 on evergreen, we'll use travis-ci to ensure the extension compiles and tests pass.

This PR also runs a single job as smoke test to avoid using excessive travis-ci build resources. Other versions and topologies will only be tested if this initial smoke test succeeds. This helps save build resources.

### DIFF
--- a/tests/cursor/cursor-tailable-001.phpt
+++ b/tests/cursor/cursor-tailable-001.phpt
@@ -22,7 +22,7 @@ function insert(MongoDB\Driver\Manager $manager, $from, $to = null)
 
     $writeResult = $manager->executeBulkWrite(NS, $bulkWrite);
 
-    printf("Inserted %d document(s): %s\n", $writeResult->getInsertedCount(), implode(range($from, $to), ', '));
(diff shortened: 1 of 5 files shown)


tests/cursor/cursor-tailable-001.phpt
```diff
@@ -22,7 +22,7 @@ function insert(MongoDB\Driver\Manager $manager, $from, $to = null)
 
     $writeResult = $manager->executeBulkWrite(NS, $bulkWrite);
 
-    printf("Inserted %d document(s): %s\n", $writeResult->getInsertedCount(), implode(range($from, $to), ', '));
+    printf("Inserted %d document(s): %s\n", $writeResult->getInsertedCount(), implode(', ', range($from, $to)));
```
jmikola: I was curious as to how this ever worked, but it appears to be noted on https://www.php.net/manual/en/function.implode.php

Did PHP 7.4 actually fix/deprecate that?

alcaeus: Yes, this is deprecated in 7.4, along with some similar „features“ in other functions.

PR #1008, opened 2019-08-12
PHPC-1355: Add the ability to specify a pipeline to an update command

alcaeus:
https://jira.mongodb.org/browse/PHPC-1355

This PR changes update document checks in `\MongoDB\Driver\BulkWrite` to ensure updates with pipelines are treated as updates and not replacements. The detection logic for this is a direct copy from libmongoc since the function is not exposed to users of libmongoc. Even though we're duplicating logic, I don't think exposing the function to the outside is a good idea.

### DIFF
--- a/tests/bulk/bulkwrite-update-003.phpt
+++ b/tests/bulk/bulkwrite-update-003.phpt
@@ -1,0 +1,65 @@
+--TEST--
+MongoDB\Driver\BulkWrite::update() with pipeline option
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_live(); ?>
+<?php skip_if_server_version('<', '4.2'); ?>
+<?php skip_if_not_clean(); ?>
+--FILE--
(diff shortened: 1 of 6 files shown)


tests/bulk/bulkwrite-update-003.phpt
```diff
@@ -0,0 +1,65 @@
...
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(URI);
+
+$bulk = new MongoDB\Driver\BulkWrite();
+
+$bulk->insert([ '_id' => 1, 'x' => 1, 'y' => 1, 't' => [ 'u' => [ 'v' => 1 ] ] ]);
+$bulk->insert([ '_id' => 2, 'x' => 2, 'y' => 1]);
+
+$manager->executeBulkWrite(DATABASE_NAME . '.' . COLLECTION_NAME, $bulk);
```
jmikola: `basic.inc` defines `NS`, which you can use here.

PR #1009, opened 2019-08-13
PHPC-1369: Upgrade libmongoc to 1.15.0

alcaeus:
https://jira.mongodb.org/browse/PHPC-1369

Also contains a test to verify the actionable exception message when using retryWrites on a storage engine that doesn't support it ([PHPC-1422](https://jira.mongodb.org/browse/PHPC-1422)) as well as two sanity tests to test retryable reads ([PHPC-1291](https://jira.mongodb.org/browse/PHPC-1291)). Spec tests for retryable reads will be implemented in [PHPLIB-463](https://jira.mongodb.org/browse/PHPLIB-463).


tests/retryable-reads/retryable-reads-001.phpt
```diff
@@ -0,0 +1,58 @@
+--TEST--
+Retryable reads: supported single-statement operations include transaction IDs
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_libmongoc_crypto(); ?>
+<?php skip_if_server_version('<', '3.6'); ?>
```
jmikola: I think this can be removed, as `skip_if_no_failcommand_failpoint()` already checks the server version. If the URI is mongos, we require 4.1.8+; otherwise, 4.0+.

tests/retryable-reads/retryable-reads-001.phpt
```diff
@@ -0,0 +1,58 @@
+--TEST--
+Retryable reads: supported single-statement operations include transaction IDs
```
jmikola: This title doesn't seem accurate. Consider "Retryable reads: find command is retried once".

EDIT: Noticed below that this was using `executeReadCommand()`. In that case, perhaps we should include the method name in the title as well if we'll have another test for `executeQuery()`.

tests/retryable-reads/retryable-reads-001.phpt
```diff
@@ -0,0 +1,58 @@
...
+configureFailPoint($manager, 'failCommand', ['times' => 1], ['failCommands' => ['find'], 'closeConnection' => true]);
+
+$observer = new Observer;
+MongoDB\Driver\Monitoring\addSubscriber($observer);
+
+$command = new MongoDB\Driver\Command([
+    'find' => COLLECTION_NAME,
+    'filter' => ['x' => 1],
+]);
+$cursor = $manager->executeReadCommand(DATABASE_NAME, $command);
```
jmikola: Any reason not to use `executeQuery()` in this test? Perhaps we can split this test into two and have the other utilize `MongoDB\Driver\Query` with `executeQuery()`? The `executeReadCommand()` test can continue to use the `find` command, but alternatively we could run an aggregate for count (effectively `countDocuments` CRUD method).

tests/retryable-reads/retryable-reads_error-001.phpt
```diff
@@ -0,0 +1,33 @@
+--TEST--
+Retryable reads: failure when disabled
```
jmikola: Similar to my comments on the other test, I think this can refer to the method being tested.

tests/retryable-reads/retryable-reads_error-001.phpt
```diff
@@ -0,0 +1,33 @@
...
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_libmongoc_crypto(); ?>
+<?php skip_if_server_version('<', '3.6'); ?>
+<?php skip_if_no_failcommand_failpoint(); ?>
+<?php skip_if_not_clean(); ?>
```
jmikola: Should not be needed if the test doesn't write.

tests/retryable-reads/retryable-reads_error-001.phpt
```diff
@@ -0,0 +1,33 @@
+--TEST--
+Retryable reads: failure when disabled
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_libmongoc_crypto(); ?>
+<?php skip_if_server_version('<', '3.6'); ?>
```
jmikola: Same comment as previous test. This may be redundant.

tests/retryable-reads/retryable-reads_error-001.phpt
```diff
@@ -0,0 +1,33 @@
...
+<?php skip_if_not_clean(); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(URI, ['retryReads' => false]);
+
+configureFailPoint($manager, 'failCommand', ['times' => 1], ['failCommands' => ['find'], 'closeConnection' => true]);
+
+throws(
```
jmikola: Asserting the exception is good, but I think it makes sense to still use APM and assert that only a single command is executed, too. This test could conceivably pass if `times: 2` was specified in the fail point and `retryReads` was enabled.

tests/retryable-writes/retryable-writes-001.phpt
```diff
@@ -32,7 +32,7 @@ class TransactionIdObserver implements MongoDB\Driver\Monitoring\CommandSubscrib
 $observer = new TransactionIdObserver;
 MongoDB\Driver\Monitoring\addSubscriber($observer);
 
-$manager = new MongoDB\Driver\Manager(URI, ['retryWrites' => true]);
+$manager = new MongoDB\Driver\Manager(URI);
```
jmikola: Looks like I missed these in https://github.com/mongodb/mongo-php-driver/pull/996. Good catch.

tests/retryable-writes/retryable-writes_error-001.phpt
```diff
@@ -0,0 +1,36 @@
...
+    \MongoDB\Driver\Exception\CommandException::class
+);
+echo "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+OK: Got MongoDB\Driver\Exception\CommandException
+This MongoDB deployment does not support retryable writes. Please add retryWrites=false to your connection string.
```
jmikola: Just to confirm, this is testing the error message raised by libmongoc, correct?

jmikola:
When upgrading the libmongoc submodule, be sure to regenerate `src/LIBMONGOC_VERSION_CURRENT`. See: https://github.com/mongodb/mongo-php-driver/blob/master/CONTRIBUTING.md#ensure-libmongoc-version-information-is-correct

Perhaps we should also reference this outside of the "Releasing" section, as that may be how you missed it.

tests/retryable-writes/retryable-writes_error-001.phpt
alcaeus: That is correct 👍

alcaeus:
> Perhaps we should also reference this outside of the "Releasing" section, as that may be how you missed it.

I looked at #996 and tried to replicate your steps, but I definitely missed this. I expanded the contribution docs to include a section on bumping libmongoc and moved the highlighted paragraph from the release docs to this new chapter since we not only bump libmongoc during release preparations.

tests/retryable-reads/retryable-reads-001.phpt
alcaeus: The test is now split, there is one each for `executeReadCommand` and `executeQuery`.

tests/retryable-reads/retryable-reads_error-001.phpt
alcaeus: Both failure tests now include a command observer to assert against the number of commands executed.

PR #1010, opened 2019-08-21
PHPC-1254: Implement test plan for ObjectID spec

alcaeus:
https://jira.mongodb.org/browse/PHPC-1254

ObjectID generation is done completely in libmongoc, which implemented the ObjectID spec with 1.14.0 (introduced in #972). This PR adds the timestamp tests from the test plan in the [ObjectID spec](https://github.com/mongodb/specifications/blob/master/source/objectid.rst).

### DIFF
--- a/tests/bson/bson-objectid-getTimestamp-002.phpt
+++ b/tests/bson/bson-objectid-getTimestamp-002.phpt
@@ -1,0 +1,29 @@
+--TEST--
+MongoDB\BSON\ObjectId::getTimestamp: Ensure that the Timestamp field is represented as an unsigned 32-bit integer
+--FILE--
+<?php
+date_default_timezone_set("UTC");
+
+function create_object_id($oid_string)
+{
+    $objectid = new MongoDB\BSON\ObjectId($oid_string);
+    $ts = sprintf("%u", $objectid->getTimestamp());
+    echo $ts, "\n";
+    echo date_create("@{$ts}")->format("Y-m-d H:i:s"), "\n";
+}
+
+create_object_id('000000000000000000000000');
+create_object_id('7FFFFFFF0000000000000000');
+create_object_id('800000000000000000000000');
+create_object_id('FFFFFFFF0000000000000000');
+
+?>
+--EXPECT--
+0
+1970-01-01 00:00:00
+2147483647
+2038-01-19 03:14:07
+2147483648
+2038-01-19 03:14:08
+4294967295
+2106-02-07 06:28:15


tests/bson/bson-objectid-getTimestamp-002.phpt
```diff
@@ -0,0 +1,29 @@
...
+
+?>
+--EXPECT--
+0
+1970-01-01 00:00:00
+2147483647
+2038-01-19 03:14:07
+2147483648
+2038-01-19 03:14:08
+4294967295
```
jmikola: I think this will fail on 32-bit platforms, since it will render as a negative value.

https://www.php.net/manual/en/mongodb-bson-objectid.gettimestamp.php talks about this and suggests using `%u` with `sprintf()` to dump the value. That said, I'm not sure how `date_create()` will handle this on a 32-bit platform.

You can try only making the `%u` change (rather than `echo $ts`) and see if this passes on our 32-bit environments in Evergreen.

If that doesn't work (or you'd rather not bother), we can just require a 64-bit platform for this test:

```
--SKIPIF--
<?php if (PHP_INT_SIZE !== 8) { die("skip Difficulty creating dates from unsigned 32-bit ints on a 32-bit platform"); } ?>
```

alcaeus: I'm taking the easy way and skip the test on 32-bit platforms: since the logic is entirely in libmongoc, I mainly added these tests as a safeguard. Since libmongoc is tested on i386 platforms I feel safe skipping these tests here.

tests/bson/bson-objectid-getTimestamp-002.phpt
```diff
@@ -0,0 +1,31 @@
+--TEST--
+MongoDB\BSON\ObjectId::getTimestamp: Ensure that the Timestamp field is represented as an unsigned 32-bit integer
+--SKIPIF--
+<?php if (PHP_INT_SIZE !== 8) { die("skip Difficulty creating dates from unsigned 32-bit ints on a 32-bit platform"); } ?>
```
jmikola: Sorry for the previous suggestion, but let's change this to simply:

```
<?php if (8 !== PHP_INT_SIZE) { die('skip Only for 64-bit platform'); } ?>
```

I was poking around in the date extension's `parse_date.re` (which is used to generate the `scan()` function `parse_date.c` that processes the string argument for `date_create` and `DateTime::__construct()`) and it doesn't look like it has any problem working with large integers. The accepted pattern is `"@" "-"? [0-9]+;` and it stores the result in an unsigned long long, which would hold a 64-bit integer even on a 32-bit platform.

Provided we used `sprintf('%u', $ts)` below for both the echo and `date_create()` argument, I actually think the test would pass on both 32-bit and 64-bit platforms, but I'm happy to skip 32-bit for simplicity.

I just don't want to throw shade at @derickr with an inaccurate skip message :)

alcaeus: Well, if we can easily have this pass on 32-bit I'm all for it.

PR #1014, opened 2019-08-23
PHPC-1290: Support mongos pinning for sharded transactions

alcaeus:
https://jira.mongodb.org/browse/PHPC-1290

Server selection in `php_phongo_manager_select_server` now checks an optional session for a pinned server and returns that. Pinning and unpinning is still done by libmongoc. Since PHPLIB also performs server selection, a new `getServer` method on the `MongoDB\Driver\Session` object exposes the server a session is pinned to. The server is also exposed in `php_phongo_session_get_debug_info` for inspection during debugging (that was quite helpful while working on the PHPLIB side of that PR).

Note: this PR bumps libmongoc to a dev version. I've created [PHPC-1428](https://jira.mongodb.org/browse/PHPC-1428) to track bumping to a stable release before we release 1.6.0.


src/LIBMONGOC_VERSION_CURRENT
```diff
@@ -1 +1 @@
-1.15.0
+1.15.1-20190823+git10f6ee9458
```
alcaeus: FWIW, the make task reported the version as `1.11.1-20190823+git10f6ee9458`. I've fixed this manually, but we may want to take a look at why it calculates the wrong version.

scripts/presets/travis/sharded_clusters/cluster.json
```diff
@@ -69,6 +69,13 @@
             "logappend": true,
             "port": 4300,
             "bind_ip_all": true
+        },
```
alcaeus: This change and all other changes in this commit are necessary because a session is not pinned when the topology contains only a single mongos node. I'm not sure if that's a bug in libmongoc, as the spec doesn't specify whether a session needs to be pinned when only connected to a single mongos. After all, that also ensures that all commands are sent to the same server during the transaction.

kevinAlbs: libmongoc considers a URI like "mongodb://foo.example.com/" to be topology type `Single`. We only pin for topology type `Sharded`:
https://github.com/mongodb/mongo-c-driver/blob/10f6ee9458449790125410a6293433a39aa9e9c1/src/libmongoc/src/mongoc/mongoc-cluster.c#L510-L512

I'm not sure if there's harm in not pinning when there's only one mongos, but I agree we should clarify in the spec.

SRV polling will only applies to topology type `Sharded` or `Unknown` (see [here](https://github.com/mongodb/mongo-c-driver/blob/59742684040c89708bfe7ed857efb38c65f30d98/src/libmongoc/src/mongoc/mongoc-topology.c#L486-L491)), so I think we might be safe from SRV polling adding more hosts in the middle of a sharded transaction. The [SRV polling spec](https://github.com/mongodb/specifications/blob/master/source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst#abstract) says:

> This feature is only available when the Server Discovery has determined that the TopologyType is Sharded, or Unknown. 

And based on:
https://github.com/mongodb/mongo-c-driver/blob/59742684040c89708bfe7ed857efb38c65f30d98/src/libmongoc/src/mongoc/mongoc-topology.c#L304-L344

If SRV lookup only returned one host, I think we would transition to topology type `Single`. That seems to abide correctly by the [Initial DNS Seedlist Discovery Spec](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst):
> Client MUST treat the host information as if the following URI was used
(i.e. we construct a URI from the records returned and proceed as normal, transitioning to `Single` as if this URI was passed directly).

So, if we used a `mongodb+srv` URI, which initially resolved to a single host like:
`mongodb://host1`, we would transition to topology type `Single` and never scan. And IIUC if we start as topology type `Sharded` and in the process of polling remove all hosts but one, we will *not* transition back to `Single`.

jmikola: > IIUC if we start as topology type Sharded and in the process of polling remove all hosts but one, we will not transition back to Single.

In this case, since the topology would remain sharded we could trust that libmongoc would still pin the session even in the case where there was only a single mongos in the topology. Is that correct? If so, I think @alcaeus' original change in this PR is correct and we should just ensure that all of our sharded cluster MO configurations define multiple mongos hosts.

src/LIBMONGOC_VERSION_CURRENT
jmikola: I wasn't able to reproduce that. Starting with an updated `master` branch of the PHPC repo:

```
$ cd src/libmongoc/
$ git fetch origin
$ git checkout r1.15
$ cd ../../
$ make libmongoc-version-current 
cd src/libmongoc/ && python build/calc_release_version.py > ../LIBMONGOC_VERSION_CURRENT
$ cat src/LIBMONGOC_VERSION_CURRENT
1.15.1-20190823+git6f73903ae0
```

https://github.com/mongodb/mongo-c-driver/commit/6f73903ae0 points to a commit in the r1.15 branch, but https://github.com/mongodb/mongo-c-driver/commit/10f6ee9458 points to a commit in the master branch (1.16-dev).

The commit hash suggests that you had actually checked out the master branch instead of r1.15, but I'm not sure where "1.11.1" came from. If I repeat the above steps to checkout the master branch and regenerate the version file I end up with `1.16.0-20190823+git10f6ee9458`.

Do you want to attempt to reproduce this?

jmikola: Upon further review, it does look like this PR is incorrectly updating the submodule to https://github.com/mongodb/mongo-c-driver/commit/10f6ee9458.

scripts/presets/travis/sharded_clusters/cluster.json
kevinAlbs: > Is that correct?

I think so, since we don't have active tests, I ran through a quick manual test (following instructions I documented when reviewing SRV polling [here](https://github.com/kevinAlbs/SRVPollingTesting)). Here's some log output of a test where the initial seedlist has two hosts, then drops to one host:

```
it has been 6 seconds
host_list:
-mongodb1.mongodb.com:27018
-mongodb2.mongodb.com:27017
topology type: Sharded
it has been 7 seconds
host_list:
-mongodb1.mongodb.com:27018
-mongodb2.mongodb.com:27017
topology type: Sharded
it has been 8 seconds
host_list:
-mongodb1.mongodb.com:27018
topology type: Sharded
it has been 9 seconds
host_list:
-mongodb1.mongodb.com:27018
topology type: Sharded
it has been 10 seconds
```

The topology type remains sharded.

Somewhat unrelated, but if the initial seedlist only resolves to one host, we do not poll. I'm not sure if that was intentional in the SRV polling spec. I created https://jira.mongodb.org/browse/SPEC-1429.

src/LIBMONGOC_VERSION_CURRENT
kevinAlbs: Note, the 1.11.1 might be related to https://jira.mongodb.org/browse/CDRIVER-3315 if the make task is checking out a branch other then master or a release branch.

src/MongoDB/Manager.c
```diff
@@ -254,20 +254,33 @@ static void php_phongo_manager_prep_uri_options(zval* options TSRMLS_DC) /* {{{
...
 	mongoc_server_description_t* selected_server;
-	bson_error_t                 error = { 0 };
+	const mongoc_read_prefs_t*   read_preference = NULL;
+	bson_error_t                 error           = { 0 };
 
 	if (!for_writes) {
 		read_preference = zreadPreference ? phongo_read_preference_from_zval(zreadPreference TSRMLS_CC) : mongoc_client_get_read_prefs(client);
 	}
 
+	if (zsession) {
```
jmikola: Do you think this should pre-empty the `if (!for_writes)` condition above? If we end up using a pinned server, there's no need to fetch `read_preference`. Swapping these would keep that related code closer together.

src/MongoDB/Manager.c
```diff
@@ -337,6 +350,7 @@ static PHP_METHOD(Manager, executeCommand)
 	bool                  free_options    = false;
 	zval*                 zreadPreference = NULL;
 	uint32_t              server_id       = 0;
+	zval*                 zsession        = NULL;
```
jmikola: Minor nit, but I'd move this above the `server_id` declaration since it'd used first. Also keeps two `zval*` declarations side-by-side, which looks nice :)

src/MongoDB/Session.c
```diff
@@ -246,6 +234,31 @@ static PHP_METHOD(Session, getOperationTime)
...
 
+/* {{{ proto MongoDB\Driver\Server|null MongoDB\Driver\Session::getServer()
+   Returns the server this session is pinned to */
+static PHP_METHOD(Session, getServer)
+{
+	php_phongo_session_t* intern;
+	uint32_t              server_id = 0;
+
+	intern = Z_SESSION_OBJ_P(getThis());
+	SESSION_CHECK_LIVELINESS(intern, "getOperationTime")
```
jmikola: This should be "getServer". If you look at the macro at the top of this file, it's supposed to refer to the current method in the exception message.

src/MongoDB/Session.c
```diff
@@ -246,6 +234,31 @@ static PHP_METHOD(Session, getOperationTime)
...
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	server_id = mongoc_client_session_get_server_id(intern->client_session);
+
+	/* For sessions without a pinned server, 0 is returned. In this case, we
+	 * return null to force a server selection round. */
```
jmikola: I think we can remove the second sentence. This could be called for a myriad of reasons, and that's just referring to what we expect PHPLIB to do.

src/MongoDB/Session.c
```diff
@@ -246,6 +234,31 @@ static PHP_METHOD(Session, getOperationTime)
...
+	if (server_id == 0) {
```
jmikola: FWIW, `php_phongo_manager_select_server()` does `if (selected_server)`, so we could just do `(!server_id)` here if you prefer. It's not as sloppy to do that in C as it would otherwise be in PHP itself :)

src/MongoDB/Session.c
```diff
@@ -618,6 +627,30 @@ static HashTable* php_phongo_session_get_debug_info(zval* object, int* is_temp T
 		ADD_ASSOC_NULL_EX(&retval, "operationTime");
 	}
 
+	if (intern->client_session) {
+		uint32_t server_id = mongoc_client_session_get_server_id(intern->client_session);
+
+		if (server_id > 0) {
```
jmikola: Since `server_id` is unsigned, `if (server_id)` would suffice.

tests/cursor/cursor-destruct-001.phpt
```diff
@@ -25,22 +25,24 @@ function getNumOpenCursors(MongoDB\Driver\Manager $manager)
 
 $manager = new MongoDB\Driver\Manager(URI);
 
+$server = $manager->selectServer(new \MongoDB\Driver\ReadPreference('primary'));
```
jmikola: I assume this was required because you added a second mongos to the sharded cluster MO config? Should we add a one-line comment here to explain? Feel free to copy-paste it to other places you had to make the same test change.

tests/retryable-reads/retryable-reads-001.phpt
```diff
@@ -26,14 +26,15 @@ class Observer implements MongoDB\Driver\Monitoring\CommandSubscriber
...
 
 $bulk = new MongoDB\Driver\BulkWrite;
 $bulk->insert(['x' => 1]);
 $bulk->insert(['x' => 2]);
 
-$manager->executeBulkWrite(NS, $bulk);
+$server->executeBulkWrite(NS, $bulk);
 
-configureFailPoint($manager, 'failCommand', ['times' => 1], ['failCommands' => ['aggregate'], 'closeConnection' => true]);
+configureTargetedFailPoint($server, 'failCommand', ['times' => 1], ['failCommands' => ['aggregate'], 'closeConnection' => true]);
```
jmikola: For my own curiosity, should this test originally have had `skip_if_multiple_mongos()`, since it set a fail point and didn't also require a replica set?

tests/manager/manager-executeBulkWrite_error-011.phpt
```diff
@@ -12,16 +12,17 @@ MongoDB\Driver\Manager::executeBulkWrite() BulkWriteException inherits labels fr
 require_once __DIR__ . "/../utils/basic.inc";
 
 $manager = new MongoDB\Driver\Manager(URI);
+$server = $manager->selectServer(new \MongoDB\Driver\ReadPreference('primary'));
```
jmikola: This test requires a replica set and wouldn't run on a sharded cluster. Is this necessary?

tests/retryable-reads/retryable-reads-002.phpt
```diff
@@ -26,19 +26,20 @@ class Observer implements MongoDB\Driver\Monitoring\CommandSubscriber
 }
 
 $manager = new MongoDB\Driver\Manager(URI, ['retryReads' => true]);
+$server = $manager->selectServer(new \MongoDB\Driver\ReadPreference('primary'));
```
jmikola: This looks necessary since we may run on a sharded cluster. Good place for a comment.

tests/session/session-debug-005.phpt
```diff
@@ -0,0 +1,88 @@
...
+  ["operationTime"]=>
+  object(MongoDB\BSON\Timestamp)#%d (%d) {
+    ["increment"]=>
+    string(%d) "%d"
+    ["timestamp"]=>
+    string(%d) "%d"
+  }
+  ["server"]=>
+  object(MongoDB\Driver\Server)#%d (%d) {
+    ["host"]=>
```
jmikola: Feel free to replace this body with `%a` (see: [EXPECTF format](https://qa.php.net/expectf_details.php)) since we really just care that the Server is non-null here. We have plenty of redundant assertions in the test suite already, but we can aim to cut down on that going forward. We can leave this for actual Server tests or cases where we care about asserting attributes of the Server.

tests/session/session-pin-001.phpt
```diff
@@ -0,0 +1,56 @@
...
+
+$bulk = new MongoDB\Driver\BulkWrite();
+$bulk->insert(['x' => 1]);
+$manager->executeBulkWrite(NS, $bulk, ['session' => $session]);
+
+var_dump($session->getServer() instanceof \MongoDB\Driver\Server);
+
+$bulk = new MongoDB\Driver\BulkWrite();
+$bulk->insert(['x' => 1]);
+$manager->executeBulkWrite(NS, $bulk, ['session' => $session]);
```
jmikola: May be worth also asserting that the pinned server remains the same. You can rely on the [Server's comparison handler](https://github.com/mongodb/mongo-php-driver/blob/1.6.0alpha3/src/MongoDB/Server.c#L523) (using `==` of course), which does a string comparison of the host and port. Likewise for other pinning tests below.

I don't quite recall why we don't simply compare numeric server IDs there, but you can open a ticket about that if you're also curious and would like to investigate at some point.

tests/session/session-pin-001.phpt
```diff
@@ -0,0 +1,56 @@
+--TEST--
+MongoDB\Driver\Session test: executeBulkWrite pins transaction to server
```
jmikola: I think we should refer to `Manager::executeBulkWrite()` here, since we're not testing the Server method. On that note, I think we'd benefit from testing that the Server methods also pin, too. That would also mean we could add error tests for Server methods to show that they still pass their own `server_id` to libmongoc instead of using a sessions pinned server. We can assert libmongoc's error message for those cases.

Rather than cram all of these into `session-pin-xxx`, what do you think about adding them under the actual method being tested? General convention for tests is to name things after the method under test, but we do deviate in some cases where it's ambiguous (e.g. APM tests, comprehensive cursor tests).

tests/standalone/executiontimeoutexception-001.phpt
```diff
@@ -10,6 +10,7 @@ ExecutionTimeoutException: exceeding $maxTimeMS (queries)
 require_once __DIR__ . "/../utils/basic.inc";
 
 $manager = new MongoDB\Driver\Manager(URI);
+$server = $manager->selectServer(new \MongoDB\Driver\ReadPreference('primary'));
```
jmikola: Good place for a comment.

tests/utils/tools.php
```diff
@@ -220,6 +220,35 @@ function is_mongos($uri)
     return get_primary_server($uri)->getType() === Server::TYPE_MONGOS;
 }
 
+/**
+ * Checks that the topology is a sharded cluster using a replica set
+ */
+function is_mongos_with_replica_set($uri)
```
jmikola: Ripped out of PHPLIB, eh? :smile:

tests/utils/tools.php
```diff
@@ -670,9 +699,23 @@ function configureFailPoint(Manager $manager, $failPoint, $mode, array $data = [
     $manager->executeCommand('admin', $cmd);
 }
 
-function failMaxTimeMS(Manager $manager)
+function configureTargetedFailPoint(Server $server, $failPoint, $mode, array $data = [])
+{
+    $doc = array(
```
jmikola: Short array syntax?

src/LIBMONGOC_VERSION_CURRENT
alcaeus: > Upon further review, it does look like this PR is incorrectly updating the submodule to [mongodb/mongo-c-driver@10f6ee9](https://github.com/mongodb/mongo-c-driver/commit/10f6ee9458).

Yep, that was the mistake: I checked out from master instead of the 1.15 release branch. Fixed now 👍

src/MongoDB/Manager.c
alcaeus: You're absolutely correct. I've switched the order to avoid parsing the read preference when we're using a pinned server anyways.

src/MongoDB/Session.c
alcaeus: Copy/paste alert 😅

src/MongoDB/Session.c
alcaeus: Ah, working with strongly typed languages is always a pleasure. You just have to remember that they work a lot cleaner than PHP does often.

tests/cursor/cursor-destruct-001.phpt
alcaeus: That is correct. All tests that relied on server selection in manager.c now select the server themselves to direct all future operations (`serverStatus` in the case of this test) to the same server.

tests/retryable-reads/retryable-reads-001.phpt
alcaeus: Correct. Since our sharded clusters only had a single mongos, there was no problem with the fail point. Now that there are more of them, we have to create it on a specific server.

tests/manager/manager-executeBulkWrite_error-011.phpt
alcaeus: Technically, you're correct. Looking at the test, I'd think that it also applies to sharded clusters backed by replica set since they also support transactions, correct?

tests/utils/tools.php
alcaeus: Your code was so good I couldn't bring myself to rewriting it 😁

tests/session/session-pin-001.phpt
alcaeus: Comparing that the correct server is pinned is a good idea, I added that.

I've created [PHPC-1429](https://jira.mongodb.org/browse/PHPC-1429) to track optimising the logic of the comparison handler.

scripts/presets/travis/sharded_clusters/cluster.json
alcaeus: > libmongoc considers a URI like "mongodb://foo.example.com/" to be topology type `Single`.

That explains why the PHP driver tests failed with a single mongos. Since it seems impossible for now that we start out with a single mongos and then discover more of them, I'm perfectly fine leaving things in the C driver as is, barring changes from SPEC-1429. This will not require any changes to the PHP driver directly, since we rely on libmongoc to pin the session to a selected server if necessary. I've closed [CDRIVER-3333](https://jira.mongodb.org/browse/CDRIVER-3333) as "Works as Designed".

tests/session/session-pin-001.phpt
alcaeus: I've moved the tests to the `server` and `manager` directories, testing each command, query, and bulkWrite methods in each.

I also followed your suggestion to add error tests for server methods to assert that the commands error out. I did not split these tests off in separate tests, as we'd be duplicating a lot of code to set up the pinned session. Let me know if that's ok with you or if you'd rather keep the two cases separate.

Testing the error case was a good call on your part, as this uncovered a case where libmongoc does not initialise the BSON reply object, causing an invalid call to `bson_destroy` in the PHP driver (see [trace in travis-ci](https://travis-ci.org/mongodb/mongo-php-driver/jobs/576703164#L1261)). I've reported this to the C team as [CDRIVER-3336](https://jira.mongodb.org/browse/CDRIVER-3336).

tests/manager/manager-executeBulkWrite_error-011.phpt
alcaeus: I've changed this on a bunch of tests that were currently only running on replica sets. These tests are now either unrestricted or they run on replica sets and sharded clusters with replica sets.

PR #1015, opened 2019-08-28
PHPC-1431, PHPC-1435: Fix Session::startTransaction() param and arginfo

jmikola:
https://jira.mongodb.org/browse/PHPC-1431
https://jira.mongodb.org/browse/PHPC-1435

### DIFF
--- a/src/MongoDB/Session.c
+++ b/src/MongoDB/Session.c
@@ -367,7 +367,7 @@ static PHP_METHOD(Session, startTransaction)
 	intern = Z_SESSION_OBJ_P(getThis());
 	SESSION_CHECK_LIVELINESS(intern, "startTransaction")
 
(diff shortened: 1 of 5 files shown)


alcaeus:
> The important data is actually the method's arg info, which I just realized is incorrect.

After double-checking, I should've seen that. I checked the `ZEND_ARG_INFO` call and didn't see anything about it being optional or not. `ZEND_ARG_ARRAY_INFO` makes more sense 👍

alcaeus:
@jmikola I've taken the liberty of amending your commit message and fixing the failing test. Since the behaviour will be different on PHP 5 and PHP 7, I've split the test. Since I currently don't have a PHP 5 setup, I'll be relying on evergreen to tell us if something is wrong.

PR #1016, opened 2019-08-29
PHPC-1290: Fix build failures on appveyor and Evergreen

alcaeus:
I was a bit trigger-happy with the merge of #1014. Turns out that while all tests pass on travis-ci, they don't pass on appveyor (three failures because of no running MongoDB) and fail on evergreen as well since we test sharded clusters on MongoDB < 4.2, which breaks the assumption that we can run sharded transactions. This PR introduces a new helper that checks for MongoDB version support depending on what topology we detected: for replicasets, 4.0 is required while we require 4.2 for sharded clusters backed by replica sets.

### DIFF
--- a/tests/manager/manager-executeBulkWrite_error-009.phpt
+++ b/tests/manager/manager-executeBulkWrite_error-009.phpt
@@ -2,6 +2,7 @@
 MongoDB\Driver\Manager::executeBulkWrite() with invalid options
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_live(); ?>
 --FILE--
 <?php
 require_once __DIR__ . "/../utils/basic.inc";
--- a/tests/manager/manager-executeQuery_error-002.phpt
+++ b/tests/manager/manager-executeQuery_error-002.phpt
@@ -1,7 +1,5 @@
 --TEST--
 MongoDB\Driver\Manager::executeQuery() with invalid options
---SKIPIF--
-<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
 --FILE--
 <?php
 require_once __DIR__ . "/../utils/basic.inc";
--- a/tests/server/server-executeBulkWrite_error-002.phpt
+++ b/tests/server/server-executeBulkWrite_error-002.phpt
@@ -2,6 +2,7 @@
 MongoDB\Driver\Server::executeBulkWrite() with invalid options
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_live(); ?>
 --FILE--
 <?php
 require_once __DIR__ . "/../utils/basic.inc";
--- a/tests/server/server-executeQuery_error-001.phpt
+++ b/tests/server/server-executeQuery_error-001.phpt
@@ -2,6 +2,7 @@
 MongoDB\Driver\Server::executeQuery() with invalid options
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_live(); ?>
 --FILE--
 <?php
 require_once __DIR__ . "/../utils/basic.inc";
--- a/tests/session/session-startTransaction_error-004.phpt
+++ b/tests/session/session-startTransaction_error-004.phpt
@@ -3,8 +3,7 @@ MongoDB\Driver\Session::startTransaction() with wrong argument for options array
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
 <?php skip_if_not_libmongoc_crypto() ?>
-<?php skip_if_not_replica_set(); ?>
-<?php skip_if_server_version('<', '4.0'); ?>
+<?php skip_if_no_transactions(); ?>
 <?php skip_if_php_version('<', '7.0.0'); ?>
 --FILE--
 <?php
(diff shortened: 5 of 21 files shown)


alcaeus:
The one appveyor build failure is unrelated, the breaking tests were fixed in all other jobs.

jmikola:
AppVeyor says "Pull request #1016 is non-mergeable" when attempting to restart, so this may require a rebase.

tests/manager/manager-executeBulkWrite_error-009.phpt
```diff
@@ -2,6 +2,7 @@
 MongoDB\Driver\Manager::executeBulkWrite() with invalid options
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_clean(); ?>
```
jmikola: This shouldn't be necessary, as the test doesn't actually insert anything. There's no risk of a duplicate key exception from the test leaving behind data.

`skip_if_not_live()` would be more appropriate here.

jmikola: I'll note that even though we're testing InvalidArgumentExceptions here, those are thrown from [`phongo_execute_bulk_write()`](https://github.com/mongodb/mongo-php-driver/blob/1.6.0alpha3/php_phongo.c#L715), which is _after_ [server selection](https://github.com/mongodb/mongo-php-driver/blob/1.6.0alpha3/src/MongoDB/Manager.c#L515) in the Manager method.

tests/manager/manager-executeQuery_error-002.phpt
```diff
@@ -1,7 +1,5 @@
 --TEST--
 MongoDB\Driver\Manager::executeQuery() with invalid options
---SKIPIF--
-<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
```
jmikola: I think we also want `skip_if_not_live()` here. Although [read preference validation](https://github.com/mongodb/mongo-php-driver/blob/1.6.0alpha3/src/MongoDB/Manager.c#L474) occurs _before_ server selection (and for some reason [again in `phongo_execute_query()`](https://github.com/mongodb/mongo-php-driver/blob/1.6.0alpha3/php_phongo.c#L852)), [session validation](https://github.com/mongodb/mongo-php-driver/blob/1.6.0alpha3/php_phongo.c#L859) only occurs _after_ server selection.

I'm not sure how this test could pass without a live server.

tests/server/server-executeBulkWrite_error-002.phpt
```diff
@@ -2,6 +2,7 @@
 MongoDB\Driver\Server::executeBulkWrite() with invalid options
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_clean(); ?>
```
jmikola: `skip_if_not_live()`

tests/server/server-executeQuery_error-001.phpt
```diff
@@ -2,6 +2,7 @@
 MongoDB\Driver\Server::executeQuery() with invalid options
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_clean(); ?>
```
jmikola: `skip_if_not_live()`. I believe this is the Server equivalent to `manager-executeQuery_error-002.phpt`.

tests/utils/skipif.php
```diff
@@ -77,6 +77,17 @@ function skip_if_not_replica_set_or_mongos_with_replica_set()
     is_replica_set(URI) or is_mongos_with_replica_set(URI) or exit('skip topology is not a replica set or sharded cluster with replica set');
 }
 
+function skip_if_no_transactions()
```
jmikola: I see you didn't remove `skip_if_not_replica_set_or_mongos_with_replica_set()`. Is that because it's still relevant for the change stream and causal consistency tests? I assume so, as they also don't need to distinguish 4.2+ for sharded clusters and thus could not utilize _this_ function.

tests/manager/manager-executeQuery_error-002.phpt
alcaeus: With the introduction of mongos pinning, the session is also validated [before server selection](https://github.com/mongodb/mongo-php-driver/blob/a85e0922c9d1c2a82fb02eef6625d75b02721a79/src/MongoDB/Manager.c#L512).

The reason the equivalent test for the server needs the `skip_if_not_live` check is that we perform server selection manually in the test.

tests/utils/skipif.php
alcaeus: That is correct - tests for causal consistency require a replica set, regardless of it being standalone or behind mongos. The reason for the separate transactions helper here is to account for different versions depending on whether we're running on mongos with replica sets or a replica set.

tests/session/session-startTransaction_error-004.phpt
```diff
@@ -0,0 +1,35 @@
+--TEST--
+MongoDB\Driver\Session::startTransaction() with wrong argument for options array on PHP 7.0
```
alcaeus: This test on PHP 7.0 was necessary because of build failures: https://evergreen.mongodb.com/task_log_raw/mongo_php_driver_tests_php7__versions~4.2_php_versions~7.0_os_php7~debian92_test_test_replicaset_auth_a85e0922c9d1c2a82fb02eef6625d75b02721a79_19_08_29_14_45_15/0?type=T#L1425

I suspect that the issue stems from nullable types being introduced in PHP 7.1 only, which is why the error message omits "or null" on PHP 7.0.

alcaeus:
Merged manually in 5eedf502.

PR #1017, opened 2019-08-29
Change order of PECL release steps in contribution docs

alcaeus:
This is a leftover after last weeks' alpha release. The PECL release instructions have been split: the `package.xml` file is still built after bumping version information, but instead of publishing right away we now publish the package file after publishing the tags to GitHub. This way errors during installation can be fixed before tagging, while at the same time ensuring that the tag is published before publishing the PECL release.


alcaeus:
Merged manually in 502247da.

PR #1022, opened 2019-09-13
PHPC-1281: Serialization support for ReadConcern, ReadPreference, and WriteConcern

alcaeus:
https://jira.mongodb.org/browse/PHPC-1281

### DIFF
--- a/src/MongoDB/ReadConcern.c
+++ b/src/MongoDB/ReadConcern.c
@@ -16,6 +16,12 @@
 
 #include <php.h>
 #include <Zend/zend_interfaces.h>
+#include <ext/standard/php_var.h>
+#if PHP_VERSION_ID >= 70000
+#include <zend_smart_str.h>
+#else
+#include <ext/standard/php_smart_str.h>
+#endif
 
 #ifdef HAVE_CONFIG_H
 #include "config.h"
@@ -197,6 +203,102 @@ static PHP_METHOD(ReadConcern, bsonSerialize)
 	convert_to_object(return_value);
 } /* }}} */
 
+/* {{{ proto string MongoDB\Driver\ReadConcern::serialize()
+*/
+static PHP_METHOD(ReadConcern, serialize)
+{
+	php_phongo_readconcern_t* intern;
+	ZVAL_RETVAL_TYPE          retval;
+	php_serialize_data_t      var_hash;
+	smart_str                 buf = { 0 };
+	const char*               level;
+
+	intern = Z_READCONCERN_OBJ_P(getThis());
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	if (!intern->read_concern) {
+		return;
+	}
+
+	level = mongoc_read_concern_get_level(intern->read_concern);
+
+	if (!level) {
+		PHONGO_RETURN_STRING("");
+	}
+
+#if PHP_VERSION_ID >= 70000
+	array_init_size(&retval, 1);
+	ADD_ASSOC_STRING(&retval, "level", level);
+#else
+	ALLOC_INIT_ZVAL(retval);
+	array_init_size(retval, 1);
+	ADD_ASSOC_STRING(retval, "level", level);
+#endif
+
+	PHP_VAR_SERIALIZE_INIT(var_hash);
+	php_var_serialize(&buf, &retval, &var_hash TSRMLS_CC);
+	smart_str_0(&buf);
+	PHP_VAR_SERIALIZE_DESTROY(var_hash);
+
+	PHONGO_RETVAL_SMART_STR(buf);
+
+	smart_str_free(&buf);
+	zval_ptr_dtor(&retval);
+} /* }}} */
+
+/* {{{ proto void MongoDB\Driver\ReadConcern::unserialize(string $serialized)
+*/
+static PHP_METHOD(ReadConcern, unserialize)
+{
+	php_phongo_readconcern_t* intern;
+	zend_error_handling       error_handling;
+	char*                     serialized;
+	phongo_zpp_char_len       serialized_len;
+#if PHP_VERSION_ID >= 70000
+	zval props;
+#else
+	zval* props;
+#endif
+	php_unserialize_data_t var_hash;
+
+	intern = Z_READCONCERN_OBJ_P(getThis());
+
+	zend_replace_error_handling(EH_THROW, phongo_exception_from_phongo_domain(PHONGO_ERROR_INVALID_ARGUMENT), &error_handling TSRMLS_CC);
+
+	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "s", &serialized, &serialized_len) == FAILURE) {
+		zend_restore_error_handling(&error_handling TSRMLS_CC);
+		return;
+	}
+	zend_restore_error_handling(&error_handling TSRMLS_CC);
+
+	if (!serialized_len) {
+		return;
+	}
+
+#if PHP_VERSION_ID < 70000
+	ALLOC_INIT_ZVAL(props);
+#endif
+	PHP_VAR_UNSERIALIZE_INIT(var_hash);
+	if (!php_var_unserialize(&props, (const unsigned char**) &serialized, (unsigned char*) serialized + serialized_len, &var_hash TSRMLS_CC)) {
+		zval_ptr_dtor(&props);
+		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "%s unserialization failed", ZSTR_VAL(php_phongo_readconcern_ce->name));
+
+		PHP_VAR_UNSERIALIZE_DESTROY(var_hash);
+		return;
+	}
+	PHP_VAR_UNSERIALIZE_DESTROY(var_hash);
+
+#if PHP_VERSION_ID >= 70000
+	php_phongo_readconcern_init_from_hash(intern, HASH_OF(&props) TSRMLS_CC);
+#else
+	php_phongo_readconcern_init_from_hash(intern, HASH_OF(props) TSRMLS_CC);
+#endif
+	zval_ptr_dtor(&props);
+} /* }}} */
+
 /* {{{ MongoDB\Driver\ReadConcern function entries */
 ZEND_BEGIN_ARG_INFO_EX(ai_ReadConcern___construct, 0, 0, 0)
 	ZEND_ARG_INFO(0, level)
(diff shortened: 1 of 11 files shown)


src/MongoDB/ReadConcern.c
```diff
@@ -197,6 +203,102 @@ static PHP_METHOD(ReadConcern, bsonSerialize)
...
+	}
+
+	if (!intern->read_concern) {
+		return;
+	}
+
+	level = mongoc_read_concern_get_level(intern->read_concern);
+
+	if (!level) {
+		PHONGO_RETURN_STRING("");
```
jmikola: I assume this was inspired by the MinKey and MaxKey methods? AFAIK, those are the only BSON types with no internal data.

src/MongoDB/ReadPreference.c
```diff
@@ -369,6 +375,31 @@ static PHP_METHOD(ReadPreference, getTagSets)
 	}
 } /* }}} */
 
+static const char* php_phongo_readpreference_get_mode_string(mongoc_read_mode_t mode TSRMLS_DC) /* {{{ */
+{
+	switch (mode) {
+		case MONGOC_READ_PRIMARY:
+			return "primary";
+			break;
```
jmikola: You can omit the `break` after a `return` within a `case` block. It's technically an unreachable statement anyway.

src/MongoDB/ReadPreference.c
```diff
@@ -369,6 +375,31 @@ static PHP_METHOD(ReadPreference, getTagSets)
 	}
 } /* }}} */
 
+static const char* php_phongo_readpreference_get_mode_string(mongoc_read_mode_t mode TSRMLS_DC) /* {{{ */
```
jmikola: I don't think this function needs to take `TSRMLS_DC`, as it doesn't call any internal PHP APIs that would require a `TSRMLS_CC`.

src/MongoDB/ReadPreference.c
```diff
@@ -387,26 +418,7 @@ static HashTable* php_phongo_readpreference_get_properties_hash(zval* object, bo
...
-		case MONGOC_READ_SECONDARY_PREFERRED:
-			modeString = "secondaryPreferred";
-			break;
-		case MONGOC_READ_NEAREST:
-			modeString = "nearest";
-			break;
-		default: /* Do nothing */
-			break;
-	}
+	modeString = php_phongo_readpreference_get_mode_string(mode TSRMLS_CC);
```
jmikola: Per my earlier comment, I don't think `TSRMLS_CC` is needed.

src/MongoDB/ReadPreference.c
```diff
@@ -471,6 +483,132 @@ static PHP_METHOD(ReadPreference, bsonSerialize)
...
+		return;
+	}
+
+	if (!intern->read_preference) {
+		return;
+	}
+
+	tags                = mongoc_read_prefs_get_tags(intern->read_preference);
+	mode                = mongoc_read_prefs_get_mode(intern->read_preference);
+	modeString          = php_phongo_readpreference_get_mode_string(mode TSRMLS_CC);
```
jmikola: Per my earlier comment, I don't think `TSRMLS_CC` is needed.

src/MongoDB/ReadPreference.c
```diff
@@ -471,6 +483,132 @@ static PHP_METHOD(ReadPreference, bsonSerialize)
...
+#else
+		ADD_ASSOC_STRING(retval, "mode", modeString);
+#endif
+	}
+
+	if (!bson_empty0(tags)) {
+		php_phongo_bson_state state = PHONGO_BSON_STATE_INITIALIZER;
+		/* Use native arrays for debugging output */
+		state.map.root_type     = PHONGO_TYPEMAP_NATIVE_ARRAY;
+		state.map.document_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
```
jmikola: If #1030 is merged first, this can be rebased to use the new macro.

src/MongoDB/WriteConcern.c
```diff
@@ -406,6 +412,134 @@ static PHP_METHOD(WriteConcern, bsonSerialize)
...
+
+	if (!intern->write_concern) {
+		return;
+	}
+
+	wtag     = mongoc_write_concern_get_wtag(intern->write_concern);
+	w        = mongoc_write_concern_get_w(intern->write_concern);
+	/* Note: PHP currently enforces that wimeoutMS is a 32-bit integer, so
+	 * casting will never truncate the value. This may change with PHPC-1411. */
+	wtimeout = (int32_t) mongoc_write_concern_get_wtimeout_int64(intern->write_concern);
```
jmikola: #1042 likely has implications for this, so whichever PR is merged first will require the other to be revised.

Since serialization is a case where we should definitely concern ourselves with portability between 32-bit and 64-bit architectures, this may warrant always serializing `wtimeout` values larger than `INT32_MAX` as an integer string and allowing `php_phongo_writeconcern_init_from_hash` to accept either type. Such an implementation would also allow us to write a single test that works on either architecture, since we could craft a serialized string that forces a `wtimeout` in the 64-bit range and then assert that it is printed accordingly by the debug handler (modified in #1042).

See [tests/functional/cursorid-001.phpt](https://github.com/mongodb/mongo-php-driver/blob/master/tests/functional/cursorid-001.phpt) for an example `EXPECTF` where we assert either string or integer output in a debug handler. I suppose an architecture-agnostic test for `wtimeout` could do the same.

src/MongoDB/WriteConcern.c
```diff
@@ -406,6 +412,134 @@ static PHP_METHOD(WriteConcern, bsonSerialize)
...
+	} else if (w != MONGOC_WRITE_CONCERN_W_DEFAULT) {
+		ADD_ASSOC_LONG_EX(&retval, "w", w);
+	}
+
+	if (mongoc_write_concern_journal_is_set(intern->write_concern)) {
+		ADD_ASSOC_BOOL_EX(&retval, "j", mongoc_write_concern_get_journal(intern->write_concern));
+	}
+
+	if (wtimeout != 0) {
+		ADD_ASSOC_LONG_EX(&retval, "wtimeout", wtimeout);
```
jmikola: Related to my comment above, this will likely need to be changed to conditionally use `ADD_ASSOC_STRING` for values larger than `INT32_MAX`. Likewise below in the PHP 5.x `#else` case.

tests/readPreference/readpreference-serialization-001.phpt
```diff
@@ -0,0 +1,187 @@
...
+
+require_once __DIR__ . '/../utils/tools.php';
+
+$tests = [
+    new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_PRIMARY),
+    new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_PRIMARY_PREFERRED),
+    new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_SECONDARY),
+    new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_SECONDARY_PREFERRED),
+    new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_NEAREST),
+    new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_PRIMARY, []),
```
jmikola: Would it make sense to also use `RP_SECONDARY` here? I realize an empty tag set is equivalent to omitting this argument entirely, so this is slightly pedantic. I also acknowledge that `mongoc_read_prefs_new()` and `mongoc_read_prefs_set_tags()` will both just call `bson_init` on the struct's `tags` field.

tests/writeConcern/writeconcern-serialization-001.phpt
```diff
@@ -0,0 +1,192 @@
...
+    new MongoDB\Driver\WriteConcern(0),
+    new MongoDB\Driver\WriteConcern(1),
+    new MongoDB\Driver\WriteConcern(MongoDB\Driver\WriteConcern::MAJORITY),
+    new MongoDB\Driver\WriteConcern('tag'),
+    new MongoDB\Driver\WriteConcern(1, 0),
+    new MongoDB\Driver\WriteConcern(1, 0, false),
+    new MongoDB\Driver\WriteConcern(1, 1000),
+    new MongoDB\Driver\WriteConcern(1, 1000, true),
+    new MongoDB\Driver\WriteConcern(-2, 0, true),
+    // Note: wtimeout is only applicable for w > 1
```
jmikola: I believe this was copied (and corrected) from the following tests, which all state "Note: wtimeout is only applicable applies for w > 1".

 * writeconcern-bsonserialize-001.phpt
 * writeconcern-bsonserialize-002.phpt
 * writeconcern-debug-003.phpt
 * writeconcern-var_export-001.phpt

Would you mind correcting those in a separate commit in this PR?

tests/writeConcern/writeconcern-serialization-001.phpt
```diff
@@ -0,0 +1,192 @@
...
+    new MongoDB\Driver\WriteConcern(-3), // MONGOC_WRITE_CONCERN_W_MAJORITY
+    new MongoDB\Driver\WriteConcern(-2), // MONGOC_WRITE_CONCERN_W_DEFAULT
+    new MongoDB\Driver\WriteConcern(-1),
+    new MongoDB\Driver\WriteConcern(0),
+    new MongoDB\Driver\WriteConcern(1),
+    new MongoDB\Driver\WriteConcern(MongoDB\Driver\WriteConcern::MAJORITY),
+    new MongoDB\Driver\WriteConcern('tag'),
+    new MongoDB\Driver\WriteConcern(1, 0),
+    new MongoDB\Driver\WriteConcern(1, 0, false),
+    new MongoDB\Driver\WriteConcern(1, 1000),
```
jmikola: Since it will be quite impossible to add an architecture-agnostic test case with a 64-bit `wtimeout` value in this file, I'd propose a new `writeconcern-serialization-002.phpt` test, where you can start with a serialized WriteConcern string, unserialized and dump it, and then re-serialize and verify that it produces the same output as the original, serialized string. That would implicitly also test the debug handler with an EXPECTF allowing either string/integer output, per my earlier comment in this PR.

src/MongoDB/ReadConcern.c
alcaeus: That is correct.

src/MongoDB/WriteConcern.c
alcaeus: I'd like to get this merged and deal with 64-bit changes in #1042.

src/MongoDB/ReadPreference.c
alcaeus: I'd like to get this merged first before dealing with the macros and rebase #1030 instead.

tests/writeConcern/writeconcern-serialization-001.phpt
alcaeus: Yep, will take care of this when adding 64-bit support in #1042.

src/MongoDB/WriteConcern.c
jmikola: Acknowledging that this will be addressed with #1042.

PR #1024, opened 2019-09-23
PHPC-991: Handle duplicates in URI options array

alcaeus:
http://jira.mongodb.org/browse/PHPC-991

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1385,7 +1385,7 @@ void php_phongo_write_concern_to_zval(zval* retval, const mongoc_write_concern_t
 } /* }}} */
 /* }}} */
 
-static mongoc_uri_t* php_phongo_make_uri(const char* uri_string, bson_t* options TSRMLS_DC) /* {{{ */
+static mongoc_uri_t* php_phongo_make_uri(const char* uri_string TSRMLS_DC) /* {{{ */
 {
 	mongoc_uri_t* uri;
 	bson_error_t  error = { 0 };
@@ -2654,7 +2654,7 @@ void phongo_manager_init(php_phongo_manager_t* manager, const char* uri_string, 
 		goto cleanup;
 	}
 
-	if (!(uri = php_phongo_make_uri(uri_string, &bson_options TSRMLS_CC))) {
+	if (!(uri = php_phongo_make_uri(uri_string TSRMLS_CC))) {
 		/* Exception should already have been thrown */
 		goto cleanup;
 	}
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1769,6 +1769,7 @@ static bool php_phongo_apply_rp_options_to_uri(mongoc_uri_t* uri, bson_t* option
 	bson_iter_t                iter;
 	mongoc_read_prefs_t*       new_rp;
 	const mongoc_read_prefs_t* old_rp;
+	bool                       ignore_slaveok = false;
 
 	if (!(old_rp = mongoc_uri_get_read_prefs_t(uri))) {
 		phongo_throw_exception(PHONGO_ERROR_MONGOC_FAILED TSRMLS_CC, "mongoc_uri_t does not have a read preference");
@@ -1786,7 +1787,7 @@ static bool php_phongo_apply_rp_options_to_uri(mongoc_uri_t* uri, bson_t* option
 	while (bson_iter_next(&iter)) {
 		const char* key = bson_iter_key(&iter);
 
-		if (!strcasecmp(key, MONGOC_URI_SLAVEOK)) {
+		if (!ignore_slaveok && !strcasecmp(key, MONGOC_URI_SLAVEOK)) {
 			if (!BSON_ITER_HOLDS_BOOL(&iter)) {
 				PHONGO_URI_INVALID_TYPE(iter, "boolean");
 				mongoc_read_prefs_destroy(new_rp);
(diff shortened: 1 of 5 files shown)


php_phongo.c
```diff
@@ -1777,140 +1778,146 @@ static bool php_phongo_apply_rp_options_to_uri(mongoc_uri_t* uri, bson_t* option
...
-			mongoc_read_prefs_destroy(new_rp);
+			if (!php_phongo_read_preference_tags_are_valid(&tags)) {
+				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Read preference tags must be an array of zero or more documents");
+				mongoc_read_prefs_destroy(new_rp);
 
-			return false;
+				return false;
+			}
+
+			mongoc_read_prefs_set_tags(new_rp, &tags);
```
jmikola: I've confirmed that it's safe to call `mongoc_read_prefs_set_tags` twice, on the off chance there are duplicate `readPreferenceTags` options.

php_phongo.c
```diff
@@ -1777,140 +1778,146 @@ static bool php_phongo_apply_rp_options_to_uri(mongoc_uri_t* uri, bson_t* option
...
-			if (max_staleness_seconds > INT32_MAX) {
-				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected maxStalenessSeconds to be <= %d, %" PRId64 " given", INT32_MAX, max_staleness_seconds);
-				mongoc_read_prefs_destroy(new_rp);
+			mongoc_read_prefs_set_max_staleness_seconds(new_rp, max_staleness_seconds);
+		}
+	}
 
-				return false;
-			}
+	if (mongoc_read_prefs_get_mode(new_rp) == MONGOC_READ_PRIMARY &&
```
jmikola: Acknowledging that these two checks need to come after the `while` loop.

php_phongo.c
```diff
@@ -1777,140 +1778,146 @@ static bool php_phongo_apply_rp_options_to_uri(mongoc_uri_t* uri, bson_t* option
...
+		mongoc_read_prefs_get_max_staleness_seconds(new_rp) != MONGOC_NO_MAX_STALENESS) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Primary read preference mode conflicts with maxStalenessSeconds");
+		mongoc_read_prefs_destroy(new_rp);
 
-		mongoc_read_prefs_set_max_staleness_seconds(new_rp, max_staleness_seconds);
+		return false;
 	}
 
 	/* This may be redundant in light of the last check (primary with tags), but
```
jmikola: While we're modifying this function, can we update this comment to "in light of the previous checks (primary with tags or maxStalenessSeconds)"?

tests/manager/manager-ctor-duplicate-option-001.phpt
```diff
@@ -0,0 +1,15 @@
+--TEST--
+MongoDB\Driver\Manager::__construct() with duplicate read preference option
+--FILE--
+<?php
+
+$manager = new MongoDB\Driver\Manager(null, ['readPreference' => 'primary', 'readpreference' => 'secondary']);
```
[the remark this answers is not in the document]
alcaeus: That's a good point. I'd assume that overwriting is indeed the behaviour we're looking for, as opposed to merging tag sets, correct?

php_phongo.c
```diff
@@ -1385,7 +1385,7 @@ void php_phongo_write_concern_to_zval(zval* retval, const mongoc_write_concern_t
 } /* }}} */
 /* }}} */
 
-static mongoc_uri_t* php_phongo_make_uri(const char* uri_string, bson_t* options TSRMLS_DC) /* {{{ */
+static mongoc_uri_t* php_phongo_make_uri(const char* uri_string TSRMLS_DC) /* {{{ */
```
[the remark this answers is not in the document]
alcaeus: That is correct, I noticed that while analysing what's going on with the options. Decided to drop it in a separate commit since it's not useful.

php_phongo.c
alcaeus: Yep, any validation we do has to be done after the fact, not within the loop as that's not what the options will look like while we're done 👍

php_phongo.c
alcaeus: That didn't occur to me, thanks for confirming. I'll add a test that overwrites `readPreferenceTags` as you suggested above, which will continuously ensure this works as expected.

tests/manager/manager-ctor-duplicate-option-001.phpt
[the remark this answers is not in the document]
alcaeus: Test added in squashed commit.

php_phongo.c
alcaeus: Updated the comment as per your suggestion.

tests/manager/manager-ctor-duplicate-option-001.phpt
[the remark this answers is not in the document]
jmikola: Correct. Merging tag sets is done for the connection string because we otherwise have no way to build an array (similar to how most frameworks interpret duplicate `key[]` form fields). But that doesn't make sense for the array options, since an entire tag set can easily be expressed in a single key. Therefore, overwriting makes sense in all cases for array options.

php_phongo.c
jmikola: It would surely be a bug in libmongoc if this leaked, but I just thought to double-check since it's a non-scalar option, unlike most of the other things we're setting in this PR (e.g. mode strings and integers).

tests/manager/manager-ctor-duplicate-option-001.phpt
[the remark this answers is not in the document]
alcaeus: That sounds good. Thanks for confirming!

PR #1030, opened 2019-10-02
PHPC-952: Move BSON type initiators to bson.c

alcaeus:
https://jira.mongodb.org/browse/PHPC-952


src/bson.c
```diff
@@ -470,14 +515,41 @@ static bool php_phongo_bson_visit_null(const bson_iter_t* iter ARG_UNUSED, const
...
+
+	intern              = Z_REGEX_OBJ_P(object);
+	intern->pattern_len = strlen(pattern);
+	intern->pattern     = estrndup(pattern, intern->pattern_len);
+	intern->flags_len   = strlen(flags);
+	intern->flags       = estrndup(flags, intern->flags_len);
+
+	/* Ensure flags are alphabetized upon initialization. This may be removed
+	 * once CDRIVER-1883 is implemented. */
+	qsort((void*) intern->flags, intern->flags_len, 1, php_phongo_regex_compare_flags);
```
jmikola: [CDRIVER-1883](https://jira.mongodb.org/browse/CDRIVER-1883) was fixed in 1.7.0. Can this and `php_phongo_regex_compare_flags` be removed?

This code was originally added in https://github.com/mongodb/mongo-php-driver/commit/e9271bf8a9919e6e85cc9372708f0841aa4fcd00 for [PHPC-885](https://jira.mongodb.org/browse/PHPC-885). Looking at that issue, the BSON corpus tests should provide coverage for this behavior.

src/bson.c
```diff
@@ -536,14 +619,33 @@ static bool php_phongo_bson_visit_symbol(const bson_iter_t* iter, const char* ke
 	return false;
 } /* }}} */
 
+static void php_phongo_bson_new_javascript_from_javascript_and_scope(int init, zval* object, const char* code, size_t code_len, const bson_t* scope TSRMLS_DC) /* {{{ */
```
jmikola: Currently, `php_phongo_new_javascript_from_javascript_and_scope` and `php_phongo_new_javascript_from_javascript` are only called with `init=1` from `bson.c`.

The `init` argument here dates back to https://github.com/mongodb/mongo-php-driver/commit/ab8d22c7ffcbace3d383a951ad0c40fbe15ac025, when `Javascript::__construct()` used to rely on this method. Since the PHP constructor initializes the object before the extension's method is even called, we needed a flag to toggle whether we should initialize in this function. That was ultimately removed in https://github.com/mongodb/mongo-php-driver/commit/41081dae61c56003a510da42484d617b5cfaad64 when we introduced serialization and `var_export` support for the Javascript BSON type, which means `init` no longer serves a purpose.

We should be able to remove it entirely, which will make this function consistent with the other BSON types.

src/MongoDB/Cursor.c
```diff
@@ -291,6 +291,16 @@ static int php_phongo_cursor_to_array_apply(zend_object_iterator* iter, void* pu
 	return ZEND_HASH_APPLY_KEEP;
 } /* }}} */
 
+static void php_phongo_cursor_id_new_from_id(zval* object, int64_t cursorid TSRMLS_DC) /* {{{ */
```
jmikola: Good catch.

This makes we wonder if we should consider adding serialization support to CursorId, since it is technically just a value object. If so, I think we would just need to always serialize the internal ID as a numeric string for portability across 32-bit and 64-bit architectures. This relates to my other comments about WriteConcern `wtimeout` handling in #1022 (related to #1042). Would you mind opening a new PHPC ticket to record this?

alcaeus: Done, see [PHPC-1459](https://jira.mongodb.org/browse/PHPC-1459).

PR #1033, opened 2019-10-07
PHPC-1074: Refactor php_phongo_bson_state struct initialization for debug output

alcaeus:
https://jira.mongodb.org/browse/PHPC-1074

I opted for the separate macro approach to convey that this is purely for debug purposes. For other purposes (there are a couple of instances where we change the `root_type` in the type map), it is still preferred to initialise the default state and update the type map manually.

### DIFF
--- a/php_bson.h
+++ b/php_bson.h
@@ -83,17 +83,16 @@ typedef struct {
 	php_phongo_field_path*  field_path;
 } php_phongo_bson_state;
 
-#if PHP_VERSION_ID >= 70000
-#define PHONGO_BSON_STATE_INITIALIZER                                                                              \
-	{                                                                                                              \
-		{ { 0 } }, { PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL }, NULL, NULL \
-	}
-#else
-#define PHONGO_BSON_STATE_INITIALIZER                                                                         \
-	{                                                                                                         \
-		NULL, { PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL }, NULL, NULL \
-	}
-#endif
+#define PHONGO_BSON_INIT_STATE(s)                       \
+	do {                                                \
+		memset(&(s), 0, sizeof(php_phongo_bson_state)); \
+	} while (0)
+#define PHONGO_BSON_INIT_DEBUG_STATE(s)                    \
+	do {                                                   \
+		memset(&(s), 0, sizeof(php_phongo_bson_state));    \
+		s.map.root_type     = PHONGO_TYPEMAP_NATIVE_ARRAY; \
+		s.map.document_type = PHONGO_TYPEMAP_NATIVE_ARRAY; \
+	} while (0)
 
 void php_phongo_zval_to_bson(zval* data, php_phongo_bson_flags_t flags, bson_t* bson, bson_t** bson_out TSRMLS_DC);
 bool php_phongo_bson_to_zval_ex(const unsigned char* data, int data_len, php_phongo_bson_state* state);
--- a/src/MongoDB/Monitoring/CommandFailedEvent.c
+++ b/src/MongoDB/Monitoring/CommandFailedEvent.c
@@ -97,7 +97,9 @@ PHP_METHOD(CommandFailedEvent, getOperationId)
 PHP_METHOD(CommandFailedEvent, getReply)
 {
 	php_phongo_commandfailedevent_t* intern;
-	php_phongo_bson_state            state = PHONGO_BSON_STATE_INITIALIZER;
+	php_phongo_bson_state            state;
+
+	PHONGO_BSON_INIT_STATE(state);
 
 	intern = Z_COMMANDFAILEDEVENT_OBJ_P(getThis());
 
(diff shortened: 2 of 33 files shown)


php_bson.h
```diff
@@ -88,11 +88,19 @@ typedef struct {
 	{                                                                                                              \
 		{ { 0 } }, { PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NONE, NULL }, NULL, NULL \
 	}
+#define PHONGO_DEBUG_BSON_STATE_INITIALIZER                                                                                        \
+	{                                                                                                                              \
+		{ { 0 } }, { PHONGO_TYPEMAP_NATIVE_ARRAY, NULL, PHONGO_TYPEMAP_NONE, NULL, PHONGO_TYPEMAP_NATIVE_ARRAY, NULL }, NULL, NULL \
```
jmikola: Even before this PR, I think the state initializer macros were out of sync with the definitions of `php_phongo_bson_state` and `php_phongo_bson_typemap`.

If we look at https://github.com/mongodb/mongo-php-driver/commit/157481e60ff4ac06161c1f7e06fe60740616ed93#diff-079eefa18a7831e711dedc5993dba146R70 from [PHPC-922](https://jira.mongodb.org/browse/PHPC-922), the trailing `0` (for `is_visiting_array`) was replaced with `NULL`, when I think a `NULL` should have been _added_ for the new `field_path` pointer field in the `php_phongo_bson_state` struct.

Similarly, https://github.com/mongodb/mongo-php-driver/commit/bdda892b2ea3163848b4c4409d3f6968a437cad9 for [PHPC-314](https://jira.mongodb.org/browse/PHPC-314) added a new `field_paths` struct to the end of `php_phongo_bson_typemap` struct without modifying the initializer macro.

I expect this hasn't been problematic in the past, as we may not access those extra fields and/or the memory happens to be cleared; however, I think this would be a good opportunity to clear this up.

Alternatively, you may want to replace all of this with two macros. One that calls `memset()` on the `php_phongo_bson_state` struct to clear it out (as in [this thread](https://stackoverflow.com/a/11152208/162228)) and another that assigns specific fields for the debug type map. Since `PHONGO_TYPEMAP_NONE` is the first element in its enum, we can rely on it being equal to zero (see: [this thread](https://stackoverflow.com/q/6434105/162228)).

This approach would also allow us to remove the conditional definitions for PHP 5 and 7, since all we really care about doing is assigning two type map fields. Consider:

```c
#define INIT_PHONGO_BSON_STATE(s) \
    do { \
        memset(&(s), 0, sizeof(php_phongo_bson_state));
    } while (0)
#define INIT_PHONGO_DEBUG_BSON_STATE(s) \
    do { \
        memset(&(s), 0, sizeof(php_phongo_bson_state)); \
        s.map.root_type     = PHONGO_TYPEMAP_NATIVE_ARRAY; \
        s.map.document_type = PHONGO_TYPEMAP_NATIVE_ARRAY; \
    } while (0)
```

Thoughts?

For extended reading, I can point out that these macros have historically been a pain in the tuchus. See:

 * https://github.com/mongodb/mongo-php-driver/commit/321c577f982e79459fdef07de0a047cfa514e741
 * https://github.com/mongodb/mongo-php-driver/commit/a869c9746a0e25306f81ca0dddffaa12bef63a7f
 * https://github.com/mongodb/mongo-php-driver/pull/591

alcaeus: That looks a lot cleaner, as it doesn't require us to continuously modify the initialiser when adding to the struct. Unfortunately, the syntax isn't as easy as previously where we could initialise the struct on declaration, but the additional call isn't that much of a problem.

src/MongoDB/Monitoring/CommandFailedEvent.c
```diff
@@ -245,6 +247,7 @@ static HashTable* php_phongo_commandfailedevent_get_debug_info(zval* object, int
 
 	sprintf(operation_id, "%" PRIu64, intern->operation_id);
 	ADD_ASSOC_STRING(&retval, "operationId", operation_id);
+	PHONGO_BSON_INIT_STATE(reply_state);
```
jmikola: Would it make sense to move this as close as possible to the declaration? It'd certainly be clearer here, but possibly also in methods earlier in the PR (before we assign `intern`). That would minimize any risk of us introduce new code that uses the state without it being properly initialized.

src/MongoDB/Monitoring/CommandStartedEvent.c
```diff
@@ -31,14 +31,15 @@ zend_class_entry* php_phongo_commandstartedevent_ce;
...
-	php_phongo_bson_state             state = PHONGO_BSON_STATE_INITIALIZER;
+	php_phongo_bson_state             state;
 
 	intern = Z_COMMANDSTARTEDEVENT_OBJ_P(getThis());
 
 	if (zend_parse_parameters_none() == FAILURE) {
 		return;
 	}
 
+	PHONGO_BSON_INIT_STATE(state);
```
jmikola: This is another case where I'd suggest moving this immediately after the declaration. Even though it's clearly not needed for the failure case, the overhead isn't much and it minimizes the risk of misuse by later commits.

alcaeus: In this case I put the initialisation in this line to avoid the `memcpy` call if we're exiting early due to a parameter parsing failure. I've done this in a few more places.

jmikola: Correction: `memset` :)

Aye, I'm fine with the early `memset` call, since the early returns are all for error cases representing an unlikely code path.

alcaeus: Whoops ;)

Fair enough, I'll move the initialisations up to be right after the declaration.

PR #1034, opened 2019-10-07
PHPC-348: Check for errors after calling bson_to_zval()

alcaeus:
https://jira.mongodb.org/browse/PHPC-348

Note: there are three occurrences of `php_phongo_bson_to_zval_ex` in `Cursor.c` ([1](https://github.com/mongodb/mongo-php-driver/blob/987fa38d7f0631bae407f122664e706b6593712f/src/MongoDB/Cursor.c#L126), [2](https://github.com/mongodb/mongo-php-driver/blob/987fa38d7f0631bae407f122664e706b6593712f/src/MongoDB/Cursor.c#L167), [3](https://github.com/mongodb/mongo-php-driver/blob/987fa38d7f0631bae407f122664e706b6593712f/src/MongoDB/Cursor.c#L255)) where I wasn't sure if freeing the result was a good idea.

When used in the `debug_info` functions, I thought it was best to add `null` values and continue assembling the output, but we may want to return without processing any further properties. I'm not quite sure what the preferred course of action is.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -244,13 +244,14 @@ void phongo_throw_exception_from_bson_error_t_and_reply(bson_error_t* error, con
 #endif
 
 		zend_throw_exception(php_phongo_commandexception_ce, error->message, error->code TSRMLS_CC);
-		php_phongo_bson_to_zval(bson_get_data(reply), reply->len, &zv);
-
+		if (php_phongo_bson_to_zval(bson_get_data(reply), reply->len, &zv)) {
 #if PHP_VERSION_ID >= 70000
-		phongo_add_exception_prop(ZEND_STRL("resultDocument"), &zv);
+			phongo_add_exception_prop(ZEND_STRL("resultDocument"), &zv);
 #else
-		phongo_add_exception_prop(ZEND_STRL("resultDocument"), zv TSRMLS_CC);
+			phongo_add_exception_prop(ZEND_STRL("resultDocument"), zv TSRMLS_CC);
 #endif
+		}
+
 		zval_ptr_dtor(&zv);
 	} else {
 		zend_throw_exception(phongo_exception_from_mongoc_domain(error->domain, error->code), error->message, error->code TSRMLS_CC);
@@ -1186,7 +1187,10 @@ void php_phongo_server_to_zval(zval* retval, mongoc_server_description_t* sd) /*
 
 		PHONGO_BSON_INIT_DEBUG_STATE(state);
 		bson_iter_document(&iter, &len, &bytes);
-		php_phongo_bson_to_zval_ex(bytes, len, &state);
+		if (!php_phongo_bson_to_zval_ex(bytes, len, &state)) {
+			zval_ptr_dtor(&state.zchild);
+			return;
+		}
 
 #if PHP_VERSION_ID >= 70000
 		ADD_ASSOC_ZVAL_EX(retval, "tags", &state.zchild);
@@ -1199,7 +1203,11 @@ void php_phongo_server_to_zval(zval* retval, mongoc_server_description_t* sd) /*
 		php_phongo_bson_state state;
 
 		PHONGO_BSON_INIT_DEBUG_STATE(state);
-		php_phongo_bson_to_zval_ex(bson_get_data(is_master), is_master->len, &state);
+
+		if (!php_phongo_bson_to_zval_ex(bson_get_data(is_master), is_master->len, &state)) {
+			zval_ptr_dtor(&state.zchild);
+			return;
+		}
 
 #if PHP_VERSION_ID >= 70000
 		ADD_ASSOC_ZVAL_EX(retval, "last_is_master", &state.zchild);
--- a/src/MongoDB/Command.c
+++ b/src/MongoDB/Command.c
@@ -194,7 +194,11 @@ static HashTable* php_phongo_command_get_debug_info(zval* object, int* is_temp T
 		zval* zv;
 #endif
 
-		php_phongo_bson_to_zval(bson_get_data(intern->bson), intern->bson->len, &zv);
+		if (!php_phongo_bson_to_zval(bson_get_data(intern->bson), intern->bson->len, &zv)) {
+			zval_ptr_dtor(&zv);
+			goto done;
+		}
+
 #if PHP_VERSION_ID >= 70000
 		ADD_ASSOC_ZVAL_EX(&retval, "command", &zv);
 #else
@@ -204,6 +208,7 @@ static HashTable* php_phongo_command_get_debug_info(zval* object, int* is_temp T
 		ADD_ASSOC_NULL_EX(&retval, "command");
 	}
 
+done:
 	return Z_ARRVAL(retval);
 
 } /* }}} */
--- a/src/MongoDB/ReadPreference.c
+++ b/src/MongoDB/ReadPreference.c
@@ -362,7 +362,12 @@ static PHP_METHOD(ReadPreference, getTagSets)
 		php_phongo_bson_state state;
 
 		PHONGO_BSON_INIT_DEBUG_STATE(state);
-		php_phongo_bson_to_zval_ex(bson_get_data(tags), tags->len, &state);
+
+		if (!php_phongo_bson_to_zval_ex(bson_get_data(tags), tags->len, &state)) {
+			zval_ptr_dtor(&state.zchild);
+			return;
+		}
+
 #if PHP_VERSION_ID >= 70000
 		RETURN_ZVAL(&state.zchild, 0, 1);
 #else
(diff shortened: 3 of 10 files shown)


src/MongoDB/ReadPreference.c
```diff
@@ -429,14 +433,29 @@ static HashTable* php_phongo_readpreference_get_properties_hash(zval* object, bo
 		php_phongo_bson_state state = PHONGO_BSON_STATE_INITIALIZER;
 		state.map.root_type         = PHONGO_TYPEMAP_NATIVE_ARRAY;
 
-		php_phongo_bson_to_zval_ex(bson_get_data(tags), tags->len, &state);
+		if (php_phongo_bson_to_zval_ex(bson_get_data(tags), tags->len, &state)) {
+#if PHP_VERSION_ID >= 70000
+			Z_ADDREF(state.zchild);
```
jmikola: I'm kind of curious as to whether there is a reason we `Z_ADDREF` here and always `zval_ptr_dtor` after updating the hash table (for either PHP version). This seems a bit different than earlier functions I reviewed in the PR.

AFAIK, `Z_ADDREF` and `zval_ptr_dtor` simply cancel out. Instead, I would think we'd simply update the hash table if BSON conversion succeeds, and call `zval_ptr_dtor` for the `else` (i.e. failure) case. Do you think it's worth changing this, or perhaps putting off the possible refactoring to another ticket? If you'd like to test this for leaks, you can run the test suite (or a portion) with `TEST_PHP_ARGS=-m` defined to enable Valgrind.

src/MongoDB/Session.c
```diff
@@ -558,13 +558,16 @@ static HashTable* php_phongo_session_get_debug_info(zval* object, int* is_temp T
...
 #if PHP_VERSION_ID >= 70000
-		ADD_ASSOC_ZVAL_EX(&retval, "logicalSessionId", &state.zchild);
+			ADD_ASSOC_ZVAL_EX(&retval, "logicalSessionId", &state.zchild);
 #else
-		ADD_ASSOC_ZVAL_EX(&retval, "logicalSessionId", state.zchild);
+			ADD_ASSOC_ZVAL_EX(&retval, "logicalSessionId", state.zchild);
 #endif
+		} else {
+			zval_ptr_dtor(&state.zchild);
+			ADD_ASSOC_NULL_EX(&retval, "logicalSessionId");
```
jmikola: Remove this line.

src/MongoDB/Session.c
```diff
@@ -580,13 +583,16 @@ static HashTable* php_phongo_session_get_debug_info(zval* object, int* is_temp T
...
 #if PHP_VERSION_ID >= 70000
-			ADD_ASSOC_ZVAL_EX(&retval, "clusterTime", &state.zchild);
+				ADD_ASSOC_ZVAL_EX(&retval, "clusterTime", &state.zchild);
 #else
-			ADD_ASSOC_ZVAL_EX(&retval, "clusterTime", state.zchild);
+				ADD_ASSOC_ZVAL_EX(&retval, "clusterTime", state.zchild);
 #endif
+			} else {
+				zval_ptr_dtor(&state.zchild);
+				ADD_ASSOC_NULL_EX(&retval, "clusterTime");
```
jmikola: Remove this line.

src/MongoDB/WriteResult.c
```diff
@@ -427,13 +428,19 @@ static HashTable* php_phongo_writeresult_get_debug_info(zval* object, int* is_te
...
-		ADD_ASSOC_ZVAL_EX(&retval, "upsertedIds", state.zchild);
+			ADD_ASSOC_ZVAL_EX(&retval, "upsertedIds", state.zchild);
 #endif
-	} else {
+		} else {
+			zval_ptr_dtor(&state.zchild);
+		}
+	}
+
+	if (!hasUpsertedIds) {
```
jmikola: If I understand correctly, this just avoids populating the "upsertedIds" array if we know an exception was thrown, correct? Was this only introduced because we refactored to get rid of the large `else` block?

No problem if that's the case (the less `else` blocks the better); I just want to make sure.

alcaeus: The refactoring was necessary to add the empty value to the array, which is obviously useless in the case of an exception (as you've noted before in this review). I've changed it back for now to avoid refactoring this without a functional need, but we may revisit this later.

php_phongo.c
```diff
@@ -244,7 +244,9 @@ void phongo_throw_exception_from_bson_error_t_and_reply(bson_error_t* error, con
 #endif
 
 		zend_throw_exception(php_phongo_commandexception_ce, error->message, error->code TSRMLS_CC);
-		php_phongo_bson_to_zval(bson_get_data(reply), reply->len, &zv);
+		if (!php_phongo_bson_to_zval(bson_get_data(reply), reply->len, &zv)) {
+			ZVAL_UNDEF(&zv);
```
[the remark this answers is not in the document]
alcaeus: Ah, I haven't thought of that. I've opted to only call `phongo_add_exception_prop` only on a successful call to not attach anything to the exception thrown if BSON parsing failed. 👍

php_phongo.c
```diff
@@ -1205,7 +1207,11 @@ void php_phongo_server_to_zval(zval* retval, mongoc_server_description_t* sd) /*
 		state.map.document_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
 
 		bson_iter_document(&iter, &len, &bytes);
-		php_phongo_bson_to_zval_ex(bytes, len, &state);
+		if (!php_phongo_bson_to_zval_ex(bytes, len, &state)) {
+			zval_ptr_dtor(&state.zchild);
+			php_phongo_bson_typemap_dtor(&state.map);
```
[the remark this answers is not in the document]
alcaeus: Ah, I see. I've noticed this being used in a few places but wasn't sure what it did. The one occurrence where I didn't add it was an oversight on my part.

Given that these calls are entirely useless unless we use `field_paths` in the type map, I've removed these calls as you suggested 👍

src/MongoDB/ReadPreference.c
alcaeus: You are correct. All other instances where we add a zval to the response simply don't call `zval_ptr_dtor` instead of adding to the refcount. I've updated this block to be consistent with other calls.

There are some more occurrences regarding `Z_ADDREF` when preparing the scope in `MongoDB\BSON\Javascript` instances for output. Given that #1036 changes the point in time when we validate scope, I've created [PHPC-1471](https://jira.mongodb.org/browse/PHPC-1471) to take care of this later and added the respective tickets as dependencies.

php_phongo.c
```diff
@@ -1205,7 +1207,11 @@ void php_phongo_server_to_zval(zval* retval, mongoc_server_description_t* sd) /*
...
+			return;
```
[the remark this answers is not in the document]
alcaeus: I don't see much value in continuing to prepare the return value if an exception has been thrown. I've created [PHPC-1472](https://jira.mongodb.org/browse/PHPC-1472) to take care of the multiple exception case.

alcaeus:
@jmikola heads up, I've changed all instances to return early to avoid subsequent operations if we're going to throw an exception anyways. I figured this is most sensible considering we don't want multiple exceptions to be thrown as they would create confusion. As mentioned, there are a few more instances in `MongoDB\BSON\Javascript`, but I'll clean those up once this PR and #1036 are merged.

src/MongoDB/Command.c
```diff
@@ -194,7 +194,11 @@ static HashTable* php_phongo_command_get_debug_info(zval* object, int* is_temp T
 		zval* zv;
 #endif
 
-		php_phongo_bson_to_zval(bson_get_data(intern->bson), intern->bson->len, &zv);
+		if (!php_phongo_bson_to_zval(bson_get_data(intern->bson), intern->bson->len, &zv)) {
+			zval_ptr_dtor(&zv);
+			return Z_ARRVAL(retval);
```
jmikola: Thoughts on using `goto done;` and adding a `done:` label above the original `return Z_ARRVAL(retval);` line at the end of this function? It'd be closer to what we do in other places with failure/cleanup jumps.

I wouldn't suggest this for cases where we're simply returning (e.g. `Javascript::getScope()`), but I think helps avoid duplication when we're returning an expression, which could get out of sync. If so, other `get_debug_info` handlers could get the same treatment.

Happy to defer to you on this. You're also allowed to berate me for publicly advocating for a `goto` statement :)

src/MongoDB/ReadPreference.c
```diff
@@ -436,15 +441,16 @@ static HashTable* php_phongo_readpreference_get_properties_hash(zval* object, bo
 		PHONGO_BSON_INIT_STATE(state);
 		state.map.root_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
 
-		php_phongo_bson_to_zval_ex(bson_get_data(tags), tags->len, &state);
+		if (!php_phongo_bson_to_zval_ex(bson_get_data(tags), tags->len, &state)) {
+			zval_ptr_dtor(&state.zchild);
+			return props;
```
jmikola: This is analogous to `return Z_ARRVAL(retval);` in the `get_debug_info` handlers, but it doesn't stand out to me as much since we're not using a macro here.

I think one concern I had with these lines getting out of sync (and suggesting `goto`) is also that we don't really test them. Some `php_phongo_bson_to_zval_ex` calls in this PR could theoretically be tested by mocking the server to return garbage data, but I don't think we could cover everything. The original motivation of PHPC-348 was just add some defensive checks since it was obvious we weren't checking for errors (not considering how likely or unlikely those errors might be).

src/MongoDB/Command.c
alcaeus: XKCD goto
([Source: XKCD](https://xkcd.com/292))

That said, use of `goto` as a replacement for `finally` and failure returns is established in the driver, so I don't mind adding those here as well 👍

PR #1035, opened 2019-10-08
PHPC-1230: Support object cloning for BSON classes

alcaeus:
https://jira.mongodb.org/browse/PHPC-1230

In most cases, I've opted to initialise the cloned object using the `php_phongo_*_init` functions. One of the few exceptions is `MongoDB\BSON\Decimal128`, where we copy the `bson_decimal128_t` struct via `memcpy` to avoid converting back and forth.

From the ticket:
> This should be a relatively straightforward task, since only Javascript contains a non-scalar property. Even then, the internal properties of each BSON class are immutable, so a deep copy may not be required there; however, we should test Javascript scopes thoroughly.

I've opted to use `bson_new` and `bson_copy_to` for the scope. This should properly copy the scope for the cloned object so we're not running into any issues.


src/BSON/Binary.c
```diff
@@ -367,6 +367,32 @@ static phongo_create_object_retval php_phongo_binary_create_object(zend_class_en
...
+	php_phongo_binary_t* intern;
+	php_phongo_binary_t* new_intern;
+	phongo_create_object_retval new_object;
+
+	intern = Z_BINARY_OBJ_P(object);
+	new_object = php_phongo_binary_create_object(Z_OBJCE_P(object) TSRMLS_CC);
+
+#if PHP_VERSION_ID >= 70000
+	new_intern = Z_OBJ_BINARY(new_object);
+	zend_objects_clone_members(&new_intern->std, &intern->std TSRMLS_CC);
```
jmikola: Is this just relevant in the event that users have assigned their own public properties to the BSON object before attempting to clone it? I found a reference to this in [`clone_obj`](https://wiki.php.net/internals/engine/objects#clone_obj) documentation, but not much else.

src/BSON/Decimal128.c
```diff
@@ -300,6 +300,34 @@ static phongo_create_object_retval php_phongo_decimal128_create_object(zend_clas
...
+		zend_object_handle handle = Z_OBJ_HANDLE_P(object);
+
+		new_intern = (php_phongo_decimal128_t*) zend_object_store_get_object_by_handle(new_object.handle TSRMLS_CC);
+		zend_objects_clone_members(&new_intern->std, new_object, &intern->std, handle TSRMLS_CC);
+	}
+#endif
+
+	// Use memcpy to copy bson value to avoid converting to string and back
+	memcpy(&new_intern->decimal, &intern->decimal, sizeof(bson_decimal128_t));
+	new_intern->initialized = true;
```
jmikola: Good catch.

src/BSON/Javascript.c
```diff
@@ -411,6 +411,34 @@ phongo_create_object_retval php_phongo_javascript_create_object(zend_class_entry
...
+		zend_object_handle handle = Z_OBJ_HANDLE_P(object);
+
+		new_intern = (php_phongo_javascript_t*) zend_object_store_get_object_by_handle(new_object.handle TSRMLS_CC);
+		zend_objects_clone_members(&new_intern->std, new_object, &intern->std, handle TSRMLS_CC);
+	}
+#endif
+
+	php_phongo_javascript_init(new_intern, intern->code, intern->code_len, NULL TSRMLS_CC);
+	new_intern->scope = bson_new();
+	bson_copy_to(intern->scope, new_intern->scope);
```
jmikola: Documentation for [`bson_copy_to()`](http://mongoc.org/libbson/current/bson_copy_to.html) states that "`dst` must be an uninitialized `bson_t` to avoid leaking memory." If you run the corresponding Javascript clone test with `TEST_PHP_ARGS=-m`, does this come up as a leak?

src/BSON/ObjectId.c
```diff
@@ -336,6 +336,32 @@ static phongo_create_object_retval php_phongo_objectid_create_object(zend_class_
...
+#else
+	{
+		zend_object_handle handle = Z_OBJ_HANDLE_P(object);
+
+		new_intern = (php_phongo_objectid_t*) zend_object_store_get_object_by_handle(new_object.handle TSRMLS_CC);
+		zend_objects_clone_members(&new_intern->std, new_object, &intern->std, handle TSRMLS_CC);
+	}
+#endif
+
+	php_phongo_objectid_init_from_hex_string(new_intern, intern->oid, 24 TSRMLS_CC);
```
jmikola: Would it make sense to also use `memcpy()` here as you did for Decimal128? That would avoid the string conversion in `php_phongo_objectid_init_from_hex_string`.

src/BSON/Undefined.c
```diff
@@ -148,6 +148,7 @@ void php_phongo_undefined_init_ce(INIT_FUNC_ARGS) /* {{{ */
 	zend_class_implements(php_phongo_undefined_ce TSRMLS_CC, 1, zend_ce_serializable);
 
 	memcpy(&php_phongo_handler_undefined, phongo_get_std_object_handlers(), sizeof(zend_object_handlers));
+	php_phongo_handler_undefined.clone_obj = zend_objects_clone_obj;
```
jmikola: Although this is defined in [`std_object_handlers`](https://lxr.room11.org/xref/php-src%407.0/Zend/zend_object_handlers.c#1684), which is returned by [`zend_get_std_object_handlers`](https://lxr.room11.org/xref/php-src%407.0/Zend/zend_objects_API.c#219), I note that we _unset_ it in `phongo_get_std_object_handlers` to disable object cloning by default. That explains why this is necessary here and in the MinKey and MaxKey classes above.

tests/bson/bson-binary-clone-001.phpt
```diff
@@ -0,0 +1,100 @@
+--TEST--
+MongoDB\BSON\Binary can be cloned
```
jmikola: Since you intentionally call `zend_objects_clone_members` in all of the BSON clone handlers, I wonder if there should be an extra test case that demonstrates an object with an application-assigned public property persists across cloning. This could be done by adding an extra line between the object creation and cloning within the `foreach` loop. For example:

```php
$binary = new MongoDB\BSON\Binary('randomBinaryData', $type);
$binary->foo = 'bar'; // Ensure zend_objects_clone_members is used
$clone = clone $binary;
```

Alternatively, an extra test after the `foreach` would do this as well. This approach would probably reduce the clutter from most of the `var_dump($clone)` calls and just let you have a final dump with the additional public property.

tests/bson/bson-javascript-clone-001.phpt
```diff
@@ -0,0 +1,33 @@
+--TEST--
+MongoDB\BSON\Javascript can be cloned
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/tools.php';
+
+$javascript = new MongoDB\BSON\Javascript("function foo(bar) {var baz = bar; var bar = foo; return bar; }", ['foo' => 42]);
```
jmikola: Is there any reason we might want separate cases for scope and no-scope? I'm fine if you think it's unnecessary.

tests/bson/bson-undefined-clone-001.phpt
```diff
@@ -0,0 +1,22 @@
+--TEST--
+MongoDB\BSON\MinKey can be cloned
```
jmikola: "MongoDB\BSON\Undefined"

tests/bson/bson-utcdatetime-clone-001.phpt
```diff
@@ -0,0 +1,30 @@
+--TEST--
+MongoDB\BSON\UTCDateTime #001
+--INI--
+date.timezone=America/Los_Angeles
```
jmikola: Is this necessary?

I realize several other tests include a `date.timezone` INI setting even though they only deal with the milliseconds and never interact with DateTime conversion. It may be redundant in these files as well:

 * bson-fromPHP-002.phpt (this doesn't even use UTCDateTime at all)
 * bson-fromPHP-003.phpt
 * bson-fromPHP_error-003.phpt
 * bson-utcdatetime-001.phpt
 * bson-utcdatetime-tostring-001.phpt

src/BSON/Binary.c
alcaeus: There is some sparse documentation to be found ([link](https://wiki.php.net/internals/engine/objects)):
```
/* Copies the properties of the old_object and calls the class entry
 * clone handler. Used in the implementation of zend_objects_clone_obj
 * In PHP > 5.3, it also initializes the properties before. */
```

This may be moot since we don't store properties on the object, but I thought it'd be good practice to call this regardless. This would indeed copy user-assigned properties when cloning BSON objects, for better or worse.

src/BSON/Decimal128.c
alcaeus: Thanks. This is one of the few instances where I didn't use the object initialisers to avoid converting values back and forth, at the downside of having the initialisation logic in multiple places. I think it's fine here, but in other cases we may not want to do this.

src/BSON/Javascript.c
alcaeus: I have to set up proper non-MacOS tooling as valgrind still doesn't work with MacOS 😔

A quick look through other usages of `bson_copy_to` reveals that `bson_destroy` is called beforehand many times. I've added the same here, especially since it does nothing if the BSON is not initialised. Great catch!

src/BSON/ObjectId.c
alcaeus: Good point 👍

src/BSON/Undefined.c
alcaeus: Correct. I've added a comment explaining why we reassign the standard handler.

tests/bson/bson-javascript-clone-001.phpt
alcaeus: I don't think it is necessary, as the scope logic covers all of the no-scope logic.

tests/bson/bson-binary-clone-001.phpt
alcaeus: Good point. I've added this. Note that these properties don't show up in `var_dump` output, but can be accessed normally on the clone. This also confirms that removing `zend_objects_clone_members` indeed stops copying those properties 😉

tests/bson/bson-utcdatetime-clone-001.phpt
alcaeus: It's not necessary in this case, so I removed it. I've created [PHPC-1470](https://jira.mongodb.org/browse/PHPC-1470) for the rest of them.

src/BSON/Javascript.c
jmikola: I can test this locally and get back to you.

tests/bson/bson-binary-clone-001.phpt
jmikola: Ah, no output in `var_dump()` means less test changes. And that makes sense since our debug handler only reads known properties from the struct. Works for me.

src/BSON/ObjectId.c
```diff
@@ -336,6 +336,34 @@ static phongo_create_object_retval php_phongo_objectid_create_object(zend_class_
...
+	{
+		zend_object_handle handle = Z_OBJ_HANDLE_P(object);
+
+		new_intern = (php_phongo_objectid_t*) zend_object_store_get_object_by_handle(new_object.handle TSRMLS_CC);
+		zend_objects_clone_members(&new_intern->std, new_object, &intern->std, handle TSRMLS_CC);
+	}
+#endif
+
+	// Use memcpy to copy bson value to avoid converting to string and back
+	memcpy(&new_intern->oid, &intern->oid, 25);
```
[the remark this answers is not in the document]
alcaeus: That's neat. I've applied that suggestion and removed hardcoded values for ObjectId lengths 👍

tests/bson/bson-utcdatetime-clone-001.phpt
```diff
@@ -0,0 +1,31 @@
+--TEST--
+MongoDB\BSON\UTCDateTime #001
```
jmikola: Replace `#001` with "can be cloned"

src/BSON/Javascript.c
jmikola: Definitely a leak here:

```
128 bytes in 1 blocks are definitely lost in loss record 25 of 30
   at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0x7E9251: __zend_malloc (zend_alloc.c:2829)
   by 0xE58C2E7: php_phongo_malloc (php_phongo.c:2743)
   by 0xE5018B4: bson_malloc (bson-memory.c:70)
   by 0xE4EF2F7: bson_new (bson.c:2039)
   by 0xE5987BD: php_phongo_javascript_clone_object (Javascript.c:440)
   by 0x8D586D: ZEND_CLONE_SPEC_CV_HANDLER (zend_vm_execute.h:33778)
   by 0x911AC6: execute_ex (zend_vm_execute.h:61991)
   by 0x91402D: zend_execute (zend_vm_execute.h:63776)
   by 0x8239AE: zend_execute_scripts (zend.c:1502)
   by 0x787656: php_execute_script (main.c:2590)
   by 0x916CD9: do_cli (php_cli.c:1011)
```

> FWIW, removing the `bson_new` call above leads to a segmentation fault.

This appears to be due to the `BSON_ASSERT(dst)`. I observed the following error when running the clone test with `bson_new` removed:

```
src/libmongoc/src/libbson/src/bson/bson.c:2194 bson_copy_to(): precondition failed: dst
Aborted (core dumped)
```

It appears [`bson_copy_to()`](http://mongoc.org/libbson/current/bson_copy_to.html) expects to receive a pointer to an uninitialized `bson_t` struct. In practice, that would be the address of a stack-allocated `bson_t`, rather than something returned by `bson_new()`. Alternatively, you could `calloc` a block of memory with the size of a `bson_t`, but that's unconventional.

[`bson_copy()`](http://mongoc.org/libbson/current/bson_copy.html) seems more appropriate here:

```c
new_intern->scope = bson_copy(intern->scope);
```

alcaeus: Ah, `bson_copy` sounds like it does what I was looking for. I saw `bson_copy_to` being used in a few places and got confused 😕

PR #1036, opened 2019-10-08
PHPC-889: Validate Javascript scope on BSON decoding

alcaeus:
https://jira.mongodb.org/browse/PHPC-889


jmikola:
I think this needs to be rebased now that #1030 has been merged.

php_phongo.c
```diff
@@ -2771,6 +2771,12 @@ void php_phongo_new_javascript_from_javascript_and_scope(int init, zval* object,
 	intern->code     = estrndup(code, code_len);
 	intern->code_len = code_len;
 	intern->scope    = scope ? bson_copy(scope) : NULL;
+
+	if (scope) {
+		php_phongo_bson_state state = PHONGO_BSON_STATE_INITIALIZER;
+		php_phongo_bson_to_zval_ex(bson_get_data(intern->scope), intern->scope->len, &state);
+		zval_ptr_dtor(&state.zchild);
```
jmikola: Once complication of using `php_phongo_bson_to_zval_ex` here is that the only exception seen by the user will refer to the field path within the Javascript's scope document. The outer-most call to `php_phongo_bson_to_zval_ex` will never get a chance to report its "Detected corrupt BSON data for field path '%s'" exception since `EG(exception)` will already have been populated. Perhaps we should change the behavior to _always_ throw a final exception with the field path and instead instruct users to look to the previous exception for more details. This would be similar to what is done in `phongo_execute_bulk_write`. Would you like to create a separate ticket for that?

It's been a long while since I thought about [PHPC-889](https://jira.mongodb.org/browse/PHPC-889), but I wonder if it'd be more efficient to use `bson_validate_with_error` (with all flags enabled, as inserts might use) and, on error, use its populated `bson_error_t` message to throw a UnexpectedValueException. I suppose one issue with that is the libbson validation errors don't report the full field path as we do in `php_phongo_bson_to_zval_ex`. That may be reason enough to keep this as-is, as we can at least guarantee that the error messages are consistent with `Javascript::getScope()` and related debug handlers.

----

An entirely separate suggestion is that `php_phongo_bson_visit_codewscope` always ends with `return false`, which means that BSON iteration is allowed to continue. The only `return true` you'll find in visitor functions is within the document and array visitors, as they give up if their recursive iteration fails due to some error thrown by libbson. Our other visitors for "scalar" BSON types never had a failure case, but we're introducing one here.

I propose that we remove the `void` return type for `php_phongo_new_javascript_from_javascript_and_scope` and instead return a success boolean. `php_phongo_bson_visit_codewscope` can then check this and conditionally `return true` if an error occurred to signal that _its_ caller should abort iteration.

`php_phongo_new_javascript_from_javascript` and `php_phongo_bson_new_javascript_from_javascript` don't technically need these changes, since they never deal with a scope. I'd be fine with having `php_phongo_bson_new_javascript_from_javascript` remaining `void` and ignoring the return value of `php_phongo_new_javascript_from_javascript_and_scope`. You can leave a comment mentioning that `php_phongo_new_javascript_from_javascript_and_scope` never fails since we aren't passing it a scope to decode.

alcaeus: I agree with your suggestion. Stopping BSON parsing makes sense as we don't know what else may be broken after reading the broken scope, so treating this like an error in a document or error seems sensible.

I've adapted both BSON visitors to abort on an error. While it's true that `php_phongo_bson_new_javascript_from_javascript` currently won't fail because it doesn't pass a scope to `php_phongo_bson_new_javascript_from_javascript_and_scope`, this assumes that there will never be a failure reason in `php_phongo_bson_new_javascript_from_javascript_and_scope` not related to the scope. To avoid having to change calls to `php_phongo_bson_new_javascript_from_javascript` when changing `php_phongo_bson_new_javascript_from_javascript_and_scope`, I decided to also return a bool in `php_phongo_bson_new_javascript_from_javascript` and handle it accordingly.

src/bson.c
```diff
@@ -605,21 +605,39 @@ static bool php_phongo_bson_visit_symbol(const bson_iter_t* iter, const char* ke
...
 	php_phongo_javascript_t* intern;
 
+	if (scope) {
+		php_phongo_bson_state state;
+
+		PHONGO_BSON_INIT_STATE(state);
+
+		if (!php_phongo_bson_to_zval_ex(bson_get_data(scope), scope->len, &state)) {
+			zval_ptr_dtor(&state.zchild);
+			php_phongo_bson_state_dtor(&state);
```
jmikola: Is this redundant? I think it relates to https://github.com/mongodb/mongo-php-driver/pull/1034#discussion_r338642376.

src/bson.c
```diff
@@ -615,11 +615,24 @@ static void php_phongo_bson_new_javascript_from_javascript_and_scope(zval* objec
...
+
+	if (scope) {
+		php_phongo_bson_state state = PHONGO_BSON_STATE_INITIALIZER;
+
+		if (!php_phongo_bson_to_zval_ex(bson_get_data(intern->scope), intern->scope->len, &state)) {
+			zval_ptr_dtor(&state.zchild);
+			return false;
+		}
+
+		zval_ptr_dtor(&state.zchild);
```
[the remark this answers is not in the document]
jmikola: What about:

```c
if (scope) {
    php_phongo_bson_state state;

    PHONGO_BSON_INIT_STATE(state);

    valid_scope = php_phongo_bson_to_zval_ex(bson_get_data(scope), scope->len, &state);
    zval_ptr_dtor(&state.zchild);

    if (!valid_scope) {
        return false;
    }
}
```

PR #1040, opened 2019-10-10
PHPC-1430: Deprecate obsolete driver options

alcaeus:
https://jira.mongodb.org/browse/PHPC-1430

Contrary to what's in the ticket, this PR adds a deprecation notice for the `context` driver option as well, as per the documentation ([source](https://www.php.net/manual/en/mongodb-driver-manager.construct.php)):
> This option [context] is supported for backwards compatibility, but should be considered deprecated.

We can then formally deprecate this in 1.7.0.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -2149,6 +2149,8 @@ static mongoc_ssl_opt_t* php_phongo_make_ssl_opt(mongoc_uri_t* uri, zval* zoptio
 	} else if (php_array_existsc(zoptions, "capath")) {
 		PHONGO_SSL_OPTION_SWAP_STRING(ssl_opt->ca_dir, "capath");
 		any_ssl_option_set = true;
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"capath\" context driver option is deprecated. Please use the \"ca_dir\" driver option instead.");
 	}
 
 	if (php_array_existsc(zoptions, "crl_file")) {
@@ -2216,6 +2218,8 @@ static bool php_phongo_apply_driver_options_to_uri(mongoc_uri_t* uri, zval* zopt
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"allow_invalid_hostname\" driver option is deprecated. Please use the \"tlsAllowInvalidHostnames\" URI option instead.");
 	}
 
 	if (php_array_existsc(zoptions, "weak_cert_validation")) {
@@ -2224,12 +2228,16 @@ static bool php_phongo_apply_driver_options_to_uri(mongoc_uri_t* uri, zval* zopt
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"weak_cert_validation\" driver option is deprecated. Please use the \"tlsAllowInvalidCertificates\" URI option instead.");
 	} else if (php_array_existsc(zoptions, "allow_self_signed")) {
 		if (!mongoc_uri_set_option_as_bool(uri, MONGOC_URI_TLSALLOWINVALIDCERTIFICATES, php_array_fetchc_bool(zoptions, "allow_self_signed"))) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Failed to parse \"%s\" driver option", "allow_self_signed");
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"allow_self_signed\" context driver option is deprecated. Please use the \"tlsAllowInvalidCertificates\" URI option instead.");
 	}
 
 	if (php_array_existsc(zoptions, "pem_file")) {
@@ -2238,12 +2246,16 @@ static bool php_phongo_apply_driver_options_to_uri(mongoc_uri_t* uri, zval* zopt
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"pem_file\" driver option is deprecated. Please use the \"tlsCertificateKeyFile\" URI option instead.");
 	} else if (php_array_existsc(zoptions, "local_cert")) {
 		if (!php_phongo_apply_driver_option_to_uri(uri, zoptions, "local_cert", MONGOC_URI_TLSCERTIFICATEKEYFILE)) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Failed to parse \"%s\" driver option", "local_cert");
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"local_cert\" context driver option is deprecated. Please use the \"tlsCertificateKeyFile\" URI option instead.");
 	}
 
 	if (php_array_existsc(zoptions, "pem_pwd")) {
@@ -2252,12 +2264,16 @@ static bool php_phongo_apply_driver_options_to_uri(mongoc_uri_t* uri, zval* zopt
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"pem_pwd\" driver option is deprecated. Please use the \"tlsCertificateKeyFilePassword\" URI option instead.");
 	} else if (php_array_existsc(zoptions, "passphrase")) {
 		if (!php_phongo_apply_driver_option_to_uri(uri, zoptions, "passphrase", MONGOC_URI_TLSCERTIFICATEKEYFILEPASSWORD)) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Failed to parse \"%s\" driver option", "passphrase");
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"passphrase\" context driver option is deprecated. Please use the \"tlsCertificateKeyFilePassword\" URI option instead.");
 	}
 
 	if (php_array_existsc(zoptions, "ca_file")) {
@@ -2266,12 +2282,16 @@ static bool php_phongo_apply_driver_options_to_uri(mongoc_uri_t* uri, zval* zopt
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"ca_file\" driver option is deprecated. Please use the \"tlsCAFile\" URI option instead.");
 	} else if (php_array_existsc(zoptions, "cafile")) {
 		if (!php_phongo_apply_driver_option_to_uri(uri, zoptions, "cafile", MONGOC_URI_TLSCAFILE)) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Failed to parse \"%s\" driver option", "cafile");
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"cafile\" context driver option is deprecated. Please use the \"tlsCAFile\" URI option instead.");
 	}
 
 	return true;
--- a/src/MongoDB/Manager.c
+++ b/src/MongoDB/Manager.c
@@ -79,6 +79,16 @@ static bool php_phongo_manager_merge_context_options(zval* zdriverOptions TSRMLS
 		return false;
 	}
 
+	/* When running PHP in debug mode, php_error_docref duplicates the current
+	 * scope, leading to a COW violation in zend_hash_merge and
+	 * zend_symtable_str_del (called by php_array_unsetc). This macro allows
+	 * that violation in debug mode and is a NOOP when in non-debug. */
+#if PHP_VERSION_ID >= 70200
+	HT_ALLOW_COW_VIOLATION(Z_ARRVAL_P(zdriverOptions));
+#endif
+
+	php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"context\" driver option is deprecated.");
+
 	/* Perform array union (see: add_function() in zend_operators.c) */
 #if PHP_VERSION_ID >= 70000
 	zend_hash_merge(Z_ARRVAL_P(zdriverOptions), Z_ARRVAL_P(zcontextOptions), zval_add_ref, 0);
@@ -90,6 +100,7 @@ static bool php_phongo_manager_merge_context_options(zval* zdriverOptions TSRMLS
 #endif
 
 	php_array_unsetc(zdriverOptions, "context");
+
 	return true;
 } /* }}} */
 
--- a/tests/manager/manager-ctor-ssl-deprecated-001.phpt
+++ b/tests/manager/manager-ctor-ssl-deprecated-001.phpt
@@ -1,0 +1,67 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): Test deprecated options
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_libmongoc_ssl(); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$deprecatedDriverOptions = [
+    ['allow_invalid_hostname' => true],
+    ['weak_cert_validation' => true],
+    ['allow_self_signed' => true],
+    ['pem_file' => 'foo'],
+    ['local_cert' => 'foo'],
+    ['pem_pwd' => 'foo'],
+    ['passphrase' => 'foo'],
+    ['ca_file' => 'foo'],
+    ['cafile' => 'foo'],
+    ['context' => stream_context_create(['ssl' => ['cafile' => 'foo']])],
+    ['context' => stream_context_create(['ssl' => ['capath' => 'foo']])],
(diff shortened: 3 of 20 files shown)


tests/manager/manager-ctor-ssl-deprecated-002.phpt
```diff
@@ -0,0 +1,22 @@
...
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_libmongoc_ssl(['OpenSSL', 'LibreSSL']); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+echo raises(
+    function () {
+        new MongoDB\Driver\Manager('mongodb://127.0.0.1/', [], ['capath' => 'foo']);
```
alcaeus: Had to extract the test for `capath` to a separate test file as it needs to be skipped when compiled against Secure Channel or Secure Transport.

php_phongo.c
```diff
@@ -2160,6 +2160,8 @@ static mongoc_ssl_opt_t* php_phongo_make_ssl_opt(mongoc_uri_t* uri, zval* zoptio
 	} else if (php_array_existsc(zoptions, "capath")) {
 		PHONGO_SSL_OPTION_SWAP_STRING(ssl_opt->ca_dir, "capath");
 		any_ssl_option_set = true;
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"capath\" driver option is deprecated. Please use the \"ca_dir\" driver option instead.");
```
jmikola: I think we should use a more explicit message for options that were [merged from the "context" option](https://github.com/mongodb/mongo-php-driver/blob/1.6.0/src/MongoDB/Manager.c#L47). Consider:

> The "capath" context driver option is deprecated...

php_phongo.c
```diff
@@ -2235,12 +2239,16 @@ static bool php_phongo_apply_driver_options_to_uri(mongoc_uri_t* uri, zval* zopt
...
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"weak_cert_validation\" driver option is deprecated. Please use the \"tlsAllowInvalidCertificates\" URI option instead.");
 	} else if (php_array_existsc(zoptions, "allow_self_signed")) {
 		if (!mongoc_uri_set_option_as_bool(uri, MONGOC_URI_TLSALLOWINVALIDCERTIFICATES, php_array_fetchc_bool(zoptions, "allow_self_signed"))) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Failed to parse \"%s\" driver option", "allow_self_signed");
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"allow_self_signed\" driver option is deprecated. Please use the \"tlsAllowInvalidCertificates\" URI option instead.");
```
jmikola: "context driver option"

php_phongo.c
```diff
@@ -2249,12 +2257,16 @@ static bool php_phongo_apply_driver_options_to_uri(mongoc_uri_t* uri, zval* zopt
...
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"pem_file\" driver option is deprecated. Please use the \"tlsCertificateKeyFile\" URI option instead.");
 	} else if (php_array_existsc(zoptions, "local_cert")) {
 		if (!php_phongo_apply_driver_option_to_uri(uri, zoptions, "local_cert", MONGOC_URI_TLSCERTIFICATEKEYFILE)) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Failed to parse \"%s\" driver option", "local_cert");
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"local_cert\" driver option is deprecated. Please use the \"tlsCertificateKeyFile\" URI option instead.");
```
jmikola: "context driver option"

php_phongo.c
```diff
@@ -2263,12 +2275,16 @@ static bool php_phongo_apply_driver_options_to_uri(mongoc_uri_t* uri, zval* zopt
...
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"pem_pwd\" driver option is deprecated. Please use the \"tlsCertificateKeyFilePassword\" URI option instead.");
 	} else if (php_array_existsc(zoptions, "passphrase")) {
 		if (!php_phongo_apply_driver_option_to_uri(uri, zoptions, "passphrase", MONGOC_URI_TLSCERTIFICATEKEYFILEPASSWORD)) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Failed to parse \"%s\" driver option", "passphrase");
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"passphrase\" driver option is deprecated. Please use the \"tlsCertificateKeyFilePassword\" URI option instead.");
```
jmikola: "context driver option"

php_phongo.c
```diff
@@ -2277,12 +2293,16 @@ static bool php_phongo_apply_driver_options_to_uri(mongoc_uri_t* uri, zval* zopt
...
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"ca_file\" driver option is deprecated. Please use the \"tlsCAFile\" URI option instead.");
 	} else if (php_array_existsc(zoptions, "cafile")) {
 		if (!php_phongo_apply_driver_option_to_uri(uri, zoptions, "cafile", MONGOC_URI_TLSCAFILE)) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Failed to parse \"%s\" driver option", "cafile");
 
 			return false;
 		}
+
+		php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"cafile\" driver option is deprecated. Please use the \"tlsCAFile\" URI option instead.");
```
jmikola: "context driver option"

src/MongoDB/Manager.c
```diff
@@ -79,6 +79,8 @@ static bool php_phongo_manager_merge_context_options(zval* zdriverOptions TSRMLS
 		return false;
 	}
 
+	php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"context\" driver option is deprecated. Please use the corresponding TLS URI options instead.");
```
jmikola: Note that the context's "capath" option is deprecated in favor of the "ca_dir" driver option, which doesn't have a URI option equivalent (along with "crl_file"). I think we can simply omit the second sentence here. If users have specified one or more relevant "context" options (a subset of the [full list of supported SSL context options](https://www.php.net/manual/en/context.ssl.php)), they'll get additional deprecation notices with more specific instructions.

tests/manager/manager-ctor-ssl-deprecated-001.phpt
```diff
@@ -0,0 +1,63 @@
...
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_libmongoc_ssl(); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$opts = array(
+    "ssl" => array(
+        "verify_peer" => false,
+        "verify_peer_name" => false,
```
jmikola: I don't think either of these two options are still used by the driver, per the context fallback table in the [`Manager::__construct()`](https://www.php.net/manual/en/mongodb-driver-manager.construct.php) documentation.

Rather, some of the options in `$deprecatedDriverOptions` below should be moved up to this array:

 * `capath`
 * `cafile`
 * `local_cert`
 * `passphrase`

Apologies if the description in [PHPC-1430](https://jira.mongodb.org/browse/PHPC-1430) didn't clarify that those were part of the context options. I likely glossed over that as they're merged into driver options by the time we start processing them in `php_phongo.c`.

tests/manager/manager-ctor-ssl-deprecated-001.phpt
```diff
@@ -0,0 +1,63 @@
...
+    ['allow_invalid_hostname' => true],
+    ['weak_cert_validation' => true],
+    ['allow_self_signed' => true],
+    ['pem_file' => 'foo'],
+    ['local_cert' => 'foo'],
+    ['pem_pwd' => 'foo'],
+    ['passphrase' => 'foo'],
+    ['ca_file' => 'foo'],
+    ['cafile' => 'foo'],
+    ['context' => stream_context_create($opts)],
```
jmikola: Ah, I now see how `$deprecatedDriverOptions` interacts with the `foreach` below. In that case, I think you'll need new test cases for each of the deprecated context options. In that case, the earlier `$opts` definition can probably be removed entirely as it will make more sense to define the five distinct variations inline in the calls to `stream_context_create()`.

tests/manager/manager-ctor-ssl-deprecated-002.phpt
jmikola: This should technically be:

```php
['context' => stream_context_create(['ssl' => ['capath' => 'foo']])]
```

tests/manager/manager-ctor-ssl-deprecated-002.phpt
```diff
@@ -0,0 +1,22 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): Test deprecated options
```
jmikola: I'd suggest adding "(capath context driver option)" or at least "(capath)" to the description, to differentiate it from the previous test's name.

tests/manager/manager-ctor-ssl-deprecated-002.phpt
alcaeus: I've added a second test for this. This way, we can separately test the behaviour of specifying the `capath` driver option and nesting it within a stream context. The latter causes a single deprecation message about the "context" parameter to be printed, after which the tests stop checking deprecation messages. Checking multiple messages (as is the case here) does not work.

src/MongoDB/Manager.c
```diff
@@ -90,6 +90,9 @@ static bool php_phongo_manager_merge_context_options(zval* zdriverOptions TSRMLS
 #endif
 
 	php_array_unsetc(zdriverOptions, "context");
+
+	php_error_docref(NULL TSRMLS_CC, E_DEPRECATED, "The \"context\" driver option is deprecated.");
```
alcaeus: I originally had this call before `zend_hash_merge` a few lines above. This resulted in an abort signal:
> [2019/10/21 09:24:53.187] 019+ php: /data/mci/8b09332019426e6646ef3090d716cfae/src/php-7.2.10/Zend/zend_hash.c:1162: zend_hash_str_del: Assertion `((ht)->gc.refcount == 1) || ((ht)->u.flags & (1<<6))' failed.
[2019/10/21 09:24:53.187] 020+
[2019/10/21 09:24:53.187] 021+ Termsig=6

I then moved it to after `zend_hash_merge` but before `php_array_unsetc`, with a similar abort being the result:
> [2019/10/21 09:24:53.187] 019+ php: /data/mci/8b09332019426e6646ef3090d716cfae/src/php-7.2.10/Zend/zend_hash.c:1162: zend_hash_str_del: Assertion `((ht)->gc.refcount == 1) || ((ht)->u.flags & (1<<6))' failed.
[2019/10/21 09:24:53.187] 020+
[2019/10/21 09:24:53.187] 021+ Termsig=6

Moving the call to its current location makes the tests pass even on evergreen. I have not been able to reproduce these failures locally.

tests/manager/manager-ctor-ssl-deprecated-002.phpt
jmikola: `capath` was never a [documented](https://www.php.net/manual/en/mongodb-driver-manager.construct.php) driver option on its own, so this is mainly testing the internal behavior where context options are merged up. That said, I don't mind having two tests, so long as there's one for the documented behavior we might expect some user(s) to still employ.

I've also acknowledged that `raises()` stops after the first notice, so we can't test for two deprecation messages at once.

src/MongoDB/Manager.c
jmikola: I assume we don't have any additional information available (e.g. full trace). I imagine the culprit here is `php_array_unsetc` rather than `zend_hash_merge`, as the unset macro calls `zend_hash_str_del` through [`zend_symtable_str_del`](https://lxr.room11.org/xref/php-src%407.3/Zend/zend_hash.h#518). Also consider that [`zend_hash_merge`](https://lxr.room11.org/xref/php-src%407.3/Zend/zend_hash.c#1998) seems to only do an add or update, but never a delete.

The `(1<<6)` flag here refers to [`HASH_FLAG_ALLOW_COW_VIOLATION`](https://lxr.room11.org/xref/php-src%407.3/Zend/zend_hash.h#41), which was introduced in https://github.com/php/php-src/commit/fd4025069d561b4196851c2e9a85b5bfdfe1465a. The actual assertion that fails is `zend_hash_str_del`'s call to [`HT_ASSERT_RC1`](https://lxr.room11.org/xref/php-src%407.2/Zend/zend_hash.c#1164), which adds a [`HASH_FLAG_ALLOW_COW_VIOLATION` check](https://lxr.room11.org/xref/php-src%407.2/Zend/zend_hash.c#27) only when PHP is compiled in debug mode. That might explain why you can't reproduce this locally.

Based on the commit that introduced this, it would seem that you could also work around this by calling `HT_ALLOW_COW_VIOLATION(Z_ARRVAL_P(zdriverOptions));`. https://github.com/php/php-src/commit/fd4025069d561b4196851c2e9a85b5bfdfe1465a does include a number of comments explaining why `HT_ALLOW_COW_VIOLATION()` is used in certain places, but I think in our case this could pertain to an additional reference to `zdriverOptions`'s hash table (remember [PHP 7's refcounts now live on the value, not the zval](https://nikic.github.io/2015/05/05/Internal-value-representation-in-PHP-7-part-1.html#zvals-in-php-7)) being created due to emitting a deprecation notice. Perhaps this is also pertinent to having set a user-defined error handler to convert this notice into an exception (as notices would not care about in-scope variables otherwise).

If you follow the rabbit hole through [`php_verror`](https://lxr.room11.org/xref/php-src%407.3/main/main.c#925), which itself is called from the `php_error_docref#` functions, you'll come to [`zend_error`](https://lxr.room11.org/xref/php-src%407.3/Zend/zend.c#1169). Therein, we see some code for the [case where a user-defined error handler has been set](https://lxr.room11.org/xref/php-src%407.3/Zend/zend.c#1286) where the current scope is captured (symbol table appears to get duplicated). If my hunch is correct, this means that we should expect `zdriverOptions`'s hash table RC to get incremented whenever a user-defined error handler is set, which is certainly the case for our tests but likewise also for most frameworks. Granted, PHP probably won't be compiled in debug mode for most users, but we shouldn't assume that.

*tl;dr*: Let's see if `HT_ALLOW_COW_VIOLATION()` fixes this and allows us to order the notice emission in the ideal location. Either way, we should add a comment explaining (a) why we call `HT_ALLOW_COW_VIOLATION` or (b) why we've deferred the call to `php_error_docref`.

----

Separate thought: let's say `$driverOptions` already had multiple references from the application (totally independent from our merging here). In that case, I expect we might also hit the RC assertion when trying to unset a field on it by calling `php_array_unsetc` from a bit of extension code. This is possibly something you can test in isolation, and combine with debugging the value of `Z_ARRVAL_P(zdriverOptions)->gc.refcount`.

alcaeus: Thank you for that excellent analysis. Using `HT_ALLOW_COW_VIOLATION` fixes the error. The error does not only originate in `zend_hash_str_del`, but also in `zend_hash_merge` where it was introduced in the same commit linked above (https://github.com/php/php-src/commit/fd4025069d561b4196851c2e9a85b5bfdfe1465a#diff-fd78a0a3f78ea28c6907f907f25b908eR1858).

I wasn't able to replicate the last thought. The refcount of `driverOptions` (analysed with `xdebug_debug_zval` for convenience) returned a refcount of 4 with no error being thrown (unless adding the `php_error_docref` call). Even when explicitly duplicating the `context` element to increase its refcount to 2 I didn't get the error in question. With the information you gathered, I'd indeed assume that this is caused by a user-defined error handler leading `zend_error` copying the symbol table.

As an additional note, this entirely behaviour is related to the `$errcontext` argument in error handlers that was deprecated in PHP 7.2. It has been removed from PHP 8 already: https://github.com/php/php-src/commit/2f1f34952e9a0dfb3adcbec82ba69f4ac82b3a3d#diff-b09edfedd835ebc4491e565c147190e7L1388. I currently don't have a setup to run debug tests on PHP 8 but will try to set one up soon.

src/MongoDB/Manager.c
```diff
@@ -79,6 +79,16 @@ static bool php_phongo_manager_merge_context_options(zval* zdriverOptions TSRMLS
 		return false;
 	}
 
+	/* When running PHP in debug mode, php_error_docref duplicates the current
+	 * scope, leading to a COW violation in zend_hash_merge and
+	 * zend_symtable_str_del (called by php_array_unsetc). This macro allows
+	 * that violation in debug mode and is a NOOP when in non-debug. */
+#ifdef HT_ALLOW_COW_VIOLATION
```
jmikola: `#if PHP_VERSION_ID >= 70200` may be more consistent with what we do elsewhere (e.g. UTCDateTime.c). This also makes it easier to grep when we bump minimum PHP versions and identify obsolete conditionals.

I don't feel strongly about this, so if you prefer the more tactical macro check I'm OK with leaving this as-is.

alcaeus: You make a good point, version numbers are easier to clean up later.

PR #1041, opened 2019-10-10
PHPC-1418: Test PHP 7.3 in Evergreen

alcaeus:
https://jira.mongodb.org/browse/PHPC-1418

@jmikola: looking at the [diff that introduced PHP 7.3](https://github.com/10gen/mongo-php-toolchain/pull/1/files#diff-a0e5b2ada9d7aee94a81a64d53055dde), I see that PHP 7.0 was removed. However, Evergreen build stages covering PHP 7.0 still complete successfully, so I'm not quite sure what's going on here. Do you know what's going on here, or where I could check to find out?

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -595,6 +595,10 @@ axes:
   - id: php-versions
     display_name: PHP Version
     values:
+      - id: "7.3"
+        display_name: "PHP 7.3"
+        variables:
+          PHP_VERSION: "7.3.8"
       - id: "7.2"
         display_name: "PHP 7.2"
         variables:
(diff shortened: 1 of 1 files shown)


jmikola:
> However, Evergreen build stages covering PHP 7.0 still complete successfully, so I'm not quite sure what's going on here. Do you know what's going on here, or where I could check to find out?

Not sure myself, but I'd suggest either asking @kryptonite303, following up on [BUILD-8750](https://jira.mongodb.org/browse/BUILD-8750), or reaching out in `#evergreen-users` in Slack. It's possible the Evergreen team was just going by [PHP: Supported Versions](https://www.php.net/supported-versions.php), but that doesn't explain why they'd have left 5.5 and 5.6 in place.

jmikola:
Also, note that the Evergreen builds in this PR are all using PHP 5.6 and 7.3, so it doesn't look like they'd have caught a removal of 7.0; however, 7.0 builds in https://evergreen.mongodb.com/waterfall/mongo-php-driver are still OK from what I can tell.

PR #1042, opened 2019-10-11
PHPC-1411: Accept 64-bit integers for wTimeoutMS

alcaeus:
https://jira.mongodb.org/browse/PHPC-1411


tests/manager/manager-ctor-write_concern-005.phpt
```diff
@@ -0,0 +1,74 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): write concern options (64-bit wtimeoutms)
+--SKIPIF--
+<?php if (4 !== PHP_INT_SIZE) { die('skip Only for 32-bit platform'); } ?>
+--FILE--
+<?php
+
+$tests = [
+    ['mongodb://127.0.0.1/?wtimeoutms=2147483648', []],
```
jmikola: [`wtimeout`](https://docs.mongodb.com/manual/reference/write-concern/#wtimeout) is only applicable when `w` is greater than 1. I realize this isn't checked during URI parsing, but we can probably remove this since 2147483648 (i.e. `INT32_MAX + 1`) is already tested in the next line. Likewise for the fifth entry in this array.

php_phongo.c
```diff
@@ -1377,10 +1377,7 @@ void php_phongo_write_concern_to_zval(zval* retval, const mongoc_write_concern_t
 	}
 
 	if (wtimeout != 0) {
-		/* Note: PHP currently enforces that wimeoutMS is a 32-bit integer, so
-		 * casting will never truncate the value. This may change with
-		 * PHPC-1411. */
-		ADD_ASSOC_LONG_EX(retval, "wtimeout", (int32_t) wtimeout);
+		ADD_ASSOC_LONG_EX(retval, "wtimeout", wtimeout);
```
jmikola: `php_phongo_write_concern_to_zval` only appears to be called from `php_phongo_bulkwrite_get_debug_info`. Given that, I think it'd make sense to use the same logic here as we do in [`php_phongo_cursorid_get_debug_info`](https://github.com/mongodb/mongo-php-driver/blob/1.6.0/src/MongoDB/CursorId.c#L100) and conditionally output `wtimeout` as a string if we're on a 32-bit architecture _and_ its value exceeds `INT32_MAX` (admittedly an edge case). `ADD_ASSOC_LONG_EX` works for all other cases.

Was your original thinking that a user debugging a WriteConcern object would have created it on the same machine, and in that case the value itself would naturally be within the range of 32-bit positive integers?

This may relate to my later comment on `PHP_METHOD(WriteConcern, getWtimeout)`.

src/MongoDB/WriteConcern.c
```diff
@@ -265,9 +252,7 @@ static PHP_METHOD(WriteConcern, getWtimeout)
 		return;
 	}
 
-	/* Note: PHP currently enforces that wimeoutMS is a 32-bit integer, so
-	 * casting will never truncate the value. This may change with PHPC-1411. */
-	RETURN_LONG((int32_t) mongoc_write_concern_get_wtimeout_int64(intern->write_concern));
+	RETURN_LONG(mongoc_write_concern_get_wtimeout_int64(intern->write_concern));
```
jmikola: I suppose there's a risk of truncation here if a user on a 32-bit platform somehow created a WriteConcern with a 64-bit `wtimeout` value; however, that doesn't seem possible so long as [`_mongoc_uri_build_write_concern()`](https://github.com/mongodb/mongo-c-driver/blob/1.15.1/src/libmongoc/src/mongoc/mongoc-uri.c#L1606) always reads it as a 32-bit integer (no matter the architecture).

tests/manager/manager-ctor-write_concern-005.phpt
```diff
@@ -0,0 +1,74 @@
...
+    ['mongodb://127.0.0.1/?w=2&wtimeoutms=2147483648', []],
```
jmikola: Looking at [`_mongoc_uri_build_write_concern()`](https://github.com/mongodb/mongo-c-driver/blob/1.15.1/src/libmongoc/src/mongoc/mongoc-uri.c#L1606), which I also linked above, I'm not sure how this is successfully parsed by libmongoc.

[`mongoc_uri_parse_options()`](https://github.com/mongodb/mongo-c-driver/blob/1.15.1/src/libmongoc/src/mongoc/mongoc-uri.c#L1173) splits the query string, but [`mongoc_uri_apply_options()`](https://github.com/mongodb/mongo-c-driver/blob/1.15.1/src/libmongoc/src/mongoc/mongoc-uri.c#L1011) does the real work of converting original string values into their final types. Since [`MONGOC_URI_WTIMEOUTMS`](https://github.com/mongodb/mongo-c-driver/blob/1.15.1/src/libmongoc/src/mongoc/mongoc-uri.c#L740) is considered an int32 option, I expect [`mongoc_uri_parse_int32()`](https://github.com/mongodb/mongo-c-driver/blob/1.15.1/src/libmongoc/src/mongoc/mongoc-uri.c#L800) will truncate it. Assuming this test is actually passing, I'm not sure why we'd see `int(2147483648)` in the expected output below (at least for the test cases using only the connection string).

FWIW, this probably highlights something that was overlooked in [CDRIVER-3087](https://jira.mongodb.org/browse/CDRIVER-3087). Perhaps that was intentional, as I believe Swift cared more about the WriteConcern struct than URI parsing.

tests/manager/manager-ctor-write_concern-005.phpt
```diff
@@ -0,0 +1,74 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): write concern options (64-bit wtimeoutms)
+--SKIPIF--
+<?php if (4 !== PHP_INT_SIZE) { die('skip Only for 32-bit platform'); } ?>
```
jmikola: Why would this test only run on 32-bit platforms? I don't see anything that would prohibit a 64-bit platform from passing it, given that the connection string parsing logic in libmongoc is the same. If anything, I'd think that a 32-bit platform wouldn't be able to deal with `2147483648` integer literals, as you have below for the URI option array test cases.

tests/manager/manager-ctor-write_concern-005.phpt
alcaeus: > FWIW, this probably highlights something that was overlooked in [CDRIVER-3087](https://jira.mongodb.org/browse/CDRIVER-3087).

I concur. FWIW, `int(2147483648)` is still a 32-bit value, but it assumes an unsigned int (with the usual conversion happening when using signed ints). I noticed this as well and changing to `int(4294967296)` (expectedly) causes the test to fail:
> Unsupported value for "wtimeoutms": "4294967296".

I've created [CDRIVER-3396](https://jira.mongodb.org/browse/CDRIVER-3396) to track this. Should be fixable in 1.15.2 as support for 64-bit `wtimeoutms` was only introduced in 1.15.

I'd hold off on working on this PR until we've fixed the underlying libmongoc issue.

jmikola: We just assigned [CDRIVER-3396](https://jira.mongodb.org/browse/CDRIVER-3396) to you this morning, so feel free to move on to it before this.

src/MongoDB/WriteConcern.c
alcaeus: With `_mongoc_uri_build_write_concern` now always reading a 64-bit integer, the question is how to proceed here: returning a string would fix this, but would technically constitute a BC break.

tests/manager/manager-ctor-write_concern-005.phpt
alcaeus: Test simplified to cover 32-bit and 64-bit systems using `EXPECTF`.

php_phongo.c
alcaeus: Updated this to reflect the logic in serialisation of CursorId and Int64: this will be exposed as string on 32-bit systems.

alcaeus:
Updated now that CDRIVER-3396 has been fixed. Contains the single commit from #1060 until that PR is merged.

tests/writeConcern/writeconcern-ctor-002.phpt
```diff
@@ -0,0 +1,20 @@
+--TEST--
+MongoDB\Driver\WriteConcern construction with 64-bit wtimeoutms
+--SKIPIF--
+<?php if (4 !== PHP_INT_SIZE) { die('skip Only for 32-bit platform'); } ?>
```
jmikola: Why does this test require a 32-bit platform? The range of signed 32-bit ints is -2147483648 through 2147483647, and this test is using a literal value of `INT32_MAX + 1`. If anything, I'd expect it to require a 64-bit environment.

tests/writeConcern/writeconcern-set_state-002.phpt
```diff
@@ -0,0 +1,19 @@
+--TEST--
+MongoDB\Driver\WriteConcern::__set_state() with 64-bit wtimeoutms
+--SKIPIF--
+<?php if (4 !== PHP_INT_SIZE) { die('skip Only for 32-bit platform'); } ?>
```
jmikola: This also seems like a test that should require 64-bit, rather than 32-bit.

php_phongo.c
alcaeus: My apologies. CursorId does this in `php_phongo_cursorid_get_debug_info`. Changing this to also return a LONG as long as it's within 32-bit boundaries works for me 👍

src/MongoDB/WriteConcern.c
jmikola: Added comment in [PHPC-1411](https://jira.mongodb.org/browse/PHPC-1411?focusedCommentId=2570617&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2570617) to note the documentation change outside of this PR.

tests/manager/manager-ctor-write_concern-005.phpt
```diff
@@ -0,0 +1,41 @@
...
+$tests = [
+    ['mongodb://127.0.0.1/?w=2&wtimeoutms=4294967296', []],
+    ['mongodb://127.0.0.1/?w=majority&wtimeoutms=4294967296', []],
+    ['mongodb://127.0.0.1/?w=customTagSet&wtimeoutms=4294967296', []],
+];
+
+foreach ($tests as $test) {
+    list($uri, $options) = $test;
+
+    $manager = new MongoDB\Driver\Manager($uri, $options);
```
jmikola: `$options` is never used, so I think you can just get rid of it and make `$tests` an array of URI strings.

src/MongoDB/WriteConcern.c
```diff
@@ -329,9 +314,7 @@ static HashTable* php_phongo_write_concern_get_properties_hash(zval* object, boo
 
 	wtag     = mongoc_write_concern_get_wtag(intern->write_concern);
 	w        = mongoc_write_concern_get_w(intern->write_concern);
-	/* Note: PHP currently enforces that wimeoutMS is a 32-bit integer, so
-	 * casting will never truncate the value. This may change with PHPC-1411. */
-	wtimeout = (int32_t) mongoc_write_concern_get_wtimeout_int64(intern->write_concern);
+	wtimeout = mongoc_write_concern_get_wtimeout_int64(intern->write_concern);
```
jmikola: Sorry for not catching this earlier, but is there a concern with silent truncation affecting BSON serialization via `bsonSerialize()`, which uses this `get_properties_hash` method? If so, we may want to consider emitting an Int64 object for an out-of-range wtimeout value, on par with emitting a string in `php_phongo_write_concern_to_zval`. AFAICT, the `to_zval` function is only used for BulkWrite's debug info.

Also, `PHP_METHOD(WriteConcern, serialize)` still includes the code being removed here:

```c
/* Note: PHP currently enforces that wimeoutMS is a 32-bit integer, so
 * casting will never truncate the value. This may change with PHPC-1411. */
wtimeout = (int32_t) mongoc_write_concern_get_wtimeout_int64(intern->write_concern);
```

I did some digging and realized that most of our value objects (BSON types and the RC/RP/WC classes) use their `init_from_hash` methods for both `_set_state` _and_ `unserialize` methods; however, the `serialize` methods _do not_ utilize the `get_properties_hash` function. `get_properties_hash` tends to only be used for the `get_debug_info` and `get_properties` handlers. This means that there's some subtle code duplication between `get_properties_hash` and `serialize`. I wonder if it's worth trying to refactor this in some future ticket.

In the meantime, I think we should ensure that WriteConcern's `serialize` method remains consistent with `get_properties_hash`. I wasn't too worried about `getWtimeout()` truncating, since the alternative is a BC break, but I think we can ensure compatibility in our debugging and serialization output (both PHP _and_ `bsonSerialize()`) by conditionally emitting an `Int64` object. Thoughts?

I'm happy to punt on this for the time being and create a follow-up ticket. It's something we could certainly address in a later minor version, since this is admittedly an edge case. I believe the only users affected by this would be those that manage to create a 64-bit wtimeout on a 32-bit system and want to have that WriteConcern serialized properly in a command document by way of `bsonSerialize()`. I'm not sure about how a 32-bit PHP instance would handle encountering a 64-bit integer in a serialized string -- presumably there'd be some error/truncation well before we even call `php_phongo_writeconcern_init_from_hash` from `PHP_METHOD(WriteConcern, unserialize)`.

alcaeus: That's a very good point.

I've ensured that we don't truncate in any place other than `getWtimeout`, where we have to do so to maintain backward compatibility. `getWtimeout` will now raise an `E_WARNING` when a value outside of the 32-bit range is encountered on a 32-bit system (see separate commit).

The logic in `serialize` and `get_properties_hash` have been adjusted: the former will dump a 64-bit value as string when encountering a 64-bit value, while the latter will create a `MongoDB\BSON\Int64` object and return that instead.

The logic in `unserialize` needs adjusting: the value is currently required to be a `long`, which we'll have to adapt to the changes in `serialize`. I just wanted to double-check that this is indeed feasible before going down that route.

alcaeus:
@jmikola: updated the logic again after our discussion yesterday.

I've added a few new symbols:
* The `php_phongo_int64_to_zval` function converts an `int64_t` to a `zval`, either as string or as a `MongoDB\BSON\Int64` instance
* The `php_phongo_parse_int64` parses a string as 64-bit int; this was mainly done to avoid duplicating that logic even more
* The `ADD_ASSOC_INT64_COMPAT` macro wraps a `php_phongo_int64_to_zval` and `ADD_ASSOC_ZVAL` call to avoid adding more compiler conditionals for PHP 5.

The `wTimeout` of a `WriteConcern` object is now always serialised as string and can also be unserialised as such. This makes sure that the timeout is truly portable between 32-bit and 64-bit architectures. `bsonSerialize` continues to return an `Int64` object for the timeout.

src/BSON/UTCDateTime.c
```diff
@@ -509,32 +491,28 @@ static HashTable* php_phongo_utcdatetime_get_properties_hash(zval* object, bool
 {
 	php_phongo_utcdatetime_t* intern;
 	HashTable*                props;
-	char                      s_milliseconds[24];
-	int                       s_milliseconds_len;
 
 	intern = Z_UTCDATETIME_OBJ_P(object);
 
-	PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_debug, intern, props, 2);
+	PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_debug, intern, props, 1);
```
jmikola: Is this just cleaning up an extra allocation, since we know there's only one property to report?

phongo_compat.h
```diff
@@ -91,6 +91,12 @@
 #define ADD_ASSOC_ZVAL(_zv, _key, _value) add_assoc_zval(_zv, _key, _value);
 #define ADD_ASSOC_NULL_EX(_zv, _key) add_assoc_null_ex(_zv, ZEND_STRL(_key));
 #define ADD_ASSOC_BOOL_EX(_zv, _key, _value) add_assoc_bool_ex(_zv, ZEND_STRL(_key), _value);
+#define ADD_ASSOC_INT64_COMPAT(_zv, _key, _value)                    \
```
jmikola: Since this forces the `int64_t` to a string (`is_bson=false`), I'd suggest we name this `ADD_ASSOC_INT64_AS_STRING`.

src/MongoDB/WriteConcern.c
```diff
@@ -358,7 +368,7 @@ static HashTable* php_phongo_write_concern_get_properties_hash(zval* object, boo
 		if (wtimeout != 0) {
 			zval z_wtimeout;
 
-			ZVAL_LONG(&z_wtimeout, wtimeout);
+			php_phongo_int64_to_zval(wtimeout, &z_wtimeout, is_bson TSRMLS_CC);
```
jmikola: Does it make sense to always call `php_phongo_int64_to_zval` here? `php_phongo_write_concern_get_properties_hash` is called from three places:

 * `get_debug_info` handler
 * `get_properties` handler
 * `bsonSerialize()` method

We know that `bsonSerialize()` needs to conditionally use an Int64 object if the value exceeds the range of a 32-bit integer. If `wtimeout` is within the range of a 32-bit integer, we can add it as an integer primitive.

The debug and properties handlers need not ensure compatibility between 32-bit and 64-bit architectures (unlike `serialize()`, so I think they'd do well to also encode as an integer whenever possible and only fall back to a string on a 32-bit architecture when `wtimeout` exceeds the range of a 32-bit integer.

src/BSON/UTCDateTime.c
alcaeus: That’s correct.

jmikola:
Note: I will follow up with a PR against your branch when I get a chance.

src/MongoDB/WriteConcern.c
alcaeus: The idea was to always export as a string, since code might be serialised using `var_export` on a 64-bit architecture and then read on a 32-bit architecture. In this case, we'd export as an int, but then be unable to read it. That was the main reason for always treating it as if we were on a 32-bit platform.

jmikola: Ah, I wasn't too concerned about `var_export` being used across platforms since it actually produces code that the user would probably manually edit (unless they're using `eval()`). I was thinking of `serialize()` as the more likely way folks might transport between platforms.

In my experience (from GitHub issues), most people using `var_export` are doing so as an alternative to `var_dump` and didn't actually rely on `__set_state()` to rebuild an object.

I'm happy to have `var_export` produce a string as needed, though.

jmikola:
@alcaeus Opened https://github.com/alcaeus/mongo-php-driver/pull/2 with proposed changes. This allowed me to revert most of your test changes here, but I also added some new tests.

alcaeus:
Thanks Jeremy! I've added your changes to this PR, so I think this is good for another review 🎉

tests/writeConcern/writeconcern-set_state-001.phpt
```diff
@@ -18,6 +16,8 @@ $tests = [
     [ 'w' => 1, 'j' => true, 'wtimeout' => 1000 ],
     [ 'j' => true ],
     [ 'wtimeout' => 1000 ],
+    // wtimeout accepts 64-bit integers as strings
+    [ 'wtimeout' => '2147483648']
```
jmikola: @alcaeus can you add a trailing comma here before merging?

PR #1043, opened 2019-10-11
PHPC-1347: Do not allow empty string for replicaSet

alcaeus:
https://jira.mongodb.org/browse/PHPC-1347

Since libmongoc rejects an URI with an empty `replicaSet` URI option, there is different behaviour depending on whether an empty replica set name is passed via the URI string or via the URI options array. The first causes an exception:
> Failed to parse MongoDB URI: 'mongodb://localhost:3000/?replicaSet='. Value for URI option "replicaset" cannot be empty string.

The latter lets users create the manager, but causes a connection exception on server selection:
> Fatal error: Uncaught MongoDB\Driver\Exception\ConnectionTimeoutException: No suitable servers found (`serverSelectionTryOnce` set)

This PR changes this and checks the `replicaSet` option when it is passed via URI options array, also throwing an exception when the option is explicitly given with an empty value. Not passing the option at all does not cause an error.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1583,6 +1583,11 @@ static bool php_phongo_apply_options_to_uri(mongoc_uri_t* uri, bson_t* options T
 				return false;
 			}
 
+			if (!strcasecmp(key, MONGOC_URI_REPLICASET) && !strcmp("", bson_iter_utf8(&iter, NULL))) {
+				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Value for URI option \"%s\" cannot be empty string.", key);
+				return false;
+			}
+
 			if (!mongoc_uri_set_option_as_utf8(uri, key, bson_iter_utf8(&iter, NULL))) {
 				/* Assignment uses mongoc_uri_set_appname() for the "appname"
 				 * option, which validates length in addition to UTF-8 encoding.
--- a/tests/manager/manager-ctor_error-004.phpt
+++ b/tests/manager/manager-ctor_error-004.phpt
@@ -1,0 +1,25 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): empty replicaSet argument
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+echo throws(function () {
+    $manager = new MongoDB\Driver\Manager('mongodb://localhost:27017/?replicaSet=');
+    $manager->executeQuery(NS, new MongoDB\Driver\Query([]));
+}, MongoDB\Driver\Exception\InvalidArgumentException::class), "\n";
+
+echo throws(function () {
+    $manager = new MongoDB\Driver\Manager('mongodb://localhost:27017', ['replicaSet' => '']);
+    $manager->executeQuery(NS, new MongoDB\Driver\Query([]));
+}, MongoDB\Driver\Exception\InvalidArgumentException::class), "\n";
+
+?>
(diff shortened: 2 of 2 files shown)


php_phongo.c
```diff
@@ -1546,6 +1546,19 @@ static bool php_phongo_apply_options_to_uri(mongoc_uri_t* uri, bson_t* options T
...
+			if (!BSON_ITER_HOLDS_UTF8(&iter)) {
+				PHONGO_URI_INVALID_TYPE(iter, "string");
+				return false;
+			}
+
+			if (!strcmp("", bson_iter_utf8(&iter, NULL))) {
+				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Value for URI option \"%s\" cannot be empty string.", key);
+				return false;
+			}
+		}
```
jmikola: I think this would be better placed within the `if (mongoc_uri_option_is_utf8(key)) {` condition further down in `php_phongo_apply_options_to_uri`. That already does the `BSON_ITER_HOLDS_UTF8` check, which you're duplicating here. Consider:

```c
if (mongoc_uri_option_is_utf8(key)) {
    if (!BSON_ITER_HOLDS_UTF8(&iter)) {
        PHONGO_URI_INVALID_TYPE(iter, "string");
        return false;
    }

    if (!strcasecmp(key, MONGOC_URI_REPLICASET) && !strcmp("", bson_iter_utf8(&iter, NULL))) {
        phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Value for URI option \"%s\" cannot be empty string.", key);
        return false;
    }

    if (!mongoc_uri_set_option_as_utf8(uri, key, bson_iter_utf8(&iter, NULL))) {
        /* Assignment uses mongoc_uri_set_appname() for the "appname"...
```

tests/connect/replicaset-no-replicaset-name-001.phpt
```diff
@@ -0,0 +1,42 @@
+--TEST--
+MongoDB\Driver\Manager: Connecting to Replica Set with an empty replicaSet argument
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_replica_set(); ?>
```
jmikola: Both errors (libmongoc URI parsing and our new exception) can be exercised without incurring I/O, so I think this test could be better rewritten like the existing `Manager::__construct()` error tests (e.g. [manager-ctor_error-003.phpt](https://github.com/mongodb/mongo-php-driver/blob/master/tests/manager/manager-ctor_error-003.phpt)). Note that the 003 test already handles the invalid type check, which you had duplicated above, so there's be no need to check that again in your new test.

With the SKIPIF removed, I think this would be as simple as two cases in their own `throws()` callable:

 * Invalid connection string with `?replicaSet=` throws libmongoc's exception
 * Valid connection string with `['replicaSet' => '']` URI options throws our exception

tests/connect/replicaset-no-replicaset-name-001.phpt
```diff
@@ -0,0 +1,25 @@
+--TEST--
+MongoDB\Driver\Manager: Connecting to Replica Set with an empty replicaSet argument
```
jmikola: Can we rename this file to `manager-ctor_error-003.phpt`? I realize we have a lot of existing tests split across various sub-directories, but the general convention is `class-method-###.phpt`, with an optional `_error` for error tests.

PR #1044, opened 2019-10-11
PHPC-1391: Prohibit starting a transaction for maxWireVersion < 7

alcaeus:
https://jira.mongodb.org/browse/PHPC-1391

The exeption itself is caused by libmongoc not allowing transactions. This commit only adds tests to protect against regressions.

### DIFF
--- a/tests/session/session-startTransaction_error-006.phpt
+++ b/tests/session/session-startTransaction_error-006.phpt
@@ -1,0 +1,26 @@
+--TEST--
+MongoDB\Driver\Session::startTransaction() throws an error on replicasets < 4.0
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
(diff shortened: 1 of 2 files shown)


tests/session/session-startTransaction_error-006.phpt
```diff
@@ -0,0 +1,26 @@
...
+$session = $manager->startSession();
+
+echo throws(function () use ($session) {
+    $session->startTransaction();
+}, MongoDB\Driver\Exception\RuntimeException::class), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
```
jmikola: If you're not using actual [EXPECTF](https://qa.php.net/expectf_details.php) patterns and merely doing an exact string match, you can use `EXPECT` here. Likewise in the second test below.

PR #1046, opened 2019-10-21
Test against multiple libmongoc versions on evergreen

alcaeus:
https://jira.mongodb.org/browse/PHPC-1002

This PR adds tests against the current stable branch (r.1.15) and upcoming minor version (master, 1.16-dev) to evergreen. This helps us ensure that builds work against the upcoming libmongoc version. For a patch build, see https://evergreen.mongodb.com/version/5dada5af1e2d17070614e379.


.evergreen/config.yml
```diff
@@ -37,6 +37,17 @@ functions:
         working_dir: "src"
         script: |
            git submodule update --init
+    # Switch to a different version of libmongoc
+    - command: shell.exec
+      params:
+        working_dir: "src/src/libmongoc"
+        script: |
+           if [ "$LIBMONGOC_VERSION" != "" ]; then
```
jmikola: Is this more portable than `[ -z "$LIBMONGOC_VERSION" ]`?

.evergreen/config.yml
```diff
@@ -668,6 +679,18 @@ axes:
...
 
+  - id: libmongoc-version
+    display_name: libmongoc version
+    values:
+      - id: "1.15-latest"
+        display_name: "1.15 latest"
+        variables:
+          LIBMONGOC_VERSION: "r1.15"
+      - id: "master"
+        display_name: "Upcoming release (1.16)"
```
jmikola: Would it make sense to leave this as "(master)" so this never has to be updated? I assume after a libmongoc release is made, we'll need to create a new entry in the `libmongoc-version` axis either way so it's a question of whether we also want to update this parenthetical while doing so.

.evergreen/config.yml
```diff
@@ -772,3 +795,9 @@ buildvariants:
   display_name: "${versions}/${php-versions}/${os-php7} — ${storage-engine}"
   tasks:
      - name: "test-standalone"
+
+- matrix_name: "libmongoc-versions-php7"
+  matrix_spec: {"os-php7": "debian92-test", "versions": "4.2", "php-versions": "7.2", "libmongoc-version": "*"}
+  display_name: "${versions}/${php-versions}/${os-php7} — ${libmongoc-version}"
```
jmikola: Does `${libmongoc-version}` just pull in the `id` from the `libmongoc-version` axis element? If so, that may only print "1.15-latest" or "master". Would it make sense to prefix that with "libmongoc: " for display purposes?

This is less of an issue with other display names (e.g. storage engines), so those IDs are pretty self-explanatory.

CONTRIBUTING.md
```diff
@@ -229,6 +229,16 @@ if $PKG_CONFIG libmongoc-1.0 --atleast-version 1.15.0; then
...
 ```
 
+### Update tested versions in evergreen configuration
+
+Evergreen tests against multiple versions of libmongoc. When updating to a newer
+libmongoc version, make sure to update the `libmongoc-version` build axis in
+`.evergreen/config.yml`. In general, we test against two additional versions of
+libmongoc:
+- The upcoming patch release of the current libmongoc minor version (e.g. the
+  `r1.x` branch)
```
jmikola: I suppose this is an argument for not simply _adding_ r1.16 to the axis, but rather replacing the previous entry for r1.15.

.evergreen/config.yml
alcaeus: It's a habit of mine, but `[ -z ... ]` works just fine and achieves the same, so I'll change it.

.evergreen/config.yml
alcaeus: We need to update the pipeline when we switch to a new version of libmongoc, but this can become outdated even when we don't make a change (e.g. once 1.16 is released, as master will be 1.17-dev then). `(master)` will work better.

.evergreen/config.yml
alcaeus: If I'm not mistaken, this pulls the display name from the axis. Prefixing with "libmongoc" makes sense.

.evergreen/config.yml
alcaeus: Correction, `-n` is what we're looking for.

CONTRIBUTING.md
alcaeus: Yes and no. Let's say 1.16 is released but we don't require it yet (for whatever reason). In this case, it would make sense adding r1.16 to the axis instead of replacing r1.15. Once we require 1.16 or newer, we can drop r1.15 from the axis.

alcaeus:
@jmikola: made the changes you suggested. I've also added a replicaset and sharded build for more resiliency. New patch build: https://evergreen.mongodb.com/version/5dafee1c1e2d170cf567f51c.

PR #1048, opened 2019-10-28
PHPC-348: Free cursor item on invalid BSON

alcaeus:
https://jira.mongodb.org/browse/PHPC-348

Follow-up to #1034, this frees cursor items when BSON conversion fails.

### DIFF
--- a/src/MongoDB/Cursor.c
+++ b/src/MongoDB/Cursor.c
@@ -123,7 +123,11 @@ static void php_phongo_cursor_iterator_move_forward(zend_object_iterator* iter T
 	}
 
 	if (mongoc_cursor_next(cursor->cursor, &doc)) {
-		php_phongo_bson_to_zval_ex(bson_get_data(doc), doc->len, &cursor->visitor_data);
+		if (!php_phongo_bson_to_zval_ex(bson_get_data(doc), doc->len, &cursor->visitor_data)) {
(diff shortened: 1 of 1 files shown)


src/MongoDB/Cursor.c
```diff
@@ -123,7 +123,11 @@ static void php_phongo_cursor_iterator_move_forward(zend_object_iterator* iter T
 	}
 
 	if (mongoc_cursor_next(cursor->cursor, &doc)) {
-		php_phongo_bson_to_zval_ex(bson_get_data(doc), doc->len, &cursor->visitor_data);
+		if (!php_phongo_bson_to_zval_ex(bson_get_data(doc), doc->len, &cursor->visitor_data)) {
+			/* Free invalid result, but don't return as we want to free the
+			 * session if the cursor is exhausted. */
+			zval_ptr_dtor(&cursor->visitor_data.zchild);
```
alcaeus: @jmikola: I've opted against adding a label and `goto`, but can add it if you think it's worth future-proofing this.

jmikola: Any reason we can't do `php_phongo_cursor_free_current` here? I think it's more sensible to ensure the zval is freed _and_ reset to undefined. Unlike the other cases you addressed in #1034 (re: return values), the cursor's current element might still be accessed after this exception is handled.

PR #1052, opened 2019-11-05
PHPC-1472: Prevent multiple exceptions from being thrown while dumping manager

alcaeus:
https://jira.mongodb.org/browse/PHPC-1472

@jmikola: I'm not sure what a good way to test this is, as it can only come up when receiving invalid data from the server. Do you have an idea how to verify this indeed works as expected?

### DIFF
--- a/php_phongo.h
+++ b/php_phongo.h
@@ -155,7 +155,7 @@ void  php_phongo_prep_legacy_option_free(zval* options TSRMLS_DC);
 void php_phongo_read_preference_prep_tagsets(zval* tagSets TSRMLS_DC);
 bool php_phongo_read_preference_tags_are_valid(const bson_t* tags);
 
-void php_phongo_server_to_zval(zval* retval, mongoc_server_description_t* sd);
+bool php_phongo_server_to_zval(zval* retval, mongoc_server_description_t* sd);
 void php_phongo_read_concern_to_zval(zval* retval, const mongoc_read_concern_t* read_concern);
 void php_phongo_write_concern_to_zval(zval* retval, const mongoc_write_concern_t* write_concern);
 void php_phongo_cursor_to_zval(zval* retval, const mongoc_cursor_t* cursor);
--- a/src/MongoDB/Manager.c
+++ b/src/MongoDB/Manager.c
@@ -893,7 +893,13 @@ static HashTable* php_phongo_manager_get_debug_info(zval* object, int* is_temp T
 	for (i = 0; i < n; i++) {
 		zval obj;
 
-		php_phongo_server_to_zval(&obj, sds[i]);
+		if (!php_phongo_server_to_zval(&obj, sds[i])) {
+			/* Exception already thrown */
+			zval_ptr_dtor(&obj);
+			zval_ptr_dtor(&cluster);
+			goto done;
+		}
+
 		add_next_index_zval(&cluster, &obj);
 	}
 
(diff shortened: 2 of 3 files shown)


src/MongoDB/Manager.c
```diff
@@ -916,6 +924,41 @@ static HashTable* php_phongo_manager_get_debug_info(zval* object, int* is_temp T
...
+failure:
+#if PHP_VERSION_ID >= 70000
+	{
+		zend_string* string_key = NULL;
+		zend_ulong   num_key    = 0;
+		zval*        server;
+
+		ZEND_HASH_FOREACH_KEY_VAL_IND(Z_ARRVAL(cluster), num_key, string_key, server)
+		{
+			zval_ptr_dtor(server);
```
jmikola: Could we get away with just calling `zval_ptr_dtor(&retval)` instead? I realize that the `ZVAL_STATIC_INIT` initializer sets everything to zero (and wouldn't identify as refcounted on its own), but [`array_init_size()`](http://lxr.jpauli.tech/source/xref/php-7.3/Zend/zend_API.h#378) will set the refcounted flag (follow `ZVAL_ARR` through to `IS_ARRAY_EX`](http://lxr.jpauli.tech/source/xref/php-7.3/Zend/zend_types.h#IS_ARRAY_EX)).

If you follow the chain from [`i_zval_ptr_dtor()`](http://lxr.jpauli.tech/source/xref/php-7.3/Zend/zend_variables.h#39) through `rc_dtor_func` (assuming we decrement the refcount to zero), you'll end up in [`zend_array_destroy()`](http://lxr.jpauli.tech/source/xref/php-7.3/Zend/zend_hash.c#1464), which will recursively call `i_zval_ptr_dtor` on each element within the array's hash table.

src/MongoDB/Manager.c
```diff
@@ -916,6 +924,41 @@ static HashTable* php_phongo_manager_get_debug_info(zval* object, int* is_temp T
 	mongoc_server_descriptions_destroy_all(sds, n);
 
 	return Z_ARRVAL(retval);
+
+failure:
```
jmikola: I think we still want to do `mongoc_server_descriptions_destroy_all(sds, n)` here, lest those leak.

src/MongoDB/Manager.c
alcaeus: Ah, good call. I tried to follow through but couldn't quite see where we end up when destroying a HashTable, so I decided to ensure we're clearing everything. I've removed the loops in favour of only calling `zval_ptr_dtor`.

src/MongoDB/Manager.c
```diff
@@ -893,7 +893,11 @@ static HashTable* php_phongo_manager_get_debug_info(zval* object, int* is_temp T
 	for (i = 0; i < n; i++) {
 		zval obj;
 
-		php_phongo_server_to_zval(&obj, sds[i]);
+		if (!php_phongo_server_to_zval(&obj, sds[i])) {
+			/* Exception already thrown */
+			goto failure;
```
jmikola: Outstanding leaks are due to fields on `php_phongo_server_to_zval`'s retval not being freed. I think we can sort that out with `zval_ptr_dtor(&obj);` before the `goto` here. Likewise for the PHP 5.6 branch.

php_phongo.h
```diff
@@ -155,7 +155,7 @@ void  php_phongo_prep_legacy_option_free(zval* options TSRMLS_DC);
 void php_phongo_read_preference_prep_tagsets(zval* tagSets TSRMLS_DC);
 bool php_phongo_read_preference_tags_are_valid(const bson_t* tags);
 
-void php_phongo_server_to_zval(zval* retval, mongoc_server_description_t* sd);
+bool php_phongo_server_to_zval(zval* retval, mongoc_server_description_t* sd);
```
jmikola: Make a point to consider other places that call `php_phongo_server_to_zval`. Those likely need logic to handle the exception. The server's `get_debug_info` handler is at least one instance.

alcaeus: As per our analysis yesterday, the only other place this is called is in [`php_phongo_server_get_debug_info`](https://github.com/mongodb/mongo-php-driver/blob/fef008ed20ee55b87e1d55686c132d03b4950942/src/MongoDB/Server.c#L600), where the resulting zval is returned. The server description is properly freed, so there's no memory leaking there.

PR #1053, opened 2019-11-06
PHPC-1459: Add serialization support for CursorId

alcaeus:
https://jira.mongodb.org/browse/PHPC-1459

Note that serialising an object will always have the identifier as string for portability between 32-bit and 64-bit platforms. This is consistent with the serialisation handling for `MongoDB\BSON\Int64`.

### DIFF
--- a/src/MongoDB/CursorId.c
+++ b/src/MongoDB/CursorId.c
@@ -16,6 +16,12 @@
 
 #include <php.h>
 #include <Zend/zend_interfaces.h>
+#include <ext/standard/php_var.h>
+#if PHP_VERSION_ID >= 70000
+#include <zend_smart_str.h>
+#else
+#include <ext/standard/php_smart_str.h>
+#endif
 
 #ifdef HAVE_CONFIG_H
 #include "config.h"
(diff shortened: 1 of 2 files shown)


tests/cursorid/cursorid-serialization-002.phpt
```diff
@@ -0,0 +1,28 @@
+--TEST--
+MongoDB\Driver\CursorID serialization on 32-bit systems
+--SKIPIF--
+<?php if (4 !== PHP_INT_SIZE) { die('skip Only for 32-bit platform'); } ?>
```
alcaeus: This needs two separate tests for 32-bit and 64-bit systems due to the different serialisation formats. I thought this was better than writing an `EXPECTF` section to account for both int and string outputs of the 64-bit value.

jmikola: Is it the serialisation formats that differ, or just the debug output?

If it's just the debug output, we do have prior art for using an EXPECTF in [tests/functional/cursorid-001.phpt](https://github.com/mongodb/mongo-php-driver/blob/master/tests/functional/cursorid-001.phpt#L35) that you could copy in.

alcaeus: Ah, yeah, that was what I wanted to avoid, but if we already use it then we should also use that in the serialisation tests.

src/MongoDB/CursorId.c
```diff
@@ -26,6 +32,48 @@
...
+	/* bson_ascii_strtoll() sets errno if conversion fails. If conversion
+	 * succeeds, we still want to ensure that the entire string was parsed. */
+	id = bson_ascii_strtoll(s_id, &endptr, 10);
+
+	if (errno || (endptr && endptr != ((const char*) s_id + s_id_len))) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Error parsing \"%s\" as 64-bit id for %s initialization", s_id, ZSTR_VAL(php_phongo_cursorid_ce->name));
+		return false;
+	}
+
+	intern->id = id;
```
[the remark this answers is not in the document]
alcaeus: See https://jira.mongodb.org/browse/PHPC-1490.

PR #1054, opened 2019-11-07
PHPC-1446: Implicitly enable TLS when any TLS-related driver option is set

alcaeus:
https://jira.mongodb.org/browse/PHPC-1446

The test assumes that the server isn't configured for SSL, which works in our test environment but may not always work. We also need to skip if the URI explicitly enables SSL as we can then no longer test that it's implicitly enabled.

### DIFF
--- a/tests/manager/manager-ctor-ssl-003.phpt
+++ b/tests/manager/manager-ctor-ssl-003.phpt
@@ -1,0 +1,28 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): Specifying a driver option implicitly enables TLS
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_libmongoc_ssl(['OpenSSL', 'LibreSSL']); ?>
+<?php skip_if_ssl(); ?>
+--FILE--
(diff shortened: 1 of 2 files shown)


tests/manager/manager-ctor-ssl-003.phpt
```diff
@@ -0,0 +1,23 @@
...
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_libmongoc_ssl(['OpenSSL', 'LibreSSL']); ?>
+<?php skip_if_ssl(); ?>
+--FILE--
+<?php
+
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = new MongoDB\Driver\Manager(URI, [], ['ca_dir' => 'foo']);
```
jmikola: I think this test should also assert that we _can_ connect and ping the URI as-is. Otherwise, this test may pass by virtue of the server being down.

You can also add a note that this test would fail if the the server was configured with either `allowSSL` or `preferSSL` for [`net.ssl.mode`](https://docs.mongodb.com/manual/reference/configuration-options/#net.ssl.mode).

PR #1027, opened 2019-10-02
[WAIT] PHPC-1274: Add function to reset manager instance

alcaeus:
https://jira.mongodb.org/browse/PHPC-1274

PR #941, opened 2018-11-29
[WAIT] PHPC-1274: Reset libmongoc client after forking

jmikola:
https://jira.mongodb.org/browse/PHPC-1274

This is _not_ suitable for merging. Just opening so we have a POC to look at.

PR #1059, opened 2019-11-12
PHPC-1274: Do not allow forked children to affect parent resources

jmikola:
https://jira.mongodb.org/browse/PHPC-1274

This PR will supersede both https://github.com/mongodb/mongo-php-driver/pull/941 and https://github.com/mongodb/mongo-php-driver/pull/1027

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -384,6 +384,7 @@ void phongo_session_init(zval* return_value, mongoc_client_session_t* client_ses
 
 	session                 = Z_SESSION_OBJ_P(return_value);
 	session->client_session = client_session;
+	session->client         = mongoc_client_session_get_client(client_session);
 }
 /* }}} */
 
@@ -2623,8 +2624,8 @@ static void php_phongo_persist_client(const char* hash, size_t hash_len, mongoc_
 {
 	php_phongo_pclient_t* pclient = (php_phongo_pclient_t*) pecalloc(1, sizeof(php_phongo_pclient_t), 1);
 
-	pclient->pid    = (int) getpid();
-	pclient->client = client;
+	pclient->created_by_pid = (int) getpid();
+	pclient->client         = client;
 
 #if PHP_VERSION_ID >= 70000
 	zend_hash_str_update_ptr(&MONGODB_G(pclients), hash, hash_len, pclient);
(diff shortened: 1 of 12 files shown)


alcaeus:
@jmikola #1060 has been merged, so you can rebase this PR on top of master once you continue working on it.

php_phongo.c
```diff
@@ -380,6 +380,7 @@ void phongo_session_init(zval* return_value, mongoc_client_session_t* client_ses
 
 	session                 = Z_SESSION_OBJ_P(return_value);
 	session->client_session = client_session;
+	session->client         = mongoc_client_session_get_client(client_session);
```
jmikola: A `client` field is needed on the cursor struct since libmongoc's public API provides no method to access a cursor's underlying client. We obviously don't need that here, but creating the field allows the macros below to universally operate on a Manager, Cursor, or Session object.

If you don't see a problem with this, I'll refactor other calls to `mongoc_client_session_get_client()` in `Session.c` to simply refer to this new field in a separate commit.

tests/session/bug1274-002.phpt
```diff
@@ -0,0 +1,102 @@
+--TEST--
+PHPC-1274: Session destruct should not end session from parent process
```
jmikola: This is currently not functioning as a regression test, as I'm having trouble getting the child process to actually kill the session when it exits if I disable the call to `mongoc_client_reset()` in the Session's `free_object` handler.

I think this may be because the session pool is really only destroyed during the call to `mongoc_client_destroy()`, and PHPC purposefully avoids calling that due to https://github.com/mongodb/mongo-php-driver/commit/600db56bd880e09502eb2e996cfd02c9aadc1ad3 ([PHPC-912](https://jira.mongodb.org/browse/PHPC-912)). If that's the case, perhaps we can't implement a regression test for this at all and will have to make do with the "does not abort" test.

jmikola: I'd like to leave this test in place despite my last comment, as it could become more relevant when we get to [PHPC-1522](https://jira.mongodb.org/browse/PHPC-1522). And even if we can't implement that (i.e. allow children to call `mongoc_client_destroy()` on clients created by parents), this may still be useful as our only test to demonstrate that child processes don't kill active sessions of a parent process.

php_phongo.c
alcaeus: > I'll refactor other calls to `mongoc_client_session_get_client()` in `Session.c`

That sounds like a good plan 👍

alcaeus:
Reset logic looks sound to me. LGTM once `mongoc_client_session_get_client` calls have been refactored.

php_phongo.c
jmikola: This is now done.

PR #1060, opened 2019-11-13
Update to libmongoc-1.16.0-dev

alcaeus:
Extracted from #1059, as this is also necessary for #1042 as well as [PHPC-1438](https://jira.mongodb.org/browse/PHPC-1438).


src/LIBMONGOC_VERSION_CURRENT
```diff
@@ -1 +1 @@
-1.16.0-20191029+gitce14000d77
+1.11.1-20191113+gitbef4cf349d
```
jmikola: This is incorrect. My original commit in #1059 updated this to `1.16.0-20191112+git622b3c8966`.

alcaeus: I updated it further to also add in support for 64-bit wTimeoutMs and exposing session state.

jmikola: Is this a problem with running [`make libmongoc-version-current`](https://github.com/mongodb/mongo-php-driver/blob/master/CONTRIBUTING.md#ensure-libmongoc-version-information-is-correct) on your machine? I remember the script had trouble inferring the correct minor version a while back, but I thought that was fixed.

jmikola: See: https://jira.mongodb.org/browse/CDRIVER-3315

Perhaps you didn't actually have the `master` branch checked out?

alcaeus: Both the submodule and the version number were updated to the last commit in libmongoc master: https://github.com/mongodb/mongo-c-driver/commit/bef4cf349d93a44be9755c4ca4921bf6f6d03f53

jmikola: Sorry for the misunderstanding. `gitbef4cf349d` is correct in this case, but `11.1` is not.

Can you attempt running `make libmongoc-version-current` once more when `src/libmongoc` is pointing to the upstream `master` branch? I think the behavior you're seeing (with `11.1`) is the result of checking out a specific commit but not `master` directly. I wouldn't expect those to be different, but perhaps they are from the script's POV (detached head?).

alcaeus: D'oh, sorry for being so daft. I have fixed this. My apologies for not seeing this earlier.

PR #1061, opened 2019-11-13
PHPC-1438: Expose session state

alcaeus:
https://jira.mongodb.org/browse/PHPC-1438

Contains the libmongoc update from #1060.

I've decided against exposing the `MONGOC_TRANSACTION_` constants directly since we're trying to get rid of that for read preference. I've also opted against introducing constants for those strings for now, but we may consider doing that to allow users to use constants instead of having to write out the strings.

### DIFF
--- a/src/MongoDB/Session.c
+++ b/src/MongoDB/Session.c
@@ -29,6 +29,12 @@
 
 zend_class_entry* php_phongo_session_ce;
 
+#define PHONGO_TRANSACTION_NONE "none"
+#define PHONGO_TRANSACTION_STARTING "starting"
+#define PHONGO_TRANSACTION_IN_PROGRESS "in_progress"
+#define PHONGO_TRANSACTION_COMMITTED "committed"
+#define PHONGO_TRANSACTION_ABORTED "aborted"
+
 #define SESSION_CHECK_LIVELINESS(i, m)                                  \
 	if (!(i)->client_session) {                                         \
 		phongo_throw_exception(                                         \
@@ -91,6 +97,25 @@ cleanup:
 	}
 
 	return retval;
+}
+
+static const char* php_phongo_get_transaction_state_string(mongoc_transaction_state_t state TSRMLS_DC)
+{
+	switch (state) {
+		case MONGOC_TRANSACTION_NONE:
+			return PHONGO_TRANSACTION_NONE;
+		case MONGOC_TRANSACTION_STARTING:
+			return PHONGO_TRANSACTION_STARTING;
+		case MONGOC_TRANSACTION_IN_PROGRESS:
+			return PHONGO_TRANSACTION_IN_PROGRESS;
+		case MONGOC_TRANSACTION_COMMITTED:
+			return PHONGO_TRANSACTION_COMMITTED;
+		case MONGOC_TRANSACTION_ABORTED:
+			return PHONGO_TRANSACTION_ABORTED;
+		default:
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Invalid transaction state %d given", (int) state);
+			return NULL;
+	}
 }
 
 /* {{{ proto void MongoDB\Driver\Session::advanceClusterTime(array|object $clusterTime)
@@ -343,6 +368,30 @@ static PHP_METHOD(Session, getTransactionOptions)
 	}
 } /* }}} */
 
+/* {{{ proto string MongoDB\Driver\Session::getTransactionState()
+   Returns the current transaction state for this session */
+static PHP_METHOD(Session, getTransactionState)
+{
+	php_phongo_session_t* intern;
+	const char*           state;
+
+	intern = Z_SESSION_OBJ_P(getThis());
+	SESSION_CHECK_LIVELINESS(intern, "getTransactionState")
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	state = php_phongo_get_transaction_state_string(mongoc_client_session_get_transaction_state(intern->client_session) TSRMLS_CC);
+	if (!state) {
+		/* Exception already thrown */
+		return;
+	}
+
+	PHONGO_RETURN_STRING(state);
+} /* }}} */
+
+
 /* Creates a opts structure from an array optionally containing an RP, RC,
  * WC object, and/or maxCommitTimeMS int. Returns NULL if no options were found,
  * or there was an invalid option. If there was an invalid option or structure,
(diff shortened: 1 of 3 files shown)


src/MongoDB/Session.c
```diff
@@ -93,6 +93,25 @@ static bool php_phongo_session_get_timestamp_parts(zval* obj, uint32_t* timestam
...
+		case MONGOC_TRANSACTION_STARTING:
+			return "starting";
+		case MONGOC_TRANSACTION_IN_PROGRESS:
+			return "in_progress";
+		case MONGOC_TRANSACTION_COMMITTED:
+			return "committed";
+		case MONGOC_TRANSACTION_ABORTED:
+			return "aborted";
+		default:
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Invalid transaction state %d given", (int) state);
```
alcaeus: Considering that libmongoc exposes transaction states as defined in the spec, I've decided to throw an exception in this case. While this requires us to pay close attention to any new transaction states, I'm not sure what other alternative would be feasible (besides returning the integer value)

src/LIBMONGOC_VERSION_CURRENT
```diff
@@ -1 +1 @@
-1.16.0-20191029+gitce14000d77
+1.11.1-20191113+gitbef4cf349d
```
jmikola: See https://github.com/mongodb/mongo-php-driver/pull/1060#discussion_r345918860.

src/MongoDB/Session.c
```diff
@@ -93,6 +93,25 @@ static bool php_phongo_session_get_timestamp_parts(zval* obj, uint32_t* timestam
 	return retval;
 }
 
+static const char* php_phongo_get_transaction_state_string(mongoc_transaction_state_t state TSRMLS_DC)
+{
+	switch (state) {
+		case MONGOC_TRANSACTION_NONE:
+			return "none";
```
jmikola: > I've decided against exposing the MONGOC_TRANSACTION_ constants directly since we're trying to get rid of that for read preference.

Totally understandable.

> I've also opted against introducing constants for those strings for now, but we may consider doing that to allow users to use constants instead of having to write out the strings.

I do think there's value in introducing constants for the strings we return here. The only problem with ReadPreference was exposing the raw numeric values (technically bit flags), which have no meaning outside of libmongoc. I have no qualms about having constants for meaningful string values, as they are useful to avoid having applications hard-code magic string literals.

As for naming, we have a few options:

 * `TRANSACTION_STATE_NONE`
 * `TRANSACTION_NONE`
 * `TXN_STATE_NONE`
 * `TXN_NONE`

I can go either way with including "state". I can't imagine what else we might need Session constants for, so I'm not too worried about naming conflicts. "TXN" is a common abbreviation for "Transaction", so maybe that's a no-brainer -- and "TXN_STATE_" is still shorter than "TRANSACTION_".

src/MongoDB/Session.c
jmikola: I think an exception is fine here. Based on how [CDRIVER-3364](https://jira.mongodb.org/browse/CDRIVER-3364) was implemented, we know that libmongoc would abort before we ever reach this point. Also, any new transaction state would likely entail a spec change, which would trickle down as CDRIVER and PHPC tickets, so I'm not worried about missing such a change.

src/MongoDB/Session.c
```diff
@@ -343,6 +362,26 @@ static PHP_METHOD(Session, getTransactionOptions)
...
+	intern = Z_SESSION_OBJ_P(getThis());
+	SESSION_CHECK_LIVELINESS(intern, "getTransactionState")
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	state = mongoc_client_session_get_transaction_state(intern->client_session);
+
+   PHONGO_RETURN_STRING(php_phongo_get_transaction_state_string(state TSRMLS_CC));
```
jmikola: Formatting.

tests/session/session-getTransactionState-001.phpt
```diff
@@ -0,0 +1,67 @@
...
+
+/* Empty transaction, and committed empty transaction */
+$session->startTransaction();
+var_dump($session->getTransactionState());
+$session->commitTransaction();
+var_dump($session->getTransactionState());
+
+/* Aborted transaction with one operation */
+$session->startTransaction();
+$bw = new \MongoDB\Driver\BulkWrite();
```
jmikola: Does it make sense to `var_dump()` here as well, so we get "starting" in the test output? That makes it easier to read the expected output, as would extra line breaks between the various test cases here.

An alternative is to replace the comments with `echo "\nTest case:\n";` statements, which we've done before in some tests that produce a bunch of output that wouldn't make much sense on its own.

src/LIBMONGOC_VERSION_CURRENT
alcaeus: Will rebase once #1060 is merged, holding off on merging before then.

src/MongoDB/Session.c
alcaeus: As for including "state", I normally would opt for including it to avoid future naming conflicts. However, `TXN_STARTING` (or `TRANSACTION_STARTING`) reads better than the other options, so I'd omit "state" for now.

As for `TXN` versus `TRANSACTION`, I normally opt against abbreviations but I'll defer to your decision here. I can live with either.

jmikola: If we're skipping `STATE`, let's use `TRANSACTION`. `TRANSACTION_STATE_` was really the only one that gave me pause due to its length.

alcaeus:
`TRANSACTION_` constants added. I've also added internal `PHONGO_TRANSACTION_*` macros to avoid using raw strings to denote transaction states.

src/MongoDB/Session.c
```diff
@@ -760,6 +806,12 @@ void php_phongo_session_init_ce(INIT_FUNC_ARGS) /* {{{ */
 	php_phongo_handler_session.free_obj = php_phongo_session_free_object;
 	php_phongo_handler_session.offset   = XtOffsetOf(php_phongo_session_t, std);
 #endif
+
+	zend_declare_class_constant_string(php_phongo_session_ce, ZEND_STRL("TRANSACTION_NONE"), PHONGO_TRANSACTION_NONE TSRMLS_CC);
+	zend_declare_class_constant_string(php_phongo_session_ce, ZEND_STRL("TRANSACTION_STARTING"), PHONGO_TRANSACTION_STARTING TSRMLS_CC);
+	zend_declare_class_constant_string(php_phongo_session_ce, ZEND_STRL("TRANSACTION_IN_PROGRESS"), PHONGO_TRANSACTION_IN_PROGRESS TSRMLS_CC);
+	zend_declare_class_constant_string(php_phongo_session_ce, ZEND_STRL("TRANSACTION_COMMITTED"), PHONGO_TRANSACTION_COMMITTED TSRMLS_CC);
+	zend_declare_class_constant_string(php_phongo_session_ce, ZEND_STRL("TRANSACTION_ABORTED"), PHONGO_TRANSACTION_ABORTED TSRMLS_CC);
```
jmikola: I suggest adding one more test for these constants, similar to https://github.com/mongodb/mongo-php-driver/blob/master/tests/readConcern/readconcern-constants.phpt.

src/MongoDB/Session.c
```diff
@@ -343,6 +368,26 @@ static PHP_METHOD(Session, getTransactionOptions)
...
+	intern = Z_SESSION_OBJ_P(getThis());
+	SESSION_CHECK_LIVELINESS(intern, "getTransactionState")
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	state = mongoc_client_session_get_transaction_state(intern->client_session);
+
+	PHONGO_RETURN_STRING(php_phongo_get_transaction_state_string(state TSRMLS_CC));
```
jmikola: Similar to my concerns in https://github.com/mongodb/mongo-php-driver/pull/1062#discussion_r353810821, I think we should only call `PHONGO_RETURN_STRING` if `php_phongo_get_transaction_state_string` returns a non-null value; otherwise we risk passing `null` to `strlen()`. Consider:

```c
const char *state;

// ...

// Or keep state and add a state_str var if this is too long for your liking
state = php_phongo_get_transaction_state_string(mongoc_client_session_get_transaction_state(intern->client_session) TSRMLS_CC);

if (!state) {
    /* Exception already thrown */
    return;
}

PHONGO_RETURN_STRING(state);
```

We won't be able to write a test for this (you can do so manually if you like), but it provides some peace of mind.

alcaeus: Logic added. The joy of working with pointers :)

PR #1062, opened 2019-11-15
PHPC-1491: Add ReadPreference::getModeString

alcaeus:
https://jira.mongodb.org/browse/PHPC-1491

### DIFF
--- a/src/MongoDB/ReadPreference.c
+++ b/src/MongoDB/ReadPreference.c
@@ -34,6 +34,12 @@
 
 zend_class_entry* php_phongo_readpreference_ce;
 
+#define PHONGO_READ_PRIMARY "primary"
+#define PHONGO_READ_PRIMARY_PREFERRED "primaryPreferred"
+#define PHONGO_READ_SECONDARY "secondary"
+#define PHONGO_READ_SECONDARY_PREFERRED "secondaryPreferred"
+#define PHONGO_READ_NEAREST "nearest"
+
 /* Initialize the object from a HashTable and return whether it was successful.
  * An exception will be thrown on error. */
 static bool php_phongo_readpreference_init_from_hash(php_phongo_readpreference_t* intern, HashTable* props TSRMLS_DC) /* {{{ */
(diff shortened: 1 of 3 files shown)


src/MongoDB/ReadPreference.c
```diff
@@ -733,6 +755,12 @@ void php_phongo_readpreference_init_ce(INIT_FUNC_ARGS) /* {{{ */
 	zend_declare_class_constant_long(php_phongo_readpreference_ce, ZEND_STRL("RP_NEAREST"), MONGOC_READ_NEAREST TSRMLS_CC);
 	zend_declare_class_constant_long(php_phongo_readpreference_ce, ZEND_STRL("NO_MAX_STALENESS"), MONGOC_NO_MAX_STALENESS TSRMLS_CC);
 	zend_declare_class_constant_long(php_phongo_readpreference_ce, ZEND_STRL("SMALLEST_MAX_STALENESS_SECONDS"), MONGOC_SMALLEST_MAX_STALENESS_SECONDS TSRMLS_CC);
+
+	zend_declare_class_constant_string(php_phongo_readpreference_ce, ZEND_STRL("PRIMARY"), PHONGO_READ_PRIMARY TSRMLS_CC);
```
alcaeus: As in #1061, I'm exposing constants here. Since the class name is `ReadPreference`, I decided to avoid using any prefix. This also resolves a naming conflict with the `RP_` constants that expose the integer values for read preference modes.

I'm not too happy about having two very similar constants (`RP_PRIMARY` and `PRIMARY`) in the same class, as this may confuse users. We don't need to expose these values as constants now and may opt to only do that in 2.0 when the `RP_` constants are dropped.

src/MongoDB/ReadPreference.c
```diff
@@ -343,6 +369,21 @@ static PHP_METHOD(ReadPreference, getMode)
...
+{
+	php_phongo_readpreference_t* intern;
+
+	intern = Z_READPREFERENCE_OBJ_P(getThis());
+
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	PHONGO_RETURN_STRING(php_phongo_readpreference_get_mode_string(mongoc_read_prefs_get_mode(intern->read_preference)));
```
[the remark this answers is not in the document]
jmikola: I would be more concerned about passing `NULL` to `strlen()` in the `ZVAL_STRING` macro:

 * http://lxr.jpauli.tech/source/xref/php-5.6/Zend/zend_API.h#ZVAL_STRING
 * http://lxr.jpauli.tech/source/xref/php-7.4/Zend/zend_API.h#ZVAL_STRING

I believe `strlen(NULL)` is undefined in most implementations:

 * https://stackoverflow.com/q/5796103/162228
 * https://en.cppreference.com/w/c/string/byte/strlen

In practice, I don't think it's possible for the read preference to have a different mode unless it was added to the enum and the PHP driver was not updated to handle it. That said, a defensive solution may be to simply throw an exception here if `php_phongo_readpreference_get_mode_string` returns `NULL` and include the numeric code in the message.

src/MongoDB/ReadPreference.c
jmikola: I'm OK with adding the constants now, as it's consistent with what we're already doing in [ReadConcern](https://www.php.net/manual/en/class.mongodb-driver-readconcern.php). If we don't already have tickets to deprecate and remove the `RP_` constants, feel free to create those.

src/MongoDB/ReadPreference.c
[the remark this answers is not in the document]
jmikola: I was just reminded of https://github.com/mongodb/mongo-php-driver/pull/1061/files#r345720246. For consistency, perhaps you'd rather throw from `php_phongo_readpreference_get_mode_string` instead of `getModeString()`?

jmikola:
I'd also suggest one additional test for the new constants, similar to I suggest adding one more test for these constants, similar to https://github.com/mongodb/mongo-php-driver/blob/master/tests/readConcern/readconcern-constants.phpt. Feel free to add the numeric constants if you like, but I'm not too concerned about those.

src/MongoDB/ReadPreference.c
alcaeus: Deprecation is tracked in [PHPC-1489](https://jira.mongodb.org/browse/PHPC-1489). I've added tests for all constants, including int ones for good measure.

alcaeus:
Oh, my apologies, I totally missed that message. `php_phongo_readpreference_get_mode_string` now throws in the `default` path.

PR #1063, opened 2019-11-19
PHPC-1471: Clean up reference handling in Javascript scope serialisation

alcaeus:
https://jira.mongodb.org/browse/PHPC-1471

This was brought up in the review for #1034. Calling `Z_ADDREF` is unnecessary if we only call `zval_ptr_dtor` when the result actually needs to be destroyed.

### DIFF
--- a/src/BSON/Javascript.c
+++ b/src/BSON/Javascript.c
@@ -215,17 +215,16 @@ static PHP_METHOD(Javascript, jsonSerialize)
 		php_phongo_bson_state state;
 
 		PHONGO_BSON_INIT_STATE(state);
-		if (php_phongo_bson_to_zval_ex(bson_get_data(intern->scope), intern->scope->len, &state)) {
-#if PHP_VERSION_ID >= 70000
-			Z_ADDREF(state.zchild);
-			ADD_ASSOC_ZVAL_EX(return_value, "$scope", &state.zchild);
-#else
-			Z_ADDREF_P(state.zchild);
-			ADD_ASSOC_ZVAL_EX(return_value, "$scope", state.zchild);
-#endif
+		if (!php_phongo_bson_to_zval_ex(bson_get_data(intern->scope), intern->scope->len, &state)) {
+			zval_ptr_dtor(&state.zchild);
+			return;
 		}
 
-		zval_ptr_dtor(&state.zchild);
+#if PHP_VERSION_ID >= 70000
+		ADD_ASSOC_ZVAL_EX(return_value, "$scope", &state.zchild);
+#else
+		ADD_ASSOC_ZVAL_EX(return_value, "$scope", state.zchild);
+#endif
 	}
 } /* }}} */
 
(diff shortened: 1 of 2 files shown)


src/BSON/Javascript.c
```diff
@@ -490,17 +487,12 @@ HashTable* php_phongo_javascript_get_properties_hash(zval* object, bool is_debug
...
-				Z_ADDREF(state.zchild);
-				zend_hash_str_update(props, "scope", sizeof("scope") - 1, &state.zchild);
-			} else {
-				zval scope;
-
-				ZVAL_NULL(&scope);
-				zend_hash_str_update(props, "scope", sizeof("scope") - 1, &scope);
+			if (!php_phongo_bson_to_zval_ex(bson_get_data(intern->scope), intern->scope->len, &state)) {
+				zval_ptr_dtor(&state.zchild);
+				return NULL;
```
jmikola: `php_phongo_javascript_get_properties_hash` never used to return early. Is this a potential problem if `props` is never freed, or did we previously discuss that PHP itself would clean it up if an exception was thrown during either of the `get_debug_info` or `get_properties` handlers?

alcaeus: What I can see from the default `get_debug_info` handler is that it's perfectly fine to return `NULL`: https://github.com/php/php-src/blob/23c65a817390d219bbe77f363cf14956c5c7119b/Zend/zend_object_handlers.c#L152. The function also accepts a `null` return from a `__debuginfo` magic function, so this seems to be ok.

Regardless, I've added a call to `FREE_HASHTABLE` to ensure we free the memory allocated for the hash table when we don't return it.

src/BSON/Javascript.c
```diff
@@ -543,6 +529,10 @@ HashTable* php_phongo_javascript_get_properties_hash(zval* object, bool is_debug
 #endif
 
 	return props;
+
+failure:
+	FREE_HASHTABLE(props);
```
jmikola: I'm not sure this is correct. If you look at the definition of [`PHONGO_GET_PROPERTY_HASH_INIT_PROPS`](https://github.com/mongodb/mongo-php-driver/blob/master/php_phongo.h#L187), you'll see that we allocate _and_ initialize if we're returning properties for debugging purposes. Otherwise, we set `props` to the address of the intern's properties hash and conditionally alloc/init it if needed.

[`FREE_HASHTABLE`](http://lxr.jpauli.tech/source/xref/php-7.4/Zend/zend_alloc.h#236) corresponds to the initial allocation, but I think we're also missing a `zend_hash_destroy` to balance out the initialization (and clean up other fields that might have been added to the hash, such as the `code` string). If that makes sense, I'd propose we do so only for the case where `is_debug` is true.

For the non-debug case when we're working with the intern's properties hash, I think it'd be preferable to leave things as-is. On a failure, `scope` hasn't been assigned at all. It's possible that `code` has been set, but a future `get_properties` call could just update the field without a risk of leaking. That's what happens all the time when calling `get_properties` on our objects. We can trust that the hash table will ultimately get cleaned up during the `free_object` handler when the Javascript object dies.

alcaeus: Very good point, I've adjusted the code and created a new `PHONGO_GET_PROPERTY_HASH_FREE_PROPS` macro that encapsulates the logic of checking `is_debug` and destroying the hash if so.

PR #1064, opened 2019-11-26
PHPC-1445: Make find commands for collecting sources portable

alcaeus:
https://jira.mongodb.org/browse/PHPC-1445

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -293,19 +293,20 @@ if test "$PHP_MONGODB" != "no"; then
     AC_SUBST(MONGOC_USER_SET_CFLAGS, [])
     AC_SUBST(MONGOC_USER_SET_LDFLAGS, [])
 
-    dnl Generated with: find src/libmongoc/src/common -name '*.c' -print0 | cut -sz -d / -f 5- | sort -z | tr '\000' ' '
+    dnl On MacOS, use gcut from the coreutils brew package instead of cut
+    dnl Generated with: find src/libmongoc/src/common -name '*.c' -print0 | cut -sz -d / -f 5- | sort -dz | tr '\000' ' '
     PHP_MONGODB_COMMON_SOURCES="common-b64.c common-md5.c"
 
-    dnl Generated with: find src/libmongoc/src/libbson/src/bson -name '*.c' -print0 | cut -sz -d / -f 7- | sort -z | tr '\000' ' '
+    dnl Generated with: find src/libmongoc/src/libbson/src/bson -name '*.c' -print0 | cut -sz -d / -f 7- | sort -dz | tr '\000' ' '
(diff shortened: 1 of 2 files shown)


config.m4
```diff
@@ -293,19 +293,19 @@ if test "$PHP_MONGODB" != "no"; then
...
 
-    dnl Generated with: find src/libmongoc/src/common -name '*.c' -print0 | cut -sz -d / -f 5- | sort -z | tr '\000' ' '
+    dnl Generated with: find src/libmongoc/src/common -name '*.c' -print0 | gcut -sz -d / -f 5- | sort -z | tr '\000' ' '
     PHP_MONGODB_COMMON_SOURCES="common-b64.c common-md5.c"
 
-    dnl Generated with: find src/libmongoc/src/libbson/src/bson -name '*.c' -print0 | cut -sz -d / -f 7- | sort -z | tr '\000' ' '
-    PHP_MONGODB_BSON_SOURCES="bcon.c bson-atomic.c bson.c bson-clock.c bson-context.c bson-decimal128.c bson-error.c bson-iso8601.c bson-iter.c bson-json.c bson-keys.c bson-md5.c bson-memory.c bson-oid.c bson-reader.c bson-string.c bson-timegm.c bson-utf8.c bson-value.c bson-version-functions.c bson-writer.c"
+    dnl Generated with: find src/libmongoc/src/libbson/src/bson -name '*.c' -print0 | gcut -sz -d / -f 7- | sort -z | tr '\000' ' '
+    PHP_MONGODB_BSON_SOURCES="bcon.c bson-atomic.c bson-clock.c bson-context.c bson-decimal128.c bson-error.c bson-iso8601.c bson-iter.c bson-json.c bson-keys.c bson-md5.c bson-memory.c bson-oid.c bson-reader.c bson-string.c bson-timegm.c bson-utf8.c bson-value.c bson-version-functions.c bson-writer.c bson.c"
```
jmikola: Is this the result of a subtle difference in macOS' `sort` command?

alcaeus: Yes. By default, `sort` on Linux will use dictionary order (invoked with `-d`), which is not what MacOS does. Making the `d` argument explicit fixes this and removes the diff.

alcaeus:
@jmikola While re-checking the sorting you pointed out above, I noticed that this needed another change.

The problem is that the regular `cut` program on MacOS does not understand the `z` flag to cut along a zero byte. Installing `coreutils` and using `gcut` solves this, but is no longer portable. I've added a note to both config.m4 and config.w32 to use `gcut` on MacOS, either by changing the command before invoking or by symlinking `cut` to invoke `gcut`. The sort order can be made consistent by using the `d` option as indicated above.

PR #1065, opened 2019-11-27
PHPC-1496: Add support for compiling with libmongocrypt

alcaeus:
https://jira.mongodb.org/browse/PHPC-1496

Commits will be squashed before merging.

Open items:
- [x] Remove symlink workaround with src/libmongocrypt-tmp (see [CDRIVER-3439](https://jira.mongodb.org/browse/CDRIVER-3439))
- [x] Generate mongocrypt-export.h automatically
- [x] Create correct mongocrypt-export.h
- [x] Compile on Windows
- [x] Double-check logic on libmongocrypt auto detection to avoid failures in auto-mode
- [x] Determine correct libmongocrypt version when compiling
- [x] Show libmongocrypt output for phpinto
- [x] Document updating libmongocrypt in contributing.md
- [x] Update PECL package generation script
- [x] Ensure libmongocrypt is built with crypto support and reports it correctly (issue found on MacOS)
- [x] Ensure libmongocrypt with crypto works properly on Windows (may currently implicitly use crypto_none)
- [x] Update to latest libmongocrypt version and remove conditional compiles


php_phongo.c
```diff
@@ -40,6 +40,10 @@
 #include <ext/standard/php_smart_str.h>
 #endif
 
+#ifdef MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION
+#include <mongocrypt/mongocrypt.h>
```
kevinAlbs: I reverted the changes of [MONGOCRYPT-213](https://jira.mongodb.org/browse/MONGOCRYPT-213) because in moving the public headers, I modified their relative #include's, which inadvertently broke the Go driver build (which is not build in the libmongocrypt evergreen project). So I'm considering other solutions to avoid any disruption.

But I'm not sure I fully understand the problem here. Is the purpose of `libmongocrypt-tmp` directory to make it possible to refer to the public header as `#include <mongocrypt/mongocrypt.h>`? Would this work without any changes to libmongocrypt if this  were `#include <mongocrypt.h>`, or is that undesirable?

If it is necessary, perhaps it would suffice if libmongocrypt have a single forwarding header in a mongocrypt subdirectory. I.e. I simply add `src/mongocrypt/mongocrypt.h`, which just includes `../mongocrypt.h`.

alcaeus: > But I'm not sure I fully understand the problem here. Is the purpose of `libmongocrypt-tmp` directory to make it possible to refer to the public header as `#include <mongocrypt/mongocrypt.h>`? Would this work without any changes to libmongocrypt if this were `#include <mongocrypt.h>`, or is that undesirable?

Changing this to `#include <mongocrypt.h>` would work for our use-case, as that would allow us to continue specifying `src/libmongocrypt/src` as include path.

A forwarding header could also work. Since we already need to ship a `mongocrypt-export.h` (PHP extensions don't use cmake), we can also add that file to avoid needing to package it. I can test that solution, maybe we can avoid making upstream changes for this, unless you want to solve this for all projects that bundle libmongocrypt.

kevinAlbs: > Changing this to #include <mongocrypt.h> would work for our use-case, as that would allow us to continue specifying src/libmongocrypt/src as include path.

If forwarding headers were added, would `src/libmongoc/src` no longer be in the include path? I had thought `#include <mongocrypt/mongocrypt.h>` intends to resolve to src/libmongoc/src/mongocrypt/mongocrypt.h, so `src/libmongoc/src` needs to be in the include path regardless.

Also, I now remember we made a similar change to libmongoc/libbson. Prior to https://jira.mongodb.org/browse/CDRIVER-2800, consumers were expected to include like `#include <bson.h>`, but that change added the `bson`/`mongoc` prefix directories, and changed the top-level files to be forwarding headers.

No other drivers bundle libmongocrypt currently. It's certainly possible to change libmongocrypt to be included via `mongocrypt/mongocrypt.h` in both the source and the installed headers, but I'm just not sure if the advantage is just convention or a real necessity (e.g. how CDRIVER-2800 was necessary for Darwin frameworks).

An easy and risk-free change in libmongocrypt would be to add forwarding headers. Let me know if you'd like me to make that change in libmongocrypt, or if you'd rather do that here.

alcaeus: I've worked around this for now by adding our own mongocrypt.h file: https://github.com/mongodb/mongo-php-driver/pull/1065/files#diff-f309cffeaf85fbe6fe0f72f5637ffe50

For now this works, but not needing compat files here would sure be better.

Another issue is mongocrypt-export.h, which normally is generated using cmake. Since PHP extensions use make, I've copied a mongocrypt-export.h into our compat folder, but ideally we wouldn't have to have this ourselves either. However, I can live with keeping it here for the time being.

config.w32
```diff
@@ -45,6 +45,7 @@ function mongodb_parse_version_file(inpath, prefix)
 
 ARG_ENABLE("mongodb", "MongoDB support", "no");
 ARG_WITH("mongodb-sasl", "MongoDB: Build against Cyrus-SASL", "yes");
+ARG_ENABLE("client-side-encryption", "MongoDB: Enable client-side encryption", "auto");
```
jmikola: From my understanding of [`ARG_ENABLE`](https://github.com/php/php-src/blob/5d6e923d46a89fe9cd8fb6c3a6da675aa67197b4/win32/build/confutils.js#L232), the default value is an arbitrary string and there's no enforcement of possible values. I've noted that within this file, we only care that the value is not `"no"`.

config.w32
```diff
@@ -68,6 +69,9 @@ if (PHP_MONGODB != "no") {
     /I" + configure_module_dirname + "/src/libmongoc/src/libbson/src \
     /I" + configure_module_dirname + "/src/libmongoc/src/libbson/src/jsonsl \
     /I" + configure_module_dirname + "/src/libmongoc/src/libmongoc/src \
+    /I" + configure_module_dirname + "/src/libmongocrypt/src \
+    /I" + configure_module_dirname + "/src/libmongocrypt/kms-message/src \
+    /I" + configure_module_dirname + "/src/libmongocrypt-compat \
```
jmikola: Should these three includes be wrapped in a `(PHP_CLIENT_SIDE_ENCRYPTION != "no")` conditional?

config.m4
```diff
@@ -282,9 +322,9 @@ if test "$PHP_MONGODB" != "no"; then
     AC_SUBST(MONGOC_ENABLE_RDTSCP, 0)
     AC_SUBST(MONGOC_ENABLE_SHM_COUNTERS, 0)
     AC_SUBST(MONGOC_TRACE, 1)
-
-    dnl TODO: Replace with detection for libmongocrypt for PHPC-1293
-    AC_SUBST(MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION, 0)
+    AC_SUBST(MONGOCRYPT_ENABLE_TRACE, 1)
+    AC_SUBST(MONGOCRYPT_IS_POSIX, 1)
```
jmikola: I'll note that you're always defining `IS_POSIX` as `1` and `IS_WIN` as `0`, so that may support my later suggestion to not check `$TARGET_OS` and always compile `PHP_MONGODB_MONGOCRYPT_POSIX_SOURCES`.

config.w32
```diff
@@ -176,7 +191,14 @@ if (PHP_MONGODB != "no") {
...
     MONGOC_CC: "",
     MONGOC_USER_SET_CFLAGS: "",
-    MONGOC_USER_SET_LDFLAGS: ""
+    MONGOC_USER_SET_LDFLAGS: "",
+    MONGOCRYPT_ENABLE_TRACE: 1,
+    MONGOCRYPT_IS_POSIX: 0,
+    MONGOCRYPT_IS_WIN: 1,
+    MONGOCRYPT_ENABLE_CRYPTO: 0,
+    MONGOCRYPT_ENABLE_CRYPTO_OPENSSL: 0,
+    MONGOCRYPT_ENABLE_CRYPTO_CNG: 0,
```
jmikola: I'd suggest prefixing this line with `// TODO: Support building with Secure Channel on Windows`, as we have earlier.

config.w32
```diff
@@ -176,7 +191,14 @@ if (PHP_MONGODB != "no") {
...
+    MONGOCRYPT_ENABLE_CRYPTO_COMMON_CRYPTO: 0
```
jmikola: I'd suggest prefixing this line with `// Secure Transport does not apply to Windows`, as we have earlier.

config.w32
```diff
@@ -176,7 +191,14 @@ if (PHP_MONGODB != "no") {
     MONGOC_HAVE_RES_SEARCH: 0,
     MONGOC_CC: "",
     MONGOC_USER_SET_CFLAGS: "",
-    MONGOC_USER_SET_LDFLAGS: ""
+    MONGOC_USER_SET_LDFLAGS: "",
+    MONGOCRYPT_ENABLE_TRACE: 1,
```
jmikola: On second thought, does it make sense to _not_ define any `MONGOCRYPT_` vars in the initial `mongoc_opts` definition and instead defer to the `if (PHP_CLIENT_SIDE_ENCRYPTION != "no")` condition later on? These options are only relevant to replacing terms in the header input files when we call `mongodb_generate_header()`, but we'll only call that from within the conditional anyway. If so, I'd suggest creating a `mongocrypt_opts` object, so we don't add unnecessary fields into `mongoc_opts`. That should also test that the replacements in the libmongocrypt configs don't require anything beyond what we'll put in `mongocrypt_opts`.

`MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION` is the only definition that needs to stay with an initial definition of `0`, since we need that to generate the libmongoc configs either way.

config.w32
```diff
@@ -224,6 +246,34 @@ if (PHP_MONGODB != "no") {
     WARNING("mongodb libsasl support not enabled, libs not found");
   }
 
+  if (PHP_CLIENT_SIDE_ENCRYPTION != "no") {
+    mongoc_opts.MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION = 1;
```
jmikola: Per my earlier comment, this is where I'd suggest defining values for all of the `MONGOCRYPT_` vars.

config.w32
```diff
@@ -86,6 +90,18 @@ if (PHP_MONGODB != "no") {
...
 
+  // Generated with: find src/libmongocrypt/src -maxdepth 1 -name '*.c' -print0 | cut -sz -d / -f 4- | sort -dz | tr '\000' ' '
```
jmikola: Would it make sense to move these source definitions to the `if (PHP_CLIENT_SIDE_ENCRYPTION != "no")` conditional? I realize that doing so would add more to the condition than most others (e.g. SSL), but this is the first case we have a conditional for bundled sources.

I think doing so would keep the additions in this PR more localized (at least in `config.w32`).

config.w32
```diff
@@ -224,6 +246,34 @@ if (PHP_MONGODB != "no") {
...
+      mongoc_opts.MONGOCRYPT_ENABLE_CRYPTO_OPENSSL = 1;
+
+      ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/src", "crypto/openssl.c", "mongodb");
+      ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/kms-message/src", "kms_crypto_openssl.c", "mongodb");
+    }
+
+    mongodb_generate_header(
+      configure_module_dirname + "/src/libmongocrypt/src/mongocrypt-config.h.in",
+      configure_module_dirname + "/src/libmongocrypt/src/mongocrypt-config.h",
+      mongoc_opts
```
jmikola: This is why I'd suggest using `mongocrypt_opts`. I don't expect anything in `mongocrypt.h.in` or `mongocrypt-config.h.in` to depend on libmongoc's own vars.

scripts/autotools/CheckSSL.m4
```diff
@@ -188,18 +189,33 @@ if test "$PHP_MONGODB_SSL" = "openssl" -o "$PHP_MONGODB_SSL" = "libressl" -o "$P
...
     AC_SUBST(MONGOC_ENABLE_SSL_OPENSSL, 1)
     AC_SUBST(MONGOC_ENABLE_SSL_LIBRESSL, 0)
     AC_SUBST(MONGOC_ENABLE_SSL_SECURE_TRANSPORT, 0)
     AC_SUBST(MONGOC_ENABLE_CRYPTO_LIBCRYPTO, 1)
     AC_SUBST(MONGOC_ENABLE_CRYPTO_COMMON_CRYPTO, 0)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO, 1)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_OPENSSL, 1)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_COMMON_CRYPTO, 0)
+
+    PHP_MONGODB_ADD_SOURCES([src/libmongocrypt/src/], "crypto/openssl.c", $PHP_MONGODB_BUNDLED_CFLAGS)
```
jmikola: @kevinAlbs: IIRC, libmongoc's own TLS-specific source files wrap themselves with compiler macros like `#ifdef MONGOC_ENABLE_SSL_OPENSSL` to ensure they're safe to compile regardless of the TLS library the project is configured to use. Is there any reason we can't do the same for libmongocrypt, given that constants like `MONGOCRYPT_ENABLE_CRYPTO_OPENSSL` exist?

@alcaeus: Were these TLS-specific source files manually excluded from the `find` commands in `config.m4` and `config.w32`? I didn't see anything in the example shell one-liner that excluded them, so I'm curious if you did so manually -- which could pose a problem if someone else were to run that command to regenerate the source file list for a particular directory.

config.w32
```diff
@@ -224,6 +246,34 @@ if (PHP_MONGODB != "no") {
...
+    ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/src", PHP_MONGODB_MONGOCRYPT_SOURCES, "mongodb");
+    ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/kms-message/src", PHP_MONGODB_MONGOCRYPT_KMS_MESSAGE_SOURCES, "mongodb");
+    ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/src/os_win", PHP_MONGODB_MONGOCRYPT_WIN_SOURCES, "mongodb");
+
+    if (mongoc_ssl_found) {
+      mongoc_opts.MONGOCRYPT_ENABLE_CRYPTO = 1;
+      mongoc_opts.MONGOCRYPT_ENABLE_CRYPTO_OPENSSL = 1;
+
+      ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/src", "crypto/openssl.c", "mongodb");
+      ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/kms-message/src", "kms_crypto_openssl.c", "mongodb");
```
jmikola: I have some questions about needing to manually include these source files separately and whether or not they're in the original list produced by `find`. Asked that in a separate thread on `CheckSSL.m4`.

php_phongo.c
```diff
@@ -3195,6 +3199,33 @@ PHP_MINFO_FUNCTION(mongodb)
...
+	php_info_print_table_row(2, "libmongocrypt crypto library", "Common Crypto");
+#elif defined(MONGOC_ENABLE_CRYPTO_CNG)
+	php_info_print_table_row(2, "libmongocrypt crypto library", "CNG");
+#else
+	php_info_print_table_row(2, "libmongocrypt crypto library", "unknown");
+#endif
+#else
+	php_info_print_table_row(2, "libmongocrypt crypto", "disabled");
+#endif
+#else
```
jmikola: I'd suggest adding `/* MONGOCRYPT_ENABLE_CRYPTO */` here, as we have above for libmongoc's crypto. Helps readability for long `#if` chains.

scripts/autotools/libmongocrypt/Version.m4
```diff
@@ -0,0 +1,4 @@
+MONGOCRYPT_CURRENT_FILE=[]PHP_EXT_SRCDIR(mongodb)[/src/LIBMONGOCRYPT_VERSION_CURRENT]
+BUILD_VERSION=$(cat $MONGOCRYPT_CURRENT_FILE)
+
+AC_SUBST(BUILD_VERSION)
```
jmikola: Is there a particular reason this differs from the libmongoc and libbson `Versions.m4` file? Does libmongocrypt not have separate constants for version parts (e.g. major, minor)?

I'm a bit concerned that we're using a non-prefixed `BUILD_VERSION` constant here. I would expect libmongocrypt to at least use `MONGOCRYPT_VERSION` for the single version string.

@kevinAlbs: Is this something we can address in libmongocrypt, or has that ship sailed?

src/libmongocrypt-compat/mongocrypt-export.h
```diff
@@ -0,0 +1,7 @@
+#ifndef MONGOCRYPT_EXPORT_H
+#define MONGOCRYPT_EXPORT_H
+
+#define MONGOCRYPT_EXPORT
+#define MONGOCRYPT_NO_EXPORT
```
jmikola: @alcaeus I assume my suggestion to define these as empty strings worked?

scripts/autotools/CheckSSL.m4
kevinAlbs: That seems like a good idea, created and scheduled https://jira.mongodb.org/browse/MONGOCRYPT-218

scripts/autotools/libmongocrypt/Version.m4
kevinAlbs: If you're referring to renaming `BUILD_VERSION` to something more specific, that's fine with me. Though libmongoc also has the same cmake option to set a specific version to build with (https://github.com/mongodb/mongo-c-driver/blob/f930d74daf3fc9125e34330110a8510fd45c3ce5/CMakeLists.txt#L78)

config.m4
```diff
@@ -230,8 +236,40 @@ if test "$PHP_MONGODB" != "no"; then
...
+      else
+        AC_SUBST(MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION, 0)
+        AC_MSG_RESULT(not found, compiling without client-side encryption)
+      fi
+    fi
+  fi
+
   if test "$PHP_LIBBSON" = "no" -a "$PHP_LIBMONGOC" = "no"; then
     PHP_MONGODB_BUNDLED_CFLAGS="$STD_CFLAGS -DBSON_COMPILATION -DMONGOC_COMPILATION"
+    PHP_MONGODB_LIBMONGOCRYPT_CFLAGS="$PHP_MONGODB_BUNDLED_CFLAGS -std=gnu99"
```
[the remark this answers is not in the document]
alcaeus: That was the only standard that got this to compile on evergreen. `-std=c99` did not work, I explicitly had to choose gnu99. Without this, the build failed because of inline declarations in `for` loops, which were only added in C99: `for (int i = 0; ...)`. c99 brought on other errors, while gnu99 worked successfully. Not sure what the difference is, but happy to change to a different standard as long as it keeps the build successful.

config.m4
```diff
@@ -306,6 +346,18 @@ if test "$PHP_MONGODB" != "no"; then
...
 
+    dnl Generated with: find src/libmongocrypt/src -maxdepth 1 -name '*.c' -print0 | cut -sz -d / -f 4- | sort -dz | tr '\000' ' '
+    PHP_MONGODB_MONGOCRYPT_SOURCES="mongocrypt-binary.c mongocrypt-buffer.c mongocrypt.c mongocrypt-cache.c mongocrypt-cache-collinfo.c mongocrypt-cache-key.c mongocrypt-ciphertext.c mongocrypt-crypto.c mongocrypt-ctx.c mongocrypt-ctx-datakey.c mongocrypt-ctx-decrypt.c mongocrypt-ctx-encrypt.c mongocrypt-key-broker.c mongocrypt-key.c mongocrypt-kms-ctx.c mongocrypt-log.c mongocrypt-marking.c mongocrypt-opts.c mongocrypt-status.c mongocrypt-traverse-util.c"
+
+    dnl Generated with: find src/libmongocrypt/src/os_posix -name '*.c' -print0 | cut -sz -d / -f 5- | sort -dz | tr '\000' ' '
+    PHP_MONGODB_MONGOCRYPT_POSIX_SOURCES="os_mutex.c os_once.c"
+
+    dnl Generated with: find src/libmongocrypt/src/os_win -name '*.c' -print0 | cut -sz -d / -f 5- | sort -dz | tr '\000' ' '
+    PHP_MONGODB_MONGOCRYPT_WIN_SOURCES="os_mutex.c os_once.c"
```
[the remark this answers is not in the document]
alcaeus: Removed the var and the `$TARGET_OS` conditional.

config.m4
```diff
@@ -385,6 +462,7 @@ Build configuration:
   Code Coverage flags (extra slow)                 : $COVERAGE_CFLAGS
   System mongoc                                    : $PHP_LIBMONGOC
   System libbson                                   : $PHP_LIBBSON
+  System libmongocrypt                             : $PHP_LIBBSON
```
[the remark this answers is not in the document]
alcaeus: Comment added.

scripts/autotools/CheckSSL.m4
alcaeus: For the `crypto/*` files, the `-maxdepth 1` argument to `find` currently excludes them. For the `kms_crypto_*` files in the other variable, we exclude the files via a `! -name 'kms_crypto_*'` argument to `find`. This makes any manual changes to the lists unnecessary, but requires maintenance if we later add any files that match the exclude pattern but should be included regardless.

I'll leave this in and have created [PHPC-1509](https://jira.mongodb.org/browse/PHPC-1509) to track this.

php_phongo.c
alcaeus: Added a comment, especially since this is not the `#else` for `MONGOCRYPT_ENABLE_CRYPTO` but rather `MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION`. It beautifully proves your point.

scripts/autotools/libmongocrypt/Version.m4
alcaeus: > Does libmongocrypt not have separate constants for version parts (e.g. major, minor)?

No, `MONGOCRYPT_VERSION` is the only version constant, containing the version as a string. This is comparable to the `MONGOC_VERSION_S` constant, but the other `MONGOC_*_VERSION` constants have no equivalent in libmongocrypt.

src/libmongocrypt-compat/mongocrypt-export.h
alcaeus: Yes, that worked beautifully 👍

scripts/autotools/CheckSSL.m4
jmikola: SGTM. Thanks!

scripts/autotools/libmongocrypt/Version.m4
jmikola: @kevinAlbs My problem is with the `@BUILD_VERSION@` replacement in libmongocrypt's header file, since that's effectively within the same namespace as other replacements when we need to process the libmongoc and libbson files with Autotools. I think the CMake example you shared is setting a `BUILD_VERSION` _environment_ variable, since libmongoc does use prefixed replacement tokens in [`mongoc-version.h.in`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-version.h.in).

Assuming we're all on the same page: I don't think there's currently a conflict with other replacements (since libmongoc and libbson all prefix), but ideally I'd like to see this fixed alongside [MONGOCRYPT-218](https://jira.mongodb.org/browse/MONGOCRYPT-218) if you can create another ticket for it (likewise for @alcaeus to create a PHPC ticket as a todo reminder).

config.m4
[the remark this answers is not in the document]
kevinAlbs: I think libmongocrypt should ideally build under `-std=c89`. I created https://jira.mongodb.org/browse/MONGOCRYPT-219 to make any necessary fixes. If this is something we should do now, I can schedule it.

scripts/autotools/libmongocrypt/Version.m4
kevinAlbs: The cmake example is referring to the same BUILD_VERSION cmake variable that is substituted in [mongocrypt.h.in](https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt.h.in#L34), it is not an environment variable, unless I'm misunderstanding.

Created and scheduled https://jira.mongodb.org/browse/MONGOCRYPT-220

jmikola: I was thinking of the CMake environment when I wrote environment variable earlier. I was mainly differentiating its context from the replacement tokens in the `.in` file. Sorry for the confusion.

scripts/autotools/CheckSSL.m4
```diff
@@ -178,6 +178,7 @@ AC_MSG_RESULT([$PHP_MONGODB_SSL])
 dnl Disable Windows SSL and crypto
 AC_SUBST(MONGOC_ENABLE_SSL_SECURE_CHANNEL, 0)
 AC_SUBST(MONGOC_ENABLE_CRYPTO_CNG, 0)
+AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_CNG, 0)
```
jmikola: Rather than modify libmongoc's `CheckSSL.m4` file, I wonder if we can split the libmongocrypt bits out into their own file. I was playing around with the following, which I included _after_ libmongoc's `CheckSSL.m4`:

```m4
if test "$PHP_CLIENT_SIDE_ENCRYPTION" != "no"; then
  AC_MSG_CHECKING([which crypto library to use for libmongocrypt])
  AC_MSG_RESULT([$PHP_MONGODB_SSL])

  dnl Disable Windows crypto
  AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_CNG, 0)

  if test "$PHP_MONGODB_SSL" = "darwin"; then
    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO, 1)
    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_OPENSSL, 0)
    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_COMMON_CRYPTO, 1)
  elif test "$PHP_MONGODB_SSL" = "openssl" -o "$PHP_MONGODB_SSL" = "libressl"; then
    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO, 1)
    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_OPENSSL, 1)
    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_COMMON_CRYPTO, 0)
  fi
fi
```

With this, we would then move the `PHP_MONGODB_ADD_SOURCES` calls back to `config.m4` (with their own `$PHP_MONGODB_SSL` conditions to handle Darwin and OpenSSL/LibreSSL. That would avoid blending libmongoc and libmongocrypt build scripts, and also ensure some consistency since no other scripts under `scripts/autotools` add sources directly (IIRC).

@kevinAlbs: As a side note, I originally worried that libmongocrypt didn't support LibreSSL until I realized both LibreSSL and OpenSSL share libcrypto. I'd propose the "openssl" terminology in libmongocrypt constants and source files get renamed to refer to libcrypto is that wouldn't be backwards breaking (otherwise, no worries). For instance, `CRYPTO_LIBCRYPTO` over `CRYPTO_OPENSSL`.

config.w32
```diff
@@ -224,6 +224,61 @@ if (PHP_MONGODB != "no") {
...
+    var PHP_MONGODB_MONGOCRYPT_WIN_SOURCES="os_mutex.c os_once.c";
+
+    // Generated with: find src/libmongocrypt/kms-message/src -name '*.c' ! -name 'kms_crypto_*' -print0 | cut -sz -d / -f 5- | sort -dz | tr '\000' ' '
+    var PHP_MONGODB_MONGOCRYPT_KMS_MESSAGE_SOURCES="hexlify.c kms_b64.c kms_caller_identity_request.c kms_decrypt_request.c kms_encrypt_request.c kms_kv_list.c kms_message.c kms_request.c kms_request_opt.c kms_request_str.c kms_response.c kms_response_parser.c sort.c";
+
+    ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/src", PHP_MONGODB_MONGOCRYPT_SOURCES, "mongodb");
+    ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/kms-message/src", PHP_MONGODB_MONGOCRYPT_KMS_MESSAGE_SOURCES, "mongodb");
+    ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/src/os_win", PHP_MONGODB_MONGOCRYPT_WIN_SOURCES, "mongodb");
+
+    if (mongoc_ssl_found) {
```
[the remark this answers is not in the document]
kevinAlbs: In the current CR for MONGOCRYPT-218, `none.c` is always compiled in, with the function definitions between an `#ifndef MONGOCRYPT_ENABLE_CRYPTO`.

If not building with TLS, can we skip defining `MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION` to 
disable FLE in libmongoc instead? That will give a better error to the user, along the lines of "Client Side Encryption not supported" if they attempt to use FLE.

scripts/autotools/CheckSSL.m4
kevinAlbs: Rename is in the CR for MONGOCRYPT-218.

config.m4
```diff
@@ -230,6 +236,37 @@ if test "$PHP_MONGODB" != "no"; then
...
+      if $PKG_CONFIG libmongocrypt-1.0 --atleast-version 1.0.0; then
+        PHP_MONGODB_MONGOCRYPT_CFLAGS=`$PKG_CONFIG libmongocrypt-1.0 --cflags`
+        PHP_MONGODB_MONGOCRYPT_LIBS=`$PKG_CONFIG libmongocrypt-1.0 --libs`
+        PHP_MONGODB_MONGOCRYPT_VERSION=`$PKG_CONFIG libmongocrypt-1.0 --modversion`
+        AC_MSG_RESULT(version $PHP_MONGODB_MONGOCRYPT_VERSION found)
+
+        PHP_MONGODB_CFLAGS="$PHP_MONGODB_CFLAGS $PHP_MONGODB_MONGOCRYPT_CFLAGS"
+        PHP_EVAL_LIBLINE($PHP_MONGODB_MONGOCRYPT_LIBS, MONGODB_SHARED_LIBADD)
+        AC_DEFINE(HAVE_SYSTEM_LIBMONGOCRYPT, 1, [Use system libmongocrypt])
+        AC_SUBST(MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION, 1)
```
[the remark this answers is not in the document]
alcaeus: Yes, this can be dropped.

config.m4
```diff
@@ -342,6 +380,41 @@ if test "$PHP_MONGODB" != "no"; then
...
+        ${ac_config_dir}/src/libmongocrypt/src/mongocrypt-config.h
+        ${ac_config_dir}/src/libmongocrypt/src/mongocrypt.h
+      ])
+
+      AC_SUBST(MONGOCRYPT_ENABLE_TRACE, 1)
+      AC_SUBST(MONGOCRYPT_IS_POSIX, 1)
+      AC_SUBST(MONGOCRYPT_IS_WIN, 0)
+      AC_SUBST(MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION, 1)
+    else
+      AC_SUBST(MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION, 0)
```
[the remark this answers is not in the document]
alcaeus: Interesting, I didn't think of that. Good to know that it works, but I can definitely set all of those flags before we call `AC_CONFIG_FILES` if it helps avoid confusion.

config.w32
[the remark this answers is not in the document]
alcaeus: We can make an SSL library a requirement for compiling with FLE support, that would be fine. Would that work for you @jmikola?

scripts/autotools/CheckSSL.m4
alcaeus: Moving to separate file sounds good 👍

php_phongo.c
alcaeus: I'll defer to @kevinAlbs - not sure what we can do about mongocrypt-export.h, but a forwarding file for mongocrypt.h would certainly be nice.

kevinAlbs: To check, the change in libmongocrypt would be to effectively have in the src directory  what you're doing with src/libmongocrypt-compat? I.e. add the directory mongocrypt and these forwarding headers?
1. src/mongocrypt/mongocrypt.h
2. src/mongocrypt/mongocrypt-export.h

As @alcaeus mentioned, mongocrypt-export.h is generated by cmake (and I believe is system dependent). Including the mongocrypt-export.h included here, with empty `MONGOCRYPT_EXPORT`/`MONGOCRYPT_NO_EXPORT` as part of libmongocrypt's src seems a little odd, but also harmless.

alcaeus: Yes, those headers would work fine for us. If cmake overwrites the default mongocrypt/mongocrypt-export.h, having it there with empty macros would indeed be harmless 👍

config.m4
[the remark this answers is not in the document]
jmikola: It looks like it'd be trivial to move `AC_CONFIG_FILES` to the end of the `if test "$PHP_CLIENT_SIDE_ENCRYPTION" != "no";` block for consistency. As for `AC_SUBST(MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION, ...)`, I think we can move that further up near the libmongoc subs with a small if/else condition on `$PHP_CLIENT_SIDE_ENCRYPTION` just to keep things localized.

config.w32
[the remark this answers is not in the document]
jmikola: Setting `MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION` to `0` when SSL isn't available seems preferable for both `config.m4` and `config.w32`.

php_phongo.c
kevinAlbs: Adding the pregenerated mongocrypt-export.h and forwarding mongocrypt.h header to libmongocrypt still seems a little odd to me. It's only necessary for this bundling, and we wouldn't be testing them within libmongocrypt. Though we can always revisit this later, I vote we leave these in the mongo-php-driver repo for now.

config.m4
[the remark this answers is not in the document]
alcaeus: Is there any risk for us building with `-std=gnu99`? `c99` doesn't support `ssize_t`, which is why that doesn't work. We just need to ensure this works on all supported platforms.

config.m4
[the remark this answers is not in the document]
alcaeus: I've moved some things around and extracted the entire libmongocrypt stuff into their include files. This moves them up before any `AC_CONFIG_FILES` calls and also makes the main config file smaller.

config.w32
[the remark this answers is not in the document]
alcaeus: I've done this. I'll have to revisit Windows support, as we currently implicitly compile with `MONGOCRYPT_ENABLE_CRYPTO=1` but no crypto library.

There's also an issue with `MONGOCRYPT_ENABLE_CRYPTO` being enabled in the configure script, but phpinfo output reporting libmongocrypt crypto as "disabled". Will have to investigate that before merging this.

scripts/autotools/CheckSSL.m4
alcaeus: Moved a whole bunch of libmongocrypt code to separate files.

config.w32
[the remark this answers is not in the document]
jmikola: > I'll have to revisit Windows support, as we currently implicitly compile with MONGOCRYPT_ENABLE_CRYPTO=1 but no crypto library.

Is that still accurate? I see that `MONGOCRYPT_ENABLE_CRYPTO` is now initialized to zero and only conditionally set to `1` below.

scripts/autotools/libmongocrypt/CheckSSL.m4
```diff
@@ -0,0 +1,35 @@
...
+    PHP_MONGODB_ADD_SOURCES([src/libmongocrypt/kms-message/src/], "kms_crypto_apple.c", $PHP_MONGODB_LIBMONGOCRYPT_CFLAGS)
+  elif test "$PHP_MONGODB_SSL" = "openssl" -o "$PHP_MONGODB_SSL" = "libressl"; then
+    AC_SUBST(MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION, 1)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO, 1)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_OPENSSL, 1)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_COMMON_CRYPTO, 0)
+
+    PHP_MONGODB_ADD_SOURCES([src/libmongocrypt/src/], "crypto/openssl.c", $PHP_MONGODB_LIBMONGOCRYPT_CFLAGS)
+    PHP_MONGODB_ADD_SOURCES([src/libmongocrypt/kms-message/src/], "kms_crypto_openssl.c", $PHP_MONGODB_LIBMONGOCRYPT_CFLAGS)
+  elif test "$PHP_CLIENT_SIDE_ENCRYPTION" = "auto"; then
```
jmikola: Can you remind me if we decided to prohibit compilation of libmongocrypt altogether if a crypto library wasn't available, rather than allow it to build and raise a runtime error?

scripts/autotools/libmongocrypt/CheckSSL.m4
```diff
@@ -0,0 +1,35 @@
+if test "$PHP_CLIENT_SIDE_ENCRYPTION" != "no"; then
+  AC_MSG_CHECKING([which crypto library to use for libmongocrypt])
+  AC_MSG_RESULT([$PHP_MONGODB_SSL])
+
+  dnl Disable Windows crypto
+  AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_CNG, 0)
+
+  if test "$PHP_MONGODB_SSL" = "darwin"; then
+    AC_SUBST(MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION, 1)
```
jmikola: I'd prefer to keep this alongside the other `AC_SUBST` calls for `MONGOC_` constants, even if it means an extra `"$PHP_CLIENT_SIDE_ENCRYPTION" != "no"` conditional.

One thing that may help is if you look in libmongoc's own `CheckSSL.m4`, you'll see that we re-assign `$PHP_MONGODB_SSL` once we know that we've selected a TLS library. If we cannot (for the auto case), we'll ultimately assign it to "no", such that the variable is only either "yes" or "no" after that include is run.

If you were to do that in this file, then we can have a definitive yes/no value for `$PHP_CLIENT_SIDE_ENCRYPTION` after all of our m4 includes when the time comes to do our `AC_SUBST`, `PHP_MONGODB_ADD_SOURCES`, and other calls from `config.m4`.

scripts/autotools/libmongocrypt/CheckSSL.m4
```diff
@@ -0,0 +1,35 @@
...
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO, 1)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_OPENSSL, 0)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_COMMON_CRYPTO, 1)
+
+    PHP_MONGODB_ADD_SOURCES([src/libmongocrypt/src/], "crypto/commoncrypto.c", $PHP_MONGODB_LIBMONGOCRYPT_CFLAGS)
```
jmikola: I'll elaborate on this in my later comment on `libmongocrypt.m4`, but I think `config.m4` is the best place for all of the `PHP_MONGODB_ADD_SOURCES` and similar calls.

scripts/autotools/libmongocrypt/libmongocrypt.m4
```diff
@@ -0,0 +1,27 @@
+if test "$PHP_CLIENT_SIDE_ENCRYPTION" != "no"; then
+  AC_SUBST(MONGOCRYPT_ENABLE_TRACE, 1)
+  AC_SUBST(MONGOCRYPT_IS_POSIX, 1)
+  AC_SUBST(MONGOCRYPT_IS_WIN, 0)
+
+  PHP_MONGODB_LIBMONGOCRYPT_CFLAGS="$PHP_MONGODB_BUNDLED_CFLAGS -std=gnu99"
```
jmikola: Hopefully we can get [MONGOCRYPT-219](https://jira.mongodb.org/browse/MONGOCRYPT-219) addressed before release and remove `-std=gnu99`. Until then, can we add a TODO comment that links to that issue?

scripts/autotools/libmongocrypt/Version.m4
```diff
@@ -0,0 +1,6 @@
+if test "$PHP_CLIENT_SIDE_ENCRYPTION" != "no"; then
+  MONGOCRYPT_CURRENT_FILE=[]PHP_EXT_SRCDIR(mongodb)[/src/LIBMONGOCRYPT_VERSION_CURRENT]
+  BUILD_VERSION=$(cat $MONGOCRYPT_CURRENT_FILE)
+
+  AC_SUBST(BUILD_VERSION)
```
jmikola: I think [MONGOCRYPT-220](https://jira.mongodb.org/browse/MONGOCRYPT-220) is merged, so this can be changed. If not, let's add a TODO.

scripts/autotools/libmongocrypt/libmongocrypt.m4
```diff
@@ -0,0 +1,27 @@
...
+
+  dnl Generated with: find src/libmongocrypt/src -maxdepth 1 -name '*.c' -print0 | cut -sz -d / -f 4- | sort -dz | tr '\000' ' '
+  PHP_MONGODB_MONGOCRYPT_SOURCES="mongocrypt-binary.c mongocrypt-buffer.c mongocrypt.c mongocrypt-cache.c mongocrypt-cache-collinfo.c mongocrypt-cache-key.c mongocrypt-ciphertext.c mongocrypt-crypto.c mongocrypt-ctx.c mongocrypt-ctx-datakey.c mongocrypt-ctx-decrypt.c mongocrypt-ctx-encrypt.c mongocrypt-key-broker.c mongocrypt-key.c mongocrypt-kms-ctx.c mongocrypt-log.c mongocrypt-marking.c mongocrypt-opts.c mongocrypt-status.c mongocrypt-traverse-util.c"
```
jmikola: Since this file is included before `CheckSSL.m4`, I think it's possible that we'd add these sources when `PHP_CLIENT_SIDE_ENCRYPTION` is "auto" and then fail to find a crypto library and disable FLE. I'll admit it's an edge case, but I'd prefer to inline this in `config.m4` after all of our includes and the existing libmongoc/libbson bundling code.

That would localize `PHP_MONGODB_ADD_SOURCES` and related macros in `config.m4` and be consistent with our existing m4 includes for libmongoc and libbson, which mainly do checks and call `AC_SUBST`. Doing so would also keep the "Generated with: find..." lines entirely in one file, too. This does mean we'd duplicate `$PHP_MONGODB_SSL` conditions in `config.m4` to add specific crypto sources, but I think it'd only be an if/elif for Darwin and OpenSSL/LibreSSL.

----

If you'd really like to keep this in a separate file, then I'd still suggest limiting libmongocrypt's `CheckSSL.m4` to definition code and moving its `ADD_SOURCES` calls to this file. Then, you could include _this_ m4 file towards the end of the the bundled condition in `config.m4` (after libbson and libmongoc sources are added). Note that doing so will likely require you to defer the `m4_popdef([_include])` call as well, since we'll need it for this.

scripts/autotools/libmongocrypt/CheckSSL.m4
alcaeus: See https://github.com/mongodb/mongo-php-driver/pull/1065#discussion_r356944874:
> If not building with TLS, can we skip defining `MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION` to
disable FLE in libmongoc instead? That will give a better error to the user, along the lines of "Client Side Encryption not supported" if they attempt to use FLE.

When compiling in auto mode, we can silently disable FLE, but when explicitly compiling with libmongocrypt we will error at configure-time as the user explicitly wants to build with FLE. We can remove the error message and rely on errors when running code, but I thought it was more prudent to cancel at compile time to make it explicit.

scripts/autotools/libmongocrypt/libmongocrypt.m4
alcaeus: Great catch. I inlined the contents of libmongocrypt.m4 into config.m4, as I was planning to clean up the bundled build config in a later step anyways. We can then decide how we want to structure the includes.

scripts/autotools/libmongocrypt/CheckSSL.m4
alcaeus: CheckSSL.m4 now re-assigns `PHP_CLIENT_SIDE_ENCRYPTION`, while an additional check in config.m4 ensures that `MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION` is set before the first `AC_CONFIG_FILES` for libmongoc. A second `PHP_CLIENT_SIDE_ENCRYPTION` conditional then adds sources for compiling with libmongocrypt to keep logic separated and grouped where appropriate.

alcaeus:
Regarding the todo item "Ensure libmongocrypt is built with crypto support and reports it correctly (issue found on MacOS)", this is due to us not including mongocrypt-config.h in php_phongo.c, so `MONGOCRYPT_ENABLE_CRYPTO` is never defined. Since the header file is not available when using system libmongocrypt, I've removed the phpinfo output for libmongocrypt crypto.

Since we are using the same library as for libmongoc crypto, it may be more appropriate to rename that output from "libmongoc crypto" to something more generic that covers both libmongoc and libmongocrypt.

config.m4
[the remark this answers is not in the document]
jmikola: I don't know how portable `gnu99` is, which is why I originally though to suggest `AC_PROG_CC_STDC`. I suppose we can keep this for now and see if any users complain after the release. I would ultimately like to see MONGOCRYPT-219 addressed, but I think we can wait on that vs. other higher priority work.

@alcaeus would you mind leaving a TODO comment above this with a link to MONGOCRYPT-219? Then I think we can resolve this thread.

config.m4
```diff
@@ -342,6 +383,34 @@ if test "$PHP_MONGODB" != "no"; then
       PHP_MONGODB_ADD_BUILD_DIR([src/libmongoc/src/zlib-1.2.11/])
       AC_CONFIG_FILES([${ac_config_dir}/src/libmongoc/src/zlib-1.2.11/zconf.h])
     fi
+
+    if test "$PHP_CLIENT_SIDE_ENCRYPTION" = "yes"; then
+      dnl TODO: MONGOCRYPT-219 makes the -std argument obsolete
```
jmikola: Oh, this might be the comment I was expecting to appear above `PHP_MONGODB_LIBMONGOCRYPT_CFLAGS`. Perhaps we should just have `PHP_MONGODB_LIBMONGOCRYPT_CFLAGS` defined here? I don't think it's relevant to anything outside of this conditional.

config.w32
```diff
@@ -224,6 +224,54 @@ if (PHP_MONGODB != "no") {
...
+      MONGOCRYPT_ENABLE_CRYPTO: 0,
+      MONGOCRYPT_ENABLE_CRYPTO_LIBCRYPTO: 0,
+      // TODO: Support building with Secure Channel on Windows
+      MONGOCRYPT_ENABLE_CRYPTO_CNG: 0,
+      // Secure Transport does not apply to Windows
+      MONGOCRYPT_ENABLE_CRYPTO_COMMON_CRYPTO: 0
+    };
+
+    // Generated with: find src/libmongocrypt/src -name '*.c' -print0 | cut -sz -d / -f 4- | sort -dz | tr '\000' ' '
+    var PHP_MONGODB_MONGOCRYPT_SOURCES="crypto/cng.c crypto/commoncrypto.c crypto/libcrypto.c crypto/none.c mongocrypt-binary.c mongocrypt-buffer.c mongocrypt.c mongocrypt-cache.c mongocrypt-cache-collinfo.c mongocrypt-cache-key.c mongocrypt-ciphertext.c mongocrypt-crypto.c mongocrypt-ctx.c mongocrypt-ctx-datakey.c mongocrypt-ctx-decrypt.c mongocrypt-ctx-encrypt.c mongocrypt-key-broker.c mongocrypt-key.c mongocrypt-kms-ctx.c mongocrypt-log.c mongocrypt-marking.c mongocrypt-opts.c mongocrypt-status.c mongocrypt-traverse-util.c os_posix/os_mutex.c os_posix/os_once.c os_win/os_mutex.c os_win/os_once.c";
```
jmikola: It looks like this was updated to include all crypto sources (per [MONGOCRYPT-218](https://jira.mongodb.org/browse/MONGOCRYPT-218)), so perhaps `config.m4` can as well?

config.m4
jmikola: Oh, this might be the comment I was expecting to appear above `PHP_MONGODB_LIBMONGOCRYPT_CFLAGS`.

I was going to ask if we should just have `PHP_MONGODB_LIBMONGOCRYPT_CFLAGS` defined here, but I see it's needed for libmongocrypt's `CheckSSL.m4`. Did we previously discuss moving the `PHP_MONGODB_ADD_SOURCES` out of that include and into this conditional?

If [MONGOCRYPT-218](https://jira.mongodb.org/browse/MONGOCRYPT-218) is now released, perhaps we don't even have to worry about this and can simply include _all_ source files all the time (as we do for libmongoc).

config.m4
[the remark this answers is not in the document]
jmikola: I don't know how portable `gnu99` is, which is why I originally though to suggest `AC_PROG_CC_STDC`. I suppose we can keep this for now and see if any users complain after the release. I would ultimately like to see MONGOCRYPT-219 addressed, but I think we can wait on that vs. other higher priority work.

@alcaeus would you mind leaving a TODO comment above this with a link to MONGOCRYPT-219? Then I think we can resolve this thread.

config.m4
```diff
@@ -273,6 +308,9 @@ if test "$PHP_MONGODB" != "no"; then
     _include([scripts/autotools/libmongoc/Versions.m4])
     _include([scripts/autotools/libmongoc/WeakSymbols.m4])
 
+    _include([scripts/autotools/libmongocrypt/CheckSSL.m4])
```
jmikola: I'd suggest a comment above this to clarify that `PHP_CLIENT_SIDE_ENCRYPTION` will be set to "yes" or "no" after this include (we can ignore the case where it fails). That will preemptively address any question of why we don't bother to consider "auto" after this point.

scripts/autotools/libmongocrypt/Version.m4
```diff
@@ -0,0 +1,6 @@
+if test "$PHP_CLIENT_SIDE_ENCRYPTION" != "no"; then
```
jmikola: Should this be changed to "yes" (as other conditionals in `config.m4`), or is it safer to leave as-is so `Version.m4` does not subtly depend on being included _after_ `CheckSSL.m4`? I'm OK with either, and there's certainly no harm in doing the `AC_SUBST` for both the "yes" and "auto" case if we did happen to include this before `CheckSSL.m4`.

scripts/autotools/libmongocrypt/CheckSSL.m4
jmikola: Compile time error works for me, especially since the default `pecl install` case will just use `auto`.

scripts/autotools/libmongocrypt/CheckSSL.m4
```diff
@@ -0,0 +1,35 @@
...
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO, 1)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_LIBCRYPTO, 1)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_COMMON_CRYPTO, 0)
+
+    PHP_MONGODB_LIBMONGOCRYPT_CFLAGS="$PHP_MONGODB_LIBMONGOCRYPT_CFLAGS -DKMS_MESSAGE_ENABLE_CRYPTO=1 -DKMS_MESSAGE_ENABLE_CRYPTO_LIBCRYPTO=1"
+  elif test "$PHP_CLIENT_SIDE_ENCRYPTION" = "auto"; then
+    PHP_CLIENT_SIDE_ENCRYPTION="no"
+
+    AC_MSG_RESULT(No SSL library found. Compiling without libmongocrypt. Please specify a library using the --with-mongodb-ssl option)
+    AC_SUBST(MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION, 0)
```
jmikola: I think this should be removed, as you will already set this in `config.m4` after this file is included with an if/else condition on `PHP_CLIENT_SIDE_ENCRYPTION`.

scripts/autotools/libmongocrypt/CheckSSL.m4
```diff
@@ -0,0 +1,35 @@
...
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO, 0)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_LIBCRYPTO, 0)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_COMMON_CRYPTO, 0)
```
jmikola: I also think we can remove these `AC_SUBST` calls, as `PHP_CLIENT_SIDE_ENCRYPTION` being "no" means libmongocrypt will be entirely disabled. Once we get back to `config.m4`, we're not going to bother adding its sources, nor will we make other `AC_SUBST` calls that might otherwise be required for a bundled build (e.g. `AC_SUBST(MONGOCRYPT_IS_POSIX, 1)`).

src/libmongocrypt-compat/mongocrypt/mongocrypt.h
```diff
@@ -0,0 +1 @@
+#include "../libmongocrypt/src/mongocrypt.h"
```
jmikola: I know [MONGOCRYPT-226](https://jira.mongodb.org/browse/MONGOCRYPT-226) refers to packaging `mongocrypt-config.h` for the shared library install, but I took a look through this PR and didn't even see an instance where we include it for our bundled builds.

Do we only generate `mongocrypt-config.h` for the benefit of compiling the bundled mongocrypt sources? If we do need to include it for any reason, would it make sense to do so from this compatibility header? If we don't actually need it, then is MONGOCRYPT-226 just so we can start making use of it to report libmongocrypt's own crypto library (independent of libmongoc)?

I assume `MONGOCRYPT_VERSION` may be the only constant we reference from PHPC, and that appears to be defined in `mongocrypt.h`. Perhaps this is not an issue because we generate that by applying substitutions to `mongocrypt.h.in` -- in that case it's just another subtle difference between the libmongoc and libmongocrypt structure (since libmongoc has a separate `mongoc-version.h` file).

scripts/autotools/libmongocrypt/CheckSSL.m4
```diff
@@ -0,0 +1,35 @@
...
+  AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_CNG, 0)
+
+  if test "$PHP_MONGODB_SSL" = "darwin"; then
+    PHP_CLIENT_SIDE_ENCRYPTION="yes"
+
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO, 1)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_LIBCRYPTO, 0)
+    AC_SUBST(MONGOCRYPT_ENABLE_CRYPTO_COMMON_CRYPTO, 1)
+
+    PHP_MONGODB_LIBMONGOCRYPT_CFLAGS="$PHP_MONGODB_LIBMONGOCRYPT_CFLAGS -DKMS_MESSAGE_ENABLE_CRYPTO=1 -DKMS_MESSAGE_ENABLE_CRYPTO_COMMON_CRYPTO=1"
```
jmikola: These lines look new, but perhaps I just didn't notice them in previous iterations of the PR. I suppose these custom defines were added to balance the logic in `kms-message/CMakeLists.txt`.

Autotools does provide an [`AC_DEFINE`](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Defining-Symbols.html) macro. Can we use that instead? Note that the value defaults to `1` if not specified, but the docs suggest not depending on that behavior, so we could do:

```
AC_DEFINE([KMS_MESSAGE_ENABLE_CRYPTO], 1)
AC_DEFINE([KMS_MESSAGE_ENABLE_CRYPTO_COMMON_CRYPTO], 1)
```

...and similar for the case below.

config.w32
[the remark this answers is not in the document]
jmikola: My last comment might pertain to the outstanding task in the PR description.

config.w32
alcaeus: config.m4 already contains these: https://github.com/mongodb/mongo-php-driver/pull/1065/files#diff-788d457a20b110cc38e571dec9ddc68cR395

scripts/autotools/libmongocrypt/Version.m4
alcaeus: I've changed this to yes: we should first check if we can compile with libmongocrypt, then handle all other libmongocrypt tasks.

scripts/autotools/libmongocrypt/CheckSSL.m4
alcaeus: I had previously tried this but it didn't work, but revisiting this the following `AC_DEFINE` worked:
```
    AC_DEFINE([KMS_MESSAGE_ENABLE_CRYPTO], 1, [ ])
    AC_DEFINE([KMS_MESSAGE_ENABLE_CRYPTO_COMMON_CRYPTO], 1, [ ])
```

config.m4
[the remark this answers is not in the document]
alcaeus: I had added this but forgot to move it around when moving the definition of `PHP_MONGODB_LIBMONGOCRYPT_CFLAGS`. This is fixed now.

scripts/autotools/libmongocrypt/CheckSSL.m4
alcaeus: On second try, this still doesn't work: while this creates `#define` instructions in `config.h`, testing FLE causes exceptions due to missing crypto:
> Operation threw unexpected MongoDB\Driver\Exception\BulkWriteException: Bulk write failed due to previous MongoDB\Driver\Exception\RuntimeException: failed to create KMS message

Reverting to `-DKMS_MESSAGE_ENABLE_CRYPTO` fixes this.

src/libmongocrypt-compat/mongocrypt/mongocrypt.h
alcaeus: > Do we only generate `mongocrypt-config.h` for the benefit of compiling the bundled mongocrypt sources? If we do need to include it for any reason, would it make sense to do so from this compatibility header? If we don't actually need it, then is MONGOCRYPT-226 just so we can start making use of it to report libmongocrypt's own crypto library (independent of libmongoc)?

We would only need it to report the crypto library, which is also why we can't rely on a forwarding header since that isn't available when compiling against system libraries. Including mongocrypt-config.h in mongocrypt.h would also solve that issue for us, since we only need to be able to check the `MONGOCRYPT_ENABLE_CRYPT` macros.

config.w32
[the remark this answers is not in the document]
alcaeus: I noticed that we only need to add support for compiling with Secure Channel but can compile with openssl. I'll update the build script to properly account for this, but we need to revisit the `KMS_MESSAGE_ENABLE_CRYPTO` macros.

scripts/autotools/libmongocrypt/CheckSSL.m4
jmikola: Ah, I forgot that `AC_DEFINE` sets stuff through the generated `config.h` file. Looking at `ax_pthread.m4` jogged my memory. In that case, CFLAGS is the way to go.

php_phongo.c
```diff
@@ -3195,6 +3199,17 @@ PHP_MINFO_FUNCTION(mongodb)
 	php_info_print_table_row(2, "libmongoc compression", "disabled");
 #endif
 
+#ifdef MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION
+#ifdef HAVE_SYSTEM_LIBMONGOCRYPT
+	php_info_print_table_row(2, "libmongocrypt headers version", MONGOCRYPT_VERSION);
+	php_info_print_table_row(2, "libmongocrypt library version", mongocrypt_version());
+#else
+	php_info_print_table_row(2, "libmongocrypt bundled version", MONGOCRYPT_VERSION);
+#endif
```
jmikola: Feel free to add libmongocrypt-specific crypto info here once you bump to 1.0.1 for MONGOCRYPT-226.

config.m4
```diff
@@ -342,6 +386,32 @@ if test "$PHP_MONGODB" != "no"; then
       PHP_MONGODB_ADD_BUILD_DIR([src/libmongoc/src/zlib-1.2.11/])
       AC_CONFIG_FILES([${ac_config_dir}/src/libmongoc/src/zlib-1.2.11/zconf.h])
     fi
+
+    if test "$PHP_CLIENT_SIDE_ENCRYPTION" = "yes"; then
+      AC_SUBST(MONGOCRYPT_ENABLE_TRACE, 1)
+      AC_SUBST(MONGOCRYPT_IS_POSIX, 1)
+      AC_SUBST(MONGOCRYPT_IS_WIN, 0)
```
jmikola: Per my [comment in MONGOCRYPT-218](https://jira.mongodb.org/browse/MONGOCRYPT-218?focusedCommentId=2653755&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2653755), I think these are obsolete. It looks like we can always compile the `os_posix/` and `os_win` sources.

config.w32
```diff
@@ -224,6 +224,52 @@ if (PHP_MONGODB != "no") {
...
+    mongoc_opts.MONGOCRYPT_ENABLE_CRYPTO = 1;
+    mongoc_opts.MONGOCRYPT_ENABLE_CRYPTO_LIBCRYPTO = 1;
+
+    mongodb_generate_header(
+      configure_module_dirname + "/src/libmongocrypt/src/mongocrypt-config.h.in",
+      configure_module_dirname + "/src/libmongocrypt/src/mongocrypt-config.h",
+      mongocrypt_opts
+    );
+
+    mongodb_generate_header(
```
jmikola: As-is, this does not replace `@MONGOCRYPT_BUILD_VERSION@`. I'll follow up with a fix for that in a PR against your branch.

config.w32
```diff
@@ -224,6 +224,52 @@ if (PHP_MONGODB != "no") {
...
+    PHP_MONGODB_CFLAGS += "\
+      /I" + configure_module_dirname + "/src/libmongocrypt/src \
+      /I" + configure_module_dirname + "/src/libmongocrypt/kms-message/src \
+      /I" + configure_module_dirname + "/src/libmongocrypt-compat \
+    ";
+
+    var mongocrypt_opts = {
+      MONGOCRYPT_ENABLE_TRACE: 1,
+      MONGOCRYPT_IS_POSIX: 0,
+      MONGOCRYPT_IS_WIN: 1,
```
jmikola: Related to my earlier comment, `IS_POSIX` and `IS_WIN` may be redundant.

config.w32
```diff
@@ -224,6 +224,52 @@ if (PHP_MONGODB != "no") {
...
   }
 
+  if (PHP_CLIENT_SIDE_ENCRYPTION != "no" && mongoc_ssl_found) {
+    mongoc_opts.MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION = 1;
+
+    PHP_MONGODB_CFLAGS += "\
+      /I" + configure_module_dirname + "/src/libmongocrypt/src \
+      /I" + configure_module_dirname + "/src/libmongocrypt/kms-message/src \
+      /I" + configure_module_dirname + "/src/libmongocrypt-compat \
+    ";
```
jmikola: `PHP_MONGODB_CFLAGS` is already passed to `EXTENSION()` by this point, so I think we need to use `ADD_FLAG` to set these. Additionally, we'll want to define the KMS constants.

config.w32
```diff
@@ -224,6 +224,52 @@ if (PHP_MONGODB != "no") {
...
+      MONGOCRYPT_ENABLE_CRYPTO: 0,
+      MONGOCRYPT_ENABLE_CRYPTO_LIBCRYPTO: 0,
+      // TODO: Support building with Secure Channel on Windows
+      MONGOCRYPT_ENABLE_CRYPTO_CNG: 0,
+      // Secure Transport does not apply to Windows
+      MONGOCRYPT_ENABLE_CRYPTO_COMMON_CRYPTO: 0
+    };
+
+    // Generated with: find src/libmongocrypt/src -name '*.c' -print0 | cut -sz -d / -f 4- | sort -dz | tr '\000' ' '
+    var PHP_MONGODB_MONGOCRYPT_SOURCES="crypto/cng.c crypto/commoncrypto.c crypto/libcrypto.c crypto/none.c mongocrypt-binary.c mongocrypt-buffer.c mongocrypt.c mongocrypt-cache.c mongocrypt-cache-collinfo.c mongocrypt-cache-key.c mongocrypt-ciphertext.c mongocrypt-crypto.c mongocrypt-ctx.c mongocrypt-ctx-datakey.c mongocrypt-ctx-decrypt.c mongocrypt-ctx-encrypt.c mongocrypt-key-broker.c mongocrypt-key.c mongocrypt-kms-ctx.c mongocrypt-log.c mongocrypt-marking.c mongocrypt-opts.c mongocrypt-status.c mongocrypt-traverse-util.c os_posix/os_mutex.c os_posix/os_once.c os_win/os_mutex.c os_win/os_once.c";
```
jmikola: I forgot that `ADD_SOURCES` doesn't like paths in the list of source files. This apparently isn't a problem for Autotools, but it causes linking problems on Windows. Since libmongocrypt differs from libmongoc in that it nests source code within directories parallel to other sources, we may need to modify the `find` command here to not recurse.

I'm inclined to just use `-maxdepth 1` instead of pursuing a [POSIX-compliant solution](https://unix.stackexchange.com/questions/275637/limit-posix-find-to-specific-depth), since it's macOS-friendly and is likely to work on both of our development machines.

config.w32
```diff
@@ -224,6 +224,52 @@ if (PHP_MONGODB != "no") {
...
+      /I" + configure_module_dirname + "/src/libmongocrypt/kms-message/src \
+      /I" + configure_module_dirname + "/src/libmongocrypt-compat \
+    ";
+
+    var mongocrypt_opts = {
+      MONGOCRYPT_ENABLE_TRACE: 1,
+      MONGOCRYPT_IS_POSIX: 0,
+      MONGOCRYPT_IS_WIN: 1,
+      MONGOCRYPT_ENABLE_CRYPTO: 0,
+      MONGOCRYPT_ENABLE_CRYPTO_LIBCRYPTO: 0,
```
jmikola: `CRYPTO` and `CRYPTO_LIBCRYPTO` can be defined to `1` here, since we know we'll be using OpenSSL at this point. This may address your earlier comment where you speculated that Windows was building with no crypto.

config.w32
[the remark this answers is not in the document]
jmikola: Two more comments here.

First, "auto" doesn't work it does in Autotools. PHP's Windows build utils expect these options to be booleans, so I think we'll just need to default it to "yes" and treat it as "auto", just as we currently do with SASL (i.e. only warn if we can't build with it).

Secondly, I realized that our suggested build command in `CONTRIBUTING.md` meant that we never actually build with SASL and would also ignore FLE:

```
$ configure --disable-all --with-openssl --enable-cli --enable-json --enable-mongodb=shared
```

The `--disable-all` turns these off by default, so I'll make sure to revise the doc to explicitly enable them. I think it still makes sense to disable everything else as it significantly speeds up the build.

jmikola:
Incoming PR: https://github.com/alcaeus/mongo-php-driver/pull/3

src/libmongocrypt-compat/mongocrypt/mongocrypt.h
alcaeus: I've updated to the latest master version and bumped the required version in config.m4. We'll wait for a release of 1.0.1 before releasing 1.7.0 👍

config.w32
```diff
@@ -224,6 +224,52 @@ if (PHP_MONGODB != "no") {
...
+
+    // Generated with: find src/libmongocrypt/kms-message/src -name '*.c' -print0 | cut -sz -d / -f 5- | sort -dz | tr '\000' ' '
+    var PHP_MONGODB_MONGOCRYPT_KMS_MESSAGE_SOURCES="hexlify.c kms_b64.c kms_caller_identity_request.c kms_crypto_apple.c kms_crypto_libcrypto.c kms_crypto_none.c kms_crypto_windows.c kms_decrypt_request.c kms_encrypt_request.c kms_kv_list.c kms_message.c kms_request.c kms_request_opt.c kms_request_str.c kms_response.c kms_response_parser.c sort.c";
+
+    ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/src", PHP_MONGODB_MONGOCRYPT_SOURCES, "mongodb");
+    ADD_SOURCES(configure_module_dirname + "/src/libmongocrypt/kms-message/src", PHP_MONGODB_MONGOCRYPT_KMS_MESSAGE_SOURCES, "mongodb");
```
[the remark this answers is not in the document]
alcaeus: That's correct, I referred to that in the comment above since I didn't get to it before. I saw you added it in your PR, so this should be resolved now.

php_phongo.c
alcaeus: Restored previous crypto output 👍

config.m4
alcaeus: I believe so, I forgot to drop these when I updated to the latest libmongocrypt version.

config.w32
alcaeus: An alternative to using `ADD_FLAG` is moving the `EXTENSION` call to the bottom of the file. I've seen this in another extension, but I haven't tested it. Calling `ADD_FLAG` seems to work at this time, so I believe this works fine for us.

config.w32
alcaeus: Ah, I didn't know that. Since I'm still unable to compile on Windows I didn't catch that. Using `-maxdepth 1` for `find` works fine on my machine, I used this previously when I excluded the crypto sources to conditionally include them depending on the crypto library chosen.

config.w32
alcaeus: That is correct. We'll have to revise this when adding support for Secure Channel on Windows, but until then this works fine.

config.m4
```diff
@@ -181,6 +181,12 @@ if test "$PHP_MONGODB" != "no"; then
                                [MongoDB: Use system libmongoc [default=no]])],
                [no],
                [no])
+  PHP_ARG_WITH([mongodb-client-side-encryption],
+                 [whether to enable client-side encryption],
+                 [AS_HELP_STRING([--with-mongodb-client-side-encryption=@<:@auto/yes/no@:>@],
+                                 [MongoDB: Enable client-side encryption [default=auto]])],
+                 [auto],
+                 [no])
```
jmikola: These five can be re-aligned before merging. I forgot to do this when switching to `PHP_ARG_WITH`.

alcaeus:
Done. I also squashed commits for a cleaner history 👍

PR #1067, opened 2019-12-04
PHPC-1503: Fix config.w32 for PHP 7.4.0

cmb69:
The VS version evaluation has been reworked[1] for PHP 7.4.0, which
includes the removal of the VC_VERSIONS array.  To avoid the JScript
error, we check whether VC_VERSIONS is defined, and otherwise fall back
to COMPILER_NAME_LONG.

[1] <http://git.php.net/?p=php-src.git;a=commit;h=6a624c1dfda0dbfaaff3e453e6cb58de12748fb3>

### DIFF
--- a/config.w32
+++ b/config.w32
@@ -248,8 +248,10 @@ if (PHP_MONGODB != "no") {
 
   if (typeof COMPILER_NAME === 'string') {
     mongoc_opts.MONGOC_CC = COMPILER_NAME;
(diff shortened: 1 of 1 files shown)


jmikola:
Cherry-picked into #1068 and merged in https://github.com/mongodb/mongo-php-driver/commit/06e45f66c9dcf8d15b925ef943775f9323b0f070.

PR #1071, opened 2019-12-06
Parse options for automatic encryption using libmongocrypt

alcaeus:
This PR covers [PHPC-1498](https://jira.mongodb.org/browse/PHPC-1498) and [PHPC-1500](https://jira.mongodb.org/browse/PHPC-1500). For now, it includes the commit of #1065 ([PHPC-1496](https://jira.mongodb.org/browse/PHPC-1496)).

The test for `autoEncryption` driver options is a dummy that ensures that the settings can be parsed. It does not include any skips since we don't have tests set up where we compile without client-side encryption support. Since we don't expose any encryption settings yet, we just ensure that the manager can be created successfully.


php_phongo.c
```diff
@@ -2503,6 +2507,64 @@ int php_phongo_set_monitoring_callbacks(mongoc_client_t* client)
...
+	*free = true;
+
+	manager = Z_MANAGER_OBJ_P(keyVaultClient);
+
+#if PHP_VERSION_ID >= 70000
+	driverOptionsClone = ecalloc(sizeof(zval), 1);
+	autoEncryptionOptsClone = &stackAutoEncryptionOptsClone;
+#else
+	ALLOC_INIT_ZVAL(driverOptionsClone);
+	MAKE_STD_ZVAL(autoEncryptionOptsClone);
```
alcaeus: @jmikola: Either this or the previous line cause a significant number of leaks on PHP 5. I'll investigate this separately once I manage to get PHP 5.6 running on my machine.

```
[2019/12/06 15:51:52.924] 051+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 052+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_hash.c(320) :  Freeing 0x7F81A23D3070 (64 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 053+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_alloc.c(2592) : Actual location (location was relayed)
[2019/12/06 15:51:52.924] 054+ Last leak repeated 21 times
[2019/12/06 15:51:52.924] 055+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 056+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_hash.c(346) :  Freeing 0x7F81A23D3108 (72 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 057+ Last leak repeated 25 times
[2019/12/06 15:51:52.924] 058+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 059+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_vm_execute.h(4077) :  Freeing 0x7F81A23D31A8 (32 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 060+ Last leak repeated 7 times
[2019/12/06 15:51:52.924] 061+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 062+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_vm_execute.h(9481) :  Freeing 0x7F81A23D32C0 (72 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 063+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_API.c(1011) : Actual location (location was relayed)
[2019/12/06 15:51:52.924] 064+ Last leak repeated 9 times
[2019/12/06 15:51:52.924] 065+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 066+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_vm_execute.h(9414) :  Freeing 0x7F81A23D3360 (32 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 067+ Last leak repeated 16 times
[2019/12/06 15:51:52.924] 068+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 069+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_vm_execute.h(11818) :  Freeing 0x7F81A23D3510 (72 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 070+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_API.c(1011) : Actual location (location was relayed)
[2019/12/06 15:51:52.924] 071+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 072+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_vm_execute.h(11751) :  Freeing 0x7F81A23D35B0 (32 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 073+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 074+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_hash.c(392) :  Freeing 0x7F81A23D3628 (64 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 075+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_alloc.c(2592) : Actual location (location was relayed)
[2019/12/06 15:51:52.924] 076+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 077+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_vm_execute.h(11802) :  Freeing 0x7F81A23D36C0 (72 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 078+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_hash.c(419) : Actual location (location was relayed)
[2019/12/06 15:51:52.924] 079+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 080+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_vm_execute.h(4138) :  Freeing 0x7F81A23D4BF0 (72 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 081+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_API.c(1011) : Actual location (location was relayed)
[2019/12/06 15:51:52.924] 082+ Last leak repeated 3 times
[2019/12/06 15:51:52.924] 083+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 084+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_API.c(314) :  Freeing 0x7F81A23D5DE8 (32 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 085+ Last leak repeated 13 times
[2019/12/06 15:51:52.924] 086+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 087+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_vm_execute.h(29216) :  Freeing 0x7F81A25A06D0 (72 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 088+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_API.c(1011) : Actual location (location was relayed)
[2019/12/06 15:51:52.924] 089+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 090+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_vm_execute.h(944) :  Freeing 0x7F81A25A28D8 (32 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 091+ [Fri Dec  6 20:51:52 2019]  Script:  '/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php'
[2019/12/06 15:51:52.924] 092+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_vm_execute.h(16156) :  Freeing 0x7F81A25A2A40 (72 bytes), script=/data/mci/0bd12b2c9b9c1e72c0226b4a04828642/src/tests/manager/manager-ctor-auto_encryption-001.php
[2019/12/06 15:51:52.924] 093+ /data/mci/0f02f769a25a722af572a3a28e934dc8/src/php-5.6.38/Zend/zend_API.c(1011) : Actual location (location was relayed)
[2019/12/06 15:51:52.924] 094+ === Total 108 memory leaks detected ===
```

alcaeus: Fixed - a tiny `Z_ADDREF_P` slipped through the cracks and was having a field day in memory. It is no more.

php_phongo.c
```diff
@@ -2550,6 +2610,11 @@ static char* php_phongo_manager_make_client_hash(const char* uri_string, zval* o
 		hash      = pestrndup(ZSTR_VAL(var_buf.s), *hash_len, 1);
 	}
 
+	if (serializable_driver_options && free_driver_options) {
+		zval_ptr_dtor(serializable_driver_options);
```
jmikola: I think this may be incorrect for PHP 5.x, as `zval_ptr_dtor()` expects a `zval**` even though it immediately dereferences it to a `zval*` when calling `i_zval_ptr_dtor()` (see: [here](http://lxr.jpauli.tech/source/xref/php-5.6/Zend/zend_execute.c#96)).

That said, I'm curious if we even need to do this since a non-null `serializable_driver_options` means we've added it to `args`, which itself is going to be passed to `zval_ptr_dtor()` later on. As-is, this may be a double-dtor (I'm not sure how that compares to a double-free).

Of course, we still need to call `efree()` to clean up the zval allocation within `php_phongo_manager_prepare_manager_for_hash`, but I think that may be something to do _after_ the call to `zval_ptr_dtor(&args);`. Does that make sense?

It may be helpful to look at `php_phongo_prep_legacy_option()` and `php_phongo_prep_legacy_option_free()` for prior art, since they do something similar.

php_phongo.c
```diff
@@ -2721,6 +2721,129 @@ static mongoc_client_t* php_phongo_find_client(const char* hash, size_t hash_len
...
+
+	if (php_array_existsc(zAutoEncryptionOpts, "bypassAutoEncryption")) {
+		zend_bool bypass_auto_encryption = php_array_fetch_bool(zAutoEncryptionOpts, "bypassAutoEncryption");
+
+		mongoc_auto_encryption_opts_set_bypass_auto_encryption(auto_encryption_opts, bypass_auto_encryption);
+	}
+
+	if (php_array_existsc(zAutoEncryptionOpts, "extraOptions")) {
+		zval   *extra_options = php_array_fetch(zAutoEncryptionOpts, "extraOptions");
+		bson_t bson_options = BSON_INITIALIZER;
```
jmikola: Alignment looks off here, so make sure you run `clang-format` on the final PR.

php_phongo.c
```diff
@@ -2799,6 +2922,13 @@ void phongo_manager_init(php_phongo_manager_t* manager, const char* uri_string,
 	}
 #endif
 
+#ifdef MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION
```
jmikola: Did we previously discuss throwing if `autoEncryption` was specified but FLE was not enabled? I believe libmongoc has logic in `mongoc_client_enable_auto_encryption()` to raise an error in that case, but perhaps that wasn't implemented when you first drafted this PR. Assuming `mongoc_client_enable_auto_encryption()` is always available, perhaps we just need to remove the `#ifdef` checks here and around `phongo_manager_set_auto_encryption_opts`.

I've acknowledged your earlier comment that we don't have an environment to actually test for such an error, since libmongocrypt is always compiled in.

phongo_compat.h
```diff
@@ -169,6 +169,14 @@
 		Z_ARRVAL_P(__z)  = __arr;    \
 		Z_TYPE_P(__z)    = IS_ARRAY; \
 	} while (0);
+#define ZVAL_DUP(z, v)        \
```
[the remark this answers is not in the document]
alcaeus: > Did you create this macro from scratch based on the PHP 7.x definition?

That's correct, I used the [current source](https://github.com/php/php-src/blob/eaa93f6b8a1f35173b0d77b573428ebc9238ef42/Zend/zend_types.h#L1117) for inspiration.

The reason for this is that while the original zval is separated, the `autoEncryption` zval is not, so modifying any members in that array would modify them persistently. Duplicating the entire ZVAL solves that problem.

php_phongo.c
alcaeus: #1080 fixes existing mistakes. All newly introduced style errors have been fixed in this PR.

php_phongo.c
alcaeus: Now that you mention it, we did indeed discuss throwing an exception in this case, since quietly disabling FLE might pose a security risk. While libmongoc would fail gracefully, the resulting error message could be confusing for users:
> libmongoc is not built with support for Client-Side Field Level Encryption. Configure with ENABLE_CLIENT_SIDE_ENCRYPTION=ON.

Given that, we may opt to throw our own error to inform users that they need to compile with the `mongodb-client-side-encryption` flag.

src/MongoDB/Manager.c
```diff
@@ -841,6 +841,10 @@ static void php_phongo_manager_free_object(phongo_free_object_arg* object TSRMLS
 		intern->client = NULL;
 	}
 
+	if (intern->client_hash) {
+		pefree(intern->client_hash, 1);
```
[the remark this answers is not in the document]
alcaeus: Changed this to no longer create persistent strings. Your assumption about hash keys being copied is correct: looking at `php_phongo_persist_client`, we persist the client using `zend_hash_str_update_ptr`, where we end up in `_zend_hash_str_add_or_update_i`. The key [is initialised using `zend_string_init`](https://github.com/php/php-src/blob/eaa93f6b8a1f35173b0d77b573428ebc9238ef42/Zend/zend_hash.c#L847), which [copies the original `char*` using `memcpy`](https://github.com/php/php-src/blob/eaa93f6b8a1f35173b0d77b573428ebc9238ef42/Zend/zend_string.h#L157). Thus, a process-scoped string shouldn't be necessary, even though we use it as key for persisting the client object.

php_phongo.c
alcaeus: > I think this may be incorrect for PHP 5.x, as `zval_ptr_dtor()` expects a `zval**` even though it immediately dereferences it to a `zval*` when calling `i_zval_ptr_dtor()` (see: [here](http://lxr.jpauli.tech/source/xref/php-5.6/Zend/zend_execute.c#96)).

The code in question is in a conditional for PHP 7. The following code block that covers PHP 5 passes the expected `zval**`:
```
	if (serializable_driver_options && free_driver_options) {
		zval_ptr_dtor(&serializable_driver_options);
	}
```

> That said, I'm curious if we even need to do this since a non-null `serializable_driver_options` means we've added it to `args`, which itself is going to be passed to `zval_ptr_dtor()` later on. As-is, this may be a double-dtor (I'm not sure how that compares to a double-free).
> 
> Of course, we still need to call `efree()` to clean up the zval allocation within `php_phongo_manager_prepare_manager_for_hash`, but I think that may be something to do _after_ the call to `zval_ptr_dtor(&args);`. Does that make sense?

Great catch, that's absolutely correct. If we duplicate the driver options, the new object will be freed when `zval_ptr_dtor(&args)` recurses into nested zvals. I didn't quite make the entire connection, but I did add a `Z_ADDREF_P` call for the case where we don't need to duplicate the driverOptions and append the original zval to `args`. I've changed the code to no longer call `zval_ptr_dtor(&serializable_driver_options)`, but only `efree` it after calling `zval_ptr_dtor(&args)`. For PHP 5, this means we don't need to clean up anything since we only called `zval_ptr_dtor` in the first place.

> It may be helpful to look at `php_phongo_prep_legacy_option()` and `php_phongo_prep_legacy_option_free()` for prior art, since they do something similar.

I originally looked at that implementation, but misinterpreted the handling of `options` vs. `args` in these two methods. Your comment helped clear up that confusion.

tests/manager/manager-ctor-auto_encryption-error-001.phpt
```diff
@@ -0,0 +1,29 @@
...
+        $manager = new MongoDB\Driver\Manager(null, [], ['autoEncryption' => $driverOptions]);
+    }, 'MongoDB\Driver\Exception\RuntimeException'), "\n\n";
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+OK: Got MongoDB\Driver\Exception\RuntimeException
+Key vault namespace option required
```
jmikola: I assume these error messages originate from libmongocrypt or libmongoc, as I don't see them in this PR. It's unfortunate that they don't actually tell users what option key is missing, but I suppose there's no way around that. I'm not going to propose that we parse out these error messages and rewrite them :stuck_out_tongue:

tests/manager/manager-ctor-auto_encryption-error-001.phpt
```diff
@@ -0,0 +1,29 @@
...
+foreach ($tests as $driverOptions) {
+    echo throws(function() use ($driverOptions) {
+        $manager = new MongoDB\Driver\Manager(null, [], ['autoEncryption' => $driverOptions]);
+    }, 'MongoDB\Driver\Exception\RuntimeException'), "\n\n";
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
```
jmikola: No patterns are used so this can just be `--EXPECT--`.

tests/manager/manager-ctor-auto_encryption-001.phpt
```diff
@@ -0,0 +1,83 @@
...
+            'bsonType' => 'object',
+        ],
+    ]],
+    ['bypassAutoEncryption' => true],
+    ['extraOptions' => ['mongocryptdBypassSpawn' => true]],
+];
+
+foreach ($tests as $driverOptions) {
+    $manager = new MongoDB\Driver\Manager(null, [], ['autoEncryption' => $driverOptions + $baseOptions]);
+    var_dump($manager);
```
jmikola: I'd propose we remove this and change the expected output block to:

```
--EXPECT--
===DONE===
```

Looking at the current `var_dump()` output, there's nothing relevant to the `autoEncryption` actually being applied, so I think all we're testing for here is that an exception hasn't been thrown. We do have other tests that do the same (i.e. no output expected to demonstrate the code runs successfully).

tests/manager/manager-ctor-auto_encryption-001.phpt
```diff
@@ -0,0 +1,83 @@
...
+                ],
+            ],
+            'bsonType' => 'object',
+        ],
+    ]],
+    ['bypassAutoEncryption' => true],
+    ['extraOptions' => ['mongocryptdBypassSpawn' => true]],
+];
+
+foreach ($tests as $driverOptions) {
```
jmikola: I think `$autoEncryptionOptions` would be a better name, as the values in `$tests` above are all options that go _within_ the `autoEncryption` key.

`$driverOptions` would be the third argument to the Manager constructor, so that name would make more sense if each value in `$tests` started with an `autoEncryption` key.

tests/manager/manager-ctor-auto_encryption-001.phpt
```diff
@@ -0,0 +1,83 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): auto encryption options
+--FILE--
```
jmikola: If `MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION` is undefined, `phongo_manager_set_auto_encryption_opts()` throws the following if it finds a `autoEncryption` key in a non-null `driverOptions`:

```c
phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Cannot enable automatic field-level encryption. Please recompile with support for libmongocrypt using the with-mongodb-client-side-encryption configure switch.");
```

This test will fail when FLE is not enabled. I propose you create a `skip_if_not_libmongocrypt()` function based on `skip_if_not_libmongoc_crypto()` and `skip_if_not_libmongoc_ssl()` in `tests/utils/skipif.php`. It will be simpler than those, as you won't need logic to check for specific libs.

tests/manager/manager-ctor-auto_encryption-error-001.phpt
```diff
@@ -0,0 +1,29 @@
+--TEST--
+MongoDB\Driver\Manager::__construct(): incomplete auto encryption options
```
jmikola: This would also benefit from a `--SKIPIF--` that calls `skip_if_not_libmongocrypt()`.

I'll defer to you if you'd like to create a separate test for the `phongo_manager_set_auto_encryption_opts()` exception message when FLE is not available. If so, that would warrant creating another function: `skip_if_libmongocrypt()`.

tests/manager/manager-ctor-auto_encryption-error-001.phpt
alcaeus: Yes, those are errors from libmongoc. If we want to make those error messages more meaningful, we need to start checking error codes or something, which is a bigger undertaking.

tests/manager/manager-ctor-auto_encryption-001.phpt
alcaeus: Added the function, also added the counterpart and an additional test for when compiling without FLE as mentioned in the comment below.

alcaeus:
@jmikola just FYI, the session/bug1274-002 test failed on the SHARDED_CLUSTER_RS deployment but passed on a subsequent run. We may want to keep an eye on this one.

jmikola:
Opened [PHPC-1526](https://jira.mongodb.org/browse/PHPC-1526) to deal with intermittent test failures. Thanks for catching that!

PR #1075, opened 2019-12-20
PHPC-1499: Add clientEncryption API

alcaeus:
https://jira.mongodb.org/browse/PHPC-1499


alcaeus:
@jmikola PR rebased to bring in changes from #1071. I've also added a build to test the driver compiled without support FLE - this helped uncover some missing symbols I had in this PR before. This is good to review now 👍

php_phongo.c
```diff
@@ -2841,18 +2841,379 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
...
+		char*     coll_name;
+		int       plen;
+		zend_bool pfree;
+
+		keyvault_namespace = php_array_fetchc_string(options, "keyVaultNamespace", &plen, &pfree);
+
+		if (!phongo_split_namespace(keyvault_namespace, &db_name, &coll_name)) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected \"keyVaultNamespace\" encryption option to contain a full collection name");
+
+			goto cleanup;
```
jmikola: If we end up jumping to `cleanup`, might we still need to conditionally call `efree(keyvault_namespace)` if `pfree` is `true`?

I checked, and you did handle that case in `phongo_manager_set_auto_encryption_opts`. Perhaps this was written before I caught that in #1071, though. Also, is there a particular reason you used `str_efree` in the auto opts, but `efree` here? Our own `phongo_compat.h` defines it as the same for PHP 7, but the [original function for PHP 5.x](http://lxr.jpauli.tech/source/xref/php-5.6/Zend/zend_string.h#53) has a check for interned strings which might be important. It looks like we use `str_efree` a few other places in the driver in a similar fashion, so that may be correct but I couldn't tell you why off the top of my head.

While looking into this, I think I found a few places where we might just be using `efree()` on something returned from `php_array_fetch_string`. `php_phongo_fetch_ssl_opt_string` may be one instance where we tend to use `str_efree` but may use `efree` directly at least once. Would you mind opening a ticket to investigate this? It's likely we won't even get to it if we drop 5.6 support first, but if so we could always close it accordingly.

php_phongo.c
```diff
@@ -2841,18 +2841,379 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
...
+	return NULL;
+} /* }}} */
+
+void phongo_clientencryption_init(php_phongo_clientencryption_t* clientencryption, mongoc_client_t* client, zval* options TSRMLS_DC) /* {{{ */
+{
+	mongoc_client_encryption_t*      ce;
+	mongoc_client_encryption_opts_t* opts;
+	bson_error_t                     error = { 0 };
+
+	opts = phongo_clientencryption_opts_from_zval(client, options TSRMLS_CC);
```
jmikola: `phongo_clientencryption_opts_from_zval` returns a newly allocated struct even if `options` is empty. If this ever returns `null`, doesn't that mean an exception was thrown? In that case, should we return immediately instead of calling `mongoc_client_encryption_new` and potentially stacking exceptions?

php_phongo.c
```diff
@@ -2841,18 +2841,379 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
...
+		uint32_t   i      = 0;
+		bool       failed = false;
+
+		if (!zkeyaltnames || Z_TYPE_P(zkeyaltnames) != IS_ARRAY) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected keyAltNames to be array, %s given", PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(zkeyaltnames));
+			goto cleanup;
+		}
+
+		ht_data           = HASH_OF(zkeyaltnames);
+		keyaltnames_count = ht_data ? zend_hash_num_elements(ht_data) : 0;
```
jmikola: Should we throw if `ht_data` is null? I'm curious if this was copied from some other iteration code in the driver.

php_phongo.c
```diff
@@ -2841,18 +2841,379 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
...
+		keyaltnames       = bson_malloc(sizeof(char*) * keyaltnames_count);
+
+#if PHP_VERSION_ID >= 70000
+		{
+			zval* keyaltname;
+
+			ZEND_HASH_FOREACH_VAL(ht_data, keyaltname)
+			{
+				if (i == keyaltnames_count) {
+					phongo_throw_exception(PHONGO_ERROR_LOGIC TSRMLS_CC, "Iterating over too many keyAltNames. Please file a bug report");
```
jmikola: How would this ever happen in practice if `keyaltnames_count` is initialized to the size of the PHP array? I know we don't use `assert()` in the extension but if we did is that what you might have used here?

php_phongo.c
```diff
@@ -2841,18 +2841,379 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
...
+					failed = true;
+				}
+
+				keyaltnames[i] = bson_strdup(Z_STRVAL_P(keyaltname));
```
jmikola: I think it'd make sense to do a type check on `keyaltname` at the start of the `ZEND_HASH_FOREACH_VAL` scope. If we find a non-string in the array, then we can raise a consistently formatted type exception.

In practice, this would probably be a non-associative array, but if you want to generate the most helpful exception message we may want to use `ZEND_HASH_FOREACH_KEY_VAL` so we can report the key. I believe both it makes the string and numeric key easily accessible. Take a look at `php_phongo_bson_state_parse_fieldpaths` for prior art if you go this route. Alternatively, I'm happy to punt on improving the error message to a future ticket after 1.7.0.

php_phongo.c
```diff
@@ -2841,18 +2841,379 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
...
+			HashPosition pos;
+			zval**       keyaltname;
+			bool         failed = false;
+
+			for (
+				zend_hash_internal_pointer_reset_ex(ht_data, &pos);
+				zend_hash_get_current_data_ex(ht_data, (void**) &keyaltname, &pos) == SUCCESS;
+				zend_hash_move_forward_ex(ht_data, &pos)) {
+
+				if (i == keyaltnames_count) {
```
jmikola: All of the earlier comments for the PHP 7 code apply here.

php_phongo.c
```diff
@@ -2841,18 +2841,379 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
...
+		bool       failed = false;
+
+		if (!zkeyaltnames || Z_TYPE_P(zkeyaltnames) != IS_ARRAY) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected keyAltNames to be array, %s given", PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(zkeyaltnames));
+			goto cleanup;
+		}
+
+		ht_data           = HASH_OF(zkeyaltnames);
+		keyaltnames_count = ht_data ? zend_hash_num_elements(ht_data) : 0;
+		keyaltnames       = bson_malloc(sizeof(char*) * keyaltnames_count);
```
jmikola: I think we can use `emalloc` here. I don't think any extension code calls `bson_malloc`. If anything, we use vtables to configure libbson to point _its_ memory functions to `emalloc` (and friends).

Edit: `ecalloc` would actually be safer than `emalloc`, to ensure the list of pointers if initialized to zero. If iteration were to get interrupted for some reason, we then wouldn't have to worry about freeing a potentially non-zero garbage value when we iterate through this again to clean up.

php_phongo.c
```diff
@@ -2841,18 +2841,379 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
...
+			}
+		}
+#endif
+
+		if (!failed) {
+			mongoc_client_encryption_datakey_opts_set_keyaltnames(opts, keyaltnames, keyaltnames_count);
+		}
+
+		for (i = 0; i < keyaltnames_count; i++) {
+			bson_free(keyaltnames[i]);
```
jmikola: Per my earlier comment, this would become `efree`.

If `failed` is `true` and iteration was aborted, I don't think we want to iterate through to `keyaltnames_count`. It should be sufficient to do something like `for (j=0; j < i; j++)` and only free pointers that we knowingly allocated.

That said, I believe `free(null)` is a no-op; however, this goes back to my earlier suggestion of using `ecalloc` instead of `emalloc` to ensure all pointers are initially zero before we start iterating through `ht_data`. Having this `for` loop only iterate through to `i` side-steps this concern entirely, though.

php_phongo.c
```diff
@@ -2841,18 +2841,379 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
...
+		goto cleanup;
+	}
+
+	if (!mongoc_client_encryption_create_datakey(clientencryption->client_encryption, kms_provider, opts, &keyid, &error)) {
+		phongo_throw_exception_from_bson_error_t(&error TSRMLS_CC);
+		goto cleanup;
+	}
+
+	if (!php_phongo_bson_value_to_zval(&keyid, return_value)) {
+		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE TSRMLS_CC, "Could not convert BSON value to ZVAL");
```
jmikola: Is this exception necessary? Looking at `php_phongo_bson_value_to_zval`, it's only error condition is when `php_phongo_bson_to_zval_ex` itself fails and in that case an exception will have been thrown. This code would end up stacking another exception on top of that and may hide the more relevant error message.

php_phongo.c
```diff
@@ -2841,18 +2841,379 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
...
+
+cleanup:
+	if (opts) {
+		mongoc_client_encryption_datakey_opts_destroy(opts);
+	}
+
+	return NULL;
+} /* }}} */
+
+void phongo_clientencryption_create_datakey(php_phongo_clientencryption_t* clientencryption, zval* return_value, char* kms_provider, zval* options TSRMLS_DC) /* {{{ */
```
jmikola: AFAIK, this function is only called from a method in `phongo_clientencryption_create_datakey`. Perhaps that's the same case for a few other functions (e.g. encrypt, decrypt, and their options helpers). Would it make sense to define these as static functions in ClientEncryption.c?

Perhaps the ClientEncryption class itself doesn't even need to exist (at runtime) unless FLE is enabled. That might be feasible if you wrap most of ClientEncryption.c itself in an `#ifdef MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION` (after including the necessary config header) and do the same for the `php_phongo_clientencryption_init_ce` call in `PHP_MINIT_FUNCTION(mongodb)` later in this file.

php_phongo_structs.h
```diff
@@ -43,6 +43,12 @@ typedef struct {
 	PHONGO_ZEND_OBJECT_POST
 } php_phongo_bulkwrite_t;
 
+typedef struct {
+	PHONGO_ZEND_OBJECT_PRE
+	mongoc_client_encryption_t* client_encryption;
```
jmikola: Sorry to bring this up, but I think we may need to consider how forking might affect FLE. Consider that this has a reference to two clients (its own and the key vault), and internal operations with the key vault client don't use our API that might conditionally call `mongoc_client_reset` when needed. I'm not sure how often the key vault client is used over the lifetime of this object, but perhaps we'll need to consider PID checks like we do for cursors and sessions when certain methods are called or this is destroyed.

There's probably a more general concern about auto encryption if the key vault is used over the lifetime of an application, too. I'm not going to insist we sort this out now (or even in 1.7.0), but would you mind opening a ticket and referring back to our FLE epic and [PHPC-1274](https://jira.mongodb.org/browse/PHPC-1274)?

src/MongoDB/ClientEncryption.c
```diff
@@ -0,0 +1,205 @@
...
+		zend_restore_error_handling(&error_handling TSRMLS_CC);
+		return;
+	}
+
+	zend_restore_error_handling(&error_handling TSRMLS_CC);
+
+	phongo_clientencryption_create_datakey(intern, return_value, kms_provider, options TSRMLS_CC);
+} /* }}} */
+
+/* {{{ proto MongoDB\BSON\Binary MongoDB\Driver\ClientEncryption::encrypt(mixed $value[, array $options])
```
jmikola: `MongoDB\BSON\Binary` looks like copypasta.

src/bson-encode.c
```diff
@@ -648,6 +648,44 @@ void php_phongo_zval_to_bson(zval* data, php_phongo_bson_flags_t flags, bson_t*
 	php_phongo_field_path_free(field_path);
 } /* }}} */
 
+/* Converts the array or object argument to a bson_value_t. If the object is an
```
jmikola: This doesn't seem to care about the type of `zval *data`. Should it be checking that and throw if it's not `IS_OBJECT` or `IS_ARRAY`? If not, should this be rephrased?

src/MongoDB/ClientEncryption.c
alcaeus: `encrypt` returns a binary BSON object with subtype 6, hence the return type.

src/MongoDB/Manager.c
```diff
@@ -349,6 +349,26 @@ static PHP_METHOD(Manager, __construct)
...
 
+/* {{{ proto MongoDB\Driver\ClientEncryption MongoDB\Driver\Manager::clientEncryption(array $options)
+   Return a ClientEncryption instance */
+static PHP_METHOD(Manager, clientEncryption)
+{
+	php_phongo_manager_t*          intern;
+	php_phongo_clientencryption_t* clientencryption;
+	zval*                          options;
+
+	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "a", &options) == FAILURE) {
```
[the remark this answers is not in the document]
alcaeus: Creating a ClientEncryption object without parameters causes an exception:
> Key vault client and namespace option required

The `null` and `IS_ARRAY` checks for `options` in `phongo_clientencryption_opts_from_zval` to prevent segmentation faults if the method is called with a non-array `options`. I can change this to an exception instead of silently returning empty options, but allowing `null` to be passed to the `clientEncryption` method is not a good idea IMO.

php_phongo.c
alcaeus: > How would this ever happen in practice if `keyaltnames_count` is initialized to the size of the PHP array? I know we don't use `assert()` in the extension but if we did is that what you might have used here?

Yes. I wanted to guard against accidental overflows and writing to memory outside of the array. I can switch to using `assert()` or remove the check entirely, depending on what's more sensible for the situation. Not sure what the usual practice for these kinds of checks is.

php_phongo.c
alcaeus: Good point. I've also noted a wrong return statement in the `cleanup` block of `php_phongo_bson_value_to_zval` and fixed that.

php_phongo.c
alcaeus: > AFAIK, this function is only called from a method in `phongo_clientencryption_create_datakey`. Perhaps that's the same case for a few other functions (e.g. encrypt, decrypt, and their options helpers). Would it make sense to define these as static functions in ClientEncryption.c?

I've put the functions into php_phongo.c to not pollute the class file with unrelated methods. We had previously discussed keeping those files clean with class-related code only. I have no problem moving the functions, but was also planning on re-organising code across the board with the goal to clean up the php_phongo file.

> Perhaps the ClientEncryption class itself doesn't even need to exist (at runtime) unless FLE is enabled.

I'm not sure that's a good idea. The class can't be instantiated by itself, and with the `Manager::clientEncryption` method throwing an exception if FLE is not compiled, the class can't be used. I prefer that over the class only being available in some instances, no matter how useless it may be without libmongocrypt.

src/MongoDB/Manager.c
```diff
@@ -349,6 +349,26 @@ static PHP_METHOD(Manager, __construct)
 	}
 } /* }}} */
 
+/* {{{ proto MongoDB\Driver\ClientEncryption MongoDB\Driver\Manager::clientEncryption(array $options)
+   Return a ClientEncryption instance */
+static PHP_METHOD(Manager, clientEncryption)
```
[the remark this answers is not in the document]
alcaeus: Other drivers allow creating a `ClientEncryption` object through the constructor. This isn't an option for us, as we'd need to unpack the `MongoDB\Driver\Manager` instance from a `keyVaultClient` option. Since this requires logic in PHPLIB, we opted to expose the class through the Manager and Client classes.

I'm perfectly fine prefixing the method, and considering that it always creates a new `ClientEncryption` instance, I'd go for `createClientEncryption`.

php_phongo.c
```diff
@@ -2841,18 +2841,379 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
 	return retval;
 }
 /* }}} */
+
+static mongoc_client_encryption_opts_t* phongo_clientencryption_opts_from_zval(mongoc_client_t* client, zval* options TSRMLS_DC) /* {{{ */
```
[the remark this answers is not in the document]
alcaeus: They are similar, but call different methods, e.g. `mongoc_auto_encryption_opts_set_*` vs. `mongoc_client_encryption_opts_set_*`. I'm not sure if we would gain anything by trying to refactor this code.

php_phongo.c
alcaeus: I changed handling of `keyvault_namespace` to be equivalent to the handling in `phongo_manager_set_auto_encryption_opts`. I've also created [PHPC-1527](https://jira.mongodb.org/browse/PHPC-1527) to investigate other instances where we may be using `efree` instead of `str_efree`.

php_phongo.c
alcaeus: Yes, this was shamelessly copied from other iterations. Looking at [the declaration of `HASH_OF`](https://github.com/php/php-src/blob/4d24f5a49453fa1449f85e16548ceeb689edb0f6/Zend/zend_API.h#L658), this will be `NULL` if the zval is neither an array nor object. Since we check for `zkeyaltnames` being an array a couple of lines earlier, I think we can keep this as-is. Changing this to check the return value of `HASH_OF` would allow users to pass objects for `keyAltNames`, which isn't desirable in my opinion.

php_phongo.c
alcaeus: Implemented type checks along with the proper error message depending on the array key. Also included tests for the behaviour.

php_phongo.c
alcaeus: Switched to using `ecalloc`, `estrdup`, and `efree` instead of the respective `bson_*` functions.

php_phongo.c
alcaeus: I noticed that as well when I added the tests the string type check above. The concern is true for `bson_free`, while `efree` ignores this. I've still changed it to only loop over those array elements that were initialised during the first loop.

src/bson-encode.c
alcaeus: Whoops. Rephrased it as the function is capable of handling any arbitrary type which can be converted to a `bson_value_t`.

src/MongoDB/Manager.c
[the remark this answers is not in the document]
jmikola: Ah, I didn't realize that two options were necessary. Technically, `keyVaultClient` isn't necessary since it defaults to the Manager's client, but `keyVaultNamespace` would definitely be required so I suppose the minimum value you could pass as the options here is `['keyVaultNamespace' => '...']`.

From an API standpoint, I wonder if `keyVaultNamespace` should be a required parameter of this method, which you then use to call `mongoc_client_encryption_opts_set_keyvault_namespace`. Is `keyVaultNamespace` also a required key in the `autoEncryption` array within `$driverOptions`? Typically, anything required would be its own parameter and `$options` would contain strictly optional values. But if you'd prefer to just take everything in one array here for consistency with `autoEncryption` (obviously we have no other approach to take there, since it's all within `$driverOptions`), then that works for me.

If we keep `keyVaultNamespace` in the options array, `zend_parse_parameters` can continue to use `a` since an array will certainly be required to specify that required option.

The only reason I suggested `|a!` is because it'd be consistent for all other options arrays in PHPC. `ZEND_ARG_ARRAY_INFO(0, options, 1)` appears numerous places in the driver. Unfortunately, I don't recall exactly why we accepted null there instead of just using `|a` and requiring an empty array at a minimum. [Registering and Using PHP Functions](http://www.phpinternalsbook.com/php7/extensions_design/php_functions.html) in the PHP 7 internals book states that `zend_parse_parameters` doesn't even touch the output variable if it wasn't provided by the caller, which is probably why we initialize all of these zval pointers to null ourselves -- that seems completely unrelated to our use of the `!` operator to accept `null` PHP values.

The alternative

tests/clientEncryption/clientEncryption-createDataKey_error-001.phpt
```diff
@@ -0,0 +1,38 @@
...
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected keyAltNames to be array, string given
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected keyAltName with index "0" to be string, array given
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected keyAltName with index "foo" to be string, array given
```
jmikola: These error messages are a good improvement. :+1:

tests/clientEncryption/clientEncryption-decrypt-001.phpt
```diff
@@ -0,0 +1,27 @@
...
+require_once __DIR__ . "/../utils/basic.inc";
+
+$key = base64_decode('Mng0NCt4ZHVUYUJCa1kxNkVyNUR1QURhZ2h2UzR2d2RrZzh0cFBwM3R6NmdWMDFBMUN3YkQ5aXRRMkhGRGdQV09wOGVNYUMxT2k3NjZKelhaQmRCZGJkTXVyZG9uSjFk');
+
+$manager = new MongoDB\Driver\Manager(URI);
+$clientEncryption = $manager->clientEncryption(['keyVaultNamespace' => 'default.keys', 'kmsProviders' => ['local' => ['key' => new MongoDB\BSON\Binary($key, 0)]]]);
+
+$key = $clientEncryption->createDataKey('local');
+
+$encrypted = $clientEncryption->encrypt('top-secret', ['keyId' => $key, 'algorithm' => 'AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic']);
```
jmikola: From the spec, there appears to be [two algorithms](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#algorithm). Should we consider defining these algorithms as constants on the ClientEncryption class? 

If so, are these only used for ClientEncryption? The EncryptOpts in the spec appear to only be for the explicit API. I imagine auto encryption can only use deterministic encryption, so perhaps it's not even an option there since there are only two supported algorithms at present (deterministic and random).

src/BSON/Binary.c
```diff
@@ -504,6 +504,7 @@ void php_phongo_binary_init_ce(INIT_FUNC_ARGS) /* {{{ */
 	zend_declare_class_constant_long(php_phongo_binary_ce, ZEND_STRL("TYPE_OLD_UUID"), BSON_SUBTYPE_UUID_DEPRECATED TSRMLS_CC);
 	zend_declare_class_constant_long(php_phongo_binary_ce, ZEND_STRL("TYPE_UUID"), BSON_SUBTYPE_UUID TSRMLS_CC);
 	zend_declare_class_constant_long(php_phongo_binary_ce, ZEND_STRL("TYPE_MD5"), BSON_SUBTYPE_MD5 TSRMLS_CC);
+	zend_declare_class_constant_long(php_phongo_binary_ce, ZEND_STRL("TYPE_ENCRYPTED"), BSON_SUBTYPE_ENCRYPTED TSRMLS_CC);
```
jmikola: I missed this earlier, but I think we should add this to `bson-binary-001.phpt`.

src/MongoDB/Manager.c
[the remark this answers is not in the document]
alcaeus: https://jira.mongodb.org/browse/PHPC-1531

php_phongo.c
alcaeus: Defensive is good for now, we can always revisit later 👍

php_phongo.c
alcaeus: Changed to `>=` to be extra-defensive and be less reliant on the loop itself 👍

tests/clientEncryption/clientEncryption-createDataKey-001.phpt
```diff
@@ -0,0 +1,28 @@
...
+
+var_dump($clientEncryption->createDataKey('local'));
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+object(MongoDB\BSON\Binary)#%d (%d) {
+  ["data"]=>
+  string(16) "%a"
```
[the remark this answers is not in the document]
alcaeus: Since the spec explicitly calls for a binary with subtype 4 (UUID) to be returned, I decided to add that to the test. I can also just assert against the return value being a binary or even anything if you prefer

tests/clientEncryption/clientEncryption-decrypt-001.phpt
alcaeus: The "encrypt" option for `jsonSchema` also allows for an `algorithm` option, which accepts both `random` and `deterministic` algorithms. That said, putting these constants in `ClientEncryption` is probably the best place to store them, as they may be confusing if added to the `Manager` class itself.

I was considering adding them as you really don't want to type these out as strings every time you need to use them, but pretty much forgot. I'll add them, but will replace the `-Deterministic` portion with `_DETERMINISTIC` and `_RANDOM` to adhere to the constant naming pattern that we use. That would make these constants `AEAD_AES_256_CBC_HMAC_SHA_512_DETERMINISTIC` and `AEAD_AES_256_CBC_HMAC_SHA_512_RANDOM`. Still a handful, but at least code completion can help the user pick an algorithm without adding typos. 👍

src/MongoDB/Manager.c
```diff
@@ -796,6 +816,10 @@ ZEND_BEGIN_ARG_INFO_EX(ai_Manager___construct, 0, 0, 0)
 	ZEND_ARG_ARRAY_INFO(0, driverOptions, 0)
 ZEND_END_ARG_INFO()
 
+ZEND_BEGIN_ARG_INFO_EX(ai_Manager_clientEncryption, 0, 0, 1)
+	ZEND_ARG_INFO(0, options)
```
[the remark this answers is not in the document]
alcaeus: Fixed. Also fixed the calls in `ClientEncryption` since use `a!` for the `$options` argument in `createDataKey` and `encrypt`.

src/MongoDB/Manager.c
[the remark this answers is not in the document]
alcaeus: I originally had `keyVaultNamespace` as separate argument, but put it into the options array for consistency with other encryption options, including those for `autoEncryption`.

src/bson-encode.c
```diff
@@ -648,6 +648,44 @@ void php_phongo_zval_to_bson(zval* data, php_phongo_bson_flags_t flags, bson_t*
...
+#else
+	zval* data_object = NULL;
+	ALLOC_INIT_ZVAL(data_object);
+#endif
+
+	array_init_size(data_object, 1);
+	add_assoc_zval(data_object, "data", data);
+
+#if PHP_VERSION_ID >= 70000
+	Z_TRY_ADDREF_P(data);
```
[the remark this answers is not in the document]
alcaeus: I noticed that both `encrypt` and `decrypt` tests fail if I switch to `Z_ADDREF_P`. Not calling any `ADDREF_P` macro also causes an error, since we add `data` to the `data_object` zval which is later freed using `zval_ptr_dtor`.

php_phongo_structs.h
jmikola: Noting that [PHPC-1529](https://jira.mongodb.org/browse/PHPC-1529) has been opened to track forking and FLE.

tests/clientEncryption/clientEncryption-createDataKey-001.phpt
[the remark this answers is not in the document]
jmikola: Oh, I'm fine with the expected output and asserting the UUID subtype on the line below. I was mainly inquiring about why we use `%a` here. Reading this again, I think that question has been answered so I'll resolve.

src/bson-encode.c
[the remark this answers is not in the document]
jmikola: Oh, I missed how `data_object` was used in this function. I think this makes sense, given that the value being encrypted in the test file is a string literal. PHP 7 will end up using an [interned string](https://wiki.php.net/rfc/performanceimprovements#interned_strings), which is _not_ ref-counted.

While `keyId` processing also uses this code path, I expect the interned string case never comes up in practice since that value originates from `createDataKey()` and will almost certainly never be a string literal.

This all LGTM.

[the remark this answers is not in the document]
alcaeus: Ah, didn't think about interned strings when using a string literal. Thank you for the explanation! ❤️

src/MongoDB/ClientEncryption.c
```diff
@@ -0,0 +1,208 @@
...
+	PHONGO_CE_DISABLE_SERIALIZATION(php_phongo_clientencryption_ce);
+
+	memcpy(&php_phongo_handler_clientencryption, phongo_get_std_object_handlers(), sizeof(zend_object_handlers));
+	php_phongo_handler_clientencryption.get_debug_info = php_phongo_clientencryption_get_debug_info;
+#if PHP_VERSION_ID >= 70000
+	php_phongo_handler_clientencryption.free_obj = php_phongo_clientencryption_free_object;
+	php_phongo_handler_clientencryption.offset   = XtOffsetOf(php_phongo_clientencryption_t, std);
+#endif
+
+	zend_declare_class_constant_string(php_phongo_clientencryption_ce, ZEND_STRL("AEAD_AES_256_CBC_HMAC_SHA_512_DETERMINISTIC"), "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic" TSRMLS_CC);
```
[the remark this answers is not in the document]
alcaeus: They were defined in libmongoc, I just missed them. Changed the code and added a test 👍

PR #1083, opened 2020-01-17
Update to libmongoc 1.16.0

alcaeus:
This updates the bundled libmongoc version to 1.16.0. config.m4 already requires this version for system libraries, so no further changes are necessary.


alcaeus:
@jmikola pushed another commit to test against libmongoc 1.16 explicitly, not only 1.17-dev.

PR #1085, opened 2020-01-22
PHPC-1497: Introduce new configure switch to use system libraries

alcaeus:
https://jira.mongodb.org/browse/PHPC-1497

The new `--with-system-libs` configure switch replaces `--with-libmongoc` and `--with-libbson`, since they both needed to have the same value. Passing `--with-system-libs=yes` will always compile against system libraries, regardless of the `--with-libmongoc` and `--with-libbson` options. When one of these two options is used, a message is printed alerting the user that the settings are deprecated and they should use `--with-system-libs` instead.

Since neither libmongoc nor libbson option took a path where to find the libraries, it is better to use this compound switch. However, if downstream packages need to supply custom paths for libbson or libmongoc, we may opt to keep the individual options instead.  Ping @remicollet and @oerdnj for feedback on this.

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -170,6 +170,12 @@ if test "$PHP_MONGODB" != "no"; then
     src/MongoDB/Monitoring/functions.c \
   "
 
+  PHP_ARG_WITH([mongodb-system-libs],
+               [whether to compile against system libraries instead of bundled],
+               [AS_HELP_STRING([--with-mongodb-system-libs=@<:@yes/no@:>@],
+                               [MongoDB: Use system libraries for libbson, libmongoc, and libmongocrypt [default=no]])],
+               [no],
+               [no])
   PHP_ARG_WITH([libbson],
                [whether to use system libbson],
                [AS_HELP_STRING([--with-libbson=@<:@yes/no@:>@],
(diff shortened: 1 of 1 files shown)


remicollet:
> if downstream packages need to supply custom paths for libbson or libmongoc,
a custom path is easily managed by setting PKG_CONFIG_PATH

LGTM

oerdnj:
lgtm too

alcaeus:
Thank you for the quick feedback! ❤️

config.m4
```diff
@@ -170,6 +170,12 @@ if test "$PHP_MONGODB" != "no"; then
     src/MongoDB/Monitoring/functions.c \
   "
 
+  PHP_ARG_WITH([system-libs],
```
jmikola: I think we should prefix this with `mongodb-` so we don't pollute the global namespace if our build process was invoked from a bundled PHP build. This was realistically never an issue for the old "libmongoc" and "libbson" names, but "system-libs" is much more generic.

config.m4
```diff
@@ -469,9 +485,9 @@ Build configuration:
   Extra CFLAGS                                     : $STD_CFLAGS $EXTRA_CFLAGS
   Developers flags (slow)                          : $MAINTAINER_CFLAGS
   Code Coverage flags (extra slow)                 : $COVERAGE_CFLAGS
-  System mongoc                                    : $PHP_LIBMONGOC
-  System libbson                                   : $PHP_LIBBSON
-  System libmongocrypt                             : $PHP_LIBBSON
+  System mongoc                                    : $PHP_SYSTEM_LIBS
```
jmikola: We might as well rewrite this as "System libmongoc" while we're changing this line.

config.m4
alcaeus: Good point 👍

config.m4
alcaeus: Done.

PR #1087, opened 2020-01-27
PHPC-1539: Create new exception classes for FLE

alcaeus:
https://jira.mongodb.org/browse/PHPC-1539


php_phongo.c
```diff
@@ -136,10 +136,16 @@ zend_class_entry* phongo_exception_from_phongo_domain(php_phongo_error_domain_t
 	MONGOC_ERROR("Resolving unknown phongo error domain: %d", domain);
 	return php_phongo_runtimeexception_ce;
 }
-zend_class_entry* phongo_exception_from_mongoc_domain(uint32_t /* mongoc_error_domain_t */ domain, uint32_t /* mongoc_error_code_t */ code)
+zend_class_entry* phongo_exception_from_mongoc_domain(mongoc_error_domain_t domain, mongoc_error_code_t code)
```
jmikola: The old signature dates back to 8710d74ccbdf59af261bd89e132c5a62db48bbd3, but I don't recall why `uint32_t` was used instead of the enums. Assuming this doesn't product any compiler warnings, it seems OK to me.

src/MongoDB/Exception/EncryptionException.c
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2014-2017 MongoDB, Inc.
```
jmikola: I think you can just put "2020" here.

php_phongo.c
alcaeus: The old logic used a `switch` statement to check the error codes, which can cause compiler warnings if not all values from `mongoc_error_domain_t` are covered. The code compiles without warning, so I assume that the reason for using `uint32_t` has gone away.

PR #1088, opened 2020-01-27
PHPC-1540: Destroy persistent clients in reverse order to prevent segmentation faults

alcaeus:
https://jira.mongodb.org/browse/PHPC-1540

When configuring a manager with a `keyVaultClient` in `autoEncryption` options, a segmentation fault can appear during shutdown. To alleviate this, we have to destroy persistent client objects in reverse order to avoid freeing a client used as `keyVaultClient` before freeing the client that's using it.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -3551,18 +3551,6 @@ static inline void php_phongo_pclient_destroy(php_phongo_pclient_t* pclient)
 	pefree(pclient, 1);
 }
 
-#if PHP_VERSION_ID >= 70000
-static void php_phongo_pclient_dtor(zval* zv)
-{
-	php_phongo_pclient_destroy((php_phongo_pclient_t*) Z_PTR_P(zv));
-}
-#else
-static void php_phongo_pclient_dtor(void* pp)
-{
-	php_phongo_pclient_destroy(*((php_phongo_pclient_t**) pp));
-}
-#endif
-
 /* {{{ PHP_RINIT_FUNCTION */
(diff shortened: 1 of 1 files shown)


php_phongo.c
```diff
@@ -3751,10 +3739,42 @@ PHP_MINIT_FUNCTION(mongodb)
...
-	/* Destroy HashTable for persistent clients. The HashTable destructor will
-	 * destroy any mongoc_client_t objects that were created by this process. */
+	/* Destroy mongoc_client_t objects in reverse order. This is necessary to
+	 * prevent segmentation faults as clients may reference other clients in
+	 * encryption settings. */
+#if PHP_VERSION_ID >= 70000
+	{
+		zval* z_ptr;
+
+		ZEND_HASH_REVERSE_FOREACH_VAL(pclients, z_ptr)
```
jmikola: I didn't realize macros and functions for reverse iteration existed. That's splendid!

alcaeus:
> Should we add a regression test in PHPC (e.g. `bug1540-001.phpt`) to cover this?

I'll take a look. Doing so requires switching to mongodb-enterprise, which might be more work considering the some build issues I ran into with the FLE spec tests in PHPLIB. I've created [PHPC-1544](https://jira.mongodb.org/browse/PHPC-1544) and scheduled it for the 1.7.1 release so it doesn't block the FLE release. The segmentation faults in the spec tests went away on my machine, so I assume the issue has been fixed.

alcaeus:
Merged manually in 69d6df10.

PR #1089, opened 2020-01-29
PHPC-1545: Require object or array type for "kmsProviders" option

alcaeus:
https://jira.mongodb.org/browse/PHPC-1545

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -2795,6 +2795,11 @@ static bool phongo_manager_set_auto_encryption_opts(php_phongo_manager_t* manage
 		zval*  kms_providers  = php_array_fetch(zAutoEncryptionOpts, "kmsProviders");
 		bson_t bson_providers = BSON_INITIALIZER;
 
+		if (Z_TYPE_P(kms_providers) != IS_OBJECT && Z_TYPE_P(kms_providers) != IS_ARRAY) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected \"kmsProviders\" encryption option to be an array or object");
+			goto cleanup;
+		}
+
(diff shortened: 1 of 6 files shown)


alcaeus:
That’s what I get for just skimming over comments. I’ll also apply those for other options in this PR 👍

php_phongo.c
```diff
@@ -2907,8 +2917,8 @@ static mongoc_client_encryption_opts_t* phongo_clientencryption_opts_from_zval(m
 		zval*  kms_providers  = php_array_fetchc(options, "kmsProviders");
 		bson_t bson_providers = BSON_INITIALIZER;
 
-		if (Z_TYPE_P(kms_providers) != IS_ARRAY) {
-			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT TSRMLS_CC, "Expected \"kmsProviders\" encryption option to be an array");
+		if (Z_TYPE_P(kms_providers) != IS_ARRAY && Z_TYPE_P(kms_providers) != IS_OBJECT) {
```
jmikola: Ah, I forgot we were already checking for an array type here. Good to know both instances of "kmsProviders" handling are consistent now!

tests/clientEncryption/clientEncryption-ctor-error-002.phpt
```diff
@@ -0,0 +1,36 @@
+--TEST--
+MongoDB\Driver\ClientEncryption::__construct() with invalid option types
```
jmikola: Shouldn't this be a test for `Manager::createClientEncryption()` since the constructor is disabled? I see this is `clientEncryption-ctor-error-002.phpt`, so perhaps move all of those over to the Manager name.

alcaeus: Ah, yes. Those were left over from when I initially built this with a public constructor. I moved them to the manager folder and renamed them accordingly.

PR #1095, opened 2020-02-04
Fix build with system libraries

remicollet:
https://jira.mongodb.org/browse/PHPC-1551

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -235,7 +235,6 @@ if test "$PHP_MONGODB" != "no"; then
     PHP_EVAL_LIBLINE($PHP_MONGODB_BSON_LIBS, MONGODB_SHARED_LIBADD)
     AC_DEFINE(HAVE_SYSTEM_LIBBSON, 1, [Use system libbson])
 
-    AC_PATH_PROG(PKG_CONFIG, pkg-config, no)
     AC_MSG_CHECKING(for libmongoc)
     if test -x "$PKG_CONFIG" && $PKG_CONFIG --exists libmongoc-1.0; then
       if $PKG_CONFIG libmongoc-1.0 --atleast-version 1.16.1; then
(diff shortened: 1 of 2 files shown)


config.m4
```diff
@@ -257,11 +257,11 @@ if test "$PHP_MONGODB" != "no"; then
       AC_PATH_PROG(PKG_CONFIG, pkg-config, no)
       AC_MSG_CHECKING(for libmongocrypt)
 
-      if test -x "$PKG_CONFIG" && $PKG_CONFIG --exists libmongocrypt-1.0; then
-        if $PKG_CONFIG libmongocrypt-1.0 --atleast-version 1.0.1; then
```
remicollet: libmongocrypt-1.0 doesn't exists

php_phongo.c
```diff
@@ -3924,7 +3924,7 @@ PHP_MINFO_FUNCTION(mongodb)
 #ifdef MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION
 #ifdef HAVE_SYSTEM_LIBMONGOCRYPT
 	php_info_print_table_row(2, "libmongocrypt headers version", MONGOCRYPT_VERSION);
-	php_info_print_table_row(2, "libmongocrypt library version", mongocrypt_version());
+	php_info_print_table_row(2, "libmongocrypt library version", mongocrypt_version(NULL));
```
remicollet: @param[out] len  An optional length of the returned string. May be NULL.

config.m4
jmikola: Thanks for catching this! I suppose this is another difference between libmongocrypt and libmongoc's packaging. /cc @kevinAlbs

remicollet: This is not related to packaging, but to what upstream .pc file defines

jmikola: Sorry, I didn't mean to imply this was OS-specific packaging that you handle. I assume the issue is that libmongocrypt's `.pc` file doesn't include "-1.0" in the name, which is inconsistent with libmongoc and libbson.

As an aside, there is also a `${PROJECT_NAME}` in [`libmongocrypt.pc.in`](https://github.com/mongodb/libmongocrypt/blob/master/cmake/libmongocrypt.pc.in), but I'm not sure where that is set. `PROJECT_VERSION` is assigned in [`CMakeLists.txt`](https://github.com/mongodb/libmongocrypt/blob/1.0.1/CMakeLists.txt#L304), but `PROJECT_NAME` is nowhere to be found. I assume that is just defaulting to the base name of the `.pc` file or perhaps the build directory.

alcaeus:
Merged manually in 433ddce6. Thanks for catching this @remicollet!

PR #1096, opened 2020-02-04
PHPC-1442: Drop support for PHP 5.6

alcaeus:
https://jira.mongodb.org/browse/PHPC-1442


php_phongo_structs.h
```diff
@@ -21,15 +21,9 @@
 
 #include "php_bson.h"
 
-#if PHP_VERSION_ID >= 70000
 #define PHONGO_ZEND_OBJECT_PRE
 #define PHONGO_ZEND_OBJECT_POST zend_object std;
 #define PHONGO_STRUCT_ZVAL zval
```
jmikola: I don't think these macros need to exist anymore. Let's just replace them with the defined value. `PHONGO_ZEND_OBJECT_PRE` can be deleted entirely.

php_phongo.c
```diff
@@ -80,11 +76,9 @@
 #define PHONGO_DEBUG_INI_DEFAULT ""
 
 ZEND_DECLARE_MODULE_GLOBALS(mongodb)
-#if PHP_VERSION_ID >= 70000
 #if defined(ZTS) && defined(COMPILE_DL_MONGODB)
 ZEND_TSRMLS_CACHE_DEFINE();
```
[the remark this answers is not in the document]
alcaeus: Ah, didn't know that these macros are empty in PHP 7. I removed all of those occurrences in a separate commit as there are a lot of them.

alcaeus:
@jmikola dropped all compat macros that are now useless. I recommend looking at the diff commit by commit as the changeset got a lot larger.

php_phongo.c
```diff
@@ -2096,23 +2096,23 @@ static mongoc_ssl_opt_t* php_phongo_make_ssl_opt(mongoc_uri_t* uri, zval* zoptio
 static void php_phongo_free_ssl_opt(mongoc_ssl_opt_t* ssl_opt)
 {
 	if (ssl_opt->pem_file) {
-		str_efree(ssl_opt->pem_file);
+		efree((char*) ssl_opt->pem_file);
```
jmikola: Is `(char*)` necessary? I noticed you didn't do it for `efree(key_vault_ns)` later on in this file.

`efree` itself takes a [`void*`](http://lxr.jpauli.tech/source/xref/php-7.4/Zend/zend_alloc.c#2541), just like `free()`, so any pointer should suffice without a cast.

phongo_compat.h
```diff
@@ -58,22 +58,14 @@
 #define PHONGO_BREAK_INTENTIONALLY_MISSING
 #endif
 
-#define phongo_char zend_string
```
jmikola: Ah, I didn't think of these but it makes sense to remove them all.

phongo_compat.h
```diff
@@ -58,22 +58,14 @@
...
 #define PHONGO_LONG_FORMAT PRId32
 #else
 #error Unsupported architecture (integers are neither 32-bit nor 64-bit)
 #endif
-#define SIZEOF_PHONGO_LONG SIZEOF_ZEND_LONG
-#define phongo_create_object_retval zend_object*
-#define phongo_get_gc_table zval**
 #define PHONGO_ALLOC_OBJECT_T(_obj_t, _class_type) (_obj_t*) ecalloc(1, sizeof(_obj_t) + zend_object_properties_size(_class_type))
-#define PHONGO_TSRMLS_FETCH_FROM_CTX(user_data)
-#define DECLARE_RETURN_VALUE_USED int return_value_used = 1;
```
jmikola: Great catch with this! I did some digging and found that it was introduced in https://github.com/mongodb/mongo-php-driver/commit/23f8316617067f44af59c027768c6470fb53e7f0 for [PHPC-370](https://jira.mongodb.org/browse/PHPC-370). That issue explains that it's meaningless for PHP 7.

src/MongoDB/BulkWrite.c
```diff
@@ -374,7 +371,7 @@ static PHP_METHOD(BulkWrite, insert)
 
 	intern->num_ops++;
 
-	if (bson_out && return_value_used) {
+	if (bson_out) {
```
jmikola: I wonder if this condition pre-dates the `cleanup` label below. Would `bson_out` only still be null if `php_phongo_zval_to_bson` failed? And in that case, would we have already jumped to `cleanup`? Either way, I'm OK leaving this as a sanity check.

Interestingly, the `cleanup` label was added in https://github.com/mongodb/mongo-php-driver/commit/1e0b1e8e290699101c21b20e324a9b154df389cd, which is separate from the `goto cleanup;` lines added in https://github.com/mongodb/mongo-php-driver/commit/7ff55ba42cb36f60836d2d83426fa909c5b54a8e (same release, though). Perhaps I just goofed in splitting up those commits.

php_phongo.c
alcaeus:
```
php_phongo.c:2083:3: error: passing 'const char *' to parameter of type 'void *' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
                efree(ssl_opt->pem_file);
                ^~~~~~~~~~~~~~~~~~~~~~~~
```

`pem_file` is declared as a `const char*` in `_mongoc_ssl_opt_t`, along with other string options. We create the `ssl_opt` member, its settings are later copied to mongoc, then we free it. The explicit cast makes sense, we just hid it behind a macro before.

src/MongoDB/BulkWrite.c
alcaeus: `bson_out` will always be set due to us using `PHONGO_BSON_RETURN_ID` a few lines earlier, so the check could definitely go away, but would most likely cause a segmentation fault in `php_phongo_bulkwrite_extract_id` if it wasn't set for some reason.

While a sanity check makes sense, it should most likely be an error condition and not silently be a no-op. I've updated this accordingly.

PR #1101, opened 2020-02-19
Enabling Travis builds for s390x

namrata-ibm:
As Travis CI [officially supports](https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z) s390x builds, adding support for same.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,14 @@ env:
     - SERVER_VERSION=4.2.0
     - DEPLOYMENT=STANDALONE
 
+linux-s390x: &linux-s390x
+  os: linux
+  arch: s390x
+  dist: bionic
+  env: &env_s390x
+    - SERVER_DISTRO=ubuntu1804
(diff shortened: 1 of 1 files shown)


jmikola:
@namrata-ibm: Thanks for the contribution. I'm not sure this is necessary since we already have test coverage for zSeries on [Evergreen](https://github.com/mongodb/mongo-php-driver/blob/656580e1103e65365643bc4bc14552009f6cda64/.evergreen/config.yml#L637), which is our internal CI. Our Evergreen matrix is much more comprehensive and we mainly rely on Travis for quicker PR feedback, so I think we'd do well to keep its matrix as small as possible (also given that it's a public resource). I'll defer to @alcaeus on this.

alcaeus:
I agree with @jmikola here - since we're using public resources I'd rather keep the travis-ci pipeline small and use our own evergreen system to test additional platforms. Thanks for the suggestion!

PR #1103, opened 2020-02-26
PHPC-1560: Fix compilation with bundled libmongocrypt for non-standard openssl directories

alcaeus:
https://jira.mongodb.org/browse/PHPC-1560

This fixes https://github.com/brefphp/bref/issues/567. Previously, config.m4 pre-initialised the `PHP_MONGODB_LIBMONGOCRYPT_CFLAGS` variable with the CFLAGS used for all bundled builds. However, these flags are changed in libmongoc/CheckSSL.m4 depending on the SSL library being used. These flags are then missing when building with bundled libmongocrypt, which leads to errors when custom directories are used.

Copying CFLAGS from `PHP_MONGODB_BUNDLED_CFLAGS` after the SSL check fixes this issue.

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -283,7 +283,7 @@ if test "$PHP_MONGODB" != "no"; then
   if test "$PHP_MONGODB_SYSTEM_LIBS" = "no"; then
     PHP_MONGODB_BUNDLED_CFLAGS="$STD_CFLAGS -DBSON_COMPILATION -DMONGOC_COMPILATION"
     dnl TODO: MONGOCRYPT-219 makes the -std argument obsolete
-    PHP_MONGODB_LIBMONGOCRYPT_CFLAGS="$PHP_MONGODB_BUNDLED_CFLAGS -DKMS_MSG_STATIC -std=gnu99"
+    PHP_MONGODB_LIBMONGOCRYPT_CFLAGS="-DKMS_MSG_STATIC -std=gnu99"
 
     dnl M4 doesn't know if we're building statically or as a shared module, so
     dnl attempt to include both paths while ignoring errors. If neither path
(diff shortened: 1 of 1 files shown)


config.m4
```diff
@@ -283,7 +283,7 @@ if test "$PHP_MONGODB" != "no"; then
   if test "$PHP_MONGODB_SYSTEM_LIBS" = "no"; then
     PHP_MONGODB_BUNDLED_CFLAGS="$STD_CFLAGS -DBSON_COMPILATION -DMONGOC_COMPILATION"
     dnl TODO: MONGOCRYPT-219 makes the -std argument obsolete
-    PHP_MONGODB_LIBMONGOCRYPT_CFLAGS="$PHP_MONGODB_BUNDLED_CFLAGS -DKMS_MSG_STATIC -std=gnu99"
+    PHP_MONGODB_LIBMONGOCRYPT_CFLAGS="-DKMS_MSG_STATIC -std=gnu99"
```
jmikola: I was going to ask why we couldn't just move this within the later conditional:

https://github.com/mongodb/mongo-php-driver/blob/1d33e149235df2b4899b9ec2d9ca5a5060f166d2/config.m4#L406-L407

However, I realized it's also set in [`libmongocrypt/CheckSSL.m4`](https://github.com/mongodb/mongo-php-driver/blob/master/scripts/autotools/libmongocrypt/CheckSSL.m4), which is included well before that point. Looking at the comment before the libmongocrypt includes, I take it this is necessary because those includes may change the value of `PHP_MONGODB_CLIENT_SIDE_ENCRYPTION` used for the conditional itself.

config.m4
```diff
@@ -421,6 +421,7 @@ if test "$PHP_MONGODB" != "no"; then
       dnl Generated with: find src/libmongocrypt/kms-message/src -maxdepth 1 -name '*.c' -print0 | cut -sz -d / -f 5- | sort -dz | tr '\000' ' '
       PHP_MONGODB_MONGOCRYPT_KMS_MESSAGE_SOURCES="hexlify.c kms_b64.c kms_caller_identity_request.c kms_crypto_apple.c kms_crypto_libcrypto.c kms_crypto_none.c kms_crypto_windows.c kms_decrypt_request.c kms_encrypt_request.c kms_kv_list.c kms_message.c kms_request.c kms_request_opt.c kms_request_str.c kms_response.c kms_response_parser.c sort.c"
 
+      PHP_MONGODB_LIBMONGOCRYPT_CFLAGS="$PHP_MONGODB_BUNDLED_CFLAGS $PHP_MONGODB_LIBMONGOCRYPT_CFLAGS"
```
jmikola: Since [`libmongocrypt/CheckSSL.m4`](https://github.com/mongodb/mongo-php-driver/blob/master/scripts/autotools/libmongocrypt/CheckSSL.m4) relies on the `$PHP_MONGODB_SSL` var set in [`libmongoc/CheckSSL.m4`](https://github.com/mongodb/mongo-php-driver/blob/master/scripts/autotools/libmongoc/CheckSSL.m4), we're always assured that libmongocrypt and libmongoc are using the _same_ TLS library. This change was just to defer reading of `$PHP_MONGODB_BUNDLED_CFLAGS` until _after_ any m4 includes might modify it.

config.m4
alcaeus: That is correct. Both libmongoc/CheckSSL.m4 and libmongocrypt/CheckSSL.m4 modify their variables, which is why I decided to only inject `PHP_MONGODB_BUNDLED_CFLAGS` just before we use it.

config.m4
alcaeus: Correct as well. libmongocrypt/CheckSSL.m4 never did its own SSL checking, it merely assigned compiler flags to ensure that libmongocrypt is built with the same SSL library we've selected for libmongoc. I'm surprised it took this long to figure out this issue, since the default includes seem to be sufficient in most cases to locate SSL headers. Brief stored those in `/opt/bref/includes`, which wasn't picked up for libmongocrypt compilation and thus caused failures even though OpenSSL was present and found.

alcaeus:
Merged manually in 75ec5790.

PR #1105, opened 2020-02-28
Update to latest libmongoc version

alcaeus:
This PR updates libmongoc to its latest version to allow working on MongoDB AWS and other 4.4 features.

The update requires some build system changes to accommodate the kms_message library bundled with libmongoc. Since both libmongoc and libmongocrypt require kms_message, the following takes place to avoid symbol conflicts:
* If compiling without libmongocrypt, kms_message bundled with libmongoc is used
* If compiling with libmongocrypt, kms_message bundled with libmongocrypt is used.

This is due to libmongocrypt having a "stronger" requirement on kms_message, thus likely requiring newer versions in the future. If issues arise due to upstream changes, we can always change the process and introduce symbol prefixes upstream.


config.w32
```diff
@@ -281,6 +286,10 @@ if (PHP_MONGODB != "no") {
     );
   } else if (PHP_MONGODB_CLIENT_SIDE_ENCRYPTION != "no") {
     WARNING("mongodb libmongocrypt support not enabled, crypto libs not found");
+
+    // Add kms_message sources bundled with libmongoc
```
jmikola: In this case `mongoc_ssl_found` is `false`. Does it make sense to compile KMS here?

Previously, the KMS defines were only set on CFLAGS when `mongoc_ssl_found` was `true`.

If I had to guess, you'd likely only want to use libmongoc's KMS sources when libmongocrypt is disabled but `mongoc_ssl_found` is `true`.

scripts/autotools/libmongoc/CheckSSL.m4
```diff
@@ -188,18 +188,24 @@ if test "$PHP_MONGODB_SSL" = "openssl" -o "$PHP_MONGODB_SSL" = "libressl" -o "$P
...
+
+    PHP_MONGODB_BUNDLED_CFLAGS="$PHP_MONGODB_BUNDLED_CFLAGS -DKMS_MESSAGE_ENABLE_CRYPTO=1 -DKMS_MESSAGE_ENABLE_CRYPTO_LIBCRYPTO=1"
   elif test "$PHP_MONGODB_SSL" = "libressl"; then
     AC_SUBST(MONGOC_ENABLE_SSL_OPENSSL, 0)
     AC_SUBST(MONGOC_ENABLE_SSL_LIBRESSL, 1)
     AC_SUBST(MONGOC_ENABLE_SSL_SECURE_TRANSPORT, 0)
     AC_SUBST(MONGOC_ENABLE_CRYPTO_LIBCRYPTO, 1)
     AC_SUBST(MONGOC_ENABLE_CRYPTO_COMMON_CRYPTO, 0)
+
+    PHP_MONGODB_BUNDLED_CFLAGS="$PHP_MONGODB_BUNDLED_CFLAGS -DKMS_MESSAGE_ENABLE_CRYPTO=1 -DKMS_MESSAGE_ENABLE_CRYPTO_LIBCRYPTO=1"
```
jmikola: Just to confirm, this will get incorporated into `PHP_MONGODB_BUNDLED_CFLAGS` _before_  `PHP_MONGODB_LIBMONGOCRYPT_CFLAGS` pulls in `PHP_MONGODB_BUNDLED_CFLAGS` (see: https://github.com/alcaeus/mongo-php-driver/blob/4df86904ea471405fd27ed50befa452bfc174671/config.m4#L436), correct?

alcaeus: Yes, correct. Those flags were only set in `PHP_MONGODB_LIBMONGOCRYPT_CFLAGS` but will now always be set if we find an SSL library.

config.w32
alcaeus: That is correct, and config.m4 was affected as well. Fixing this in #1108.

PR #1107, opened 2020-03-02
PHPC-1483: MONGODB-AWS Support

alcaeus:
https://jira.mongodb.org/browse/PHPC-1483

Note that this PR does not add any spec tests for now: authentication is handled entirely by libmongoc and all spec tests are implemented in the client-facing `mongodb/mongodb` library, which implements all spec tests. I've created [PHPLIB-535](https://jira.mongodb.org/browse/PHPLIB-535) to track this.


php_phongo.c
```diff
@@ -1375,7 +1375,8 @@ static bool php_phongo_uri_finalize_auth(mongoc_uri_t* uri) /* {{{ */
 		}
 
 		/* MONGODB-X509 is the only mechanism that doesn't require username */
-		if (strcasecmp(mongoc_uri_get_auth_mechanism(uri), "MONGODB-X509")) {
+		if (strcasecmp(mongoc_uri_get_auth_mechanism(uri), "MONGODB-X509") &&
+			strcasecmp(mongoc_uri_get_auth_mechanism(uri), "MONGODB-AWS")) {
```
jmikola: I suppose the comment above should be updated, since MONGODB-X509 is no longer the only mechanism to not require a username.

tests/manager/manager-ctor-auth_mechanism-001.phpt
```diff
@@ -7,9 +7,11 @@ $tests = [
     ['mongodb://username@127.0.0.1/?authMechanism=MONGODB-X509', []],
     ['mongodb://127.0.0.1/?authMechanism=MONGODB-X509', []],
     ['mongodb://username@127.0.0.1/?authMechanism=GSSAPI', []],
+    ['mongodb://127.0.0.1/?authMechanism=MONGODB-AWS', []],
```
jmikola: This is just testing that the mechanism is not rejected by libmongoc, correct?

Does acceptance of this mechanism depend on whether `MONGOC_ENABLE_MONGODB_AWS_AUTH` is enabled at compile time, or is it similar to GSSAPI as I describe below?

I don't think we have any conditional builds of the driver without SASL, but it looks like GSSAPI support is conditional on `MONGOC_ENABLE_SASL` being defined at compile time. Perhaps that only extends to the authentication attempt at runtime and not URI option validation, though, since `_mongoc_cluster_auth_node_sasl` still gets defined when SASL is not available and simply raises an error.

If the AWS mechanism behaves similarly, this LGTM. I assume we'll need more specific checks in the PHPLIB integration tests to skip if the mechanism (and KMS message) is not available.

config.m4
```diff
@@ -336,7 +336,7 @@ if test "$PHP_MONGODB" != "no"; then
     AC_SUBST(BSON_OS, 1)
 
     AC_SUBST(MONGOC_NO_AUTOMATIC_GLOBALS, 1)
-    AC_SUBST(MONGOC_ENABLE_MONGODB_AWS_AUTH, 0)
+    AC_SUBST(MONGOC_ENABLE_MONGODB_AWS_AUTH, 1)
```
jmikola: Does this PR depend on https://github.com/mongodb/mongo-php-driver/pull/1108? Presumably we cannot enable `MONGOC_ENABLE_MONGODB_AWS_AUTH` if KMS itself cannot be compiled (from either libmongoc or libmongocrypt).

I believe the most direct dependency would be on whether TLS is available.

alcaeus: I'll run some tests without SSL. I'm under the impression that libmongoc will throw an error if enabling MONGODB-AWS without an SSL library. In that case, we may want to disable support for MONGODB-AWS as it's not an essential feature for most people.

tests/manager/manager-ctor-auth_mechanism-001.phpt
alcaeus: > This is just testing that the mechanism is not rejected by libmongoc, correct?

Yes, that's correct. Since libmongoc handles authentication for us, we don't even have a way of checking whether it picks the right credentials, as is necessary for the auth spec tests. Thus, the test is only there to make sure that we
a) don't reject the auth mechanism
b) accept the auth mechanism without a username

> Does acceptance of this mechanism depend on whether `MONGOC_ENABLE_MONGODB_AWS_AUTH` is enabled at compile time, or is it similar to GSSAPI as I describe below?
> 
> I don't think we have any conditional builds of the driver without SASL, but it looks like GSSAPI support is conditional on `MONGOC_ENABLE_SASL` being defined at compile time. Perhaps that only extends to the authentication attempt at runtime and not URI option validation, though, since `_mongoc_cluster_auth_node_sasl` still gets defined when SASL is not available and simply raises an error.

I'll test this on a special build without MONGODB-AWS to see how it behaves and will adjust accordingly. Thanks for bringing it up 👍

> If the AWS mechanism behaves similarly, this LGTM. I assume we'll need more specific checks in the PHPLIB integration tests to skip if the mechanism (and KMS message) is not available.

I'm not sure we can add a lot of integration tests: we can only blackbox-test authentication against a MONGODB-AWS system, and I have yet to check how much effort it is to set it up. Given how we're not really involved in the auth process I'd rather only test that we handle the arguments appropriately and let the libmongoc test suite take care of the rest.

alcaeus:
@jmikola building without SSL support revealed compile-time errors due to missing kms-library headers. I've changed the configure scripts to only enable support for MongoDB-AWS if an SSL library was found.

The test also revealed a couple of other issues, for which I've created [PHPC-1565](https://jira.mongodb.org/browse/PHPC-1565) to take care of in a separate PR.

config.w32
```diff
@@ -213,6 +213,7 @@ if (PHP_MONGODB != "no") {
     mongoc_opts.MONGOC_ENABLE_CRYPTO_LIBCRYPTO = 1;
     mongoc_opts.MONGOC_ENABLE_SSL = 1;
     mongoc_opts.MONGOC_ENABLE_CRYPTO = 1;
+    mongoc_opts.MONGOC_ENABLE_MONGODB_AWS_AUTH = 1;
```
jmikola: Noting that this is always safe to enable, since we require TLS on Windows.

config.m4
```diff
@@ -390,7 +390,8 @@ if test "$PHP_MONGODB" != "no"; then
 
     dnl If compiling without libmongocrypt, use kms_message sources bundled with libmongoc.
     dnl If compiling with libmongocrypt, kms_message bundled with libmongocrypt is used as it is most likely newer.
-    if test "$PHP_MONGODB_CLIENT_SIDE_ENCRYPTION" != "yes" && "$PHP_MONGODB_SSL" != "no"; then
+    if test "$PHP_MONGODB_CLIENT_SIDE_ENCRYPTION" != "yes" && test "$PHP_MONGODB_SSL" != "no"; then
+      AC_SUBST(MONGOC_ENABLE_MONGODB_AWS_AUTH, 1)
```
jmikola: Noting that `MONGOC_ENABLE_MONGODB_AWS_AUTH` earlier defaults to `0`, so these are the two instances where we'd actually enable it (i.e. both KMS and TLS are available).

config.w32
alcaeus: Ah, good to know. I'll keep it within the conditional to avoid nasty surprises down the road if we ever loosen the requirement.

jmikola: For reference, the dependency is [here](https://github.com/mongodb/mongo-php-driver/blob/aceebe78004910d1b85f8577c6672a846a5fd68c/config.w32#L59).

PR #1108, opened 2020-03-03
Only enable kms-message support if an SSL library was found

alcaeus:
This PR ensures we only enable kms-message compilation in libmongoc if any SSL library was found (see https://github.com/mongodb/mongo-php-driver/pull/1105#discussion_r386732019).

### DIFF
--- a/config.w32
+++ b/config.w32
@@ -285,8 +285,11 @@ if (PHP_MONGODB != "no") {
       mongocrypt_opts
     );
   } else if (PHP_MONGODB_CLIENT_SIDE_ENCRYPTION != "no") {
+    // No SSL library found, we can't enable libmongocrypt
     WARNING("mongodb libmongocrypt support not enabled, crypto libs not found");
+  }
 
+  if (PHP_MONGODB_CLIENT_SIDE_ENCRYPTION == "no" && mongoc_ssl_found) {
     // Add kms_message sources bundled with libmongoc
     ADD_SOURCES(configure_module_dirname + "/src/libmongoc/src/kms_message/src", PHP_MONGODB_KMS_MESSAGE_SOURCES, "mongodb");
     ADD_FLAG("CFLAGS_MONGODB", "/I" + configure_module_dirname + "/src/libmongoc/src/kms-message/src");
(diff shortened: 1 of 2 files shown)


config.w32
```diff
@@ -232,61 +232,64 @@ if (PHP_MONGODB != "no") {
...
-    mongodb_generate_header(
-      configure_module_dirname + "/src/libmongocrypt/src/mongocrypt.h.in",
-      configure_module_dirname + "/src/libmongocrypt/src/mongocrypt.h",
-      mongocrypt_opts
-    );
-  } else if (PHP_MONGODB_CLIENT_SIDE_ENCRYPTION != "no") {
-    WARNING("mongodb libmongocrypt support not enabled, crypto libs not found");
-
+  if (PHP_MONGODB_CLIENT_SIDE_ENCRYPTION != "no") {
+    if (mongoc_ssl_found) {
```
jmikola: I think this logic is correct, but I would have suggested the following to minimize the diff and conditional nesting:

```
if (PHP_MONGODB_CLIENT_SIDE_ENCRYPTION != "no" && mongoc_ssl_found) {
  // the big block of directives to enable libmongocrypt
} else if (PHP_MONGODB_CLIENT_SIDE_ENCRYPTION != "no") {
  // No SSL library found, we can't enable libmongocrypt
  WARNING("mongodb libmongocrypt support not enabled, crypto libs not found");
}

if (PHP_MONGODB_CLIENT_SIDE_ENCRYPTION == "no" && mongoc_ssl_found) {
  // Add kms_message sources bundled with libmongoc
  ADD_SOURCES(configure_module_dirname + "/src/libmongoc/src/kms_message/src", PHP_MONGODB_KMS_MESSAGE_SOURCES, "mongodb");
  ADD_FLAG("CFLAGS_MONGODB", "/I" + configure_module_dirname + "/src/libmongoc/src/kms-message/src");
}
```

Either works for me, and I appreciate the extra comment you added ("No SSL library found, we can't enable libmongocrypt") before the warning.

alcaeus:
Merged manually in 922f6a6a.

PR #1112, opened 2020-03-09
Expand use of RetryableWriteError labels

alcaeus:
This PR covers [PHPC-1510](https://jira.mongodb.org/browse/PHPC-1510) and [PHPC-1530](https://jira.mongodb.org/browse/PHPC-1530).

Most of the logic is handled in libmongoc. Since error labels that we expose in `MongoDB\Driver\Exception\RuntimeException` may be nested in the `writeConcernError` field of the reply, we also expose those.

Spec tests will be updated in a follow-up PR to https://github.com/mongodb/mongo-php-library.


php_phongo.c
```diff
@@ -193,34 +193,56 @@ void phongo_throw_exception(php_phongo_error_domain_t domain, const char* format
 	va_end(args);
 }
 
+static int phongo_exception_append_error_labels(zval* labels, const bson_iter_t* iter)
```
jmikola: If I understand correctly, you split this logic into a separate function since it is called for both `errorLabels` and `writeConcernError.errorLabels`.

alcaeus: Correct 👍

PR #1114, opened 2020-03-11
PHPC-1494: Add client metadata support for wrapping libraries

alcaeus:
https://jira.mongodb.org/browse/PHPC-1494

This adds a new "driver" driver option, which can be used to pass custom driver metadata to be used in the server handshake.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -2459,6 +2459,16 @@ static char* php_phongo_manager_make_client_hash(const char* uri_string, zval* o
 	return hash;
 }
 
+static void php_phongo_set_handshake_data()
+{
+	char* php_version_string = malloc(4 + sizeof(PHP_VERSION) + 1);
+
+	snprintf(php_version_string, 4 + sizeof(PHP_VERSION) + 1, "PHP %s", PHP_VERSION);
+	mongoc_handshake_data_append("ext-mongodb:PHP", PHP_MONGODB_VERSION, php_version_string);
+
+	free(php_version_string);
+}
+
 static mongoc_client_t* php_phongo_make_mongo_client(const mongoc_uri_t* uri) /* {{{ */
 {
 	const char *mongoc_version, *bson_version;
@@ -2484,6 +2494,8 @@ static mongoc_client_t* php_phongo_make_mongo_client(const mongoc_uri_t* uri) /*
 		BSON_VERSION_S,
 		bson_version,
 		PHP_VERSION);
+
+	php_phongo_set_handshake_data();
 
 	return mongoc_client_new_from_uri(uri);
 } /* }}} */
@@ -3326,20 +3338,12 @@ static zend_class_entry* php_phongo_fetch_internal_class(const char* class_name,
 /* {{{ PHP_MINIT_FUNCTION */
 PHP_MINIT_FUNCTION(mongodb)
 {
-	char* php_version_string;
-
 	(void) type; /* We don't care if we are loaded via dl() or extension= */
 
 	REGISTER_INI_ENTRIES();
 
 	/* Initialize libmongoc */
 	mongoc_init();
-
-	/* Set handshake options */
-	php_version_string = malloc(4 + sizeof(PHP_VERSION) + 1);
-	snprintf(php_version_string, 4 + sizeof(PHP_VERSION) + 1, "PHP %s", PHP_VERSION);
-	mongoc_handshake_data_append("ext-mongodb:PHP", PHP_MONGODB_VERSION, php_version_string);
-	free(php_version_string);
 
 	/* Initialize libbson */
 	bson_mem_set_vtable(&MONGODB_G(bsonMemVTable));
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -74,6 +74,8 @@
 
 #define PHONGO_DEBUG_INI "mongodb.debug"
 #define PHONGO_DEBUG_INI_DEFAULT ""
+#define PHONGO_METADATA_SEPARATOR " / "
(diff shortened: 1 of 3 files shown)


php_phongo.c
```diff
@@ -2459,7 +2459,138 @@ static char* php_phongo_manager_make_client_hash(const char* uri_string, zval* o
...
+		}
+
+		ht_data = HASH_OF(driver);
+
+		{
+			zend_string* string_key = NULL;
+			zend_ulong   num_key    = 0;
+			zval*        property;
+
+			ZEND_HASH_FOREACH_KEY_VAL_IND(ht_data, num_key, string_key, property)
```
jmikola: Any reason to use an iteration loop here instead of individual `php_array_fetch` lookups?

To contrast with a couple of other places that fully iterate through an options array:

 * `phongo_clientencryption_datakey_opts_from_zval` uses `ZEND_HASH_FOREACH_KEY_VAL` to check for non-string values and warn about extra keys.
 * `php_phongo_apply_options_to_uri` also does iteration through all keys because it needs to handle case variation.

I don't think this has any of those concerns, so I would have expected something more like what we do in `php_phongo_apply_driver_options_to_uri` or the Query constructor.

php_phongo.c
```diff
@@ -2459,7 +2459,138 @@ static char* php_phongo_manager_make_client_hash(const char* uri_string, zval* o
...
+
+static void php_phongo_handshake_data_append(const char* name, size_t name_len, const char* version, size_t version_len, const char* platform, size_t platform_len)
+{
+	char* php_version_string;
+	char* driver_name;
+	char* driver_version;
+	char* full_platform;
+
+	php_version_string = malloc(4 + sizeof(PHP_VERSION));
+	sprintf(php_version_string, "PHP %s", PHP_VERSION);
```
jmikola: I realize `sizeof(PHP_VERSION)` includes the null byte and we don't have to worry about overflow, but I think it'd be safer to calculate the length first and then use `snprintf` with the same limit.

Also, can we consider using `ecalloc` and `efree` (or the persistent versions if you prefer)? We don't technically need PHP's memory management, but this will ensure we use consistent allocation functions. As-is, I think this might be the only instance of C's memory allocation functions being used directly. Using PHP's `e` functions would also ensure debug mode catches leaks for us without having to rely on Valgrind for raw leaks.

php_phongo.c
```diff
@@ -2459,7 +2459,138 @@ static char* php_phongo_manager_make_client_hash(const char* uri_string, zval* o
...
+	return true;
+}
+
+static char* php_phongo_concat_handshake_data(const char* default_value, const char* custom_value, size_t custom_value_len)
+{
+	char*  ret;
+	size_t ret_len = strlen(default_value) + 1;
+
+	if (custom_value) {
+		ret_len += custom_value_len + 1;
```
jmikola: `ret_len` already covers the null byte, so the `+1` here is for the pipe separator character, correct?

php_phongo.c
```diff
@@ -2459,7 +2459,138 @@ static char* php_phongo_manager_make_client_hash(const char* uri_string, zval* o
...
+	}
+
+	ret = malloc(ret_len);
+
+	if (custom_value) {
+		snprintf(ret, ret_len, "%s|%s", default_value, custom_value);
```
jmikola: IIRC, `_append_and_truncate` uses ` / ` to concatenate extra strings passed to `mongoc_handshake_data_append`. If we use `|` here, does that mean we'll end up with strings containing something like:

```
mongoc / ext-mongodb:PHP|customDriverOption
```

If so, would it make more sense to use ` / ` here as well? Ideally, the separator string could be a constant in `mongoc-handshake.h` to assure consistency, but that can always be done down the line.

php_phongo.c
alcaeus: The original idea was to have these handled case-insensitive as well, which isn't possible with the `php_array_exists` functions. Since other options (e.g. "context" are case-sensitive as well) I've changed this to no longer use a loop.

php_phongo.c
alcaeus: Ah, good point. I've switched to using `ecalloc` and `efree` throughout and am now using `snprintf` and `strlen` to ensure we're not overflowing the previously allocated buffer. Thanks for pointing this out!

php_phongo.c
alcaeus: Yes, that's correct. I've added a comment to clarify.

php_phongo.c
alcaeus: I based this on the spec but missed the part that allowed deviations from the pipe as delimiter. I've extracted it to our own macro for now until the separator string is defined in libmongoc. As a bonus, this also makes the line above where we add the separator length more understandable. I've changed this to use ` / ` as separator (with spaces) for consistency with libmongoc.

php_phongo.c
```diff
@@ -74,6 +74,8 @@
 
 #define PHONGO_DEBUG_INI "mongodb.debug"
 #define PHONGO_DEBUG_INI_DEFAULT ""
+#define PHONGO_METADATA_SEPARATOR " / "
+#define PHONGO_METADATA_SEPARATOR_LEN strlen(PHONGO_METADATA_SEPARATOR)
```
jmikola: Since `PHONGO_METADATA_SEPARATOR` is a string literal, `(sizeof(PHONGO_METADATA_SEPARATOR) - 1)` would be more commonly used in C. I believe PHP itself does that for various length constants.

This is definitely a nit, as I expect a compiler might optimize call to `strlen` out anyway.

PR #1119, opened 2020-03-30
PHPC-1561: Allow hinting for delete commands

alcaeus:
https://jira.mongodb.org/browse/PHPC-1561

Adds support for adding hints to the delete commands, including findAndDelete. Spec tests will be added in a PHPLIB pull request.


tests/bulk/bulkwrite-delete_error-001.phpt
```diff
@@ -17,4 +21,7 @@ echo throws(function() use ($bulk) {
 --EXPECTF--
 OK: Got MongoDB\Driver\Exception\InvalidArgumentException
 Expected "collation" option to be array or object, int%S given
+
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected "hint" option to be string, array, or object, int%S given
```
jmikola: Was `int%S` for PHP 5.x compatibility? I think PHP 7.x always uses the same term ("int" or "integer"), so we might be able to do without the `%S` and just replace these across the board.

tests/bulk/bulkwrite-update_error-008.phpt
```diff
@@ -22,5 +23,5 @@ echo throws(function() use ($manager, $bulk) {
 <?php exit(0); ?>
 --EXPECT--
 OK: Got MongoDB\Driver\Exception\BulkWriteException
-Bulk write failed due to previous MongoDB\Driver\Exception\RuntimeException: The selected server does not support hint for update
+BSON field 'update.updates.hint' is an unknown field.
```
jmikola: Why does "Bulk write failed due to previous" no longer apply here? Did some logic change in libmongoc to no longer detect this?

In that case, what happens if the server version is <= 3.6.0 (I realize the test is skipped, just curious)?

alcaeus: 3.4.0 has a different exception message, which I decided to ignore by only checking for 3.6 and newer.

Since this is no longer a client-side error in libmongoc, `error.domain` will be `MONGOC_ERROR_SERVER`, which means that we won't throw an exception for the original error ([see source](https://github.com/mongodb/mongo-php-driver/blob/0f7fb930d83f28cc07ef885696b3fc33b94be137/php_phongo.c#L756..L758)). This also means that we won't wrap that previous exception but rather throw a BulkWriteException with the original error message ([see source](https://github.com/mongodb/mongo-php-driver/blob/0f7fb930d83f28cc07ef885696b3fc33b94be137/php_phongo.c#L773..L775)).

This was changed in libmongoc to satisfy spec tests which make sure that server-side errors are thrown on servers > 3.4 that don't support the `hint` option.

tests/bulk/bulkwrite-delete_error-001.phpt
alcaeus: I'll check and replace these in a separate pull request. Good point 👍

PR #1125, opened 2020-04-22
PHPC-1591: Add tests for allowed cases when authSource is given but username is not

alcaeus:
https://jira.mongodb.org/browse/PHPC-1591

Note that the suggested test in the ticket was not added as the C driver currently does not pass that test.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1426,7 +1426,7 @@ static bool php_phongo_uri_finalize_auth(mongoc_uri_t* uri) /* {{{ */
 			}
 		}
 	} else if (require_auth) {
-		if (!username || strcmp(username, "") == 0) {
+		if (source && strcmp(source, "$external") != 0 && (!username || strcmp(username, "") == 0)) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Failed to parse URI options: Default authentication mechanism requires username.");
 			return false;
 		}
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -1382,11 +1382,21 @@ static const char* php_phongo_bson_type_to_string(bson_type_t type) /* {{{ */
 
 static bool php_phongo_uri_finalize_auth(mongoc_uri_t* uri) /* {{{ */
 {
+	const bson_t* credentials = mongoc_uri_get_credentials(uri);
+	bson_iter_t   iter;
+	const char*   source       = NULL;
+	const char*   username     = mongoc_uri_get_username(uri);
+	bool          require_auth = username != NULL;
+
+	if (bson_iter_init_find_case(&iter, credentials, MONGOC_URI_AUTHSOURCE)) {
+		source       = bson_iter_utf8(&iter, NULL);
+		require_auth = true;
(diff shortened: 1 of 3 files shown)


tests/manager/manager-ctor-auth_mechanism-error-001.phpt
```diff
@@ -33,6 +33,14 @@ echo throws(function() {
     new MongoDB\Driver\Manager('mongodb://localhost:27017/', ['password' => 'password', 'authMechanism' => 'MONGODB-X509']);
 }, "MongoDB\Driver\Exception\InvalidArgumentException"), "\n";
 
+echo throws(function() {
+    new MongoDB\Driver\Manager('mongodb://localhost:27017/?authSource=foo');
+}, "MongoDB\Driver\Exception\InvalidArgumentException"), "\n";
```
jmikola, line 38: I don't understand how this passes. The validation logic was added in `php_phongo_uri_finalize_auth`, but that is only called from `php_phongo_apply_options_to_uri`, which itself returns early if there are no options to process. That suggests that this case would depend on [CDRIVER-3618](https://jira.mongodb.org/browse/CDRIVER-3618) where libmongoc would apply the validation to the final `mongoc_uri_t`.

Do let me know if I'm missing something.

php_phongo.c
```diff
@@ -1382,11 +1382,21 @@ static const char* php_phongo_bson_type_to_string(bson_type_t type) /* {{{ */
...
 {
+	const bson_t* credentials = mongoc_uri_get_credentials(uri);
+	bson_iter_t   iter;
+	const char*   source       = NULL;
+	const char*   username     = mongoc_uri_get_username(uri);
+	bool          require_auth = username != NULL;
+
+	if (bson_iter_init_find_case(&iter, credentials, MONGOC_URI_AUTHSOURCE)) {
+		source       = bson_iter_utf8(&iter, NULL);
+		require_auth = true;
```
jmikola, line 1393: [DRIVERS-949](https://jira.mongodb.org/browse/DRIVERS-949) includes a test case where a username is _not_ required if `authSource` is `$external`. I don't think this accounts for that, so I'd suggest manually adding that test (since we can't use the auth spec tests directly) and revising this to relax validation.

While the username is optional for both MONGODB-X509 and MONGODB-IAM mechanisms, it looks like the test only cares about `authSource` since the mechanism is never specified.

tests/manager/manager-ctor-auth_mechanism-error-001.phpt
alcaeus: The exception is thrown in `php_phongo_make_uri` due to not receiving a `mongoc_uri_t` from `mongoc_uri_new_with_error`. The "Default authentication mechanism requires username" messages originates [in libmongoc](https://github.com/mongodb/mongo-c-driver/blob/2307b56f4f50d6d65409af2f2cfb7ab51bff1124/src/libmongoc/src/mongoc/mongoc-uri.c#L1392..L1397). This particular code was added fairly recently (September 2019), but was never copied over to PHPC.

PR #1128, opened 2020-04-29
PHPC-1131, PHPC-1614, PHPC-1615, PHPC-1616: Optimize AppVeyor and Windows build fixes

jmikola:
https://jira.mongodb.org/browse/PHPC-1131
https://jira.mongodb.org/browse/PHPC-1614
https://jira.mongodb.org/browse/PHPC-1615
https://jira.mongodb.org/browse/PHPC-1616


php_phongo.c
```diff
@@ -1291,7 +1291,7 @@ void php_phongo_write_concern_to_zval(zval* retval, const mongoc_write_concern_t
 	}
 
 	if (wtimeout != 0) {
-#if SIZEOF_LONG == 4
+#if SIZEOF_ZEND_LONG == 4
```
jmikola, line 1294: @sgolemon @auroraeosrose Am I correct in thinking we should have never relied on SIZEOF_LONG in PHP 7? This could explain why some of our tests involving 64-bit integers never quite worked correctly on Windows, since we skip based on `PHP_INT_SIZE` (to determine if we're dealing with 32-bit or 64-bit ints), and that constant is defined as `SIZEOF_ZEND_LONG` instead of `SIZEOF_LONG`.

tests/manager/manager-ctor-auto_encryption-001.phpt
```diff
@@ -2,6 +2,7 @@
 MongoDB\Driver\Manager::__construct(): auto encryption options
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_appveyor(); /* AppVeyor does not have mongocryptd installed */ ?>
```
alcaeus, line 5: On travis-ci we install MongoDB Enterprise, and I thought I had changed that on appveyor as well. If we're not already installing the enterprise package, can you please create a tracking ticket for this?

config.w32
```diff
@@ -91,15 +108,15 @@ if (PHP_MONGODB != "no") {
...
-  ADD_SOURCES(configure_module_dirname + "/src/MongoDB/Exception", "AuthenticationException.c BulkWriteException.c CommandException.c ConnectionException.c ConnectionTimeoutException.c EncryptionException.c Exception.c ExecutionTimeoutException.c InvalidArgumentException.c LogicException.c RuntimeException.c ServerException.c SSLConnectionException.c UnexpectedValueException.c WriteException.c", "mongodb");
-  ADD_SOURCES(configure_module_dirname + "/src/MongoDB/Monitoring", "CommandFailedEvent.c CommandStartedEvent.c CommandSubscriber.c CommandSucceededEvent.c Subscriber.c functions.c", "mongodb");
-  ADD_SOURCES(configure_module_dirname + "/src/libmongoc/src/common", PHP_MONGODB_COMMON_SOURCES, "mongodb");
-  ADD_SOURCES(configure_module_dirname + "/src/libmongoc/src/libbson/src/bson", PHP_MONGODB_BSON_SOURCES, "mongodb");
-  ADD_SOURCES(configure_module_dirname + "/src/libmongoc/src/libbson/src/jsonsl", PHP_MONGODB_JSONSL_SOURCES, "mongodb");
-  ADD_SOURCES(configure_module_dirname + "/src/libmongoc/src/libmongoc/src/mongoc", PHP_MONGODB_MONGOC_SOURCES, "mongodb");
+  MONGODB_ADD_SOURCES("/src", "bson.c bson-encode.c");
```
alcaeus, line 111: Nice!

tests/manager/manager-ctor-auto_encryption-001.phpt
jmikola: IIRC, AppVeyor never had `mongod` running at all. I don't see any reference to it in the [previous build scripts](https://github.com/mongodb/mongo-php-driver/tree/master/.appveyor). If you look back at our old builds I think you'll find that only tests with no server requirement were actually executed.

config.w32
jmikola: The real impetus for this was needing to recursively create the object path. It was much easier to just create our own function :)

tests/manager/manager-ctor-auto_encryption-001.phpt
alcaeus: Ah, yeah. That explains it 👍

php_phongo.c
sgolemon: There's sort of two parts to that answer.

1/ `sizeof(long) == sizeof(int)` on Windows as an implementation detail.  The standard allows long to be as big _or larger than_ int.  Windows makes them equal.

It's worth noting that pre PHP 7, PHP would use `long` as the underlying PHP integer type. We changed it to `long long` to make this more consistent across platforms and architectures.

2/ `sizeof(long long)` is guaranteed to be `>= 8` by the C specification, so this `#if` block should never succeed.

http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf `§5.2.4.2.1` (Top of page 23)

— minimum value for an object of type long long int
LLONG_MIN -9223372036854775807 // −(2^63 − 1)
— maximum value for an object of type long long int
LLONG_MAX +9223372036854775807 // (2^63 − 1)

That said, `ext/mysqlnd/mysqlnd_portability.h` has a similar check, so either my understanding is incorrect, or the author of that statement's understanding is.

What's the observed issue trying to be fixed here?

sgolemon: > What's the observed issue trying to be fixed here?

Nevermind, I just looked at the larger patch.  This is about serializing data into PHP types.
In this case, use: `#if SIZEOF_ZEND_LONG >= 8` to be more precise about the type being checked.  This will cover the `long long` case of PHP 7 and the `long` case of PHP 5.

jmikola: The root objective is determining the size of PHP's native integer type, and I wanted to confirm that `SIZEOF_ZEND_LONG` was the most reliable constant for doing so. In the past, I recall we've had some occasional test failures in Windows related to integer precision (e.g. thinking we had 64-bit integers but finding out PHP was limited to 32-bit), and I think that was probably due to us consulting `SIZEOF_LONG` and the platform's size not matching what PHP was actually configured to use.

Based on [`ZEND_ENABLE_ZVAL_LONG64`](http://lxr.jpauli.tech/source/xref/php-7.4/Zend/zend_long.h#29), I don't see any evidence that PHP 7 is using `long long` (at least not directly). The definition branches on either using `int64_t` or `int32_t`. I would expect `SIZEOF_ZEND_LONG` to either be 8 or 4, respectively; however, I don't see any harm in using `>= 8` for detecting the "not 32-bit case".

jmikola: > This will cover the `long long` case of PHP 7 and the `long` case of PHP 5.

Quick follow-up: I don't believe `SIZEOF_ZEND_LONG` exists in PHP 5, and we also dropped support for PHP 5 recently.

sgolemon: > we also dropped support for PHP 5 recently

Then I'd say forget the `#if` checks altogether.  As of PHP 7, we guarantee that PHP integers are at least 64bit (signed).

https://wiki.php.net/rfc/size_t_and_int64_next

> I don't believe SIZEOF_ZEND_LONG exists in PHP 5

I could have sworn it did, but looking again I see you're right.

sgolemon: > Then I'd say forget the #if checks altogether. 

Per slack, keep the check.  I have a tiny doubt now, and worst case keeping it is a little dead code.

jmikola: Just wanted to address this for the benefit of anyone else following along:

> In this case, use: #if SIZEOF_ZEND_LONG >= 8 to be more precise about the type being checked.

In `phongo_compat.h`, we have an `#else` case when `SIZEOF_ZEND_LONG` is neither 8 nor 4, which raises a build error:

```
#error Unsupported architecture (integers are neither 32-bit nor 64-bit)
```

Given that, I'm going to leave the `== 4` and `== 8` conditions as-is for now. We can revisit when/if PHP gets support for larger integers.

PR #1130, opened 2020-05-12
PHPC-1313: Support Zstandard compression

jmikola:
https://jira.mongodb.org/browse/PHPC-1313

### DIFF
--- a/scripts/autotools/libmongoc/CheckCompression.m4
+++ b/scripts/autotools/libmongoc/CheckCompression.m4
@@ -1,6 +1,7 @@
 found_snappy="no"
 found_zlib="no"
 bundled_zlib="no"
+found_zstd="no"
 
 PKG_CHECK_MODULES([PHP_MONGODB_SNAPPY],[snappy],[
   PHP_MONGODB_BUNDLED_CFLAGS="$PHP_MONGODB_BUNDLED_CFLAGS $PHP_MONGODB_SNAPPY_CFLAGS"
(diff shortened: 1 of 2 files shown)


scripts/autotools/libmongoc/CheckCompression.m4
```diff
@@ -47,7 +48,27 @@ AS_IF([test "$found_zlib" != "yes"],[
...
 ])
 
-if test "$found_snappy" = "yes" -o "$found_zlib" = "yes" -o "$bundled_zlib" = "yes"; then
+PKG_CHECK_MODULES([PHP_MONGODB_ZSTD],[libzstd],[
+  PHP_MONGODB_BUNDLED_CFLAGS="$PHP_MONGODB_BUNDLED_CFLAGS $PHP_MONGODB_ZSTD_CFLAGS"
+  PHP_EVAL_LIBLINE([$PHP_MONGODB_ZSTD_LIBS],[MONGODB_SHARED_LIBADD])
+  found_zstd="yes"
+],[
+  PHP_CHECK_LIBRARY([zstd],
+                    [ZSTD_compress],
```
jmikola, line 57: It was trivial to copy the fallback detection when `pkg-config` is not available. This API comes from the [zstd documentation](http://facebook.github.io/zstd/zstd_manual.html) and you'll find the example closely resembles what we do for snappy above.

One difference is that the `pkg-config` name is "libzstd", while the shared library itself is simply "zstd".

scripts/autotools/libmongoc/CheckCompression.m4
```diff
@@ -59,10 +80,14 @@ if test "$found_snappy" = "yes" -o "$found_zlib" = "yes" -o "$bundled_zlib" = "y
   else
     AC_SUBST(MONGOC_ENABLE_COMPRESSION_ZLIB, 0)
   fi
+  if test "$found_zstd" = "yes"; then
+    AC_SUBST(MONGOC_ENABLE_COMPRESSION_ZSTD, 1)
```
jmikola, line 84: While we don't have any build flags for compression (we just link whatever is available) nor do we have Evergreen environments w/ and w/o these libs, I did both detection branches (`pkg-config` and the fallback) locally.

On Ubuntu, the relevant packages are `libzstd-dev` and `libzstd1`. I first ran `configure` and verified that the `pkg-config` branch picked up "libzstd" and added `-lzstd` to our link flags. Relevant `configure` output is simply:

```
checking for PHP_MONGODB_ZSTD... yes
```

I then modified the M4 script to fail the `pkg-config` check by adding a typo in the "libzstd" package name to verify that the `PHP_CHECK_LIBRARY` and `AC_CHECK_HEADER` fallback successfully added the same link flag. The `configure` output in this case changes to:

```
checking for PHP_MONGODB_ZSTD... no
checking for ZSTD_compress in -lzstd... yes
checking zstd.h usability... yes
checking zstd.h presence... yes
checking for zstd.h... yes
```

PR #1133, opened 2020-05-14
PHPC-1598: Objects with get_properties handlers should have get_gc delegate to zend_std_get_properties

jmikola:
https://jira.mongodb.org/browse/PHPC-1598

The standard `get_gc` handler [defers to `get_properties`](http://lxr.jpauli.tech/source/xref/php-7.4/Zend/zend_object_handlers.c#118) to collect other zvals to inspect for garbage collection. ReadPreference has no properties beyond a `mongoc_read_concern_t` pointer, which is not a zval, but we do use `get_properties` to report fields for `var_export()` (since [PHPC-850](https://jira.mongodb.org/browse/PHPC-850)). This is problematic if `get_properties` is invoked during GC, since it will return properties that will already be freed by ReadPreference's `free_object` handler (via `FREE_HASHTABLE`). This can be avoided by defining our own `get_gc` handler and correctly reporting that there are no internal properties for garbage collection. Instead, `get_gc` should delegate to `zend_std_get_properties` (also relevant to our BSON classes).


jmikola:
@alcaeus: I updated the PR with an additional regression test to demonstrate that the current implementation of the BSON `get_gc` handlers was incorrect. This took a bit longer than expected as I had to wrap my head around PHP's garbage collector.

Please give the revised commit a look when you get a chance and re-review.

src/BSON/Binary.c
```diff
@@ -352,9 +352,8 @@ static int php_phongo_binary_compare_objects(zval* o1, zval* o2) /* {{{ */
 static HashTable* php_phongo_binary_get_gc(zval* object, zval** table, int* n) /* {{{ */
```
alcaeus, line 352: Given that all `*_get_gc` functions do exactly the same, would it make sense to create a generic `php_phongo_get_gc` function and use that for all classes that need this functionality?

jmikola: Good idea. I've amended the patch, which reduced quite a bit of copypasta.

This change now has us define the common `get_gc` handler for all of our objects, which I don't expect will be a problem. It's also more future proof in the event that other classes define a `get_properties` handler down the line.

PR #1135, opened 2020-05-19
Bump to latest libmongoc version

alcaeus:
This brings in support for speculative authentication and allows us to work on server hedged reads.


src/LIBMONGOC_VERSION_CURRENT
```diff
@@ -1 +1 @@
-1.17.0-beta
+1.18.0-20200519+gitc63f92c305
```
alcaeus, line 1: Switching to 1.18-dev is necessary as features still need to be back ported into 1.17 from master. Since we'll require 1.17 when this is released, I'm not changing the requirement in config.m4 for the time being.

jmikola:
https://ci.appveyor.com/project/jmikola/mongo-php-driver/builds/32981488/job/4qvaal6ebs9fvugm is showing link failures. You may need to resync the source files in both `config.m4` and `config.w32`.

Travis also has various failures, but those look to be related to "Unable to locate package lcov" failing very early in the build process.

alcaeus:
"It worked on my machine" 😂

I'll check out the failures tomorrow.

PR #1136, opened 2020-05-20
PHPC-1571: Support server hedged reads via read preference

alcaeus:
https://jira.mongodb.org/browse/PHPC-1571

### DIFF
--- a/src/MongoDB/ReadPreference.c
+++ b/src/MongoDB/ReadPreference.c
@@ -40,7 +40,7 @@ zend_class_entry* php_phongo_readpreference_ce;
  * An exception will be thrown on error. */
 static bool php_phongo_readpreference_init_from_hash(php_phongo_readpreference_t* intern, HashTable* props) /* {{{ */
 {
-	zval *mode, *tagSets, *maxStalenessSeconds;
+	zval *mode, *tagSets, *maxStalenessSeconds, *hedge;
 
 	if ((mode = zend_hash_str_find(props, "mode", sizeof("mode") - 1)) && Z_TYPE_P(mode) == IS_STRING) {
 		if (strcasecmp(Z_STRVAL_P(mode), PHONGO_READ_PRIMARY) == 0) {
@@ -107,6 +107,31 @@ static bool php_phongo_readpreference_init_from_hash(php_phongo_readpreference_t
 			mongoc_read_prefs_set_max_staleness_seconds(intern->read_preference, Z_LVAL_P(maxStalenessSeconds));
 		} else {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "%s initialization requires \"maxStalenessSeconds\" field to be integer", ZSTR_VAL(php_phongo_readpreference_ce->name));
+			goto failure;
+		}
+	}
+
+	if ((hedge = zend_hash_str_find(props, "hedge", sizeof("hedge") - 1))) {
+		if (Z_TYPE_P(hedge) == IS_ARRAY || Z_TYPE_P(hedge) == IS_OBJECT) {
+			bson_t* hedge_doc = bson_new();
+
+			if (mongoc_read_prefs_get_mode(intern->read_preference) == MONGOC_READ_PRIMARY) {
+				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "%s initialization requires \"hedge\" field to not be present with \"primary\" mode", ZSTR_VAL(php_phongo_readpreference_ce->name));
+				bson_destroy(hedge_doc);
+				goto failure;
+			}
+
+			php_phongo_zval_to_bson(hedge, PHONGO_BSON_NONE, hedge_doc, NULL);
+
+			if (EG(exception)) {
+				bson_destroy(hedge_doc);
+				goto failure;
+			}
+
+			mongoc_read_prefs_set_hedge(intern->read_preference, hedge_doc);
+			bson_destroy(hedge_doc);
+		} else {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "%s initialization requires \"hedge\" field to be an array or object", ZSTR_VAL(php_phongo_readpreference_ce->name));
 			goto failure;
 		}
 	}
@@ -238,6 +263,33 @@ static PHP_METHOD(ReadPreference, __construct)
 		mongoc_read_prefs_set_max_staleness_seconds(intern->read_preference, maxStalenessSeconds);
 	}
 
+	if (options && php_array_exists(options, "hedge")) {
+		zval* hedge = php_array_fetchc(options, "hedge");
+
+		if (Z_TYPE_P(hedge) == IS_ARRAY || Z_TYPE_P(hedge) == IS_OBJECT) {
+			bson_t* hedge_doc = bson_new();
+
+			if (mongoc_read_prefs_get_mode(intern->read_preference) == MONGOC_READ_PRIMARY) {
+				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "hedge may not be used with primary mode");
+				bson_destroy(hedge_doc);
+				return;
+			}
+
+			php_phongo_zval_to_bson(hedge, PHONGO_BSON_NONE, hedge_doc, NULL);
+
+			if (EG(exception)) {
+				bson_destroy(hedge_doc);
+				return;
+			}
+
+			mongoc_read_prefs_set_hedge(intern->read_preference, hedge_doc);
+			bson_destroy(hedge_doc);
+		} else {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "%s initialization requires \"hedge\" field to be an array or object", ZSTR_VAL(php_phongo_readpreference_ce->name));
+			return;
+		}
+	}
+
 	if (!mongoc_read_prefs_is_valid(intern->read_preference)) {
 		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Read preference is not valid");
 		return;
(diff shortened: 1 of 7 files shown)


src/MongoDB/ReadPreference.c
```diff
@@ -264,6 +312,38 @@ static PHP_METHOD(ReadPreference, __set_state)
...
+		return;
+	}
+
+	hedge = mongoc_read_prefs_get_hedge(intern->read_preference);
+
+	if (!bson_empty0(hedge)) {
+		php_phongo_bson_state state;
+
+		PHONGO_BSON_INIT_STATE(state);
+		state.map.root_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
```
alcaeus, line 334: Not sure if there's prior art for this: we want to return a hedge document. Do we use:
* a PHP `array`, which is the current solution:
```
array(1) {
  ["enabled"]=>
  bool(true)
}
```
* an object since it's a document?
```
array(stdClass)#%d (%d) {
  ["enabled"]=>
  bool(true)
}
```

Type map support is definitely not warranted here, and with arrays being what the average PHP developer would use here I would keep the current array return. PHONGO_BSON_INIT_DEBUG_STATE would use that for arrays and documents as well, in the off-chance that the hedge document would get more complex in the future.

src/MongoDB/ReadPreference.c
```diff
@@ -217,25 +242,48 @@ static PHP_METHOD(ReadPreference, __construct)
...
-		if (maxStalenessSeconds != MONGOC_NO_MAX_STALENESS) {
-			if (maxStalenessSeconds < MONGOC_SMALLEST_MAX_STALENESS_SECONDS) {
-				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected maxStalenessSeconds to be >= %d, %" PHONGO_LONG_FORMAT " given", MONGOC_SMALLEST_MAX_STALENESS_SECONDS, maxStalenessSeconds);
-				return;
-			}
-			if (maxStalenessSeconds > INT32_MAX) {
-				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected maxStalenessSeconds to be <= %" PRId32 ", %" PHONGO_LONG_FORMAT " given", INT32_MAX, maxStalenessSeconds);
-				return;
+	if (options) {
+		if (php_array_exists(options, "maxStalenessSeconds")) {
```
jmikola, line 246: For the record, I'm fine with adding an extra `if (options && php_array_exists(options, "hedge"))` if you'd rather keep the diff small, given it's only two instances and a non-zero check is cheap. No objections if you prefer this, though.

src/MongoDB/ReadPreference.c
```diff
@@ -217,25 +242,48 @@ static PHP_METHOD(ReadPreference, __construct)
...
+
+				if (EG(exception)) {
+					bson_destroy(hedge_doc);
+					return;
+				}
+
+				mongoc_read_prefs_set_hedge(intern->read_preference, hedge_doc);
+				bson_destroy(hedge_doc);
+			} else {
+				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "%s initialization requires \"hedge\" field to be an array of object", ZSTR_VAL(php_phongo_readpreference_ce->name));
```
jmikola, line 283: "array or object"

src/MongoDB/ReadPreference.c
jmikola: I think returning an object is fine. On the off chance someone actually needs to use this method and work with its return value, it's not too difficult to cast to an array or apply their own type map. Prior art would be `Javascript::getScope()` and the `getInfo()` methods for WriteConcernError and WriteError. I believe all of those code paths use `PHONGO_BSON_INIT_STATE` via `php_phongo_bson_to_zval`. 

`Server::getInfo()` may be an exception for returning an array, but we've historically never documented that it is a copy of the most recent `isMaster` document. It's always been thought of as a debug method.

src/MongoDB/ReadPreference.c
```diff
@@ -264,6 +312,38 @@ static PHP_METHOD(ReadPreference, __set_state)
...
+	hedge = mongoc_read_prefs_get_hedge(intern->read_preference);
+
+	if (!bson_empty0(hedge)) {
+		php_phongo_bson_state state;
+
+		PHONGO_BSON_INIT_STATE(state);
+		state.map.root_type = PHONGO_TYPEMAP_NATIVE_ARRAY;
+
+		if (!php_phongo_bson_to_zval_ex(bson_get_data(hedge), hedge->len, &state)) {
+			zval_ptr_dtor(&state.zchild);
```
jmikola, line 337: Exception is already thrown in this case, correct?

src/MongoDB/ReadPreference.c
```diff
@@ -398,6 +480,19 @@ static HashTable* php_phongo_readpreference_get_properties_hash(zval* object, bo
 		zend_hash_str_update(props, "maxStalenessSeconds", sizeof("maxStalenessSeconds") - 1, &z_max_ss);
 	}
 
+	if (!bson_empty0(hedge)) {
+		php_phongo_bson_state state;
+
+		PHONGO_BSON_INIT_DEBUG_STATE(state);
```
jmikola, line 486: This can also use `PHONGO_BSON_INIT_STATE`, like `php_phongo_javascript_get_properties_hash` does for `scope`.

src/MongoDB/ReadPreference.c
```diff
@@ -465,6 +562,19 @@ static PHP_METHOD(ReadPreference, serialize)
 		ADD_ASSOC_LONG_EX(&retval, "maxStalenessSeconds", maxStalenessSeconds);
 	}
 
+	if (!bson_empty0(hedge)) {
+		php_phongo_bson_state state;
+
+		PHONGO_BSON_INIT_DEBUG_STATE(state);
```
jmikola, line 568: Serialization would also benefit from using a document (even more so than `get_properties`) to guarantee type fidelity when round-tripping. I don't expect this is an issue for the current structure of the `hedge` document but it could become one down the line.

tests/readPreference/readpreference-ctor_error-007.phpt
```diff
@@ -0,0 +1,18 @@
...
+echo throws(function() {
+    new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_PRIMARY, null, ['hedge' => ['enabled' => true]]);
+}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Read preference is not valid
```
jmikola, line 17: I think it'd be user-friendly to copy over the exception you have in `php_phongo_readpreference_init_from_hash` so we get a more meaningful error message. This would be consistent with what the constructor was already doing for `maxStalenessSeconds` and `tags`.

I consider the exception from `mongoc_read_prefs_is_valid` to be more of a fail safe in case we missed anything (or someone is using an older PHPC with a newer libmongoc).

While writing this comment, I also realized that `php_phongo_readpreference_init_from_hash` doesn't include a final check for `mongoc_read_prefs_is_valid`. We also don't appear to use the WC valid check at all in WriteConcern.c. I've opened [PHPC-1629](https://jira.mongodb.org/browse/PHPC-1629) to track this.

tests/readPreference/readpreference-set_state_error-001.phpt
```diff
@@ -53,4 +62,8 @@ OK: Got MongoDB\Driver\Exception\InvalidArgumentException
 MongoDB\Driver\ReadPreference initialization requires "maxStalenessSeconds" integer field to be >= 90
 OK: Got MongoDB\Driver\Exception\InvalidArgumentException
 MongoDB\Driver\ReadPreference initialization requires "maxStalenessSeconds" array field to not be present with "primary" mode
```
jmikola, line 64: `"maxStalenessSeconds" array field` looks to be old copypasta from the `tags` exception. Would you mind correcting this in a separate commit? I don't think we need to track it in an issue.

src/MongoDB/ReadPreference.c
alcaeus: `php_phongo_bson_to_zval_ex` throws an exception and returns `false` on error.

src/MongoDB/ReadPreference.c
alcaeus: Changed to return objects in all cases.

src/MongoDB/ReadPreference.c
alcaeus: Reverted for a smaller diff 👍

src/MongoDB/ReadPreference.c
```diff
@@ -269,6 +269,12 @@ static PHP_METHOD(ReadPreference, __construct)
 		if (Z_TYPE_P(hedge) == IS_ARRAY || Z_TYPE_P(hedge) == IS_OBJECT) {
 			bson_t* hedge_doc = bson_new();
 
+			if (mongoc_read_prefs_get_mode(intern->read_preference) == MONGOC_READ_PRIMARY) {
+				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "hedge may not be used with primary mode");
+				bson_destroy(hedge_doc);
+				return;
```
jmikola, line 275: Does this belong in the other commit? This commit is for `maxStalenessSeconds`.

src/MongoDB/ReadPreference.c
```diff
@@ -91,6 +91,10 @@ static bool php_phongo_readpreference_init_from_hash(php_phongo_readpreference_t
 
 	if ((maxStalenessSeconds = zend_hash_str_find(props, "maxStalenessSeconds", sizeof("maxStalenessSeconds") - 1))) {
 		if (Z_TYPE_P(maxStalenessSeconds) == IS_LONG) {
+			if (mongoc_read_prefs_get_mode(intern->read_preference) == MONGOC_READ_PRIMARY) {
+				phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "%s initialization requires \"maxStalenessSeconds\" field to not be present with \"primary\" mode", ZSTR_VAL(php_phongo_readpreference_ce->name));
+				goto failure;
+			}
```
jmikola, line 97: Do we allow `-1` or `0` to be specified for `maxStalenessSeconds` as an "unset" value?

Reading https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#maxstalenessseconds, an error is only thrown for a positive value with mode `primary`. Obviously, any positive value less than `smallestMaxStalenessSeconds` would be rejected, which is why I'm asking about handling a negative/zero for "unset".

It looks like our construct has a guard for `MONGOC_NO_MAX_STALENESS` (as do the functions that populate properties), so I wonder if we should add that somewhere above this block as well for consistency. This is admittedly a nit, as normal operation wouldn't trigger this unless users were crafting their own serialization or set_state stuff. Perhaps this can go in the condition for `Z_TYPE_P(maxStalenessSeconds) == IS_LONG` with logical OR?

src/MongoDB/ReadPreference.c
```diff
@@ -353,6 +436,7 @@ static HashTable* php_phongo_readpreference_get_properties_hash(zval* object, bo
 	HashTable*                   props;
 	const char*                  modeString = NULL;
 	const bson_t*                tags;
+	const bson_t*                hedge;
```
jmikola, line 439: Should `PHONGO_GET_PROPERTY_HASH_INIT_PROPS` use `4` instead of `3` below? I have some recollection of seeing that in an earlier version of this PR, but I could be confusing it with something else.

EDIT: Looks like I was confusing this with `array_init_size` in the `serialize` method.

src/MongoDB/ReadPreference.c
alcaeus: You are correct. Fixed.

src/MongoDB/ReadPreference.c
alcaeus: That `if` won't work, as the `else` path throws an exception. I've added a separate check for `MONGOC_NO_MAX_STALENESS` instead.

src/MongoDB/ReadPreference.c
alcaeus: Ah, yes. Not sure if it makes a difference, but I updated it for the sake of consistency.

jmikola: IIRC, it just serves as a hint for the initial allocation. You can add more elements and PHP will re-size the HashTable as needed.

PR #1140, opened 2020-05-29
Remove dead ZEND_ENGINE_3 checks

alcaeus:
These checks are no longer relevant as we've dropped support for PHP 5.6, which was the last version to use Zend Engine 2. PHP 7 always uses ZEND_ENGINE_3, making this check useless.

### DIFF
--- a/src/contrib/php_array_api.h
+++ b/src/contrib/php_array_api.h
@@ -25,19 +25,11 @@
 #include "zend_hash.h"
 #include "zend_list.h"
 
-#ifdef ZEND_ENGINE_3
 # define PAA_LENGTH_ADJ(l) (l)
 # define PAA_SYM_EXISTS zend_symtable_str_exists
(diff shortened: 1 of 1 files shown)


src/contrib/php_array_api.h
```diff
@@ -25,19 +25,11 @@
 #include "zend_hash.h"
 #include "zend_list.h"
 
-#ifdef ZEND_ENGINE_3
```
jmikola, line 28: `php_array_api.h` is vendored from https://github.com/sgolemon/php-array-api, so I'm not too concerned with this code lingering. If you want to contribute these changes upstream first (since this isn't an urgent bug fix), that would allow us to keep things in sync. Sara may also want to make a tag before removing PHP 5.x macros.

alcaeus: In that case, we can keep it in for the time being. Closing this PR in favour of updating the file later.

PR #1144, opened 2020-06-30
PHPC-1639: Manager::executeCommand should not inherit read preference

alcaeus:
https://jira.mongodb.org/browse/PHPC-1639

~I've fixed this in 1.8 for the time being, but can re-target this PR on 1.7 for a July patch release. I've implemented this in 1.8 for the time being, since merging up from 1.7 to 1.8 would cause additional work due to the removal of `TSRMLS` macros in function signatures.~

### DIFF
--- a/src/MongoDB/Manager.c
+++ b/src/MongoDB/Manager.c
@@ -271,7 +271,7 @@ static void php_phongo_manager_prep_uri_options(zval* options TSRMLS_DC) /* {{{ 
  *
  * On success, server_id will be set and the function will return true;
  * otherwise, false is returned and an exception is thrown. */
-static bool php_phongo_manager_select_server(bool for_writes, zval* zreadPreference, zval* zsession, mongoc_client_t* client, uint32_t* server_id TSRMLS_DC) /* {{{ */
+static bool php_phongo_manager_select_server(bool for_writes, bool inherit_read_preference, zval* zreadPreference, zval* zsession, mongoc_client_t* client, uint32_t* server_id TSRMLS_DC) /* {{{ */
 {
 	mongoc_server_description_t* selected_server;
 	const mongoc_read_prefs_t*   read_preference = NULL;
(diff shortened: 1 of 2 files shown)


src/MongoDB/Manager.c
```diff
@@ -289,7 +289,11 @@ static bool php_phongo_manager_select_server(bool for_writes, zval* zreadPrefere
...
 	if (!for_writes) {
-		read_preference = zreadPreference ? phongo_read_preference_from_zval(zreadPreference TSRMLS_CC) : mongoc_client_get_read_prefs(client);
+		if (zreadPreference) {
+			read_preference = phongo_read_preference_from_zval(zreadPreference TSRMLS_CC);
+		} else if (inherit_read_preference) {
+			read_preference = mongoc_client_get_read_prefs(client);
+		}
 	}
 
 	selected_server = mongoc_client_select_server(client, for_writes, read_preference, &error);
```
jmikola, line 299: According to the [`mongoc_client_select_server` docs](http://mongoc.org/libmongoc/current/mongoc_client_select_server.html), the client's read preference is used if `read_preference` is `NULL`. That said, I traced the function calls through to `mongoc_topology_description_suitable_servers`:

 * mongoc_client_select_server
 * mongoc_topology_select
 * mongoc_topology_select_server_id
 * mongoc_topology_description_select
 * mongoc_topology_description_suitable_servers

Therein, we have:

```c
mongoc_read_mode_t read_mode = mongoc_read_prefs_get_mode (read_pref);
```

If `read_pref` is `NULL` here, a primary RP is used. This would seem to conflict with the `mongoc_client_select_server` docs, unless I'm missing something. Maybe @kevinAlbs can chime in.

Either libmongoc has a bug in its implementation or the docs need to be corrected, but I think the safest thing to do in PHPC is to default to a primary read preference explicitly.

tests/manager/manager-executeCommand-006.phpt
```diff
@@ -0,0 +1,26 @@
+--TEST--
+MongoDB\Driver\Manager::executeCommand() takes a read preference in options array
```
jmikola, line 2: Test name should be "does not inherit read preference"

src/MongoDB/Manager.c
kevinAlbs: I agree with your assessment. `mongoc_client_select_server` does not inherit the client's  read preference, but defaults to primary read preference. The documentation does not match the implementation. I created https://jira.mongodb.org/browse/CDRIVER-3732 to update the documentation and validate with a test.

jmikola: @alcaeus I'll defer to you if you want to hard-code a primary RP here or leave the code as-is. Either way, I'm confident that the included test ensures correctness.

PR #1156, opened 2020-07-29
PHPC-1167: Avoid dangling session pointer in bulk writes

alcaeus:
PHPC-1167

This solution is similar to the cursor solution implemented for PHPC-1151. The bulk write keeps a copy of the session object to ensure it won't go out of scope directly after executing the bulk write. Since we don't have a use-after-free (since the session is not accessed after the bulk write has been executed), there are no tests to guard against regressions.

### DIFF
--- a/src/MongoDB/BulkWrite.c
+++ b/src/MongoDB/BulkWrite.c
@@ -566,6 +566,10 @@ static void php_phongo_bulkwrite_free_object(zend_object* object) /* {{{ */
 	if (intern->collection) {
 		efree(intern->collection);
 	}
+
+	if (!Z_ISUNDEF(intern->session)) {
+		zval_ptr_dtor(&intern->session);
+	}
 } /* }}} */
 
 static zend_object* php_phongo_bulkwrite_create_object(zend_class_entry* class_type) /* {{{ */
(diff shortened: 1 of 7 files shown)


src/MongoDB/BulkWrite.c
```diff
@@ -614,6 +618,13 @@ static HashTable* php_phongo_bulkwrite_get_debug_info(zval* object, int* is_temp
 	ADD_ASSOC_BOOL_EX(&retval, "executed", intern->executed);
 	ADD_ASSOC_LONG_EX(&retval, "server_id", mongoc_bulk_operation_get_hint(intern->bulk));
 
+	if (!Z_ISUNDEF(intern->session)) {
+		ADD_ASSOC_ZVAL_EX(&retval, "session", &intern->session);
+		Z_ADDREF(intern->session);
```
jmikola, line 623: Do we not have any tests that execute a bulk write with an explicit session and then dumps the bulk object? If not, I think we should that as `bulkwrite-debug-002.phpt` (splitting it off since it will require crypto and a live server). Perhaps you want to create a complementary test for the cursor as well, that dumps a cursor after executing a query with an explicit session.

alcaeus: There's a test to check if a cursor prints an explicit session (cursor-session-001.phpt) and omits an implicit session (cursor-session-002.phpt). I've added one for the bulk write object as well.

PR #1165, opened 2020-10-08
[WAIT] PHPC-1645: Allow disabling libmongoc client persistence

alcaeus:
PHPC-1645

I've chosen `disableClientPersistence` as name as it implies that the option should be omitted entirely if persistence is to be enabled. Since we don't canonicalise options before creating the client hash, a manager with driverOptions `[]` will create a different client from a manager with driverOptions `['disableClientPersistence' => false]` due to the different driver options. This is intentional at this time.

### DIFF
--- a/src/MongoDB/Manager.c
+++ b/src/MongoDB/Manager.c
@@ -798,7 +798,13 @@ static void php_phongo_manager_free_object(zend_object* object) /* {{{ */
 	zend_object_std_dtor(&intern->std);
 
 	if (intern->client) {
-		MONGOC_DEBUG("Not destroying persistent client for Manager");
+		if (intern->use_persistent_client) {
+			MONGOC_DEBUG("Not destroying persistent client for Manager");
+		} else {
+			MONGOC_DEBUG("Destroying private client for Manager");
+			mongoc_client_destroy(intern->client);
+		}
+
 		intern->client = NULL;
 	}
 
--- a/tests/manager/manager-ctor-009.phpt
+++ b/tests/manager/manager-ctor-009.phpt
@@ -1,0 +1,38 @@
+--TEST--
+MongoDB\Driver\Manager::__construct() allows disabling client persistence
+--FILE--
+<?php
+
+ini_set('mongodb.debug', 'stderr');
+new MongoDB\Driver\Manager();
+
+// Won't reuse first client instance due to different options
+new MongoDB\Driver\Manager(null, [], ['disableClientPersistence' => false]);
+// Will reuse the previously created client
+new MongoDB\Driver\Manager(null, [], ['disableClientPersistence' => false]);
+
+// Will create a new private client
+new MongoDB\Driver\Manager(null, [], ['disableClientPersistence' => true]);
+
+// Will create another new private client
+new MongoDB\Driver\Manager(null, [], ['disableClientPersistence' => true]);
+ini_set('mongodb.debug', '');
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+%A
+[%s]     PHONGO: DEBUG   > Created client with hash: %s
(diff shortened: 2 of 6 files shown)


tests/manager/manager-ctor-009.phpt
```diff
@@ -0,0 +1,38 @@
...
+ini_set('mongodb.debug', '');
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+%A
+[%s]     PHONGO: DEBUG   > Created client with hash: %s
+[%s]     PHONGO: DEBUG   > Persisted client with hash: %s
+[%s]     mongoc: DEBUG   > Not destroying persistent client for Manager%A
```
alcaeus, line 28: Note that the log domain is `mongoc` here, despite this message originating in ext-mongodb. I'll have to investigate that separately.

jmikola:
> Since we don't canonicalise options before creating the client hash, a manager with driverOptions [] will create a different client from a manager with driverOptions ['disableClientPersistence' => false] due to the different driver options. This is intentional at this time.

I think this makes sense, as silently removing it would mean that we might have to change how we document client persistence. If we ultimately plan to strip out irrelevant options (e.g. `typeMap`) down the line, that will be a better time to address that.

Also, there's no reason at all to specify `false` for this instead of omitting it entirely -- although I'm sure that won't stop someone from doing so.

src/MongoDB/Manager.c
```diff
@@ -798,7 +798,13 @@ static void php_phongo_manager_free_object(zend_object* object) /* {{{ */
 	zend_object_std_dtor(&intern->std);
 
 	if (intern->client) {
-		MONGOC_DEBUG("Not destroying persistent client for Manager");
+		if (intern->use_persistent_client) {
+			MONGOC_DEBUG("Not destroying persistent client for Manager");
+		} else {
+			MONGOC_DEBUG("Destroying private client for Manager");
```
jmikola, line 804: Admittedly a nit because this is internal logging, but I'd suggest removing "private" and rephrasing as "Destroying client for Manager with disableClientPersistence=true" or "Destroying non-persistent client for Manager".

tests/manager/manager-ctor-009.phpt
jmikola: Looks like this has always been the case. You can fix it by adding this at the top of the file _after_ including headers (just in case it conflicts with something in the libmongoc/libbson headers):

```
#undef MONGOC_LOG_DOMAIN
#define MONGOC_LOG_DOMAIN "PHONGO"
```

jmikola (approved):

src/MongoDB/Manager.c
```diff
@@ -798,7 +798,13 @@ static void php_phongo_manager_free_object(zend_object* object) /* {{{ */
...
+			mongoc_client_destroy(intern->client);
```
jmikola, line 805: We've never had to deal with this before, but I think actually need test(s) to ensure that this isn't ever called before other resources are freed. Any struct that hold a `mongoc_client_t*` will likely need to be changed to hold a reference to the Manager object to ensure that this object isn't garbage-collected before those dependent objects are freed.

This should be straightforward for most things (e.g. WriteResult, Cursor, Session, Server), but I think the command monitoring events will be trickier. Currently, they hold a client reference so that they can construct Server objects on-demand. You'd need to refactor that so they hold an actual Server object, which in turn has a reference to a Manager. One problem is that you won't know what Manager to refer to when doing this for persistent clients. We presumably already have this issue today, but the Server objects we create are totally independent from a Manager so it never comes up. That said, this referencing-holding is only need for non-persistent Managers so maybe it doesn't matter how you handle this (or if you do at all, assuming you want to support two code paths).

jmikola:
Continuing work in https://github.com/mongodb/mongo-php-driver/pull/1187.

PR #1166, opened 2020-10-10
PHPC-1631: Support PHP 8

jmikola:
https://jira.mongodb.org/browse/PHPC-1631

This is rebased on https://github.com/mongodb/mongo-php-driver/pull/1167 for a memory leak fix that turned up while testing.

@alcaeus: When merging, please squash commits into one per issue so various bug fixes get cross-referenced accordingly. The bulk of changes for PHPC-1689 can be rolled up into one commit, though.

### DIFF
--- a/src/MongoDB/Cursor.c
+++ b/src/MongoDB/Cursor.c
@@ -333,6 +333,17 @@ static PHP_METHOD(Cursor, isDead)
 	RETURN_BOOL(!mongoc_cursor_more(intern->cursor));
 } /* }}} */
 
+#if PHP_VERSION_ID >= 80000
+static PHP_METHOD(Cursor, getIterator)
+{
+	if (zend_parse_parameters_none() == FAILURE) {
+		return;
+	}
+
+	zend_create_internal_iterator_zval(return_value, getThis());
+}
+#endif
+
 /* {{{ MongoDB\Driver\Cursor function entries */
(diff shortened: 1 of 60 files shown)


alcaeus (approved):

src/MongoDB/Cursor.c
```diff
@@ -502,6 +516,9 @@ void php_phongo_cursor_init_ce(INIT_FUNC_ARGS) /* {{{ */
 	PHONGO_CE_DISABLE_SERIALIZATION(php_phongo_cursor_ce);
 	php_phongo_cursor_ce->get_iterator = php_phongo_cursor_get_iterator;
 
+#if PHP_VERSION_ID >= 80000
```
stof, line 519: why not implementing IteratorAggregate on all PHP versions ?

alcaeus: Since the internal iterator only exists on PHP 8, we decided to only implement this on PHP 8 for the time being. You can find some more motivation in the original ticket: https://jira.mongodb.org/browse/PHPC-1690.

I may end up revisiting this sooner than later depending on how PHPLIB reacts to these changes.

stof: hmm, I forgot that the internal iterator is a PHP 8 feature. That's a good reason then :smile:

PR #1172, opened 2020-10-22
Don't try linking against libresolv on AIX

### LINKED ISSUE #1171: AIX platforms shouldn't try linking with libresolv
### Bug Report

mongodb+srv:// URIs don't work on AIX platforms (including IBM i PASE). This is because the configure script for vendored libmongoc hardcodes `-lresolv` linker flags, which prevents the checks for the libresolv functions from working. On AIX, the resolver functions are in libc, so checking for libresolv should be optional. (`AC_SEARCH_LIBS` would be appropriate, but the script in `CheckResolv.m4` is more complicated than that.)

Removing the hardcoded `-lresolv` worked, but I suspect you might want a more graceful solution that doesn't need to hardcode per platform. My "remove hardcoded library, add hardcoded platform workaround" patch is:

```diff
diff --git a/scripts/autotools/CheckHost.m4 b/scripts/autotools/CheckHost.m4
index 52448385..e33ebce1 100644
--- a/scripts/autotools/CheckHost.m4
+++ b/scripts/autotools/CheckHost.m4
@@ -9,6 +9,7 @@ os_linux=no
 os_solaris=no
 os_darwin=no
 os_gnu=no
+os_aix=no
 
 case "$host" in
     *-mingw*|*-*-cygwin*)
@@ -44,6 +45,10 @@ case "$host" in
         os_darwin=yes
         TARGET_OS=unix
         ;;
+    *-*-aix*|*-*-os400*)
+        os_aix=yes
+        TARGET_OS=unix
+        ;;
     gnu*|k*bsd*-gnu*)
         os_gnu=yes
         TARGET_OS=unix
diff --git a/scripts/autotools/libmongoc/CheckResolv.m4 b/scripts/autotools/libmongoc/CheckResolv.m4
index 5cfa238c..15bbdfd5 100644
--- a/scripts/autotools/libmongoc/CheckResolv.m4
+++ b/scripts/autotools/libmongoc/CheckResolv.m4
@@ -2,9 +2,13 @@ dnl Disable Windows DNSAPI
 AC_SUBST(MONGOC_HAVE_DNSAPI, 0)
 
 found_resolv="no"
+need_libresolv="no"
 
 old_LIBS="$LIBS"
-LIBS="$LIBS -lresolv"
+dnl On AIX, resolv functions are in libc.
+if test "x$os_aix" = "xno"; then
+  LIBS="$LIBS -lresolv"
+fi
 
 dnl Thread-safe DNS query function for _mongoc_client_get_srv.
 dnl Could be a macro, not a function, so check with AC_LINK_IFELSE.
@@ -80,7 +84,9 @@ AC_LINK_IFELSE([AC_LANG_PROGRAM([[
    ]])], [
       AC_MSG_RESULT([yes])
       AC_SUBST(MONGOC_HAVE_RES_SEARCH, 1)
-      found_resolv="yes"
+      if test "x$os_aix" = "xno"; then
+        need_libresolv="yes"
+      fi
    ], [
       AC_MSG_RESULT([no])
       AC_SUBST(MONGOC_HAVE_RES_SEARCH, 0)
@@ -89,6 +95,6 @@ AC_LINK_IFELSE([AC_LANG_PROGRAM([[
 
 LIBS="$old_LIBS"
 
-AS_IF([test "$found_resolv" = "yes"],[
+AS_IF([test "$need_libresolv" = "yes"],[
   PHP_ADD_LIBRARY([resolv],,[MONGODB_SHARED_LIBADD])
 ])
```

### Environment

IBM i PASE (but as mentioned, any AIX platform will be the same)

### Test Script

Anything that connects with a mongodb+srv:// URL.

### Expected and Actual Behavior

Error with broken libresolv support (from @jwoehr):

```
Uncaught MongoDB\Driver\Exception\ConnectionTimeoutException: libresolv unavailable, cannot use mongodb+srv URI in /home/JWOEHR/work/PHP/vendor/mongodb/mongodb/src/functions.php:431
Stack trace:
#0 /home/JWOEHR/work/PHP/vendor/mongodb/mongodb/src/functions.php(431): MongoDB\Driver\Manager->selectServer(Object(MongoDB\Driver\ReadPreference))
#1 /home/JWOEHR/work/PHP/vendor/mongodb/mongodb/src/Collection.php(651): MongoDB\select_server(Object(MongoDB\Driver\Manager), Array)
#2 php shell code(1): MongoDB\Collection->find()
#3 {main}
  thrown in /home/JWOEHR/work/PHP/vendor/mongodb/mongodb/src/functions.php on line 431
```

### Debug Log

Unnecessary.

NattyNarwhal:
On AIX, the resolver functions are in libc, so trying libresolv
will cause the checks to fail. The behaviour should be instead
like AC_SEARCH_LIBS.

This hardcodes a check not to use libresolv on AIX. Better than
nothing, but it may not compensate for other systems without a
libresolv.

Fixes  #1171

### DIFF
--- a/scripts/autotools/libmongoc/CheckResolv.m4
+++ b/scripts/autotools/libmongoc/CheckResolv.m4
@@ -1,10 +1,13 @@
 dnl Disable Windows DNSAPI
 AC_SUBST(MONGOC_HAVE_DNSAPI, 0)
 
-found_resolv="no"
+need_libresolv="no"
 
 old_LIBS="$LIBS"
-LIBS="$LIBS -lresolv"
+dnl On AIX, resolv functions are in libc.
+if test "x$os_aix" = "xno"; then
+  LIBS="$LIBS -lresolv"
+fi
 
 dnl Thread-safe DNS query function for _mongoc_client_get_srv.
 dnl Could be a macro, not a function, so check with AC_LINK_IFELSE.
@@ -24,7 +27,6 @@ AC_LINK_IFELSE([AC_LANG_PROGRAM([[
    AC_MSG_RESULT([yes])
    AC_SUBST(MONGOC_HAVE_RES_SEARCH, 0)
    AC_SUBST(MONGOC_HAVE_RES_NSEARCH, 1)
-   found_resolv="yes"
 
    dnl We have res_nsearch. Call res_ndestroy (BSD/Mac) or res_nclose (Linux)?
    AC_MSG_CHECKING([for res_ndestroy])
@@ -80,7 +82,9 @@ AC_LINK_IFELSE([AC_LANG_PROGRAM([[
    ]])], [
       AC_MSG_RESULT([yes])
       AC_SUBST(MONGOC_HAVE_RES_SEARCH, 1)
-      found_resolv="yes"
+      if test "x$os_aix" = "xno"; then
+        need_libresolv="yes"
+      fi
    ], [
       AC_MSG_RESULT([no])
       AC_SUBST(MONGOC_HAVE_RES_SEARCH, 0)
@@ -89,6 +93,6 @@ AC_LINK_IFELSE([AC_LANG_PROGRAM([[
 
 LIBS="$old_LIBS"
 
-AS_IF([test "$found_resolv" = "yes"],[
+AS_IF([test "$need_libresolv" = "yes"],[
   PHP_ADD_LIBRARY([resolv],,[MONGODB_SHARED_LIBADD])
 ])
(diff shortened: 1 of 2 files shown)


scripts/autotools/libmongoc/CheckResolv.m4
```diff
@@ -80,7 +84,9 @@ AC_LINK_IFELSE([AC_LANG_PROGRAM([[
    ]])], [
       AC_MSG_RESULT([yes])
       AC_SUBST(MONGOC_HAVE_RES_SEARCH, 1)
-      found_resolv="yes"
```
alcaeus, line 83: This might create an inconsistency where `found_resolv` should be `yes`, but we're not setting it. However, I believe we can completely remove the variable since we're not checking for it anywhere else.

alcaeus (changes requested):

alcaeus (approved):

PR #1175, opened 2020-10-29
PHPC-1691: Provide Iterator implementation for cursors

alcaeus:
PHPC-1691

While working on PHP 8 compatibility for the library, I noticed that implementing `IteratorAggregate` has side-effects, as `IteratorIterator` unwraps such iterators, which causes a few issues with change streams. Furthermore, the `InternalIterator` we use on PHP 8 always rewinds the cursor at the beginning, which can lead to unnecessary `getMore` requests. Implementing Iterator logic in the cursor directly works around this problem and ensures a consistent API across all supported PHP versions.


src/MongoDB/Cursor.c
```diff
@@ -366,9 +309,72 @@ static PHP_METHOD(Cursor, getIterator)
...
+
+	zend_replace_error_handling(EH_THROW, phongo_exception_from_phongo_domain(PHONGO_ERROR_INVALID_ARGUMENT), &error_handling);
+	if (zend_parse_parameters_none() == FAILURE) {
+		zend_restore_error_handling(&error_handling);
+		return;
+	}
+	zend_restore_error_handling(&error_handling);
+
+	php_phongo_cursor_move_forward(intern);
+}
```
sgolemon, line 347: next() should return the value as well.

    php_phongo_cursor_move_forward(intern);
    zval* data = php_phongo_cursor_get_current_data(intern);
    if (data) {
        ZVAL_COPY_DEREF(return_value, data);
    }

sgolemon (approved):

src/MongoDB/Cursor.c
sgolemon: Ignore this comment. I was wrong about it.

PR #1178, opened 2020-11-09
PHPC-1704: Check for standard and date extensions on Windows builds

alcaeus:
PHPC-1704

In PHP 8, PHP 7.3.25+, and PHP 7.4.13+, ADD_EXTENSION_DEP can correctly check for the date and standard extensions. On older versions, we have to continue to assume that these extensions are in fact installed.

Note: I'll hold off on merging this until the release candidates for 7.3.25 and 7.4.13 are available, which they should be in a couple of days. Then I can confirm that these checks work as they should.

### DIFF
--- a/config.w32
+++ b/config.w32
@@ -65,8 +65,18 @@ ARG_WITH("mongodb-sasl", "MongoDB: Build against Cyrus-SASL", "yes");
 ARG_WITH("mongodb-client-side-encryption", "MongoDB: Enable client-side encryption", "yes");
 
 if (PHP_MONGODB != "no") {
-  /* Note: ADD_EXTENSION_DEP() cannot be used to declare that we depend on the
-   * date and standard extensions. Assume that they're always enabled. */
+  /* Note: ADD_EXTENSION_DEP() only reports the date and standard extensions as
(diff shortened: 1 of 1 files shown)


config.w32
```diff
@@ -65,8 +65,18 @@ ARG_WITH("mongodb-sasl", "MongoDB: Build against Cyrus-SASL", "yes");
 ARG_WITH("mongodb-client-side-encryption", "MongoDB: Enable client-side encryption", "yes");
 
 if (PHP_MONGODB != "no") {
-  /* Note: ADD_EXTENSION_DEP() cannot be used to declare that we depend on the
-   * date and standard extensions. Assume that they're always enabled. */
+  /* Note: ADD_EXTENSION_DEP() only reports the date and standard extensions as
+   * installed in PHP 7.3.25+, 7.4.13+, and 8.0.0+). On other versions, assume
+   * that they're always enabled. */
+  if (
+    PHP_VERSION == 8 ||
```
cmb69, line 72: Consider to be more "adventurous":
```suggestion
    PHP_VERSION >= 8 ||
```

cmb69 (approved):
LGTM, and builds fine with PHP 7.4.12 and a recent 7.4.13-dev.

PR #1186, opened 2020-12-07
PHPC-1713: Ensure Cursor::current returns null on invalid positions

### LINKED ISSUE #1183: [Cursor] the current method does not return anything!
After the last fetch, the `current` method of `MongoDB\Driver\Cursor` must be return `null` but not returns anything.([according to the doc](https://www.php.net/manual/en/mongodb-driver-cursor.current.php#refsect1-mongodb-driver-cursor.current-returnvalues)).
| Q | A
| ------------- | ---
| OS | CentOS 8
| PHP | 7.4.12
| MongoDB extension version | 1.9.0
| MongoDB server | MongoDB Enterprise 4.4.1
| MongoDB environment | Replica Set

**Test.php**
```php
<?php
$mng = new \MongoDB\Driver\Manager('mongodb://user:pass@127.0.0.1:27017/myDB');
$mng->selectServer($mng->getReadPreference());
$cursor = $mng->executeQuery('myDB.myCollection',new \MongoDB\Driver\Query([],['batchSize' => 100]));
$cursor->SetTypeMap([
    'root' => 'array',
    'document' => 'array'
]);
var_dump($cursor->isDead());
var_dump($cursor->getId());
$cursor->rewind();
while(true){
    $row = $cursor->current();
    var_dump($row);
    $cursor->next();
}
```
```sh
root# php Test.php
```
**output**
```text
bool(false)
object(MongoDB\Driver\CursorId)#4 (1) {
  ["id"]=>
  int(0)
}
array(13) {
...
}
array(13) {
...
}
PHP Notice:  Undefined variable: row in Test.php on line 14
NULL
PHP Fatal error:  Uncaught MongoDB\Driver\Exception\RuntimeException: Cannot advance a completed or failed cursor. in Test.php:15
Stack trace:
#0 Test.php(15): MongoDB\Driver\Cursor->next()
#1 {main}
  thrown in Test.php on line 15
```

alcaeus:
PHPC-1713

Fixes #1183.

### DIFF
--- a/src/MongoDB/Cursor.c
+++ b/src/MongoDB/Cursor.c
@@ -312,7 +312,11 @@ PHP_METHOD(Cursor, current)
 	data = php_phongo_cursor_get_current_data(intern);
 
 	if (data) {
-		ZVAL_COPY_DEREF(return_value, data);
+		if (Z_ISUNDEF_P(data)) {
+			RETURN_NULL();
+		} else {
+			ZVAL_COPY_DEREF(return_value, data);
+		}
 	}
 }
 
--- a/src/MongoDB/Cursor.c
+++ b/src/MongoDB/Cursor.c
@@ -311,12 +311,10 @@ PHP_METHOD(Cursor, current)
 
 	data = php_phongo_cursor_get_current_data(intern);
 
-	if (data) {
-		if (Z_ISUNDEF_P(data)) {
-			RETURN_NULL();
-		} else {
-			ZVAL_COPY_DEREF(return_value, data);
-		}
+	if (Z_ISUNDEF_P(data)) {
+		RETURN_NULL();
+	} else {
+		ZVAL_COPY_DEREF(return_value, data);
 	}
 }
 
--- a/tests/cursor/bug1713-001.phpt
+++ b/tests/cursor/bug1713-001.phpt
@@ -1,9 +1,8 @@
 --TEST--
-MongoDB\Driver\Cursor PHPC-171: The cursor current method does not return anything
+PHPC-1713: MongoDB\Driver\Cursor::current() does not return anything
 --SKIPIF--
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
 <?php skip_if_not_live(); ?>
-<?php skip_if_not_clean(); ?>
 --FILE--
 <?php
 require_once __DIR__ . "/../utils/basic.inc";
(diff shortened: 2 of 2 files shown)


src/MongoDB/Cursor.c
```diff
@@ -312,7 +312,11 @@ PHP_METHOD(Cursor, current)
 	data = php_phongo_cursor_get_current_data(intern);
 
 	if (data) {
-		ZVAL_COPY_DEREF(return_value, data);
+		if (Z_ISUNDEF_P(data)) {
```
jmikola, line 315: At first glance, I wondered if it would be possible for this function to still leave `return_value` unset.

`php_phongo_cursor_get_current_data` always returns a non-null pointer to a block of memory within `intern`, which itself is never null (we'd hit a segfault if so). In that case, I think can should drop the outer `if (data)` for readability -- optionally leave a comment that "data points to some memory within intern and is always non-null".

tests/cursor/bug1713-001.phpt
```diff
@@ -0,0 +1,24 @@
+--TEST--
+MongoDB\Driver\Cursor PHPC-171: The cursor current method does not return anything
```
jmikola, line 2: By convention, the JIRA ticket is used as the test name prefix. Non-ticket tests start with a class/method name. I'd suggest:

> PHPC-1713: MongoDB\Driver\Cursor::current() does not return anything

tests/cursor/bug1713-001.phpt
```diff
@@ -0,0 +1,24 @@
...
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_live(); ?>
+<?php skip_if_not_clean(); ?>
```
jmikola, line 6: Technically, I don't think this matters since we don't actually care if the collection is dropped. Whether it exists or not, we're just testing that `php_phongo_cursor_t.visitor_data.zchild` is zeroed out before the first call to `rewind()` or `next()`.

That said, no objections to keeping this in place.

jmikola (changes requested):

src/MongoDB/Cursor.c
alcaeus: Removed the check 👍

tests/cursor/bug1713-001.phpt
alcaeus: Noted. I've dropped the call, no reason to clear the database if we don't care about its contents.

jmikola (approved):

alcaeus:
@jmikola realised I (once again) targeted the wrong branch with the PR. Will rebase onto v1.9 and merge after that.

PR #1189, opened 2021-01-18
PHPC-1529: Reset libmongocrypt key vault client after forking

alcaeus:
PHPC-1529

Fixing this in master to take advantage of the new disableClientPersistence option for tests. The test was a bit tricky to write, but I managed to verify that the keyVaultClient is reset by triggering a reset on the client (by running a query in the fork), then trying to iterate a cursor that was previously opened on the keyVaultClient. With the additional call to `php_phongo_client_reset_once` removed, this test fails as expected.

### DIFF
--- a/tests/cursor/bug1529-001.phpt
+++ b/tests/cursor/bug1529-001.phpt
@@ -5,6 +5,7 @@ PHPC-1529: Resetting a client should also reset the keyVaultClient
 <?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
 <?php skip_if_not_live(); ?>
 <?php skip_if_not_clean(); ?>
+<?php skip_if_not_libmongocrypt(); ?>
 --FILE--
 <?php
 require_once __DIR__ . "/../utils/basic.inc";
--- a/tests/cursor/bug1529-001.phpt
+++ b/tests/cursor/bug1529-001.phpt
@@ -64,7 +64,7 @@ $cursor = $keyVaultClient->executeQuery(NS, $query);
 $childPid = pcntl_fork();
 
 if ($childPid === 0) {
-    /* Executing a query with the parent's client resets this client as well as
+    /* Executing any operation with the parent's client resets this client as well as
      * the keyVaultClient. Continuing iteration of the cursor opened on the
      * keyVaultClient before resetting it should then result in an error due to
      * the client having been reset. */
--- a/tests/cursor/bug1529-001.phpt
+++ b/tests/cursor/bug1529-001.phpt
@@ -1,0 +1,106 @@
+--TEST--
+PHPC-1529: Resetting a client should also reset the keyVaultClient
+--SKIPIF--
+<?php if (!function_exists('pcntl_fork')) { die('skip pcntl_fork() not available'); } ?>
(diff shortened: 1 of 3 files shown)


tests/cursor/bug1529-001.phpt
```diff
@@ -0,0 +1,106 @@
...
+     * keyVaultClient before resetting it should then result in an error due to
+     * the client having been reset. */
+    $childCursor = $manager->executeQuery(NS, $query);
+
+    echo throws(
+        function () use ($cursor) { iterator_count($cursor); },
+        MongoDB\Driver\Exception\RuntimeException::class
+    ), "\n";
+
+    unset($childCursor);
```
jmikola, line 78: Is it relevenat for `$childCursor` to remain alive this long? Just curious if `$manager->executeQuery()` above could just as easily be changed to executing a ping command.

tests/cursor/bug1529-001.phpt
```diff
@@ -0,0 +1,106 @@
...
+
+    unset($cursor);
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+Parent executes find with batchSize: 2
+Child executes listCollections
```
jmikola, line 98: Where does this come from, and why does it come between both `find` operations? Does libmongocrypt execute this when a Manager is initialized upon executing the first operation?

tests/cursor/bug1529-001.phpt
alcaeus: I replaced the `find` with a single `ping`. This also removes the `listCollections` call you pointed out below.

That `listCollections` command is run by the keyVaultClient to fetch collection metadata, which is needed to prepare the query if automatic encryption was enabled on the collection.

jmikola: Ah, so libmongocrypt only decides it's necessary to do so when running a query (perhaps because automatic decryption may be needed), but a `ping` command has nothing to do with that so it avoids it altogether? If so, SGTM.

tests/cursor/bug1529-001.phpt
```diff
@@ -0,0 +1,103 @@
...
+
+MongoDB\Driver\Monitoring\addSubscriber(new CommandLogger);
+
+$query = new MongoDB\Driver\Query([], ['batchSize' => 2]);
+$cursor = $keyVaultClient->executeQuery(NS, $query);
+
+$childPid = pcntl_fork();
+
+if ($childPid === 0) {
+    /* Executing a query with the parent's client resets this client as well as
```
jmikola, line 67: This should probably be "Executing a command" or "Executing any operation"

jmikola (approved):
LGTM with change to the comment above `$manager->executeCommand()`

tests/cursor/bug1529-001.phpt
alcaeus: That is correct. When running a query against a collection that uses automatic encryption, it may be necessary to encrypt values in the query to ensure querying against encrypted values still produces expected results (this supports equality checks on fields that use a deterministic encryption algorithm).

jmikola (changes requested):
Please remove `mongocryptd.pid` and feel free to add it to `.gitignore` if you think that will help.

jmikola (approved):

PR #1190, opened 2021-01-19
PHPC-1722: Destroy persistent clients in GSHUTDOWN

jmikola:
https://jira.mongodb.org/browse/PHPC-1722
https://jira.mongodb.org/browse/PHPC-1366
https://jira.mongodb.org/browse/PHPC-1367

This ensures persistent clients in all threads are freed, not just those in the main thread. Doing so also required that we call mongoc_cleanup and bson_mem_restore_vtable in the final GSHUTDOWN (tracked by an atomic counter).

This also moves the vtable to a local variable in MINIT and swaps the order of mongoc_init and bson_mem_set_vtable (and the corresponding teardown functions) so that PHP's allocators will be used for all of libmongoc.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -3798,11 +3798,12 @@ PHP_MINIT_FUNCTION(mongodb)
 
 	REGISTER_INI_ENTRIES();
 
-	/* Initialize libmongoc */
-	mongoc_init();
-
-	/* Initialize libbson */
+	/* Assign our custom vtable to libbson, so all memory allocation in libbson
+	 * (and libmongoc) will use PHP's persistent memory API. After doing so,
+	 * initialize libmongoc. Later, we will shutdown libmongoc and restore
+	 * libbson's vtable in the final GSHUTDOWN. */
 	bson_mem_set_vtable(&bson_mem_vtable);
+	mongoc_init();
 
 	/* Prep default object handlers to be used when we register the classes */
 	memcpy(&phongo_std_object_handlers, zend_get_std_object_handlers(), sizeof(zend_object_handlers));
@@ -3982,8 +3983,8 @@ PHP_GSHUTDOWN_FUNCTION(mongodb)
 	 * all threads have been destroyed, and it is now safe to shutdown libmongoc
 	 * and restore libbson's original vtable. */
 	if (bson_atomic_int_add(&phongo_num_threads, -1) == 0) {
-		bson_mem_restore_vtable();
 		mongoc_cleanup();
+		bson_mem_restore_vtable();
 	}
 }
 /* }}} */
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -3504,8 +3504,8 @@ static void* php_phongo_calloc(size_t num_members, size_t num_bytes) /* {{{ */
 	return pecalloc(num_members, num_bytes, 1);
 } /* }}} */
 
-static void* php_phongo_realloc(void* mem, size_t num_bytes)
-{ /* {{{ */
+static void* php_phongo_realloc(void* mem, size_t num_bytes) /* {{{ */
+{
 	return perealloc(mem, num_bytes, 1);
 } /* }}} */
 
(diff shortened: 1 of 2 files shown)


jmikola:
We don't have a way to test threaded SAPIs from our test suite, but I was able to manually test this using [pconn-sapi](https://github.com/johannes/pconn-sapi) (thanks @johannes!) and PHP 7.0.33. For the following terminal session runs two requests in each of five threads:

```
./php-pconn -n 2 -t 5 foo.php
GINIT ++phongo_num_threads: 1
GINIT ++phongo_num_threads: 2
GINIT ++phongo_num_threads: 3
float(1)
GINIT ++phongo_num_threads: 4
float(1)
float(1)
float(1)
GINIT ++phongo_num_threads: 5
float(1)
float(1)
GINIT ++phongo_num_threads: 6
float(1)
float(1)
float(1)
float(1)
GSHUTDOWN --phongo_num_threads: 5
GSHUTDOWN --phongo_num_threads: 4
GSHUTDOWN --phongo_num_threads: 3
GSHUTDOWN --phongo_num_threads: 2
GSHUTDOWN --phongo_num_threads: 1
GSHUTDOWN --phongo_num_threads: 0
GSHUTDOWN mongoc_cleanup
```

I added some `stderr` logging to report when the atomic counter was incremented and decremented, as well as when the final call to `mongoc_cleanup` was made. The `foo.php` script constructs a Manager and executes a ping command (using a persistent client) and reads as follows:

```php
<?php

$m = new MongoDB\Driver\Manager;
var_dump($m->executeCommand('admin', new MongoDB\Driver\Command(['ping' => 1]))->toArray()[0]->ok);
```

php_phongo.c
```diff
@@ -3497,8 +3504,8 @@ static void* php_phongo_calloc(size_t num_members, size_t num_bytes) /* {{{ */
 	return pecalloc(num_members, num_bytes, 1);
 } /* }}} */
 
-static void* php_phongo_realloc(void* mem, size_t num_bytes)
-{ /* {{{ */
+static void* php_phongo_realloc(void* mem, size_t num_bytes) /* {{{ */
+{
```
jmikola, line 3508: This is an old code formatting issue I came across while working on the vtable commits.

php_phongo.c
```diff
@@ -3781,16 +3787,24 @@ static HashTable* php_phongo_std_get_gc(phongo_compat_object_handler_type* objec
...
 	(void) type; /* We don't care if we are loaded via dl() or extension= */
 
 	REGISTER_INI_ENTRIES();
 
-	/* Initialize libmongoc */
+	/* Assign our custom vtable to libbson, so all memory allocation in libbson
+	 * (and libmongoc) will use PHP's persistent memory API. After doing so,
+	 * initialize libmongoc. Later, we will shutdown libmongoc and restore
+	 * libbson's vtable in the final GSHUTDOWN. */
+	bson_mem_set_vtable(&bson_mem_vtable);
```
jmikola, line 3805: `bson_mem_set_vtable` does a struct assignment, which is sufficient to get our function pointers copied into libbson's own internal global (`gMemVtable`). We don't actually need this parameter to stick around as a global.

php_phongo.c
```diff
@@ -3962,11 +3957,35 @@ PHP_RSHUTDOWN_FUNCTION(mongodb)
...
+	ZEND_HASH_REVERSE_FOREACH_PTR(&mongodb_globals->persistent_clients, pclient)
+	{
+		php_phongo_pclient_destroy(pclient);
+	}
+	ZEND_HASH_FOREACH_END();
+
+	/* Destroy HashTable for persistent clients. mongoc_client_t objects have
+	 * already been destroyed. */
+	zend_hash_destroy(&mongodb_globals->persistent_clients);
+
```
jmikola, line 3974: This entire block will be refactored by [PHPC-1737](https://jira.mongodb.org/browse/PHPC-1737) in a subsequent PR, unless you'd prefer I lump it in to this one.

alcaeus (approved):
LGTM!

php_phongo.c
alcaeus: Subsequent PR sounds good, no need to lump it all into one PR.

PR #1192, opened 2021-01-21
PHPC-1748 Fix wrong return value of Cursor::key when iterator is not valid

alcaeus:
PHPC-1748

### DIFF
--- a/src/MongoDB/Cursor.c
+++ b/src/MongoDB/Cursor.c
@@ -52,101 +52,6 @@ static void php_phongo_cursor_free_current(php_phongo_cursor_t* cursor) /* {{{ *
 	}
 } /* }}} */
 
-/* {{{ MongoDB\Driver\Cursor iterator handlers */
-static int php_phongo_cursor_valid(php_phongo_cursor_t* cursor) /* {{{ */
-{
-	if (!Z_ISUNDEF(cursor->visitor_data.zchild)) {
-		return SUCCESS;
-	}
-
-	return FAILURE;
-} /* }}} */
-
-static void php_phongo_cursor_get_current_key(php_phongo_cursor_t* cursor, zval* key) /* {{{ */
-{
-	if (Z_ISUNDEF(cursor->visitor_data.zchild)) {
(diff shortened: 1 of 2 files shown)


src/MongoDB/Cursor.c
```diff
@@ -64,6 +64,11 @@ static int php_phongo_cursor_valid(php_phongo_cursor_t* cursor) /* {{{ */
 
 static void php_phongo_cursor_get_current_key(php_phongo_cursor_t* cursor, zval* key) /* {{{ */
 {
+	if (Z_ISUNDEF(cursor->visitor_data.zchild)) {
+		ZVAL_NULL(key);
+		return;
+	}
+
```
jmikola, line 71: Since you introduced the Iterator methods, it looks like `php_phongo_cursor_get_current_key` and `php_phongo_cursor_get_current_data` are only called from `key()` and `current()`, respectively. Can we just delete these two functions?

That would allow you to move this logic into `key()`, which would be consistent with what you did for `current()` in https://github.com/mongodb/mongo-php-driver/commit/70a097e5a145b6eff9931171a86b4c738488bfe8.

alcaeus: Sure, I can refactor this 👍

src/MongoDB/Cursor.c
```diff
@@ -62,21 +62,6 @@ static int php_phongo_cursor_valid(php_phongo_cursor_t* cursor) /* {{{ */
 	return FAILURE;
 } /* }}} */
 
-static void php_phongo_cursor_get_current_key(php_phongo_cursor_t* cursor, zval* key) /* {{{ */
```
jmikola, line 65: I just realized `php_phongo_cursor_valid` is also only called once from `valid()`. Do you want to include that in a separate commit? If not, we can get to it down the line at some point.

If so, I'd also suggest just removing the "MongoDB\Driver\Cursor iterator handlers" comment (and the corresponding `}}}`) since we really don't have iterator handlers anymore.

jmikola (approved):

PR #1199, opened 2021-02-12
PHPC-1739 Move from travis-ci to GitHub Actions

alcaeus:
PHPC-1739

This PR drops travis-ci.yml and adds GitHub Actions workflows for tests and clang-format. Tests currently only run on PHP 7.4 and 8.0, with the rest being tested on Evergreen.

Due to this change, we also had to remove custom mo configs, as the GitHub action leverages drivers-evergreen-tools. Tests are now stable both on GitHub Actions and Evergreen, with only selected config files needing overriding. Due to this, this PR includes cleanup scheduled in PHPC-1242, but I can extract this to a separate PR if necessary.

Since this PR targets v1.9 which isn't tested by Evergreen, here's a patch build running ALL tasks and variants: https://spruce.mongodb.com/version/6026605bc9ec4448427d4f1f/tasks


.evergreen/compile-unix.sh
```diff
@@ -66,8 +66,6 @@ echo "PHP_VERSION: $PHP_VERSION"
 OLD_PATH=$PATH
 PATH=/opt/php/${PHP_VERSION}-64bit/bin:$OLD_PATH
 
-#cat `which phpize` | sed 's@/data/mci/.*/src@/opt@' > ./phpize
-#chmod +x ./phpize
```
jmikola, line 70: Merge conflict with #1201?

.evergreen/orchestration/configs/replica_sets/replicaset-old.json
```diff
@@ -12,10 +12,9 @@
...
             },
             "rsParams": {
-                "priority": 99,
                 "tags": {
                     "ordinal": "one",
-                    "dc": "pa"
+                    "dc": "ny"
                 }
             },
             "server_id": "RS-OLD-one"
```
jmikola, line 20: Any reason not to change the `server_id` values as well?

src/BSON/DBPointer.c
```diff
@@ -46,7 +46,7 @@ static bool php_phongo_dbpointer_init(php_phongo_dbpointer_t* intern, const char
 	intern->ref     = estrndup(ref, ref_len);
 	intern->ref_len = ref_len;
 
-	strncpy(intern->id, id, sizeof(intern->id));
+	strncpy(intern->id, id, sizeof(intern->id) - 1);
```
jmikola, line 49: This looks like a bug fix, even though we might never hit this thanks to the `bson_oid_is_valid` check above. Should it have a separate ticket, though?

bin/prep-release.php
```diff
@@ -71,9 +71,6 @@ function get_files() {
         "Vagrantfile",
 
         "scripts/*/*.{sh}",
-        "scripts/*/ldap/*.{conf,ldif,py,sh}",
-        "scripts/centos/ldap/users",
-        "scripts/presets/*.{json}",
         "scripts/ssl/*.pem",
```
jmikola, line 74: Our old MO configs referred to these certs and our test files still read these. If these certs are no longer in sync with the MO configs, perhaps we should remove these and modify those tests to rely on certs from drivers-evergreen-tools.

I'll note that https://github.com/mongodb-labs/drivers-evergreen-tools/tree/master/.evergreen/orchestration/lib only contains a `client.pem` file. I'm not sure where the other files are.

CONTRIBUTING.md
```diff
@@ -89,14 +89,15 @@ run in as many environments as possible. To paraphrase the
...
 Mongo Orchestration is an HTTP server that provides a REST API for maintaining
-MongoDB configurations. These configurations are located in ``scripts/presets``
-and for Travis CI in ``scripts/presets/travis``. The presets for Travis CI can
-also be spun-up locally, and that is the preferred testing method. An older way
-using a specific VM is also still available (see further down).
+MongoDB configurations. These configurations are provided by the
+[drivers-evergreen-tools](https://github.com/mongodb-labs/drivers-evergreen-tools)
+repository. These configurations can be run locally. Alternatively, you can use
+the GitHub Actions workflow if you don't want to set up a local development
+environment.
```
jmikola, line 100: This section may be a good place to mention that the `MONGO_ORCHESTRATION_URI` environment variable can be defined if the default `http://localhost:8889/v1` is not suitable. Perhaps you can add a line below the shell command that starts MO.

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          extensions: ":mongodb"
+          tools: "phpize"
+
+      - name: "Configure driver"
+        run: .github/workflows/configure.sh
```
jmikola, line 37: I assume `make format-check` is the main reason you had to split out `configure.sh`?

.evergreen/run-tests.sh
```diff
@@ -26,7 +26,7 @@ echo "Running $AUTH tests, connecting to $MONGODB_URI"
 # Run the tests, and store the results in a Evergreen compatible JSON results file
 case "$OS" in
    *)
-      TEST_PHP_ARGS="-q -x --show-diff -g FAIL,XFAIL,BORK,WARN,LEAK,SKIP" make test TESTS=$TESTS
+      TEST_PHP_ARGS="-q -x --show-diff -g FAIL,BORK,WARN,LEAK,SKIP" make test TESTS=$TESTS
```
jmikola, line 29: Why was XFAIL removed? Is this just to suppress them from output?

Will this prevent us from noticing when an XFAIL test unexpectedly passes?

.evergreen/config.yml
```diff
@@ -363,7 +366,7 @@ functions:
         working_dir: "src"
         script: |
           ${PREPARE_SHELL}
-          TESTS=${TESTS} SSL=${SSL} MONGODB_URI="${MONGODB_URI}" sh ${PROJECT_DIRECTORY}/.evergreen/run-tests.sh
+          TESTS=${TESTS} SSL=${SSL} MONGODB_URI="${MONGODB_URI}${APPEND_URI}" sh ${PROJECT_DIRECTORY}/.evergreen/run-tests.sh
```
jmikola, line 369: Will `${MONGODB_URI}` always omit the trailing slash and URI options, such that `${APPEND_URI}` is safe to concatenate?

I'm not sure where this was being handled previously. It looks like it was previously a `var` for `bootstrap mongo-orchestration`, but I didn't find how it was referenced in either this file or drivers-evergreen-tools.

.evergreen/config.yml
```diff
@@ -72,13 +72,15 @@ functions:
            fi
 
            export MONGO_ORCHESTRATION_HOME="$DRIVERS_TOOLS/.evergreen/orchestration"
+           export PROJECT_ORCHESTRATION_HOME="$PROJECT_DIRECTORY/.evergreen/orchestration"
```
jmikola, line 75: Does our `config.yml` inherit the copy in drivers-evergreen-tools, or is that version just a base that gets copied to create the config for a new driver? If we evaluate this in addition to the drivers-evergreen-tools file, I notice this is already defined there: https://github.com/mongodb-labs/drivers-evergreen-tools/blob/92dc624d1c712fa02283f14f675997b89ceb8ef2/.evergreen/config.yml#L58

src/BSON/DBPointer.c
alcaeus: If you want to include this in the changelog for tracking purposes, I'm happy to file a ticket for this 👍

jmikola: https://jira.mongodb.org/browse/PHPC-1775

.evergreen/compile-unix.sh
alcaeus: Looking forward to it :D

.evergreen/orchestration/configs/replica_sets/replicaset-old.json
alcaeus: Must've missed this. I noticed the configs provided by drivers-evergreen-tools don't include this, is there any reason to keep these around?

bin/prep-release.php
alcaeus: I'll take another look at those certs and report back.

.github/workflows/clang-format.yml
alcaeus: Correct. I wanted to extract as much as I could to scripts to avoid having to maintain multiple workflows with the same stuff.

.evergreen/run-tests.sh
alcaeus: I removed this to no longer get the diff printed for tests that are expected to fail, but it doesn't achieve what I wanted. Will revert and revisit.

For the record, tests that have an `XFAIL` section but don't fail are printed as warnings:
```
=====================================================================
WARNED TEST SUMMARY
---------------------------------------------------------------------
Test xfail [tests/xfail-pass.phpt] (warn: XFAIL section but test passes)
=====================================================================
```

.evergreen/config.yml
alcaeus: There's no inheritance for evergreen.yml, so any changes made in drivers-evergreen-tools have to be duplicated in all repositories that are based on drivers-evergreen-tools

.evergreen/config.yml
alcaeus: Not entirely. This was previously handled in run-orchestration.sh, but the default run-orchestration script in drivers-evergreen-tools does not support it. The main reason for this variable is to append SSL data, which is not included in the mongodb_uri provided by mongo-orchestration (see https://github.com/10gen/mongo-orchestration/issues/268). I plan to revisit this in mo directly in the coming weeks so we can get rid of this

.evergreen/orchestration/configs/replica_sets/replicaset-old.json
jmikola: This is only useful if we want to manage servers in a replica set directly (see https://github.com/10gen/mongo-orchestration/pull/168). None of our tests rely on that, so I think we can remove this. MO will generate a UUID if this field is omitted (see [here](https://github.com/10gen/mongo-orchestration/blob/066ec58c817947ca136c25690a705fd4e4db2a7b/mongo_orchestration/servers.py#L512)).

For historical context, this was likely related to https://github.com/estolfo/mo-tests.

.evergreen/run-tests.sh
jmikola: If a passing XFAIL test yields a warning, then doesn't this work as intended?

alcaeus: Sorry, I was referring to the `--show-diff` portion not working as intended. Currently, a successful XFAIL test will print its diff output. I was hoping that removing the `XFAIL` portion would remove this, as I'm not interested in a diff from a test that's expected to fail anyways. Even then, a diff is shown, meaning that I might as well undo this change for now until I figure out how to accomplish this:
```
$ TEST_PHP_ARGS="-q -x --show-diff -g FAIL,BORK,WARN,LEAK,SKIP" make test TESTS=tests/connect/standalone-plain-0001.phpt

[...]
=====================================================================
Running selected tests.

========DIFF========
001+ Notice: Undefined index: port in /Users/alcaeus/Code/mongodb/mongo-php-driver/tests/connect/standalone-plain-0001.php on line 9
002+ 
003+ Fatal error: Uncaught MongoDB\Driver\Exception\InvalidArgumentException: Failed to parse MongoDB URI: 'mongodb://root:toor@127.0.0.1:0/admin'. Invalid host string in URI. in /Users/alcaeus/Code/mongodb/mongo-php-driver/tests/connect/standalone-plain-0001.php:10
004+ Stack trace:
001- User Created
002- string(9) "important"
003- User deleted
004- ===DONE===
005+ #0 /Users/alcaeus/Code/mongodb/mongo-php-driver/tests/connect/standalone-plain-0001.php(10): MongoDB\Driver\Manager->__construct('mongodb://root:...')
006+ #1 {main}
007+   thrown in /Users/alcaeus/Code/mongodb/mongo-php-driver/tests/connect/standalone-plain-0001.php on line 10
========DONE========
=====================================================================
Number of tests :    1                 1
Tests skipped   :    0 (  0.0%) --------
Tests warned    :    0 (  0.0%) (  0.0%)
Tests failed    :    0 (  0.0%) (  0.0%)
Expected fail   :    1 (100.0%) (100.0%)
Tests passed    :    0 (  0.0%) (  0.0%)
---------------------------------------------------------------------
Time taken      :    0 seconds
=====================================================================

=====================================================================
EXPECTED FAILED TEST SUMMARY
---------------------------------------------------------------------
Connect to MongoDB with using PLAIN auth mechanism [tests/connect/standalone-plain-0001.phpt]  XFAIL REASON: authMechanism=PLAIN (LDAP) tests must be reimplemented (PHPC-1172)
parse_url() tests must be reimplemented (PHPC-1177)
=====================================================================
```

.evergreen/orchestration/configs/replica_sets/replicaset-old.json
alcaeus: Removed the `server_id` config from this file.

src/BSON/DBPointer.c
alcaeus: I'll extract this in a separate PR that includes the other sanity checks you mentioned in the ticket.

bin/prep-release.php
alcaeus: I've removed the certificates entirely. We need to test against the certificates from mongo-orchestration, so the path will now be read through the `SSL_DIR` environment variable. If that doesn't point to a valid path, SSL tests will be skipped. Users that test the driver against a server with SSL enabled need to set this environment variable to the path where the certificates rest. This is done automatically for evergreen but currently skipped on GitHub Actions (since we don't test SSL variants there anyways).

CONTRIBUTING.md
alcaeus: Added 👍

.evergreen/run-tests.sh
alcaeus: I reverted this change entirely.

tests/utils/skipif.php
```diff
@@ -198,6 +198,7 @@ function skip_if_ssl()
 function skip_if_not_ssl()
 {
     is_ssl(URI) or exit('skip URI is not using SSL');
+    realpath(getenv('SSL_DIR')) !== false or exit('skip SSL_DIR environment not set');
```
jmikola, line 201: Is this necessary? Not all SSL tests use the certs: https://github.com/mongodb/mongo-php-driver/blob/master/tests/connect/standalone-ssl-no_verify-001.phpt

jmikola: Side note: [`getenv()`](https://www.php.net/getenv) is actually what returns `false`, so I don't think you'd want to call `realpath()` here even if you do keep this line.

tests/connect/standalone-x509-extract_username-002.phpt
```diff
@@ -11,7 +11,7 @@ parse_url() tests must be reimplemented (PHPC-1177)
 <?php
 require_once __DIR__ . "/../utils/basic.inc";
 
-$SSL_DIR = realpath(__DIR__ . '/../../scripts/ssl/');
+$SSL_DIR = realpath(getenv('SSL_DIR'));
```
jmikola, line 14: While you're modifying these lines, what do you think about defining an SSL_DIR constant in `basic.inc` and just getting rid of this? Something like:

```php
define('SSL_DIR', realpath(getenv('SSL_DIR') ?: ''));
```

`realpath()` is happy to take `false` so maybe the ternary is superfluous. It will all resolve to the CWD and ultimately lead to a bad path when we actually try to refer to certs in a test. For that reason, I think we're probably OK doing without the change to `skip_if_not_ssl` below.

tests/utils/skipif.php
alcaeus: Ah, I hadn't considered `getenv` returning `false` as well. `realpath` returns false if the path doesn't exist, but per your comment below this might not entirely cover everything. In light of this and your comment below, I'll create a separate `skip_if_no_ssl_dir` function and include that in all tests that actually do something with the certificates to avoid skipping more tests than necessary.

src/BSON/DBPointer.c
jmikola: Did you mean to pull this commit out? Looks like it's still here.

tests/connect/standalone-x509-extract_username-002.phpt
alcaeus: Refactored to introduce `skip_if_no_ssl_dir`, which made the repeated calls to `getenv` in these tests obsolete.

CONTRIBUTING.md
```diff
@@ -107,69 +108,18 @@ Once installed, Mongo Orchestration can be started with
...
-#### Restarting Mongo Orchestration
-
-If something goes awry in the test VM, you can reload it by running:
+If Mongo Orchestration is running on non-default settings, the
+``MONGO_ORCHESTRATION_URI`` environment variable can be used to specify where
+mongo-orchestration is listening:
 
 ```
-make test-bootstrap
+MONGO_ORCHESTRATION_URI=http://some-host:1234/v1 make test
```
jmikola, line 122: nit: either use example.com (what it's intended for) for localhost. In practice, I can't imagine MO would run somewhere _other_ than localhost. This is generally just needed if the the port is changed. I think `localhost:1234` would be fine.

tests/utils/skipif.php
```diff
@@ -200,6 +200,18 @@ function skip_if_not_ssl()
     is_ssl(URI) or exit('skip URI is not using SSL');
 }
 
+/**
+ * Skips the test if no SSL directory has been defined.
+ */
+function skip_if_no_ssl_dir()
+{
+    $sslDir = getenv('SSL_DIR');
+    $sslDir !== false or exit('skip SSL_DIR environment not set');
```
jmikola, line 209: "environment variable is not set"

jmikola (approved):
Two nits, but LGTM.

PR #1204, opened 2021-03-01
PHPC-1716 Allow configuring server API version in manager

alcaeus:
PHPC-1716

This PR is built on the current WIP state of versioned API support in the C driver to allow designing the PHP driver interface and verify that it passes all spec tests. Note that the changes to submodules (including the libmongoc version) will be re-done once versioned API support lands in the main libmongoc development branch.


src/MongoDB/ServerApi.c
```diff
@@ -0,0 +1,391 @@
...
+	mongoc_server_api_version_t server_api_version;
+
+	if (!mongoc_server_api_version_from_string(ZSTR_VAL(version), &server_api_version)) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Server API version \"%s\" is not supported in this driver version", ZSTR_VAL(version));
+		return false;
+	}
+
+	if (*server_api) {
+		mongoc_server_api_destroy(*server_api);
+	}
```
jmikola, line 42: Is this ever reached? If not, would it be better to throw an exception and return `false`? AFAICT, this is only called from the constructor and `php_phongo_serverapi_init_from_hash`, both of which should use freshly initialized objects.

src/MongoDB/ServerApi.c
```diff
@@ -0,0 +1,391 @@
...
+	zend_bool               strict = 0;
+	zend_bool               strict_null = 1;
+	zend_bool               deprecation_errors = 0;
+	zend_bool               deprecation_errors_null = 1;
+
+	intern = Z_SERVERAPI_OBJ_P(getThis());
+
+	zend_replace_error_handling(EH_THROW, phongo_exception_from_phongo_domain(PHONGO_ERROR_INVALID_ARGUMENT), &error_handling);
+
+	ZEND_PARSE_PARAMETERS_START(1, 3)
```
jmikola, line 112: Does PHP core indent these blocks? I realize that might not play nice with `clang-format`.

src/MongoDB/ServerApi.c
```diff
@@ -0,0 +1,391 @@
...
+	}
+
+	return props;
+} /* }}} */
+
+/* {{{ proto array MongoDB\Driver\ServerApi::bsonSerialize()
+*/
+static PHP_METHOD(ServerApi, bsonSerialize)
+{
+	ZEND_PARSE_PARAMETERS_NONE();
```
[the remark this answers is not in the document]
jmikola: Looking now at what you did in the `__construct` method, perhaps this is OK and you were just missing the error handling override?

src/MongoDB/ServerApi.c
```diff
@@ -0,0 +1,391 @@
...
+
+	intern = Z_SERVERAPI_OBJ_P(getThis());
+
+	zend_replace_error_handling(EH_THROW, phongo_exception_from_phongo_domain(PHONGO_ERROR_INVALID_ARGUMENT), &error_handling);
+
+	ZEND_PARSE_PARAMETERS_START(1, 1)
+	Z_PARAM_STRING(serialized, serialized_len)
+	ZEND_PARSE_PARAMETERS_END_EX(
+		zend_restore_error_handling(&error_handling);
+		return );
```
jmikola, line 269: Is the semicolon after `return` purposely omitted? I know the macro adds its own, but for our readability perhaps it makes sense to write this as:

```c
ZEND_PARSE_PARAMETERS_END_EX(
    zend_restore_error_handling(&error_handling);
    return;
);
```

tests/utils/tools.php
```diff
@@ -88,6 +88,15 @@ function get_module_info($row)
     return $matches[1];
 }
 
+function create_test_manager(string $uri = null, array $options = [], array $driverOptions = [])
+{
+    if (getenv('API_VERSION') && ! isset($driverOptions['serverApi'])) {
+        $driverOptions['serverApi'] = new \MongoDB\Driver\ServerApi(getenv('API_VERSION'));
+    }
+
+    return new MongoDB\Driver\Manager($uri ?? URI, $options, $driverOptions);
```
jmikola, line 97: Top of the file has `use MongoDB\Driver\Manager;` so this can just be Manager.

tests/utils/tools.php
```diff
@@ -88,6 +88,15 @@ function get_module_info($row)
     return $matches[1];
 }
 
+function create_test_manager(string $uri = null, array $options = [], array $driverOptions = [])
+{
+    if (getenv('API_VERSION') && ! isset($driverOptions['serverApi'])) {
+        $driverOptions['serverApi'] = new \MongoDB\Driver\ServerApi(getenv('API_VERSION'));
```
jmikola, line 94: Consider a `use` statement at the top, as we do for other classes.

.evergreen/config.yml
```diff
@@ -585,6 +585,19 @@ tasks:
...
             TESTS: "tests/atlas.phpt"
 
+    - name: "test-requireApiVersion"
+      tags: ["versioned_api"]
+      commands:
+        - func: "compile driver"
+        - func: "bootstrap mongo-orchestration"
+          vars:
+            TOPOLOGY: "server"
+            AUTH: "auth"
```
jmikola, line 595: Is this related to https://github.com/mongodb/mongo-php-driver/pull/1206?

tests/bson-corpus/array-decodeError-001.phpt
```diff
@@ -7,7 +7,7 @@ DO NOT EDIT THIS FILE
 --FILE--
 <?php
 
-require_once __DIR__ . '/../utils/tools.php';
```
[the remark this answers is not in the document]
alcaeus: Yes, I replaced all instances of `tools.php` to require the `.inc` file for consistency. I can revert this change if you think that's unwise.

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "src/libmongoc"]
 	path = src/libmongoc
-	url = https://github.com/mongodb/mongo-c-driver.git
+	url = https://github.com/alcaeus/mongo-c-driver.git
```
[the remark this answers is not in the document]
alcaeus: I had not managed to create the PR before. https://github.com/mongodb/mongo-c-driver/pull/765 contains the changes necessary, most notably fixing a wrong function name and adding a getter for the `version` field in `mongoc_version_api_t`.

.evergreen/config.yml
alcaeus: Yes, this is how I discovered this. I'll rebase this PR once the original has merged, which will also take care of the remaining test failures.

src/MongoDB/ServerApi.c
```diff
@@ -0,0 +1,391 @@
+/*
+ * Copyright 2015-2017 MongoDB, Inc.
```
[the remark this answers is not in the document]
alcaeus: Done and created PHPC-1792 to correct this across all files.

src/MongoDB/ServerApi.c
alcaeus: Unfortunately, clang-format does not allow us to indent these blocks and changes them right back.

src/MongoDB/ServerApi.c
alcaeus: I omitted it on purpose since the macro appends its own, but after seeing what clang-format does when we do have a semicolon (the resulting line is `return;);` I decided to extract all this functionality into corresponding `PHONGO_PARSE_PARAMETERS` macros which take care of calling `zend_replace_error_handling` appropriately.

src/MongoDB/ServerApi.c
alcaeus: Replaced this with an exception instructing users to file a bug report as it shouldn't happen.

tests/utils/tools.php
alcaeus: You are correct...the joy of not having PHP code assist in my IDE.

tests/utils/tools.php
alcaeus: Fixed

src/MongoDB/ServerApi.c
jmikola: Should we standardize on American English spelling ("initialized") here?

phongo_compat.h
```diff
@@ -169,6 +169,44 @@
...
+
+/* Compatibility macros to override error handling logic */
+#define PHONGO_PARSE_PARAMETERS_START(min_num_args, max_num_args)               \
+	do {                                                                        \
+		zend_error_handling error_handling;                                     \
+		zend_replace_error_handling(                                            \
+			EH_THROW,                                                           \
+			phongo_exception_from_phongo_domain(PHONGO_ERROR_INVALID_ARGUMENT), \
+			&error_handling);                                                   \
+	ZEND_PARSE_PARAMETERS_START(min_num_args, max_num_args)
```
jmikola, line 187: Should this be indented consistently with the lines above since it falls within `do {`?

phongo_compat.h
```diff
@@ -169,6 +169,44 @@
...
+
+#define PHONGO_PARSE_PARAMETERS_END                   \
+	ZEND_PARSE_PARAMETERS_END_EX(                     \
```
jmikola, line 190: Same question about indentation here.

tests/bson-corpus/array-decodeError-001.phpt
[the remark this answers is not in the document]
jmikola: It's not necessary but I can see how it'd make maintenance easier by not having to think about which file to include. No objections to leaving as-is.

jmikola (approved):
Some questions on code formatting for macros, but LGTM otherwise.

phongo_compat.h
alcaeus: You're right, totally missed that. Thanks!

alcaeus: Ah...I was wondering about this: it seems like clang-format immediately wants to undo my changes and restore this broken formatting, so I'm afraid we're stuck with this for the time being :(

src/MongoDB/ServerApi.c
alcaeus: I forgot we use Simplified English as opposed to Traditional English. I'll update it.

alcaeus:
> Beyond that, take a look at the CI failures. AppVeyor reports a compilation failure in some macro.

This was due to me using `ZEND_STRL` in `zend_hash_str_add`, which is implemented as a macro in PHP < 7.3. I've removed this usage for the time being. According to @cmb69, `ZEND_STRL` should not be used; this will most likely also affect other usages of `ZEND_STRL` in similar instances.

> The Evergreen failure appears to be related to what you fixed in #1206.

Yes, those are fixed as well now.

alcaeus:
Note: holding off on merge until https://github.com/mongodb/mongo-c-driver/pull/765 is merged.

jmikola:
> This was due to me using ZEND_STRL in zend_hash_str_add, which is implemented as a macro in PHP < 7.3. I've removed this usage for the time being. According to @cmb69, ZEND_STRL should not be used; this will most likely also affect other usages of ZEND_STRL in similar instances.

I'm familiar with that. I assume it's because `ZEND_STRL` may not be expanded before the preprocessor checks arguments for the outer macro (`zend_hash_str_add` in this case).

Our previous issues with `ZEND_STRL` had to do with it only being suitable for literals (see: https://github.com/mongodb/mongo-php-driver/search?q=ZEND_STRL&type=commits). That seems unrelated to this.

I don't think we need to stop using `ZEND_STRL` altogether, though. The build failures when we do something wrong should be obvious :)

PR #1206, opened 2021-03-18
Fix running auth tests

alcaeus:
While changing test setup to rely on drivers-evergreen-tools, I missed that the `AUTH` variable needs to be "auth" for drivers-evergreen-tools while "yes" was used for our own scripts. This PR updates the build config and fixes a test that can be flaky due to potential elections in the replica set.

### DIFF
--- a/tests/server/server-getTags-002.phpt
+++ b/tests/server/server-getTags-002.phpt
@@ -12,20 +12,29 @@ $manager = new MongoDB\Driver\Manager(URI);
 $command = new MongoDB\Driver\Command(['ping' => 1]);
 $manager->executeCommand(DATABASE_NAME, $command);
 
-foreach ($manager->getServers() as $server) {
-    $tags = $server->getTags();
-    echo "dc: ", array_key_exists('dc', $tags) ? $tags['dc'] : 'not set', "\n";
-    echo "ordinal: ", array_key_exists('ordinal', $tags) ? $tags['ordinal'] : 'not set', "\n";
+function assertSomeServerHasTags(array $servers, array $expectedTags) {
+    foreach ($servers as $server) {
+        /* Using a non-strict comparison guards against tags being returned in
+         * a different order than expected. */
+        if ($expectedTags == $server->getTags()) {
+            printf("Found server with tags: %s\n", json_encode($expectedTags));
+            return;
+        }
+    }
+
+    printf("No server has tags: %s\n", json_encode($expectedTags));
 }
+
+$servers = $manager->getServers();
+assertSomeServerHasTags($servers, ['dc' => 'ny', 'ordinal' => 'one']);
+assertSomeServerHasTags($servers, ['dc' => 'pa', 'ordinal' => 'two']);
+assertSomeServerHasTags($servers, []);
 
 ?>
 ===DONE===
 <?php exit(0); ?>
 --EXPECT--
-dc: ny
-ordinal: one
-dc: pa
-ordinal: two
-dc: not set
-ordinal: not set
+Found server with tags: {"dc":"ny","ordinal":"one"}
+Found server with tags: {"dc":"pa","ordinal":"two"}
+Found server with tags: []
 ===DONE===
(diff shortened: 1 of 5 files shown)


tests/server/server-getTags-002.phpt
```diff
@@ -9,21 +9,23 @@ MongoDB\Driver\Server::getTags() with replica set
...
-echo "ordinal: ", array_key_exists('ordinal', $tags) ? $tags['ordinal'] : 'not set', "\n";
-
-$tags = $manager->selectServer(new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_SECONDARY))->getTags();
-echo "dc: ", array_key_exists('dc', $tags) ? $tags['dc'] : 'not set', "\n";
-echo "ordinal: ", array_key_exists('ordinal', $tags) ? $tags['ordinal'] : 'not set', "\n";
+foreach ($manager->getServers() as $server) {
+    $tags = $server->getTags();
+    echo "dc: ", array_key_exists('dc', $tags) ? $tags['dc'] : 'not set', "\n";
+    echo "ordinal: ", array_key_exists('ordinal', $tags) ? $tags['ordinal'] : 'not set', "\n";
+}
```
jmikola, line 19: Is this still a potentially flaky test if the order of servers returned by `getServers()` is non-deterministic? If so, perhaps this needs to be rewritten to instead have a function that asserts any server has certain tags. For example:

```php
function assertSomeServerHasTags($servers, $expectedTags) {
    foreach ($servers as $server) {
        if (array_intersect_assoc($expectedTags, $server->getTags()) === $expectedTags) {
            printf("Found server with tags: %s\n", json_encode($expectedTags));
            return;
        }
    }
    printf("No server has tags: %s\n", json_encode($expectedTags));
}

$servers = $manager->getServers();
assertSomeServerHasTags($servers, ['dc' => 'ny', 'ordinal' => 'one']);
assertSomeServerHasTags($servers, ['dc' => 'pa', 'ordinal' => 'two']);
assertSomeServerHasTags($servers, []);
```

My understanding of [`array_intersect_assoc()`](https://www.php.net/manual/en/function.array-intersect-assoc.php) is that its return value uses the same key ordering as the first argument, which makes it possible to do the `===` check above.

alcaeus: Good point. I've kept the loop but removed the use of `array_intersect_assoc`, as the last check will match for any server:
```
php > var_dump(array_intersect_assoc([], ['dc' => 'pa', 'ordinal' => 'two']));
array(0) {
}
```

A loose equality check works around this, and since we're expecting the tags in the order that we create them I think that should be safe but still protect us against non-deterministic order of the server returned from `getServers()`.

tests/server/server-getTags-002.phpt
```diff
@@ -12,20 +12,27 @@ $manager = new MongoDB\Driver\Manager(URI);
 $command = new MongoDB\Driver\Command(['ping' => 1]);
 $manager->executeCommand(DATABASE_NAME, $command);
 
-foreach ($manager->getServers() as $server) {
-    $tags = $server->getTags();
-    echo "dc: ", array_key_exists('dc', $tags) ? $tags['dc'] : 'not set', "\n";
-    echo "ordinal: ", array_key_exists('ordinal', $tags) ? $tags['ordinal'] : 'not set', "\n";
+function assertSomeServerHasTags(array $servers, array $expectedTags) {
+    foreach ($servers as $server) {
+        if ($expectedTags == $server->getTags()) {
```
jmikola, line 17: If `getTags()` returns an array, is there any reason we can't do a strict equality check here?

Both are going to enforce the same order AFAIK, and the associative array itself just has string keys and values (no concern with mismatched types).

jmikola (approved):
Suggestion to change to strict equality in the test, but LGTM either way.

tests/server/server-getTags-002.phpt
alcaeus: I changed to a strict comparison, which then caused tests to fail again. With some debug output, I was able to capture the following diff:
```patch
003+ No server has tags: {"ordinal":"two","dc":"pa"}
004+ Server tags: {"localhost:27017":{"ordinal":"one","dc":"ny"},"localhost:27018":{"dc":"pa","ordinal":"two"},"localhost:27019":[]}
003- Found server with tags: {"ordinal":"two","dc":"pa"}
```

As you can see, the tags returned for the second server has a different order from the first, which causes a strict equality check to fail. I've added a comment to clarify why we're using `==` in the test for future reference.

PR #1213, opened 2021-04-04
PHPC-1804: Manager::addSubscriber and removeSubscriber

jmikola:
https://jira.mongodb.org/browse/PHPC-1804


src/MongoDB/Monitoring/functions.c
```diff
@@ -25,77 +25,50 @@
 
 ZEND_EXTERN_MODULE_GLOBALS(mongodb)
 
-static char* php_phongo_make_subscriber_hash(zval* subscriber)
-{
-	char* hash;
-	int   hash_len;
-
-	hash_len = spprintf(&hash, 0, "SUBS-%09d", Z_OBJ_HANDLE_P(subscriber));
```
jmikola, line 33: After reading through SplObjectStorage's implementation, I realized [`spl_object_storage_get_hash()`](https://github.com/php/php-src/blob/php-8.0.3/ext/spl/spl_observer.c#L78) just uses the object's handle as a numeric hash key. I don't recall why we originally had this logic to use a string key, but it seems unnecessary.

src/MongoDB/Manager.c
```diff
@@ -281,6 +281,34 @@ static PHP_METHOD(Manager, __construct)
...
+	php_phongo_manager_t* intern;
+	zval*                 subscriber;
+
+	PHONGO_PARSE_PARAMETERS_START(1, 1)
+	Z_PARAM_OBJECT_OF_CLASS(subscriber, php_phongo_subscriber_ce)
+	PHONGO_PARSE_PARAMETERS_END();
+
+	intern = Z_MANAGER_OBJ_P(getThis());
+
+	/* Lazily initialize the subscriber HashTable */
```
jmikola, line 297: I considered moving the implementation of the add/removeSubscriber methods into functions within `phongo_apm.c`. If so, these would just be responsible for parsing arguments and then delegate to another function. Let me know what you think.

In the case of the Manager class, we'd still need to leave the freeing of `intern->subscribers` in `php_phongo_manager_free_object` so we can't get rid of _all_ of the APM-related code.

src/phongo_apm.c
```diff
@@ -0,0 +1,297 @@
...
+		if (Z_TYPE_P(subscriber) != IS_OBJECT || !instanceof_function(Z_OBJCE_P(subscriber), subscriber_ce)) {
+			continue;
+		}
+
+		if (zend_hash_index_exists(to, Z_OBJ_HANDLE_P(subscriber))) {
+			continue;
+		}
+
+		zend_hash_index_update(to, Z_OBJ_HANDLE_P(subscriber), subscriber);
+		Z_ADDREF_P(subscriber);
```
jmikola, line 57: I originally wasn't sure if it was absolutely necessary to add ref-counts while building this HashTable, but I think it's reasonable given a possible edge case where a subscriber could otherwise get garbage collected while dispatching a single event. Consider:

 * Register a global subscriber, such that the only reference to it is in the internal, global HashTable
 * Register a Manager subscriber
 * Execute a command on the Manager, which dispatches an event
 * The Manager subscriber gets notified first and unregisters the global subscriber, which immediately frees the first subscriber
 * Segfault while attempting to notify the global subscriber

Granted, `phongo_apm_get_subscribers` collects global subscribers before Manager subscribers so I'd expect the global subscribers will always be notified first. If the edge case above is not possible as written, I assume we could just swap the example for two different global subscribers.

src/phongo_apm.c
```diff
@@ -0,0 +1,297 @@
...
+}
+
+/* Search for a Manager associated with the given client in the request-scoped
+ * registry. If any Manager is found, copy it to @out, increment its ref-count,
+ * and return true; otherwise, set @out to undefined and return false. */
+static bool phongo_apm_copy_manager_for_client(mongoc_client_t* client, zval* out)
+{
+	php_phongo_manager_t* manager;
+
+	ZVAL_UNDEF(out);
```
jmikola, line 97: While moving this function from `php_phongo.c` to here, I moved this line up. Previously, it was only called for the final `return false` and never for the `return false` in the condition that immediately follows. That said, I don't think this changes anything I believe `php_phongo_commandstartedevent_t.manager` will be initialized to zero anyway.

src/phongo_apm.c
```diff
@@ -0,0 +1,297 @@
...
+#include <Zend/zend_exceptions.h>
+#include <Zend/zend_interfaces.h>
+#include <Zend/zend_operators.h>
+
+#include "phongo_compat.h"
+#include "php_phongo.h"
+#include "phongo_apm.h"
+
+#undef MONGOC_LOG_DOMAIN
+#define MONGOC_LOG_DOMAIN "PHONGO"
```
jmikola, line 35: I'm not sure this is necessary, as we don't log anything below. Let me know if you think we can remove it.

Docs for `mongoc_client_set_apm_callbacks` say "If any arguments are invalid, returns false and logs an error", but I don't think this would have any affect on the log domain used within.

tests/manager/manager-addSubscriber-005.phpt
```diff
@@ -0,0 +1,45 @@
...
+$m = create_test_manager();
+
+$pingCommand = new MongoDB\Driver\Command(['ping' => 1]);
+
+$subscriber = new MySubscriber;
+
+MongoDB\Driver\Monitoring\addSubscriber($subscriber);
+$m->addSubscriber($subscriber);
+
+printf("ping: %d\n", $m->executeCommand(DATABASE_NAME, $pingCommand)->toArray()[0]->ok);
```
alcaeus, line 39: It may be worth removing the manager subscriber and ensuring that the command still turns up in the global subscriber.

jmikola: Added this below. Resolve if it looks good.

src/MongoDB/Manager.c
jmikola: This is now done.

alcaeus (approved):
LGTM!

PR #1217, opened 2021-04-16
PHPC-1792: Change all copyrights to be <year>-present

agolin95:
Some of these were in the form of `Copyright <year> MongoDB, Inc.` as opposed to `Copyright <year>=<year> MongoDB, Inc.` - I changed those as well assuming we would want equivalence across files, but please lmk if that was an incorrect assumption!

I also opted for "Present" instead of "present" (template updated accordingly) - please lmk if you aren't a fan of that!

### DIFF
--- a/template.rc
+++ b/template.rc
@@ -38,7 +38,7 @@ BEGIN
             VALUE "FileDescription", FILE_DESCRIPTION
             VALUE "FileVersion", PHP_MONGODB_VERSION
             VALUE "InternalName", FILE_NAME
-            VALUE "LegalCopyright", "Copyright © 2014-present MongoDB, Inc."
+            VALUE "LegalCopyright", "Copyright 2014-Present MongoDB, Inc."
             VALUE "OriginalFilename", FILE_NAME
             VALUE "ProductName", "MongoDB Driver for PHP"
             VALUE "ProductVersion", PHP_MONGODB_VERSION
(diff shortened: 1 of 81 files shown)


jmikola:
I'm in favor of lowercase "present". That was in the original template from legal and I believe it is what is used in other projects (server for sure).

As for changing `<year>` to `<year>-present`, I think that's correct. Before we started using "present", files that were introduced only received the current year. If they were updated in a later year, that line was then updated with a range. So all copyright lines in the driver should now be in the form "<first year introduced>-present". 

> template updated accordingly

FYI, `template.rc` is just a file for the Windows DLL produced by PECL. It's serves no purpose for the driver itself.

agolin95:
@jmikola thank you for the review and sorry for the delay! Pushed a commit that changes titlecase "Present" to lowercase "present"

template.rc
```diff
@@ -38,7 +38,7 @@ BEGIN
             VALUE "FileDescription", FILE_DESCRIPTION
             VALUE "FileVersion", PHP_MONGODB_VERSION
             VALUE "InternalName", FILE_NAME
-            VALUE "LegalCopyright", "Copyright � 2014-present MongoDB, Inc."
+            VALUE "LegalCopyright", "Copyright � 2014-present MongoDB, Inc."
```
alcaeus, line 41: This looks jumbled, I think we can revert the change to this file.

alcaeus (approved):
LGTM besides the template.rc change

template.rc
alcaeus:
```suggestion
            VALUE "LegalCopyright", "Copyright © 2014-present MongoDB, Inc."
```

alcaeus (changes requested):

alcaeus (approved):

PR #1221, opened 2021-05-21
PHPC-1756: Remove oppressive terminology from source code

alcaeus:
The libmongoc update takes care of PHPC-1840 and PHPC-1803. This PR also brings in changes for PHPC-1836 and PHPC-1842.


php_phongo.c
```diff
@@ -1811,19 +1810,6 @@ static bool php_phongo_apply_rp_options_to_uri(mongoc_uri_t* uri, bson_t* option
 	while (bson_iter_next(&iter)) {
 		const char* key = bson_iter_key(&iter);
 
-		if (!ignore_slaveok && !strcasecmp(key, MONGOC_URI_SLAVEOK)) {
```
alcaeus, line 1814: See PHPC-1836

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "src/libmongoc"]
 	path = src/libmongoc
-	url = https://github.com/mongodb/mongo-c-driver.git
+	url = https://github.com/alcaeus/mongo-c-driver.git
```
alcaeus, line 3: I'll update this PR once all the pull requests in libmongoc have been merged.

php_phongo.c
```diff
@@ -1199,12 +1199,12 @@ bool php_phongo_server_to_zval(zval* retval, mongoc_server_description_t* sd) /*
...
 		PHONGO_BSON_INIT_DEBUG_STATE(state);
 
-		if (!php_phongo_bson_to_zval_ex(bson_get_data(is_master), is_master->len, &state)) {
+		if (!php_phongo_bson_to_zval_ex(bson_get_data(hello_response), hello_response->len, &state)) {
 			zval_ptr_dtor(&state.zchild);
 			return false;
 		}
 
-		ADD_ASSOC_ZVAL_EX(retval, "last_is_master", &state.zchild);
+		ADD_ASSOC_ZVAL_EX(retval, "last_hello_response", &state.zchild);
```
alcaeus, line 1207: Since we never documented the returned fields, I decided to drop the old field entirely. Let me know if you think we should keep it in for now.

jmikola: This is just used for debug output, since Manager and Server are not serializable. No objections to changing this, since we don't document debug output at all.

php_phongo.c
```diff
@@ -1553,7 +1553,7 @@ static bool php_phongo_apply_options_to_uri(mongoc_uri_t* uri, bson_t* options)
 			!strcasecmp(key, MONGOC_URI_READPREFERENCE) ||
 			!strcasecmp(key, MONGOC_URI_READPREFERENCETAGS) ||
 			!strcasecmp(key, MONGOC_URI_SAFE) ||
-			!strcasecmp(key, MONGOC_URI_SLAVEOK) ||
+			!strcasecmp(key, MONGOC_URI_SECONDARYOK) ||
```
jmikola, line 1556: If `slaveOk` handling was removed entirely from `php_phongo_apply_rp_options_to_uri`, what's the reason for renaming it to a new constant here?

This seems like it's related to [PHPC-1836](https://jira.mongodb.org/browse/PHPC-1836) and [CDRIVER-3989](https://jira.mongodb.org/browse/CDRIVER-3989) (removing `slaveOk` URI handling), but I don't see this constant in the corresponding libmongoc PR: https://github.com/mongodb/mongo-c-driver/pull/794

I found a reference to this in your [cdriver-3997](https://github.com/mongodb/mongo-c-driver/compare/master...alcaeus:cdriver-3997) branch, but it's still defined to "slaveok". Was there a more recent decision to continue parsing the legacy URI option and only rename the constant?

I didn't take a look at libmongoc's URI string parsing (not sure where to look as this work is split between so many tickets), but if PHP needs to explicitly ignore it here that suggests that libmongoc would otherwise still accept it by `mongoc_uri_option_is_bool` in the code down below.

What's the reason for libmongoc continuing to parse this if we're dropping it entirely in PHP? I realize they might not be able to drop the old constant due to ABI, but I expect they could stop using it and allow `slaveOk` in a URI string to start raising warnings for unsupported URI options (something we don't surface through PHP).

src/MongoDB/Server.c
```diff
@@ -295,7 +295,7 @@ static PHP_METHOD(Server, getTags)
 } /* }}} */
 
 /* {{{ proto array MongoDB\Driver\Server::getInfo()
-   Returns the last isMaster result document for this Server */
+   Returns the last hello response for this Server */
```
jmikola, line 298: Reminder: we'll need a separate PR to update the example in https://www.php.net/manual/en/mongodb-driver-server.getinfo.php. Users will certainly continue to get `ismaster` from older server versions, but I assume we can update the example to include the response from a 5.0+ server.

tests/connect/standalone-ssl-verify_cert-error-001.phpt
```diff
@@ -28,5 +28,5 @@ Deprecated: MongoDB\Driver\Manager::__construct(): The "allow_invalid_hostname"
 
 Deprecated: MongoDB\Driver\Manager::__construct(): The "weak_cert_validation" driver option is deprecated. Please use the "tlsAllowInvalidCertificates" URI option instead.%s
 OK: Got MongoDB\Driver\Exception\ConnectionTimeoutException thrown from executeCommand
-No suitable servers found (`serverSelectionTryOnce` set): [%s calling ismaster on '%s:%d']
+No suitable servers found (`serverSelectionTryOnce` set): [%s calling %s on '%s:%d']
```
jmikola, line 31: AFAIK, this error message comes from `mongoc-topology-scanner.c`. The template is always going to refer to either `ismaster` or `hello`, so I assume it was preferable to just use `%s` here instead of a regex pattern like `(ismaster|hello)`, correct?

tests/manager/manager-set-uri-options-002.phpt
```diff
@@ -50,7 +50,7 @@ Deprecated: MongoDB\Driver\Manager::__construct(): The "context" driver option i
 
 Deprecated: MongoDB\Driver\Manager::__construct(): The "allow_self_signed" context driver option is deprecated. Please use the "tlsAllowInvalidCertificates" URI option instead.%s
 OK: Got MongoDB\Driver\Exception\ConnectionTimeoutException
-No suitable servers found (`serverSelectionTryOnce` set): [%scalling ismaster on '%s']
+No suitable servers found (`serverSelectionTryOnce` set): [%scalling %s on '%s']
```
jmikola, line 53: I realize this was in the older diff, but is there any reason this is `%scalling` instead of `%s calling`? Might be a good chance to change it.

tests/server/server-executeBulkWrite-004.phpt
```diff
@@ -28,11 +28,11 @@ foreach ($writeConcerns as $wc) {
 ?>
 ===DONE===
 <?php exit(0); ?>
---EXPECT--
+--EXPECTF--
 OK: Got MongoDB\Driver\Exception\RuntimeException
-not master
+not %s
```
jmikola, line 33: `not %s` seems pretty generic. Would it make sense to use an error code here, or are we concerned that it might not be standardized across all server versions?

I will note that this test is expecting RuntimeException instead of ServerException. That may be because it predates the introduction of ServerException.

Assuming this only affects `server-executeBulkWrite-004.phpt` and `server-executeBulkWrite-007.phpt`, we can punt this to a separate ticket or address it here if you're confident that the code is reliable for any 3.0+ server.

src/MongoDB/Server.c
jmikola: Just saw https://github.com/mongodb/mongo-php-library/pull/828/files#diff-17b7c449b8afb0e8b5e94d80a7a42dd52689f9347aa1d18bd40f6ec30020b56dR65.

Perhaps the doc change is just a matter of swapping `ismaster` for `isWritablePrimary`.

.gitmodules
alcaeus: Yes, the change also includes my work done for CDRIVER-3997, which is ready but was already building on 2 PRs that were in review, so I held off on creating the PR for that.

php_phongo.c
alcaeus: Yes, this can be removed as libmongoc ignores the option completely

src/MongoDB/Server.c
alcaeus: PHP docs are being handled in https://github.com/alcaeus/doc-en/pull/2

tests/connect/standalone-ssl-verify_cert-error-001.phpt
alcaeus: It was the easiest, but `(ismaster|hello)` is better.

tests/manager/manager-set-uri-options-002.phpt
alcaeus: Ah, I missed that. Will change that as well.

tests/server/server-executeBulkWrite-004.phpt
alcaeus: Updated to check for `(primary|master)` to be more resilient.

jmikola: I logged [PHPC-1851](https://jira.mongodb.org/browse/PHPC-1851) to revisit changing this to error code checks.

tests/server/server-executeBulkWrite-004.phpt
```diff
@@ -28,11 +28,11 @@ foreach ($writeConcerns as $wc) {
...
+--EXPECTF--
 OK: Got MongoDB\Driver\Exception\RuntimeException
-not master
+not (primary|master)
 OK: Got MongoDB\Driver\Exception\RuntimeException
-not master
+not (primary|master)
 OK: Got MongoDB\Driver\Exception\RuntimeException
-not master
+not (primary|master)
```
jmikola, line 37: Using a regex within [EXPECTF](https://qa.php.net/phpt_details.php#expectf_section) requires wrapping the pattern with `%r`. I think this might be responsible for the CI test failures.

tests/server/server-executeBulkWrite-007.phpt
```diff
@@ -32,11 +32,11 @@ foreach ($writeConcerns as $wc) {
...
+--EXPECTF--
 OK: Got MongoDB\Driver\Exception\RuntimeException
-not master
+not (primary|master)
 OK: Got MongoDB\Driver\Exception\RuntimeException
-not master
+not (primary|master)
 OK: Got MongoDB\Driver\Exception\RuntimeException
-not master
+not (primary|master)
```
jmikola, line 41: Same comment as https://github.com/mongodb/mongo-php-driver/pull/1221/files#r638907540

jmikola (commented):
LGTM once regex patterns in EXPECTF are fixed.

tests/server/server-executeBulkWrite-004.phpt
alcaeus: Oh. I assumed they could be added directly, so I didn't bother to run them. Thanks!

alcaeus:
Test failures are due to 5.0 behaving differently. I'm fixing those in PHPC-1848.

alcaeus:
@jmikola test failures are being fixed in #1225 - should we merge this and fix the build in your PR?

jmikola (approved):
OK to sort out test failures in #1225.

PR #1222, opened 2021-05-26
PHPC-1848: Add 5.0 to the server testing matrix

alcaeus:
PHPC-1848

This also updates the config file to use a `latest-stable` alias to avoid having to update multiple build variants when it changes.

### DIFF
--- a/tests/manager/manager-executeQuery_error-003.phpt
+++ b/tests/manager/manager-executeQuery_error-003.phpt
@@ -9,7 +9,7 @@ MongoDB\Driver\Manager::executeQuery() exposes error document via CommandExcepti
 require_once __DIR__ . "/../utils/basic.inc";
 
 $manager = create_test_manager();
-$query = new MongoDB\Driver\Query(['$foo' => 1]);
+$query = new MongoDB\Driver\Query(['field' => ['$unsupportedOperator' => true]]);
 
 try {
     $manager->executeQuery(NS, $query);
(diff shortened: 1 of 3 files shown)


tests/manager/manager-executeQuery_error-003.phpt
```diff
@@ -24,7 +24,7 @@ try {
 ===DONE===
 <?php exit(0); ?>
 --EXPECT--
-MongoDB\Driver\Exception\CommandException(2): unknown top level operator: $foo
```
jmikola, line 27: Why was this change needed? Testing with 5.0 locally, this error message/code appears to still be returned for a query with `{$foo: 1}`.

While we're changing this, I would suggest using `$unsupportedOperator` to make the test more self-documenting.

jmikola (approved):
One question about the test change, but LGTM otherwise.

tests/manager/manager-executeQuery_error-003.phpt
alcaeus: Here's a diff from the test run:
```diff
- BulkWriteException: unknown top level operator: $foo
+ BulkWriteException: unknown top level operator: $foo. If you have a field name that starts with a '$' symbol, consider using $getField or $setField.
```

I'll use `$unsupportedOperator` instead.

jmikola: Oh, interesting. That must have been added in the rc0 release. I was still running an earlier 5.0 version locally.

PR #1226, opened 2021-06-21
PHPC-1721: Fix argument order for some ecalloc function calls

tanlisu:
https://jira.mongodb.org/browse/PHPC-1721

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -2471,7 +2471,7 @@ static char* php_phongo_concat_handshake_data(const char* default_value, const c
 		ret_len += custom_value_len + PHONGO_METADATA_SEPARATOR_LEN;
 	}
 
-	ret = ecalloc(ret_len, sizeof(char*));
+	ret = ecalloc(ret_len, sizeof(char));
 
 	if (custom_value) {
 		snprintf(ret, ret_len, "%s%s%s ", default_value, PHONGO_METADATA_SEPARATOR, custom_value);
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -933,7 +933,7 @@ static zval* phongo_create_implicit_session(zval* manager) /* {{{ */
 		return NULL;
 	}
 
-	zsession = ecalloc(sizeof(zval), 1);
+	zsession = ecalloc(1, sizeof(zval));
 
 	phongo_session_init(zsession, manager, cs);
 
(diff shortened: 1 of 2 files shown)


php_phongo.c
```diff
@@ -2471,7 +2471,7 @@ static char* php_phongo_concat_handshake_data(const char* default_value, const c
 		ret_len += custom_value_len + PHONGO_METADATA_SEPARATOR_LEN;
 	}
 
-	ret = ecalloc(sizeof(char*), ret_len);
+	ret = ecalloc(ret_len, sizeof(char*));
```
jmikola, line 2474: I think the second argument should be `sizeof(char)`, since we're allocating a string here and not an array of string pointers. In practice, this just means we've been allocating more memory than necessary (likely 8x as much on a 64-bit system). It looks like this mistake dates back to https://github.com/mongodb/mongo-php-driver/commit/895fcf4bbc0f70eb7537fb8ac62610840849af7b#diff-14fe3819169577796c8c63f3bea25bcf0504df8c6e131f7beb916d255c3d49d8R2499.



@alcaeus can you confirm?

php_phongo.c
```diff
@@ -2491,7 +2491,7 @@ static void php_phongo_handshake_data_append(const char* name, size_t name_len,
 	char*  full_platform;
 
 	php_version_string_len = strlen(PHP_VERSION) + PHONGO_METADATA_PHP_VERSION_PREFIX_LEN + 1;
-	php_version_string     = ecalloc(sizeof(char*), php_version_string_len);
+	php_version_string     = ecalloc(php_version_string_len, sizeof(char*));
```
jmikola, line 2494: Similar to the previous comment, I think the second arg should be `sizeof(char)` here.

jmikola (commented):
I quickly grepped for "alloc(size" in the code base and these were the only results that turned up, so I think that covers "we would do well to check other allocation functions as well" in the JIRA ticket description.

Let's wait for @alcaeus to confirm the handshake-related lines before merging, but you can make those changes now so we have this ready to go (I think it's likely he'll confirm it was an old mistake).

php_phongo.c
alcaeus: You are correct (in both cases). This should indeed by `sizeof(char)`.
```suggestion
	ret = ecalloc(ret_len, sizeof(char));
```

php_phongo.c
alcaeus: Correct.

```suggestion
	php_version_string     = ecalloc(php_version_string_len, sizeof(char));
```

PR #1229, opened 2021-06-22
PHPC-1762: CRUD spec prose tests for WriteError and WriteConcernError.errInfo

jmikola:
https://jira.mongodb.org/browse/PHPC-1762

Note: the prose test for WriteConcernError.info was originally in PHPLIB but was inadvertently removed in mongodb/mongo-php-library@a9e29770ea14df2a6f49ab18fbd3c5f588ab8b23 (mongodb/mongo-php-library#820). The prose tests don't actually depend on PHPLIB, so we can add them to PHPC directly rather than re-introduce a CrudSpecTest file in PHPLIB.

As noted in [PHPC-1762](https://jira.mongodb.org/browse/PHPC-1762), PHPC has always exposed WriteError.errInfo (since [PHPC-154](https://jira.mongodb.org/browse/PHPC-154) so no code changes were necessary (much like [CDRIVER-3892](https://jira.mongodb.org/browse/CDRIVER-3892)).


jmikola:
@alcaeus: Looking at AppVeyor, all of the builds fail due to [tests/manager/manager-ctor-serverApi-001.phpt](https://github.com/mongodb/mongo-php-driver/blob/master/tests/manager/manager-ctor-serverApi-001.phpt) attempting to execute a `hello` command on a 4.0.1 server. That test was introduced in https://github.com/mongodb/mongo-php-driver/commit/deead9687e4c294f74af0bc66f0813dcc54e3285 for [PHPC-1716](https://jira.mongodb.org/browse/PHPC-1716).

I'm not sure why that test isn't reported as a failure on other CI platforms (e.g. GitHub using 4.4 clusters). Shouldn't that test require a particular server version if it's using server API 1?

alcaeus:
@jmikola yes, that test is only expected to run on 5.0+, there's no need to test the Versioned API against an older server since we expect this to not work correctly anyways.

alcaeus (approved):
LGTM

PR #1231, opened 2021-06-23
PHPC-1490: Add support for var_export() and __set_state() in CursorId.c

tanlisu:
https://jira.mongodb.org/browse/PHPC-1490


src/MongoDB/CursorId.c
```diff
@@ -193,24 +222,16 @@ static zend_object* php_phongo_cursorid_create_object(zend_class_entry* class_ty
...
 	*is_temp = 1;
-	intern   = Z_OBJ_CURSORID(PHONGO_COMPAT_GET_OBJ(object));
-
-	array_init(&retval);
-
-#if SIZEOF_ZEND_LONG == 4
-	ADD_ASSOC_INT64_AS_STRING(&retval, "id", intern->id);
-#else
-	ADD_ASSOC_LONG_EX(&retval, "id", intern->id);
-#endif
```
jmikola, line 208: We should keep this around so `var_dump()` behavior isn't changed by this PR. That should entail copying this to `php_phongo_cursorid_get_properties_hash` and placing it within a conditional dependent on `is_debug`.

tests/cursorid/cursorid-set_state-001.phpt
```diff
@@ -0,0 +1,21 @@
...
+    'id' => "$cursorId",
+]));
+echo "\n\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+MongoDB\Driver\CursorId::__set_state(array(
+%w'id' => '7250031947823432848',
```
jmikola, line 18: You can use three spaces here (as you did in `cursorid-var_export-001.phpt`). A wildcard wouldn't really be needed unless different PHP versions started using varying indentation levels (not an issue AFAIK). Removing `%w` also means we can change `EXPECTF` to `EXPECT`.

tests/cursorid/cursorid-set_state-001.phpt
```diff
@@ -0,0 +1,21 @@
...
+MongoDB\Driver\CursorId::__set_state()
+--FILE--
+<?php
+
+$cursorId = unserialize('C:23:"MongoDB\Driver\CursorId":42:{a:1:{s:2:"id";s:19:"7250031947823432848";}}');
+
+var_export(MongoDB\Driver\CursorId::__set_state([
+    'id' => "$cursorId",
+]));
+echo "\n\n";
```
jmikola, line 11: A single newline is fine. The original tests used two just to add more space between each successive `var_export` call (assuming this was lifted from the ReadConcern tests, with multiple cases in a single phpt file).

If you remove the newline, you can adjust the expected output accordingly.

tests/cursorid/cursorid-var_export-001.phpt
```diff
@@ -0,0 +1,19 @@
+--TEST--
+MongoDB\Driver\CursorID: var_export()
```
jmikola, line 2: Small nit, but the class name is "CursorId". I realize some of the existing test files had this wrong and were using "ID". Feel free to fix that anywhere so "CursorId" is always used.

See [PHPC-1004](https://jira.mongodb.org/browse/PHPC-1004) for related discussion on this. At one point, the driver was inconsistent and used "ID" for ObjectId and "Id" for CursorId. This was never an issue in practice because class names in PHP are case-insensitive, but we standardized on "Id" for consistency.

tests/cursorid/cursorid-debug-003.phpt
```diff
@@ -0,0 +1,22 @@
+--TEST--
+MongoDB\Driver\CursorID debug output on 64-bit platform
+--SKIPIF--
+<?php if (4 == PHP_INT_SIZE) { die('skip Only for 64-bit platform'); } ?>
```
jmikola, line 4: For the sake of consistency with other tests, I'd suggest `8 !== PHP_INT_SIZE`.

As a side note, the PHP driver/library also aims to use strict equality operators (i.e. `===` and `!==`) whenever possible instead of the basic `==` and `!=` operators. The strict operators require types to match, while the basic operators allow for type juggling. This is discussed a bit more in [Comparison Operators](https://www.php.net/manual/en/language.operators.comparison.php). For some insight into why type juggling can be a bad thing, you can read [PHP Sadness: Comparison Operators](http://phpsadness.com/sad/52).

One exception is when we're dealing with comparing stdClass objects, since strict equality requires both operands to be the same object in memory. See [this paste](https://3v4l.org/2B1vV) for an example. In cases where we're using separate stdClass instances to represent BSON documents and we just want to compare them, we might defer to `==` to do a match since the objects are certainly not the same in memory.

tests/cursorid/cursorid-set_state_error-001.phpt
```diff
@@ -0,0 +1,18 @@
+--TEST--
+MongoDB\Driver\CursorId::__set_state() requires "id" string field
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/basic.inc';
+
+echo throws(function() {
+    MongoDB\Driver\CursorId::__set_state(['id' => 0]);
+}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
```
jmikola, line 10: This can be `MongoDB\Driver\Exception\InvalidArgumentException::class` instead. Many of the older test files didn't use that since the syntax was not supported in all versions of PHP that the driver itself supported. This is no longer an issue.

See: https://www.php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class for more context

Another benefit to this approach is that a typo in the class name leads to a more helpful error that the class cannot be found, rather than `throws()` simply failing some internal string comparison.

tests/cursorid/cursorid-set_state-001.phpt
```diff
@@ -0,0 +1,21 @@
+--TEST--
+MongoDB\Driver\CursorId::__set_state()
+--FILE--
+<?php
+
+$cursorId = unserialize('C:23:"MongoDB\Driver\CursorId":42:{a:1:{s:2:"id";s:19:"7250031947823432848";}}');
+
+var_export(MongoDB\Driver\CursorId::__set_state([
+    'id' => "$cursorId",
```
jmikola, line 9: Out of curiosity, how/why did you decide to use `"$cursorId"` here?

This happens to work because double quoted strings will convert the `$cursorId` variable to a string per PHP's [string parsing semantics](https://www.php.net/manual/en/language.types.string.php#language.types.string.parsing). We typically use single-quoted strings in the driver/library whenever possible to avoid such behavior (and avoid any chance of it happening inadvertently).

If you look at the [`__toString` magic method](https://www.php.net/manual/en/language.oop5.magic.php#object.tostring) on the CursorId class, you'll see that it returns the internal 64-bit integer as a string value.

There's really no need for this test to exercise `CursorId::__toString` and unserialization. You could simply this by using `'7250031947823432848'` here directly.

----

Having said all that, it looks like we might not have another test that covers `CursorId::__toString` so it'd be worth adding one (e.g. `cursorid-tostring-001.phpt`) that does something like:

```php
$cursorId = unserialize('C:23:"MongoDB\Driver\CursorId":42:{a:1:{s:2:"id";s:19:"7250031947823432848";}}');
var_dump((string) $cursorId);
```

tanlisu: Ah ok, I didn't have any clear reasoning behind using `"$cursorId"` - just figured that it would work as a string. I will add the `__toString` test!

tests/cursorid/cursorid-set_state_error-001.phpt
```diff
@@ -7,12 +7,12 @@ require_once __DIR__ . '/../utils/basic.inc';
 
 echo throws(function() {
     MongoDB\Driver\CursorId::__set_state(['id' => 0]);
-}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
+}, 'MongoDB\Driver\Exception\InvalidArgumentException::class'), "\n";
```
jmikola, line 10: Apologies if I didn't clarify in https://github.com/mongodb/mongo-php-driver/pull/1231#discussion_r658990539, but `::class` is actually a built-in class constant that resolves to a string value. Here, you're actually expecting an exception with the class name "MongoDB\Driver\Exception\InvalidArgumentException::class", which doesn't exist. Please remove the quotes.

The "ALMOST" below is telling us that an exception _was_ thrown but it didn't match the class we expected. That output is just to make the test diffs more readable and isn't something we should ever expect. For assertions such as these, we'd only ever want to see "OK" in the test output for an exact match.

jmikola (changes requested):

tests/cursorid/cursorid-set_state_error-001.phpt
tanlisu: Oh sorry got it, just fixed!

jmikola (approved):
LGTM.

When squashing and merging, you can disregard all of the individual commit messages and just use the top-level commit with "PHPC-1490: Add support for var_export() and __set_state() in CursorId.c" and the PR number.

I noticed most of the commits repeat the same message and issue number. In the future, it'd be clearer to use messages that reflect the changes within since we expect to squash everything at merge time. For example, the message for https://github.com/mongodb/mongo-php-driver/pull/1231/commits/7fb5706aa74b4b41e2a580197e78454112fb4c50 could have just mentioned that the expected exception class was being fixed. If the entire PR pertains to a single issue, prefixing each commit with an issue also shouldn't be necessary since we're going to squash everything.

This is in contrast to https://github.com/mongodb/mongo-php-library/pull/834, which will end up having multiple commits specific tickets and _not_ be squashed into one at merge time.

PR #1233, opened 2021-06-29
PHPC-1474 and PHPC-1879: Expose transaction information in debug info for session and fix Session::getTransactionOptions() leak

tanlisu:
https://jira.mongodb.org/browse/PHPC-1474
https://jira.mongodb.org/browse/PHPC-1879

### DIFF
--- a/src/MongoDB/Session.c
+++ b/src/MongoDB/Session.c
@@ -98,12 +98,20 @@ static const char* php_phongo_get_transaction_state_string(mongoc_transaction_st
 	}
 }
 
-static void php_phongo_transaction_options_to_zval(mongoc_transaction_opt_t* opts, zval* retval)
+static void php_phongo_transaction_options_to_zval(mongoc_client_session_t* cs, zval* retval)
 {
+	mongoc_transaction_opt_t*     opts;
 	int64_t                       max_commit_time_ms;
 	const mongoc_read_concern_t*  read_concern;
 	const mongoc_read_prefs_t*    read_preference;
 	const mongoc_write_concern_t* write_concern;
+
+	if (!cs) {
+		ZVAL_NULL(retval);
+		return;
+	}
+
+	opts = mongoc_session_opts_get_transaction_opts(cs);
 
 	if (!opts) {
 		ZVAL_NULL(retval);
@@ -141,6 +149,8 @@ static void php_phongo_transaction_options_to_zval(mongoc_transaction_opt_t* opt
 		phongo_writeconcern_init(&zwrite_concern, write_concern);
 		ADD_ASSOC_ZVAL_EX(retval, "writeConcern", &zwrite_concern);
 	}
+
+	mongoc_transaction_opts_destroy(opts);
 }
 
 /* {{{ proto void MongoDB\Driver\Session::advanceClusterTime(array|object $clusterTime)
@@ -332,8 +342,8 @@ static PHP_METHOD(Session, getServer)
    Returns options for the currently running transaction */
 static PHP_METHOD(Session, getTransactionOptions)
 {
-	zend_error_handling       error_handling;
-	php_phongo_session_t*     intern;
+	zend_error_handling   error_handling;
+	php_phongo_session_t* intern;
 
 	intern = Z_SESSION_OBJ_P(getThis());
 	SESSION_CHECK_LIVELINESS(intern, "getTransactionOptions")
@@ -345,7 +355,7 @@ static PHP_METHOD(Session, getTransactionOptions)
 	}
 	zend_restore_error_handling(&error_handling);
 
-	php_phongo_transaction_options_to_zval(mongoc_session_opts_get_transaction_opts(intern->client_session), return_value);
+	php_phongo_transaction_options_to_zval(intern->client_session, return_value);
 } /* }}} */
 
 /* {{{ proto string MongoDB\Driver\Session::getTransactionState()
@@ -728,8 +738,7 @@ static HashTable* php_phongo_session_get_debug_info(phongo_compat_object_handler
 	}
 
 	if (intern->client_session) {
-		const mongoc_session_opt_t* cs_opts;
-		cs_opts = mongoc_client_session_get_opts(intern->client_session);
+		const mongoc_session_opt_t* cs_opts = mongoc_client_session_get_opts(intern->client_session);
 		ADD_ASSOC_BOOL_EX(&retval, "causalConsistency", mongoc_session_opts_get_causal_consistency(cs_opts));
 	} else {
 		ADD_ASSOC_NULL_EX(&retval, "causalConsistency");
@@ -776,10 +785,10 @@ static HashTable* php_phongo_session_get_debug_info(phongo_compat_object_handler
 
 	if (intern->client_session) {
 		const char* state = php_phongo_get_transaction_state_string(mongoc_client_session_get_transaction_state(intern->client_session));
-		
+
 		if (!state) {
 			/* Exception should already have been thrown */
-    		goto done;
+			goto done;
 		}
 
 		ADD_ASSOC_STRING(&retval, "transactionState", state);
@@ -788,17 +797,9 @@ static HashTable* php_phongo_session_get_debug_info(phongo_compat_object_handler
 	}
 
 	if (intern->client_session) {
-		mongoc_transaction_opt_t* txn_opts = mongoc_session_opts_get_transaction_opts(intern->client_session);
-		
-		if (txn_opts) {
-
-			zval transaction;
-
-			php_phongo_transaction_options_to_zval(txn_opts, &transaction);
-			ADD_ASSOC_ZVAL_EX(&retval, "transactionOptions", &transaction);
-		} else {
-			ADD_ASSOC_NULL_EX(&retval, "transactionOptions");
-		}
+		zval txn_opts;
+		php_phongo_transaction_options_to_zval(intern->client_session, &txn_opts);
+		ADD_ASSOC_ZVAL_EX(&retval, "transactionOptions", &txn_opts);
 	} else {
 		ADD_ASSOC_NULL_EX(&retval, "transactionOptions");
 	}
--- a/src/MongoDB/Session.c
+++ b/src/MongoDB/Session.c
@@ -98,6 +98,51 @@ static const char* php_phongo_get_transaction_state_string(mongoc_transaction_st
 	}
 }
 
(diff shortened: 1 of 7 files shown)


tests/session/session-debug-002.phpt
```diff
@@ -55,5 +55,11 @@ object(MongoDB\Driver\Session)#%d (%d) {
   }
   ["server"]=>
   NULL
+  ["inTransaction"]=>
+  bool(true)
+  ["transactionState"]=>
+  string(9) "committed"
```
jmikola, line 61: Relevant test failure [task log](https://evergreen.mongodb.com/task_log_raw/mongo_php_driver_mongo_php7__versions~4.0_php_versions~7.3_os_php7~rhel70_test_test_replicaset_auth_patch_f317976f7487641ce355fdb101805b388920bcb8_60db6d2b7742ae77ab6f4a4b_21_06_29_18_57_48/0?type=T#L1097):

```
[2021/06/29 19:42:23.086] ========DIFF========
 [2021/06/29 19:42:23.086] 023+       ["hash"]=>
 [2021/06/29 19:42:23.086] 024+       object(MongoDB\BSON\Binary)#6 (2) {
 [2021/06/29 19:42:23.086] 025+         ["data"]=>
 [2021/06/29 19:42:23.086] 026+         string(20) "�<2�^��s��� ?!���@е"
 [2021/06/29 19:42:23.086] 027+         ["type"]=>
 [2021/06/29 19:42:23.086] 028+         int(0)
 [2021/06/29 19:42:23.086] 029+       }
 [2021/06/29 19:42:23.086] 030+       ["keyId"]=>
 [2021/06/29 19:42:23.086] 031+       int(6979303250201673730)
 [2021/06/29 19:42:23.086] 032+     }
 [2021/06/29 19:42:23.086] 036-   bool(true)
 [2021/06/29 19:42:23.086] 038-   string(9) "committed"
 [2021/06/29 19:42:23.086] 046+   bool(false)
 [2021/06/29 19:42:23.086] 048+   string(4) "none"
 [2021/06/29 19:42:23.086] ========DONE========
 [2021/06/29 19:42:23.086] FAIL MongoDB\Driver\Session debug output (after an operation) [tests/session/session-debug-002.phpt]
```

I think the problem is that you're expecting to be in a transaction and a "committed" state. The test itself never starts a transaction, so I think you should be expecting the same output as `session-debug-001.phpt`.

IIRC, the earlier part of the diff is just `run-tests.php` (the script used to execute PHPT files) outputting all of the content that matches a wildcard `%a` expectation for debugging purposes. That doesn't mean that output is incorrect, though. We typically see this when tests use _multiple_ wildcard patterns and something fails (diff output is very noisy), so I wouldn't worry about it here. Fixing the expectation for you two fields should resolve everything.

tests/session/session-debug-004.phpt
```diff
@@ -34,5 +34,11 @@ object(MongoDB\Driver\Session)#%d (%d) {
   NULL
   ["server"]=>
   NULL
+  ["inTransaction"]=>
+  NULL
+  ["transactionState"]=>
+  NULL
```
jmikola, line 40: I forgot that `Session::endSession` actually sets `intern->client_session` to null, so this is a good test that you added `if (client-session) {` around your new fields.

tests/session/session-debug-006.phpt
```diff
@@ -0,0 +1,74 @@
+--TEST--
+MongoDB\Driver\Session debug output (with transaction options)
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_libmongoc_crypto(); ?>
+<?php skip_if_not_live(); ?>
+<?php skip_if_server_version('<', '3.6'); ?>
```
jmikola, line 7: This SKIPIF was copied from a previous test file, but it's not sufficient for this case. Since you're calling `startTransaction` here, you should look to the SKIPIF logic in that method's corresponding tests:

```
<?php skip_if_not_libmongoc_crypto() ?>
<?php skip_if_no_transactions(); ?>
```

Note: this differs from the SKIPIF logic in `session-debug-005.phpt`, which uses:

```
<?php skip_if_not_mongos_with_replica_set(); ?>
<?php skip_if_server_version('<', '4.1.6'); ?>
<?php skip_if_not_clean(); ?>
```

That file is specifically testing transactions on a sharded cluster. The combination of `skip_if_not_mongos_with_replica_set` and `skip_if_server_version('<', '4.1.6');` is actually more restrictive than `skip_if_no_transactions`, which you can verify by reading through `skipif.php` and `tools.php`. That said, `session-debug-005.phpt` _does_ appear to be missing `skip_if_not_libmongoc_crypto();` so if you could add that to its SKIPIF that'd be great. In practice, we don't test libmongoc w/o crypto (needed to generate session IDs), but it's reasonable to have that checked for consistency (and self-documentation).

src/MongoDB/Session.c
```diff
@@ -764,6 +770,39 @@ static HashTable* php_phongo_session_get_debug_info(phongo_compat_object_handler
...
+			ADD_ASSOC_STRING(&retval, "transactionState", state);
+		} else {
+			ADD_ASSOC_NULL_EX(&retval, "transactionState");
+		}
+	} else {
+		ADD_ASSOC_NULL_EX(&retval, "transactionState");
+	}
+
+	if (intern->client_session) {
+		mongoc_transaction_opt_t* txn_opts = mongoc_session_opts_get_transaction_opts(intern->client_session);
```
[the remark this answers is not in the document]
tanlisu: I refactored the helper function `php_phongo_transaction_options_to_zval` and freed `opts` in it as you said to do, but I wasn't able to verify the memory leak that you noted in https://jira.mongodb.org/browse/PHPC-1879 since the `session-getTransactionOptions-001.phpt` test gets skipped when I run it. How should I test that?

src/MongoDB/Session.c
```diff
@@ -141,6 +149,10 @@ static void php_phongo_transaction_options_to_zval(mongoc_transaction_opt_t* opt
 		phongo_writeconcern_init(&zwrite_concern, write_concern);
 		ADD_ASSOC_ZVAL_EX(retval, "writeConcern", &zwrite_concern);
 	}
+
+	if (opts) {
```
jmikola, line 153: `opts` is always non-null by this point, so you don't need the conditional.

src/MongoDB/Session.c
```diff
@@ -778,27 +788,21 @@ static HashTable* php_phongo_session_get_debug_info(phongo_compat_object_handler
...
-		if (txn_opts) {
-
-			zval transaction;
-
-			php_phongo_transaction_options_to_zval(txn_opts, &transaction);
-			ADD_ASSOC_ZVAL_EX(&retval, "transactionOptions", &transaction);
-		} else {
-			ADD_ASSOC_NULL_EX(&retval, "transactionOptions");
-		}
+		zval transaction;
```
jmikola, line 803: `txn_opts` would be a more descriptive variable name here.

tests/session/session-debug-005.phpt
```diff
@@ -68,6 +69,12 @@ object(MongoDB\Driver\Session)#%d (%d) {
...
   string(11) "in_progress"
   ["transactionOptions"]=>
-  NULL
+  array(1) {
+    ["readPreference"]=>
+    object(MongoDB\Driver\ReadPreference)#%d (%d) {
+      ["mode"]=>
+      string(7) "primary"
+    }
+  }
```
jmikola, line 78: FYI: Even though the test doesn't specify a primary read preference, this is ultimately inherited from the Manager itself (where "primary" is the default). The rules for this are discussed in the [Transactions spec](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#readpreference).

src/MongoDB/Session.c
```diff
@@ -98,6 +98,63 @@ static const char* php_phongo_get_transaction_state_string(mongoc_transaction_st
...
+
+	if (!mongoc_write_concern_is_default(write_concern)) {
+		zval zwrite_concern;
+
+		phongo_writeconcern_init(&zwrite_concern, write_concern);
+		ADD_ASSOC_ZVAL_EX(retval, "writeConcern", &zwrite_concern);
+	}
+
+	if (opts) {
+		mongoc_transaction_opts_destroy(opts);
```
jmikola, line 154: I believe this change was made through the "Refactor php_phongo_transaction_options_to_zval, free the mongoc_transaction_opt_t" commit. Please amend that commit to reference PHPC-1879 so we can ensure it gets linked up in JIRA when this is ultimately merged. You'll also want to ensure that we keep that line in the final message when squashing/merging the PR.

I suppose we could just wait until merge time to add a reference to PHPC-1879, but this would give you a good reason to learn how to use `git rebase` to change messages of existing commits (among other things). See https://docs.github.com/en/github/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message#amending-older-or-multiple-commit-messages for docs on that.

src/MongoDB/Session.c
```diff
@@ -725,6 +740,7 @@ static HashTable* php_phongo_session_get_debug_info(phongo_compat_object_handler
 	}
 
 	if (intern->client_session) {
+		const mongoc_session_opt_t* cs_opts;
```
jmikola, line 743: This declaration can be combined with the subsequent line.

src/MongoDB/Session.c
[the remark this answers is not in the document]
jmikola: Why is `session-getTransactionOptions-001.phpt` skipped? libmongoc should definitely have crypto available and your local environment should be a replica set (if only a single primary) running at least MongoDB 4.0. Both of those should satisfy the SKIPIF.

If you happened to wipe your data directory for `mongod`, did you reinitialize the replica set after re-launching it? The command we used when first setting it up was [`rs.initiate()`](https://docs.mongodb.com/manual/reference/method/rs.initiate/). And recall that `mongod` should be started with the [`--replSet`](https://docs.mongodb.com/manual/reference/program/mongod/#std-option-mongod.--replSet) command line option.

If `mongod` is definitely a replica set, does your `MONGODB_URI` environment variable still specify the replica set name in the connection string?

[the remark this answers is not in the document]
tanlisu: Hmm yes, I think I am following everything correctly with `rs.initiate()`, `--replSet`, and the `MONGODB_URI`. The test is being skipped because of the  `<?php skip_if_no_transactions(); ?>`  (the message says "topology does not support transactions").

PR #1234, opened 2021-07-01
PHPC-1685 Drop support for PHP 7.0

alcaeus:
PHPC-1685

While we're only removing the (now deprecated) `SetUpTearDownTrait` at this time, more significant changes will happen once we drop PHP 7.1, as this allows us to start introducing argument types.


jmikola (approved):

jmikola:
> While we're only removing the (now deprecated) `SetUpTearDownTrait` at this time...

/me looks at https://github.com/mongodb/mongo-php-library/pull/836

/me looks back at this PHPC PR

PR #1239, opened 2021-07-19
PHPC-1805 and PHPC-1910: Implement ServerDescription class and Server::getServerDescription()

tanlisu:
https://jira.mongodb.org/browse/PHPC-1805
https://jira.mongodb.org/browse/PHPC-1910


src/MongoDB/Server.c
```diff
@@ -384,6 +385,23 @@ static PHP_METHOD(Server, getPort)
...
+	php_phongo_server_t* intern;
+
+	intern = Z_SERVER_OBJ_P(getThis());
+
+	zend_replace_error_handling(EH_THROW, phongo_exception_from_phongo_domain(PHONGO_ERROR_INVALID_ARGUMENT), &error_handling);
+	if (zend_parse_parameters_none() == FAILURE) {
+		zend_restore_error_handling(&error_handling);
+		return;
+	}
+	zend_restore_error_handling(&error_handling);
```
alcaeus, line 400: This was only first used in ServerApi and not changed in existing files, but the `PHONGO_PARSE_PARAMETERS_NONE` macro simplifies this:
```suggestion
	php_phongo_server_t* intern = Z_SERVER_OBJ_P(getThis());

	PHONGO_PARSE_PARAMETERS_NONE();
```

src/MongoDB/ServerDescription.c
```diff
@@ -0,0 +1,151 @@
...
+	const bson_t*                   getHelloResponse;
+
+	intern = Z_SERVERDESCRIPTION_OBJ_P(getThis());
+
+	zend_replace_error_handling(EH_THROW, phongo_exception_from_phongo_domain(PHONGO_ERROR_INVALID_ARGUMENT), &error_handling);
+	if (zend_parse_parameters_none() == FAILURE) {
+		zend_restore_error_handling(&error_handling);
+		return;
+	}
+	zend_restore_error_handling(&error_handling);
```
alcaeus, line 46: Same comment here about using `PHONGO_PARSE_PARAMETERS_NONE`.

tests/server/server-getServerDescription-001.phpt
```diff
@@ -0,0 +1,21 @@
...
+$manager = create_test_manager();
+
+var_dump($manager->selectServer(new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::RP_PRIMARY))->getServerDescription());
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+array(%d) {
+  %a
```
alcaeus, line 19: We may want to replace the `EXPECTF` block with some more detailed assertions. Due to the different formats it may be problematic to use this way of matching the returned data, but we could work around this by inspecting the resulting array in the test (as is done in other tests where we have the same restrictions.

src/MongoDB/ServerDescription.c
```diff
@@ -0,0 +1,357 @@
...
+
+	php_phongo_server_to_zval(&retval, intern->server_description);
+
+	return Z_ARRVAL(retval);
+} /* }}} */
+
+static HashTable* php_phongo_serverdescription_get_properties(phongo_compat_object_handler_type* object) /* {{{ */
+{
+	// return php_phongo_serverdescription_get_properties_hash(object, false);
+} /* }}} */
```
tanlisu, line 328: Is the get_properties() method needed for this class / how should I write it?

src/MongoDB/ServerDescription.c
```diff
@@ -0,0 +1,347 @@
...
+	PHP_ME(ServerDescription, getLastUpdateTime, ai_ServerDescription_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(ServerDescription, getPort, ai_ServerDescription_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(ServerDescription, getRoundTripTime, ai_ServerDescription_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(ServerDescription, getTags, ai_ServerDescription_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(ServerDescription, getType, ai_ServerDescription_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(ServerDescription, isArbiter, ai_ServerDescription_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(ServerDescription, isHidden, ai_ServerDescription_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(ServerDescription, isPassive, ai_ServerDescription_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(ServerDescription, isPrimary, ai_ServerDescription_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
+	PHP_ME(ServerDescription, isSecondary, ai_ServerDescription_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
```
jmikola, line 259: I realize these were lifted from the Server class, but since this is a new API I'd prefer to keep it minimal until someone actually requests these. Let's limit it to the relevant methods on [mongoc_server_description_t](http://mongoc.org/libmongoc/current/mongoc_server_description_t.html):

 * `getHelloResponse`
 * `getHost`
 * `getPort`
 * `getLastUpdateTime`
 * `getRoundTripTime`
 * `getType`

We can omit `getId` since that's just the internal identifier libmongoc uses and not really relevant to our users.

As far as reporting this info in debug output, we can use the following fields:

 * `hello_response`
 * `host`
 * `port`
 * `last_update_time`
 * `round_trip_time`
 * `type`

Since we've yet to drop support for 32-bit, I think you can lift logic we use elsewhere to conditionally report `last_update_time` and `round_trip_time` as strings if needed (using existing macros). I'm not too worried about the methods themselves -- those can always return a `zend_long` type.

tests/serverDescription/serverDescription-getType-001.phpt
```diff
@@ -0,0 +1,20 @@
...
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_live(); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = create_test_manager();
+$server = $manager->selectServer(new MongoDB\Driver\ReadPreference('primary'));
+
+var_dump($server->getServerDescription()->getType());
```
jmikola, line 13: If you'd like to improve the expectation, you can assert that type is either a standalone, mongos, or primary (per constants in https://www.php.net/manual/en/class.mongodb-driver-server.php). Those are the only things we'd expect given the read preference above.

src/MongoDB/ServerDescription.c
```diff
@@ -0,0 +1,234 @@
...
+{
+	php_phongo_serverdescription_t* intern = NULL;
+	zval                            retval = ZVAL_STATIC_INIT;
+
+	intern = Z_OBJ_SERVERDESCRIPTION(PHONGO_COMPAT_GET_OBJ(object));
+
+	php_phongo_server_description_to_zval(&retval, intern->server_description);
+
+	return Z_ARRVAL(retval);
+} /* }}} */
```
tanlisu, line 186: I'm not completely sure I did this part correctly - I'm getting the expected outputs in the `debug` and `var_export` tests, but I'm no longer using the `is_debug` parameter in this function.

tanlisu:
The GitHub CI test failure for the PHP 8.1 build environment (https://github.com/mongodb/mongo-php-driver/pull/1239/checks?check_run_id=3183644508) is expected I believe - see https://jira.mongodb.org/browse/PHPC-1849

jmikola:
> The GitHub CI test failure for the PHP 8.1 build environment

Please rebase on (or merge in) master to pull in https://github.com/mongodb/mongo-php-driver/pull/1241. I think it makes sense to just disable 8.1 until we address [PHPC-1849](https://jira.mongodb.org/browse/PHPC-1849).

src/MongoDB/ServerDescription.c
jmikola: Contrast your implementation here with `get_properties_hash` implementations for other classes. It looks like you modeled this function after `php_phongo_server_get_debug_info`. Server objects don't have `get_properties` handlers, which is why there is no direct equivalent of this function there.

Since the `get_debug_info` and `get_properties` handlers both expect to return `HashTable*`, you're not going to be able to use `php_phongo_server_description_to_zval` as written. That's wouldn't work with the `PHONGO_GET_PROPERTY_HASH_INIT_PROPS` macro, which I'll note is also the only thing that utilizes the `properties` field on `php_phongo_serverdescription_t` (not being used elsewhere in this PR AFAICT).

There are two different design models going on in PHPC:

1. For classes that implement both `get_debug_info` and `get_properties`, we use a common `get_properties_hash` function and a `PHONGO_GET_PROPERTY_HASH_INIT_PROPS` macro. This macro either allocates a new HashTable (for debug output) or gets a pointer to the internal HashTable on the `intern` struct (initializing it if needed). The logic there is that debug output (i.e. `get_debug_info`) always get a fresh HashTable, which is temporary and can be freed by PHP (also why we set `*is_temp = 1;` in various `get_debug_info` handlers). For `get_properties`, the HashTable doesn't need to be temporary (we can retain ownership of it) so we just return our own internal HashTable.
2. For classes that skip `get_properties`, everything happens in `get_debug_info` and there's no need for a shared utility function. In the case of Server, we _could_ allocate a HashTable to populate and return but since we already needed a `server_to_zval` function to call from Manager's own debug handler, we instead declare a local zval and use the same function. At the end of `php_phongo_server_get_debug_info` we use `Z_ARRVAL` to just return the HashTable within the zval since the top-level zval itself isn't relevant. We're not leaking memory by doing so since the dynamic memory allocation all pertains to the HashTable anyway (zval is just a local struct otherwise).

Since it'd be nice to have both handlers for ServerDescription, I think it makes sense to alter `php_phongo_server_description_to_zval` to instead populate a HashTable directly instead of taking a zval and operating on _its_ HashTable. That will mean TopologyDescription's debug handler will differ a bit from Manager's (since they're analogous to one another in the sends that they both contain many server descriptions).

Let me know if you have any questions above the above or would like me to elaborate further on something.

tanlisu: Got it! I've updated the code according to the first design model you described already, except that I populated the hashtable directly in `get_properties_hash` and removed the call to `php_phongo_server_description_to_zval`. Should I refactor this to populate the hashtable in `php_phongo_server_description_to_zval` then?

jmikola: Populating it directly in `get_properties_hash` makes sense for this PR. If you end up needing to re-use the code for TopologyDescription you can factor it out in that PR.

php_phongo.c
```diff
@@ -1162,6 +1173,35 @@ php_phongo_server_description_type_t php_phongo_server_description_type(mongoc_s
 	return PHONGO_SERVER_UNKNOWN;
 }
 
+bool php_phongo_server_description_to_zval(zval* retval, mongoc_server_description_t* sd) /* {{{ */
```
jmikola, line 1176: This is no longer used. I'd suggest removing this from the current PR and keeping a copy of it for the TopologyDescription PR if you decide to reintroduce it.

I realize that may sound futile, but it will keep the commits more focused on the necessary changes if/when we need to refer back to them.

src/MongoDB/ServerDescription.c
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	{
+		const bson_t*         hello_response = mongoc_server_description_hello_response(intern->server_description);
+		php_phongo_bson_state state;
+
+		PHONGO_BSON_INIT_DEBUG_STATE(state);
+
+		if (!php_phongo_bson_to_zval_ex(bson_get_data(hello_response), hello_response->len, &state)) {
+			return false;
```
jmikola, line 206: Since `false` is a zero value in C, like `NULL`, this likely doesn't raise a compiler warning; however, it's inconsistent since we do document the function as returning a pointer.

This may actually leak since you're not cleaning up `&state.zchild)`, as is done in other places where `php_phongo_bson_to_zval_ex` may return false (including this PR) earlier in this PR).

contrast this with `php_phongo_readpreference_get_properties_hash`, which also handles an error case. There, we clean up and immediately `goto done;` to `return props` (in whatever state it happens to be in). I believe it does get disregarded on account of the exception. Offhand, I'm not sure if there's a reason we do that vs. returning `NULL` -- but rather than dig through the git history (or PHP internals) to unearth the reason for doing so I think we can just follow convention here.

jmikola (commented):
Some comments on the code. I'll get to the tests afterwards when those points are addressed and I run this locally.

src/MongoDB/ServerDescription.c
```diff
@@ -0,0 +1,269 @@
...
+	php_phongo_serverdescription_t* intern;
+	const bson_t*                   helloResponse;
+
+	intern = Z_SERVERDESCRIPTION_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	helloResponse = mongoc_server_description_hello_response(intern->server_description);
+
+	if (helloResponse->len) {
```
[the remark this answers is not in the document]
tanlisu: Ah ok, I got it from the `getTagSets()` function in `ReadPreference`. Why is it used there?
https://github.com/mongodb/mongo-php-driver/blob/80f88146f82d5d3f52be16b60742820b2b55abdb/src/MongoDB/ReadPreference.c#L456

src/MongoDB/ServerDescription.c
```diff
@@ -0,0 +1,269 @@
...
+   Returns the server’s last update time, in microseconds */
+static PHP_METHOD(ServerDescription, getLastUpdateTime)
+{
+	php_phongo_serverdescription_t* intern;
+
+	intern = Z_SERVERDESCRIPTION_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	RETVAL_LONG((zend_long) mongoc_server_description_last_update_time(intern->server_description));
```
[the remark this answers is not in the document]
tanlisu: Ok makes sense. I used the cast because I saw that it was used in couple other places for the round trip time: https://github.com/mongodb/mongo-php-driver/blob/80f88146f82d5d3f52be16b60742820b2b55abdb/php_phongo.c#L1209
and https://github.com/mongodb/mongo-php-driver/blob/80f88146f82d5d3f52be16b60742820b2b55abdb/src/MongoDB/Server.c#L353
Should I change it in those functions as well since the explicit cast isn't necessary?

src/MongoDB/ServerDescription.c
```diff
@@ -0,0 +1,269 @@
...
+	intern = Z_SERVERDESCRIPTION_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	RETVAL_LONG((zend_long) mongoc_server_description_round_trip_time(intern->server_description));
+} /* }}} */
+
+/* {{{ proto integer MongoDB\Driver\ServerDescription::getType()
+   Returns the server’s node type */
+static PHP_METHOD(ServerDescription, getType)
```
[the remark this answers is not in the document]
tanlisu: Ok, I'll make the ticket to have `getType()` return a string. I think I already defined the string value constants in `php_phongo_serverdescription_init_ce()` - this is what you meant right?
https://github.com/mongodb/mongo-php-driver/blob/a844179bc8ddc94a936ec631d24b54f29bacd436/src/MongoDB/ServerDescription.c#L251

src/MongoDB/ServerDescription.c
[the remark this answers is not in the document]
jmikola: Looks like a mistake dating back to the first version of the driver (https://github.com/mongodb/mongo-php-driver/commit/27ee04ff888ed5e86680fef5fa3709f4c6a3a587). Would you mind opening a separate ticket as a reminder to clean that up?

Doing so wouldn't require any API changes as the method is [documented](https://www.php.net/manual/en/mongodb-driver-readpreference.gettagsets.php) as only returning an array. So that's just dead code to be removed.

src/MongoDB/ServerDescription.c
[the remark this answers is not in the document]
jmikola: Looks like both of those instances date back to https://github.com/mongodb/mongo-php-driver/commit/c795d20517944518ca592d12f6c318f8dc3abafd when we had a `phongo_long` type that was conditionally defined as a `zend_long` or `long` for PHP 7 and 5, respectively (see https://github.com/mongodb/mongo-php-driver/blob/c795d20517944518ca592d12f6c318f8dc3abafd/phongo_compat.h#L163).

That was removed in https://github.com/mongodb/mongo-php-driver/commit/b7e4a8e718d294e8e68d81c8595ad3e5d12394c2 but it looks like the commit made a broad search and replace of `phongo_long` with `zend_long`, which led to this being left over. I'd suggest making a new JIRA ticket for that and marking it as related to PHPC-1442. Feel free to open a separate PR on master if you like.

src/MongoDB/ServerDescription.c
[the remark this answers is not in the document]
jmikola: You did create constants on ServerDescription but the values for those constants are integers derived from the [`php_phongo_server_description_type_t` enum definition] in https://github.com/mongodb/mongo-php-driver/blob/1.10.0/php_phongo.h#L58.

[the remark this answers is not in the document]
tanlisu: https://jira.mongodb.org/browse/PHPC-1925

src/MongoDB/Server.c
```diff
@@ -384,6 +384,20 @@ static PHP_METHOD(Server, getPort)
...
+/* {{{ proto MongoDB\Driver\ServerDescription MongoDB\Driver\Server::getServerDescription()
+   Returns the server description for this Server */
+static PHP_METHOD(Server, getServerDescription)
+{
+	mongoc_server_description_t* server_description;
+	php_phongo_server_t*         intern = Z_SERVER_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	server_description = mongoc_client_get_server_description(Z_MANAGER_OBJ_P(&intern->manager)->client, intern->server_id);
```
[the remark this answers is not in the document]
tanlisu: https://jira.mongodb.org/browse/PHPC-1926

src/MongoDB/ServerDescription.c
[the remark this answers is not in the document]
tanlisu: https://jira.mongodb.org/browse/PHPC-1927

src/MongoDB/ServerDescription.c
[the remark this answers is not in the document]
tanlisu: https://jira.mongodb.org/browse/PHPC-1928

src/MongoDB/ServerDescription.c
```diff
@@ -0,0 +1,268 @@
...
+	php_phongo_serverdescription_t* intern;
+	const bson_t*                   helloResponse;
+
+	intern = Z_SERVERDESCRIPTION_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	helloResponse = mongoc_server_description_hello_response(intern->server_description);
+
+	php_phongo_bson_state state;
```
jmikola, line 44: For consistency we should declare all variables at the top of the scope. Perhaps you don't see this using clang on macOS, but I get the following error compiling with GCC:

```
src/MongoDB/ServerDescription.c: In function ‘zim_ServerDescription_getHelloResponse’:
src/MongoDB/ServerDescription.c:44:2: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
```

Note that newer C standards would permit this (see: https://stackoverflow.com/q/8474100).

src/MongoDB/ServerDescription.c
```diff
@@ -0,0 +1,268 @@
...
+		/* Exception should already have been thrown */
+		zval_ptr_dtor(&state.zchild);
+		return;
+	}
+
+	RETURN_ZVAL(&state.zchild, 0, 1);
+} /* }}} */
+
+/* {{{ proto string MongoDB\Driver\ServerDescription::getHost()
+   Returns the server’s hostname */
```
jmikola, line 58: Didn't realize these were curly quotes as well.

```suggestion
   Returns the server's hostname */
```

src/MongoDB/ServerDescription.c
```diff
@@ -0,0 +1,268 @@
...
+} /* }}} */
+
+HashTable* php_phongo_serverdescription_get_properties_hash(phongo_compat_object_handler_type* object, bool is_debug) /* {{{ */
+{
+	php_phongo_serverdescription_t* intern = NULL;
+	HashTable*                      props;
+
+	intern = Z_OBJ_SERVERDESCRIPTION(PHONGO_COMPAT_GET_OBJ(object));
+
+	PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_debug, intern, props, 2);
```
jmikola, line 180: The final parameter here is the initial size of the HashTable. This should agree with the number of properties we expect to fill in below to avoid PHP having to dynamically resize the HashTable.

```suggestion
	PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_debug, intern, props, 6);
```

tests/serverDescription/serverDescription-debug-001.phpt
```diff
@@ -0,0 +1,36 @@
+--TEST--
+MongoDB\Driver\ServerDescription debug output
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_live(); ?>
+<?php skip_if_not_clean(); ?>
```
jmikola, line 6: IIRC, this is only needed if the test is going to read/write data. Since you're just selecting a server here, we don't need to concern ourselves with the collection under test being empty.

tests/serverDescription/serverDescription-debug-001.phpt
```diff
@@ -0,0 +1,36 @@
...
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = create_test_manager();
+$server = $manager->selectServer(new MongoDB\Driver\ReadPreference('primary'));
```
jmikola, line 12: You can use 'primary' in other tests as well if you like, instead of the constants.

tests/serverDescription/serverDescription-getType-001.phpt
```diff
@@ -0,0 +1,36 @@
...
+
+$expected_types = array(
+    MongoDB\Driver\ServerDescription::TYPE_STANDALONE,
+    MongoDB\Driver\ServerDescription::TYPE_MONGOS,
+    MongoDB\Driver\ServerDescription::TYPE_RS_PRIMARY
+);
+
+function isExpectedType(int $type, array $expected_types) {
+    foreach ($expected_types as $expected_type) {
+        if ($expected_type == $type) {
```
jmikola, line 18: Whenever possible, use the strict equality operator. See https://www.php.net/manual/en/language.operators.comparison.php for more info.
```suggestion
        if ($expected_type === $type) {
```

jmikola: Alternatively, you can replace the entire `isExpectedType` function with [`in_array`](https://www.php.net/in_array) and use the strict option. That would let you inline a call to `in_array` in the `var_dump()` below.

tests/serverDescription/serverDescription-var_export-001.phpt
```diff
@@ -0,0 +1,31 @@
...
+<?php exit(0); ?>
+--EXPECTF--
+MongoDB\Driver\ServerDescription::__set_state(array(
+   'host' => '%s',
+   'port' => %d,
+   'type' => %d,
+   'hello_response' => 
+  array (
+    %a
+  ),
```
jmikola, line 27: It's very bizarre that `var_export` uses a three-space indent for keys, but a two-space indent for this `array (` output. I can confirm that's the case locally for PHP 7.4, though, so this is fine.

tests/serverDescription/serverDescription-getType-001.phpt
```diff
@@ -0,0 +1,36 @@
...
+$manager = create_test_manager();
+$server = $manager->selectServer(new MongoDB\Driver\ReadPreference('primary'));
+$type = $server->getServerDescription()->getType();
+
+var_dump(isExpectedType($type, $expected_types));
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
```
jmikola, line 34: No need for EXPECTF if you're not using patterns (e.g. `%d`). You can double-check other tests in case they also were using EXPECTF unnecessarily.

jmikola (commented):
Should be good to go after this round of feedback is addressed.

tests/serverDescription/serverDescription-debug-001.phpt
tanlisu: Ok, the same SKIPIF is included in `server-debug.phpt` - should I make a ticket to remove that?

jmikola: You can just edit the file in GitHub and generate a PR (instead of directly committing). I don't think we need to track that with a ticket.

tanlisu: https://github.com/mongodb/mongo-php-driver/pull/1242

jmikola:
> I'm not sure why this wasn't caught by CI. Looking at the build status for the PR, only AppVeyor (Windows) appears to have run on the most recent iteration.

I think this is due to the GitHub CI config only expecting certain branches for PRs (see: [`tests.yml`](https://github.com/mongodb/mongo-php-driver/blob/master/.github/workflows/tests.yml#L4)). I just addressed that in https://github.com/mongodb/mongo-php-driver/commit/69cc2b5c7418ef4bf320a160b3d6b0918ef7102d and merged it into the `feature/sdam-monitoring` branch. Please rebase and force-push this again to try and trigger a CI build.

The Evergreen config likely has something similar, but it's less flexible as I think we'd need an entirely different project created just for the feature branch. We can do without that for now.

tanlisu:
I rebased on the `feature/sdam-monitoring` branch but I think the GitHub CI is still missing the Coding Standards test.

tests/serverDescription/serverDescription-debug-001.phpt
jmikola: Same comment as https://github.com/mongodb/mongo-php-driver/pull/1242#discussion_r680176434

I was only suggesting that `skip_if_not_clean` be removed.

jmikola:
> I rebased on the feature/sdam-monitoring branch but I think the GitHub CI is still missing the Coding Standards test.

I didn't realize we had multiple YML configs. Fixed in https://github.com/mongodb/mongo-php-driver/commit/f3bf5a37fa05425b3cf9ffce809fca650d559245 and just merged into the feature branch.

jmikola:
The PR description had a reference to [PHPC-1849](https://jira.mongodb.org/browse/PHPC-1849), which I think was incorrect. I changed that to [PHPC-1910](https://jira.mongodb.org/browse/PHPC-1910).

Also, one of the commits says "Implement ServerDescription class and add to monitoring events". Since this PR didn't actually integrate anything with the APM events please make sure that's removed from the message when merging. In fact, I'm not sure any of the individual commit messages need to be kept when squashing, since they don't have any relevance beyond the code review. The PR title sufficiently describes everything in this case.

jmikola (approved):

PR #1240, opened 2021-07-23
PHPC-1922 Update PHONGO_CE_DISABLE_SERIALIZATION for PHP 8.1

remicollet:
From UPGRADING.INTERNALS 

```
 f. Non-serializable classes should be indicated using the
     ZEND_ACC_NOT_SERIALIZABLE (@not-serializable in stubs) rather than the
     zend_class_(un)serialize_deny handlers which are removed.

```

### DIFF
--- a/php_phongo.h
+++ b/php_phongo.h
@@ -167,19 +167,19 @@ bool php_phongo_manager_unregister(php_phongo_manager_t* manager);
 #define PHONGO_CE_FINAL(ce)             \
 	do {                                \
 		ce->ce_flags |= ZEND_ACC_FINAL; \
-	} while (0);
+	} while (0)
 
 #if PHP_VERSION_ID < 80100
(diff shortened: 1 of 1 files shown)


remicollet:
Looks like 8.1 failure is related to outdated version available (some alpha) as the change is for 8.1.0beta1

alcaeus:
@remicollet thanks for the patch. I'll take a look after the weekend and see where I need to kick the build to fix it

jmikola:
Note: this may be related to [PHPC-1849](https://jira.mongodb.org/browse/PHPC-1849), although that pertains to the Serializable interface. We may want a new ticket for this PR, which can be added to the same epic for 8.1 support.

alcaeus (approved):
Verified locally on PHP 8.1, we'll figure out the 8.1 build environment when we pick up PHPC-1850. I've also created PHPC-1922 to track this.

alcaeus:
Thanks @remicollet!

PR #1244, opened 2021-08-03
PHPC-1891, PHPC-1892, PHPC-1893: Implement TopologyDescription, SDAMSubscriber, and TopologyChangedEvent

tanlisu:
https://jira.mongodb.org/browse/PHPC-1891
https://jira.mongodb.org/browse/PHPC-1892
https://jira.mongodb.org/browse/PHPC-1893


php_phongo_structs.h
```diff
@@ -127,6 +127,12 @@ typedef struct {
 	zend_object              std;
 } php_phongo_session_t;
 
+typedef struct {
+	mongoc_topology_description_t* topology_description;
+	HashTable*                     properties;
+	zend_object                    std;
+} php_phongo_topologydescription_t;
+
```
tanlisu, line 135: Should I leave this as it is now (modeled after `php_phongo_server_description_t`) or should I still change it to hold the topology `type` and `server_descriptions` as we discussed yesterday?

jmikola: @kevinAlbs confirmed in Slack that `mongoc_topology_description_t` _doesn't_ actually depend on a libmongoc client (ignoring the APM callbacks/context property, which we can guarantee won't be used for PHP's TopologyDescription object). This can remain as-is.

That means your init method for the TopologyDescription class will be more straightforward, and complexity essentially shifts to the getType() and getServerDescriptions() methods.

tanlisu:
@jmikola For updating the libmongoc submodule, how do I determine what to update LIBMONGOC_VERSION_CURRENT to?

jmikola:
> For updating the libmongoc submodule, how do I determine what to update LIBMONGOC_VERSION_CURRENT to?

This was already covered in [Ensure libmongoc version information is correct](https://github.com/mongodb/mongo-php-driver/blob/master/CONTRIBUTING.md#ensure-libmongoc-version-information-is-correct) within the contributing docs.

I've also amended the [Update libmongoc submodule](https://github.com/mongodb/mongo-php-driver/blob/master/CONTRIBUTING.md#update-libmongoc-submodule) section to outline the process for pointing the submodule to a specific fork, which covers your case where you need to point to the commit in https://github.com/mongodb/mongo-c-driver/pull/785.

jmikola:
Note: another step for bumping libmongoc submodules is to [Update sources in build configurations](https://github.com/mongodb/mongo-php-driver/blob/master/CONTRIBUTING.md#update-sources-in-build-configurations). I recently did this in https://github.com/mongodb/mongo-php-driver/commit/ac4935338a65d7f3370befd5c245e1be68f9eeba and have since merged master into the `feature/sdam-monitoring` branch so you may want to update `feature/sdam-monitoring` and rebase this PR to incorporate those changes.

tests/apm/monitoring-topologyChanged-001.phpt
```diff
@@ -0,0 +1,38 @@
...
+
+MongoDB\Driver\Monitoring\addSubscriber( $subscriber );
+
+$cursor = $m->executeQuery( "demo.test", $query );
+?>
+--EXPECTF--
+type: Unknown
+array(%d) {%A
+}
+type: Single
```
jmikola, line 36: Did this test pass locally for you? If so, what was your `MONGODB_URI` environment variable defined as and what topology was the server/cluster it was referring to?

tests/apm/monitoring-topologyChanged-001.phpt
```diff
@@ -0,0 +1,38 @@
...
+    }
+}
+
+$query = new MongoDB\Driver\Query( [] );
+$subscriber = new MySubscriber;
+
+MongoDB\Driver\Monitoring\addSubscriber( $subscriber );
+
+$cursor = $m->executeQuery( "demo.test", $query );
+?>
```
jmikola, line 31: I suggest ending tests with:

```
===DONE===
<?php exit(0); ?>
```

This is beneficial because it allows us to catch failures that may occur after the script intentionally exits. In the case of this test, it looks like a segfault is encountered when PHP is shutting down. After modify these lines locally, I observe the test emit the following output:

```
...

===DONE===
Segmentation fault (core dumped)

Termsig=11
```

tests/apm/monitoring-topologyChanged-001.phpt
```diff
@@ -0,0 +1,38 @@
...
+--EXPECTF--
+type: Unknown
+array(%d) {%A
+}
+type: Single
+array(%d) {%A
+}
```
jmikola, line 38: Please terminate each file with a newline for consistency. I believe most other tests do so, since it's rare that we see this symbol in GitHub diffs.

src/MongoDB/Monitoring/TopologyChangedEvent.c
```diff
@@ -0,0 +1,181 @@
...
+#endif
+
+#include "phongo_compat.h"
+#include "php_phongo.h"
+
+zend_class_entry* php_phongo_topologychangedevent_ce;
+
+/* {{{ proto MongoDB\Driver\Monitoring\TopologyChangedEvent TopologyChangedEvent::getNewDescription()
+   Returns this event's new description */
+PHP_METHOD(TopologyChangedEvent, getNewDescription)
```
jmikola, line 31: Note: These methods are only referenced within this file, so you can add a `static` modifier as you did in the previous PR for ServerDescription.

src/MongoDB/Monitoring/TopologyChangedEvent.c
```diff
@@ -0,0 +1,181 @@
...
+	php_phongo_topologychangedevent_t* intern = Z_TOPOLOGYCHANGEDEVENT_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	topology_description = mongoc_topology_description_new_copy(intern->old_topology_description);
+
+	phongo_topologydescription_init(return_value, topology_description);
+} /* }}} */
+
+/* {{{ proto string TopologyChangedEvent::getTopologyId()
```
jmikola, line 57: The [SDAM Monitoring spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#topology-ids) doesn't mandate a particular type for topology IDs, but since libmongoc uses ObjectIds I think this method should return an ObjectId as well.

Most functions for initializing an ObjectId are internal to ObjectId.c for the type class itself, but you should find another example of constructing a `MongoDB\BSON\ObjectId` instance in `src/bson.c`, which contains all of the code responsible for converting BSON data to PHP types. The corresponding code therein should serve as a model for what you'll need to do here to prepare the `return_value` zval.

src/MongoDB/Monitoring/TopologyChangedEvent.c
```diff
@@ -0,0 +1,181 @@
...
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	bson_oid_to_string(&intern->topology_id, topology_id);
+	RETVAL_STRING(topology_id);
+} /* }}} */
+
+/**
+ * Event thrown when the driver observes a change in any of the servers
+ * it is connected to or a change in the overall server topology.
```
jmikola, line 72: What documentation inspired this comment? It doesn't agree with the definition of a TopologyChangedEvent in the [SDAM Monitoring spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#events).

src/MongoDB/TopologyDescription.c
```diff
@@ -0,0 +1,207 @@
...
+zend_class_entry* php_phongo_topologydescription_ce;
+
+#define PHONGO_TOPOLOGY_UNKNOWN "Unknown"
+#define PHONGO_TOPOLOGY_SINGLE "Single"
+#define PHONGO_TOPOLOGY_SHARDED "Sharded"
+#define PHONGO_TOPOLOGY_REPLICA_SET_NO_PRIMARY "ReplicaSetNoPrimary"
+#define PHONGO_TOPOLOGY_REPLICA_SET_WITH_PRIMARY "ReplicaSetWithPrimary"
+
+/* {{{ proto array MongoDB\Driver\TopologyDescription::getServers()
+   Returns an array of mongoc_server_description_t structs for all known servers in the topology */
```
jmikola, line 38: As a PHP method, this method should be returning an array of _ServerDescription objects_.

Remember that a TopologyDescription object is only created via a TopologyChangedEvent, and that event has no dependency on an active libmongoc client or PHPC Manager object.

src/MongoDB/TopologyDescription.c
```diff
@@ -0,0 +1,207 @@
...
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	sds = mongoc_topology_description_get_servers(intern->topology_description, &n);
+
+	array_init_size(return_value, n);
+
+	for (i = 0; i < n; i++) {
+		zval obj;
+
+		phongo_server_init(&obj, getThis(), mongoc_server_description_id(sds[i]));
```
jmikola, line 56: The second argument to `phongo_server_init` is `zval* manager`, but `getThis()` evaluates to the zval for the current `TopologyDescription` object. Consider what this function is actually supposed to return vs. what's being attempted here.

src/MongoDB/TopologyDescription.c
```diff
@@ -0,0 +1,207 @@
...
+} /* }}} */
+
+HashTable* php_phongo_topologydescription_get_properties_hash(phongo_compat_object_handler_type* object, bool is_debug) /* {{{ */
+{
+	php_phongo_topologydescription_t* intern = NULL;
+	HashTable*                        props;
+
+	intern = Z_OBJ_TOPOLOGYDESCRIPTION(PHONGO_COMPAT_GET_OBJ(object));
+
+	PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_debug, intern, props, 2);
```
jmikola, line 128: In other places where `PHONGO_GET_PROPERTY_HASH_INIT_PROPS` is called, a subsequent line returns early if `intern` is uninitialized. Depending on the struct definition, those lines either consult a pointer that should be non-zero or a boolean (in cases where the `intern` is so simple it doesn't contain a pointer). That's missing here, and it looks like it was also missed in the PR for ServerDescription.c.

You can address this here, but I'd suggest a separate PR to go back and revise ServerDescription.c.

We don't test for this, but the early return is a sanity check to avoid making further access to `intern` if it happens to be uninitialized (e.g. only zero-allocated).

src/phongo_apm.c
```diff
@@ -262,6 +262,36 @@ static void phongo_apm_command_failed(const mongoc_apm_command_failed_t* event)
...
+	client      = mongoc_apm_topology_changed_get_context(event);
+	subscribers = phongo_apm_get_subscribers_to_notify(php_phongo_sdamsubscriber_ce, client);
+
+	/* Return early if there are no APM subscribers to notify */
+	if (zend_hash_num_elements(subscribers) == 0) {
+		goto cleanup;
+	}
+
+	object_init_ex(&z_event, php_phongo_topologychangedevent_ce); // todo
+	p_event = Z_TOPOLOGYCHANGEDEVENT_OBJ_P(&z_event);             // todo
```
jmikola, line 281: What are the `// todo` comments referring to?

src/phongo_apm.c
```diff
@@ -262,6 +262,36 @@ static void phongo_apm_command_failed(const mongoc_apm_command_failed_t* event)
...
+
+	mongoc_apm_topology_changed_get_topology_id(event, &p_event->topology_id);
+	p_event->new_topology_description = mongoc_topology_description_new_copy(mongoc_apm_topology_changed_get_new_description(event));
+	p_event->new_topology_description = mongoc_topology_description_new_copy(mongoc_apm_topology_changed_get_previous_description(event));
```
jmikola, line 285:
```suggestion
	p_event->old_topology_description = mongoc_topology_description_new_copy(mongoc_apm_topology_changed_get_previous_description(event));
```

tests/apm/monitoring-topologyChanged-001.phpt
```diff
@@ -0,0 +1,38 @@
...
+
+$query = new MongoDB\Driver\Query( [] );
+$subscriber = new MySubscriber;
+
+MongoDB\Driver\Monitoring\addSubscriber( $subscriber );
+
+$cursor = $m->executeQuery( "demo.test", $query );
+?>
+--EXPECTF--
```
jmikola, line 32: Does this test pass for you locally?

src/MongoDB/Monitoring/TopologyChangedEvent.c
tanlisu: I got the description from the [libmongoc docs](http://mongoc.org/libmongoc/current/mongoc_apm_topology_changed_t.html) - which should we follow?

src/MongoDB/TopologyDescription.c
tanlisu: https://jira.mongodb.org/browse/PHPC-1934

src/phongo_apm.c
tanlisu: Sorry, forgot to clean that up!

tests/apm/monitoring-topologyChanged-001.phpt
tanlisu: Ah, it did pass locally but I realize now that my `MONGODB_URI` wasn't set correctly.

src/MongoDB/Monitoring/TopologyChangedEvent.c
jmikola: I'm not aware of any other class files that have a comment like this documenting the class itself. I'd just suggest we remove this. If anything, this is the kind of documentation that we'd ultimately want in the PHP.net API documentation.

tanlisu: Ok - I got it from the other event classes like `CommandStartedEvent` which contain a similar comment https://github.com/mongodb/mongo-php-driver/blob/6a91e0c8cfc23f5fb0ba90ea125856eaa30ea9db/src/MongoDB/Monitoring/CommandStartedEvent.c#L156

jmikola: Noted. I still think we can do without it here.

src/MongoDB/TopologyDescription.c
```diff
@@ -0,0 +1,236 @@
...
+/* {{{ proto boolean MongoDB\Driver\TopologyDescription::hasReadableServer()
+    Returns whether the topology has a readable server available */
+static PHP_METHOD(TopologyDescription, hasReadableServer)
+{
+	php_phongo_topologydescription_t* intern          = Z_TOPOLOGYDESCRIPTION_OBJ_P(getThis());
+	const mongoc_read_prefs_t*        read_preference = NULL;
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	RETVAL_BOOL(mongoc_topology_description_has_readable_server(intern->topology_description, read_preference));
```
tanlisu, line 72: Should `read_preference` be set to something here or is `NULL` fine (the [docs](http://mongoc.org/libmongoc/current/mongoc_topology_description_has_readable_server.html) say it will just use the default read preferences)? I remember you mentioned looking at `php_phongo_manager_select_server` from the `Manager` class, but that function seems to get the read preference from a provided zval which I don't think I have here.

tests/topologyDescription/topologyDescription-debug-001.phpt
```diff
@@ -0,0 +1,40 @@
...
+object(MongoDB\Driver\TopologyDescription)#%d (%d) {
+  %a
+}
+object(MongoDB\Driver\TopologyDescription)#%d (%d) {
+  %a
+}
+object(MongoDB\Driver\TopologyDescription)#%d (%d) {
+  %a
+}
+===DONE===
```
tanlisu, line 40: I wrote the expected outputs for all the tests according to the results I got from running them, but I don't really understand why each test results in three outputs and am not sure whether I should make the expected outputs more specific.

tanlisu:
The GitHub CI build failures (https://github.com/mongodb/mongo-php-driver/pull/1244/checks?check_run_id=3256294983) say "implicit declaration of function ‘mongoc_topology_description_new_copy’" - is this because I didn't update the libmongoc submodule correctly? I updated `src/LIBMONGOC_VERSION_CURRENT` using `make libmongoc-version-current` and checked that `PHP_MONGODB_MONGOC_SOURCES` was updated in both `config.m4` and `config.w32`.

jmikola:
> Is this because I didn't update the libmongoc submodule correctly?

Most likely. I don't see any commits in this PR updating `.gitmodules` or the `src/libmongoc` submodule. If you look at previous commits in PHPC where we update the libmongoc submodule, you'll see that we change `src/LIBMONGOC_VERSION_CURRENT` and the submodule concurrently. In your case, since you also need to update the submodule to point to a different repository, you also need to update `.gitmodules` (as discussed in https://github.com/mongodb/mongo-php-driver/pull/1244#issuecomment-893017866 and the recently updated docs linked therein).

tests/topologyDescription/topologyDescription-debug-001.phpt
jmikola: We can take a look at the output later, but it's natural for there to be multiple TopologyChanged events emitted. That occurs in many of the existing [SDAM Monitoring specs tests](https://github.com/mongodb/specifications/tree/master/source/server-discovery-and-monitoring/tests/monitoring).

In this particular test, the only purpose of the event subscriber is to provide a TopologyDescription object. We don't actually care about what or how many events are dispatched. The subscriber can be changed to only capture a single TopologyDescription object. Some other points to consider:

 * The add/removeSubscriber functions can also be called from anywhere.
 * You can use Manager methods instead of global functions. That would actually be make more sense, since those methods were created with SDAM Monitoring in mind (see: [PHPC-1666](https://jira.mongodb.org/browse/PHPC-1666)).
 * The subscriber is an object, so you can structure it as you like with properties and extra methods if you like. There is no requirement to do everything from within the event handling method. If you look through PHPC tests for existing CommandSubscriber implementations, you'll see a few different variations.

src/MongoDB/TopologyDescription.c
jmikola: The read preference should be an optional parameter to this method. When we reviewed this earlier in the week I suggested making a note of `Manager::selectServer()`, which also takes a read preference argument for calling [mongoc_client_select_server|http://mongoc.org/libmongoc/1.18.0/mongoc_client_select_server.html].

`Manager::selectServer()`'s argument happens to be required (although it need not be -- see: [PHPC-1937](https://jira.mongodb.org/browse/PHPC-1937)), but there are other examples of parsing optional arguments throughout PHPC. When using `zend_parse_parameters`, everything to the right of a pipe character (`|`) is considered optional. Since this is a new function, I think it'd be a better opportunity to forgo `zend_parse_parameters` and instead use the new PHP 7+ macros, which are documented in [The PHP function structure and API, in C](https://www.phpinternalsbook.com/php7/extensions_design/php_functions.html#the-php-function-structure-and-api-in-c) in the internals book. Andreas also used those when implementing the ServerApi class for Versioned API (see: `ServerApi::__construct`).

tanlisu:
Hmm I've updated `.gitmodules` and `src/libmongoc` (https://github.com/mongodb/mongo-php-driver/pull/1244/commits/20ca7fc84a27d95bf28954328692b3c7b013635b) but the `src/libmongoc` submodule doesn't seem to be updating - do you know what might be wrong?

.gitmodules
```diff
@@ -1,6 +1,7 @@
 [submodule "src/libmongoc"]
 	path = src/libmongoc
-	url = https://github.com/mongodb/mongo-c-driver.git
+	url = https://github.com/jmikola/mongo-c-driver.git
+	branch = cdriver-3909
```
jmikola, line 4: Adding `branch` here shouldn't be necessary.

[Updating the submodule](https://github.com/mongodb/mongo-php-driver/blob/master/CONTRIBUTING.md#update-libmongoc-submodule) should only entail entering the `src/libmongoc` directory, checking out the appropriate commit (in this case, the commit on my fork), and then going back to the root directory of the mongo-php-driver repository and adding `src/libmongoc` to your commit.

Your PR is missing a change to `src/libmongoc`, so I assume you missed the last step.

For example, if you look at https://github.com/mongodb/mongo-php-driver/commit/ac4935338a65d7f3370befd5c245e1be68f9eeba you'll see a change for "Submodule libmongoc updated 155 files". That is the result of me checking out a different commit inside of `src/libmongoc` and then going back to the root PHPC repository and adding that path to a commit. Before doing so `git status` should also indicate that `src/libmongoc` has changes, which indicates that you're able to add it to a commit.

jmikola: I also want to confirm that you were able to successfully add my libmongoc fork as a remote (e.g. `git remote add...`) within `src/libmongoc`, fetch from it (to pull in references to its branches/commits), and checkout the `cdriver-3909` branch on your local machine.

If so, you should be left with `src/libmongoc` pointing to https://github.com/jmikola/mongo-c-driver/commit/be7016760646e0ca129956da1933e3f5983db23b. The branch name itself is just a pointer to that commit, and the commit hash is all that is used by git when updating the submodule.

I don't recall ever using a `branch` option in `.gitmodules`. Changing the `url` should be all that is necessary as it will ensure git refers to the correct repository (my fork in this case) that contains the commit hash you're updating `src/libmongoc` to reference.

tanlisu: Oh got it! Yes, I added your libmongoc fork as a remote; the part I was missed was adding the `src/libmongoc` to my commit.

tests/topologyDescription/topologyDescription-debug-001.phpt
tanlisu: I updated the tests to use the `Manager::addSubscriber()` method. How would I change the subscriber to only capture a single TopologyDescription object? I think the tests are failing on the GitHub CI (https://github.com/mongodb/mongo-php-driver/pull/1244/checks?check_run_id=3263852984) right now because only two TopologyChanged events are emitted, but when I run locally I get three.

jmikola: Sorry, missed this last comment in my GitHub notifications.

> How would I change the subscriber to only capture a single TopologyDescription object?

We know the `topologyChanged()` _method_ is going to be invoked for each TopologyChangedEvent emitted by libmongoc's SDAM layer; however you control the implementation of the `topologyChanged()` method and have full control over what to do (or not do) with the event object(s) you receive.

tanlisu:
@jmikola I realize I didn't add arginfo for the `TopologyDescription` class as in https://github.com/mongodb/mongo-php-driver/pull/1245 - does it sound right to add that for just the `hasReadableServer` function which takes a read preference as argument?

jmikola:
> I realize I didn't add arginfo for the TopologyDescription class as in #1245 - does it sound right to add that for just the hasReadableServer function which takes a read preference as argument?

Yes. Arginfo should definitely be correct for any method. The arginfo for `hasReadableServer` will resemble that for `Manager::selectServer()`, but you'll need to change some parameters in both macros to convey that the ReadPreference argument is optional and can be null.

Consider the following definitions of both macros from LXR:

 * [ZEND_BEGIN_ARG_INFO_EX](http://lxr.jpauli.tech/source/xref/php-7.4/Zend/zend_API.h#121)
 * [ZEND_ARG_OBJ_INFO](http://lxr.jpauli.tech/source/xref/php-7.4/Zend/zend_API.h#100)

tests/topologyDescription/topologyDescription-getType-001.phpt
```diff
@@ -0,0 +1,44 @@
...
+$manager->addSubscriber($subscriber);
+
+$command = new MongoDB\Driver\Command(['ping' => 1]);
+$manager->executeCommand(DATABASE_NAME, $command);
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+bool(true)
```
jmikola, line 43: Related to my earlier comment in this file: Rather than use `bool(true)` here, I think it'd be clearer to just dump `getType()` directly and use `%r` to match one of three possible output strings. You can read [EXPECTF Section](https://qa.php.net/phpt_details.php#expectf_section) for reference on the operator and should also find various instances of it being used in the PHPC test suite by grepping for `%r(`. In this case, we'd use a fairly basic regex pattern and should only need to use `|` operator to enumerate the three possibilities.

tests/topologyDescription/topologyDescription-hasReadableServer-001.phpt
```diff
@@ -0,0 +1,36 @@
...
+$manager->addSubscriber($subscriber);
+
+$command = new MongoDB\Driver\Command(['ping' => 1]);
+$manager->executeCommand(DATABASE_NAME, $command);
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+bool(%s)
```
jmikola, line 35: This test would be improved by capturing the final TopologyDescription and asserting that `hasReadableServer()` returns true.

tests/topologyDescription/topologyDescription-hasReadableServer-002.phpt
```diff
@@ -0,0 +1,37 @@
...
+$manager->addSubscriber($subscriber);
+
+$command = new MongoDB\Driver\Command(['ping' => 1]);
+$manager->executeCommand(DATABASE_NAME, $command);
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+bool(%s)
```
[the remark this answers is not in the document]
jmikola: Related to my earlier comment in `topologyDescription-hasReadableServer-002.phpt`, I think this test could benefit from the same change of expecting `hasReadableServer()` to return `true`.

That would also set us up to add future tests down the line, such as one that requires connecting to a standalone server (not a replica set or sharded cluster) and expects `hasReadableServer(new ReadPreference('secondary'))` to return `false`.

tests/topologyDescription/topologyDescription-hasWritableServer-001.phpt
```diff
@@ -0,0 +1,36 @@
...
+$manager->addSubscriber($subscriber);;
+
+$command = new MongoDB\Driver\Command(['ping' => 1]);
+$manager->executeCommand(DATABASE_NAME, $command);
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+bool(%s)
```
jmikola, line 35: Related to my earlier comments, we can change this test to capture the last TopologyDescription object and expect `true` here.

tests/topologyDescription/topologyDescription-var_export-001.phpt
```diff
@@ -0,0 +1,41 @@
+--TEST--
+MongoDB\Driver\TopologyDescription: var_export()
```
jmikola, line 2: Since `__set_state` is not implemented on this class (for the same reason it's not serializable), I'd suggest we avoid testing `var_export` and stick to `var_dump` and the "debug" test(s). This file can be deleted.

tests/topologyDescription/topologyDescription-debug-001.phpt
```diff
@@ -0,0 +1,38 @@
...
+
+$command = new MongoDB\Driver\Command(['ping' => 1]);
+$manager->executeCommand(DATABASE_NAME, $command);
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+object(MongoDB\Driver\TopologyDescription)#%d (%d) {
+  %a
```
jmikola, line 36: In the `var_export` test below you asserted that the debug output at least contained "servers" and "type" fields. Why is this empty?

The `get_debug_info` and `get_properties` handlers both call `php_phongo_topologydescription_get_properties_hash`, so there isn't much technical difference between `var_dump` and `var_export` apart from how PHP will present returned HashTable. But using `%a` here essentially isn't testing `php_phongo_topologydescription_get_properties_hash` at all, since we're not expecting _any_ fields.

This relates my other other comment about `%a` and `%A` being very liberal wildcards and generally not something to use this high up. If you look at other uses of them within the test suite I think you'll generally find them used for matching arbitrary binary data (which could include newline characters) or the body of `Server::getInfo()` (which originates from the server).

If you end up deleting the `var_export` test as I suggest, this should be revised in turn to expect those fields. Also, if you change the test to capture the final TopologyDescription instead of the first, you should be able to assert at least one element in the server object (since that'd be common to any topology we connect to and run a `ping` command). And the `type` field could use `%r` to expect one of the three possible strings we'd expect as well.

php_phongo.c
```diff
@@ -1222,6 +1262,38 @@ bool php_phongo_server_to_zval(zval* retval, mongoc_server_description_t* sd) /*
...
+		array_init_size(&servers, n);
+
+		for (i = 0; i < n; i++) {
+			zval obj;
+
+			if (!php_phongo_server_description_to_zval(&obj, sds[i])) {
+				/* Exception already thrown */
+				zval_ptr_dtor(&obj);
+				zval_ptr_dtor(&servers);
+				mongoc_server_descriptions_destroy_all(sds, n);
```
jmikola, line 1283: Is `mongoc_server_descriptions_destroy_all` part of the general cleanup done when recovering from an error, why isn't is also called during the non-error code path? I believe this leaks; however, it's hidden by the fact that the TopologyDescription debug test currently dumps a description with no servers (i.e. `n=0`).

Changing `topologyDescription-debug-001.phpt` to dump the final TopologyDescription results in the following in `topologyDescription-debug-001.mem` when running with `TEST_PHP_ARGS=-m` (Valgrind):

```
==1239422== 8 bytes in 1 blocks are indirectly lost in loss record 5 of 33
==1239422==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==1239422==    by 0x784468: __zend_malloc (zend_alloc.c:3030)
==1239422==    by 0x7844D5: __zend_calloc (zend_alloc.c:3042)
==1239422==    by 0x8379BE6: php_phongo_calloc (php_phongo.c:3434)
==1239422==    by 0x82B1B92: bson_malloc0 (bson-memory.c:108)
==1239422==    by 0x82FD6AE: mongoc_generation_map_new (mongoc-generation-map.c:61)
==1239422==    by 0x82FD6D2: mongoc_generation_map_copy (mongoc-generation-map.c:71)
==1239422==    by 0x8323991: mongoc_server_description_new_copy (mongoc-server-description.c:830)
==1239422==    by 0x833BA73: mongoc_topology_description_get_servers (mongoc-topology-description.c:2281)
==1239422==    by 0x83A8581: php_phongo_topologydescription_get_properties_hash (TopologyDescription.c:172)
==1239422==    by 0x83A8705: php_phongo_topologydescription_get_debug_info (TopologyDescription.c:208)
==1239422==    by 0x8A0A6A: zend_std_get_properties_for (zend_object_handlers.c:1779)
==1239422== 
==1239422== 64 bytes in 1 blocks are indirectly lost in loss record 22 of 33
==1239422==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==1239422==    by 0x784468: __zend_malloc (zend_alloc.c:3030)
==1239422==    by 0x8379BBD: php_phongo_malloc (php_phongo.c:3429)
==1239422==    by 0x82B1B11: bson_malloc (bson-memory.c:70)
==1239422==    by 0x829FA85: bson_copy_to (bson.c:2222)
==1239422==    by 0x8324A0F: mongoc_server_description_set_topology_version (mongoc-server-description.c:1252)
==1239422==    by 0x83234F8: mongoc_server_description_handle_hello (mongoc-server-description.c:728)
==1239422==    by 0x8323923: mongoc_server_description_new_copy (mongoc-server-description.c:814)
==1239422==    by 0x833BA73: mongoc_topology_description_get_servers (mongoc-topology-description.c:2281)
==1239422==    by 0x83A8581: php_phongo_topologydescription_get_properties_hash (TopologyDescription.c:172)
==1239422==    by 0x83A8705: php_phongo_topologydescription_get_debug_info (TopologyDescription.c:208)
==1239422==    by 0x8A0A6A: zend_std_get_properties_for (zend_object_handlers.c:1779)
==1239422== 
==1239422== 1,024 bytes in 1 blocks are indirectly lost in loss record 29 of 33
==1239422==    at 0x483DFAF: realloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==1239422==    by 0x78451B: __zend_realloc (zend_alloc.c:3049)
==1239422==    by 0x8379C0F: php_phongo_realloc (php_phongo.c:3439)
==1239422==    by 0x82B1C28: bson_realloc (bson-memory.c:154)
==1239422==    by 0x82B1C97: bson_realloc_ctx (bson-memory.c:194)
==1239422==    by 0x829AF1F: _bson_impl_alloc_grow (bson.c:188)
==1239422==    by 0x829AF8A: _bson_grow (bson.c:222)
==1239422==    by 0x829B0CC: _bson_append_va (bson.c:326)
==1239422==    by 0x829B43B: _bson_append (bson.c:412)
==1239422==    by 0x829CCA5: bson_append_int32 (bson.c:1200)
==1239422==    by 0x829D6AB: bson_append_iter (bson.c:1399)
==1239422==    by 0x829FC28: bson_copy_to_excluding_noinit_va (bson.c:2263)
==1239422== 
==1239422== 2,144 bytes in 1 blocks are indirectly lost in loss record 31 of 33
==1239422==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==1239422==    by 0x784468: __zend_malloc (zend_alloc.c:3030)
==1239422==    by 0x7844D5: __zend_calloc (zend_alloc.c:3042)
==1239422==    by 0x8379BE6: php_phongo_calloc (php_phongo.c:3434)
==1239422==    by 0x82B1B92: bson_malloc0 (bson-memory.c:108)
==1239422==    by 0x83237D1: mongoc_server_description_new_copy (mongoc-server-description.c:795)
==1239422==    by 0x833BA73: mongoc_topology_description_get_servers (mongoc-topology-description.c:2281)
==1239422==    by 0x83A8581: php_phongo_topologydescription_get_properties_hash (TopologyDescription.c:172)
==1239422==    by 0x83A8705: php_phongo_topologydescription_get_debug_info (TopologyDescription.c:208)
==1239422==    by 0x8A0A6A: zend_std_get_properties_for (zend_object_handlers.c:1779)
==1239422==    by 0x8A0B57: zend_get_properties_for (zend_object_handlers.c:1808)
==1239422==    by 0x6CBC89: php_var_dump (var.c:154)
==1239422== 
==1239422== 3,248 (8 direct, 3,240 indirect) bytes in 1 blocks are definitely lost in loss record 33 of 33
==1239422==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==1239422==    by 0x784468: __zend_malloc (zend_alloc.c:3030)
==1239422==    by 0x7844D5: __zend_calloc (zend_alloc.c:3042)
==1239422==    by 0x8379BE6: php_phongo_calloc (php_phongo.c:3434)
==1239422==    by 0x82B1B92: bson_malloc0 (bson-memory.c:108)
==1239422==    by 0x833BA11: mongoc_topology_description_get_servers (mongoc-topology-description.c:2272)
==1239422==    by 0x83A8581: php_phongo_topologydescription_get_properties_hash (TopologyDescription.c:172)
==1239422==    by 0x83A8705: php_phongo_topologydescription_get_debug_info (TopologyDescription.c:208)
==1239422==    by 0x8A0A6A: zend_std_get_properties_for (zend_object_handlers.c:1779)
==1239422==    by 0x8A0B57: zend_get_properties_for (zend_object_handlers.c:1808)
==1239422==    by 0x6CBC89: php_var_dump (var.c:154)
==1239422==    by 0x6CC192: zif_var_dump (var.c:218)
==1239422== 
```

Note: Unlike the other leak I mentioned, PHP doesn't report _this_ leak itself since it exists entirely in libmongoc. Leaks reported by PHP's debug mode are mainly with zval usage and internal PHP memory.

src/MongoDB/Monitoring/TopologyChangedEvent.c
```diff
@@ -0,0 +1,177 @@
...
+	INIT_NS_CLASS_ENTRY(ce, "MongoDB\\Driver\\Monitoring", "TopologyChangedEvent", php_phongo_topologychangedevent_me);
+	php_phongo_topologychangedevent_ce                = zend_register_internal_class(&ce);
+	php_phongo_topologychangedevent_ce->create_object = php_phongo_topologychangedevent_create_object;
+	PHONGO_CE_FINAL(php_phongo_topologychangedevent_ce);
+	PHONGO_CE_DISABLE_SERIALIZATION(php_phongo_topologychangedevent_ce);
+
+	memcpy(&php_phongo_handler_topologychangedevent, phongo_get_std_object_handlers(), sizeof(zend_object_handlers));
+	php_phongo_handler_topologychangedevent.get_debug_info = php_phongo_topologychangedevent_get_debug_info;
+	php_phongo_handler_topologychangedevent.free_obj       = php_phongo_topologychangedevent_free_object;
+	php_phongo_handler_topologychangedevent.offset         = XtOffsetOf(php_phongo_commandstartedevent_t, std);
```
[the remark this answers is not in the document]
tanlisu: Ah, sorry for that mistake. I did install Valgrind earlier but we weren't able to get it working correctly - right now when I run any test with `TEST_PHP_ARGS=-m`, I get the following error in the `.mem` file:

> --95198-- UNKNOWN mach_msg unhandled MACH_SEND_TRAILER option
--95198-- UNKNOWN mach_msg unhandled MACH_SEND_TRAILER option (repeated 2 times)
--95198-- UNKNOWN mach_msg unhandled MACH_SEND_TRAILER option (repeated 4 times)
--95198-- UNKNOWN mach_msg unhandled MACH_SEND_TRAILER option (repeated 8 times)
--95198-- WARNING: unhandled amd64-darwin syscall: unix:228
--95198-- You may be able to write your own handler.
--95198-- Read the file README_MISSING_SYSCALL_OR_IOCTL.
--95198-- Nevertheless we consider this a bug.  Please report
--95198-- it at http://valgrind.org/support/bug_reports.html.

src/MongoDB/TopologyDescription.c
```diff
@@ -0,0 +1,248 @@
...
+	php_phongo_topologydescription_t* intern = Z_TOPOLOGYDESCRIPTION_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	RETVAL_STRING(mongoc_topology_description_type(intern->topology_description));
+} /* }}} */
+
+/* {{{ MongoDB\Driver\TopologyDescription function entries */
+/* clang-format off */
+ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(ai_TopologyDescription_hasReadableServer, 0, 0, _IS_BOOL, 0)
```
tanlisu, line 107: I saw that `ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_TYPE_INFO_EX` is used in https://github.com/mongodb/mongo-php-driver/pull/1245 but for some reason the tests wouldn't compile when I used that instead of `ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX`. What is the difference between the two / what does it mean for the return type to be tentative?

tests/topologyDescription/topologyDescription-debug-001.phpt
tanlisu: Sorry, I must have missed updating this test to be similar to the `var_export` one, but was thinking that all the tests would probably need to be improved so I would come back to it eventually.

src/MongoDB/TopologyDescription.c
jmikola: I believe `ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_TYPE_INFO_EX` is the PHP internal equivalent of the `ReturnTypeWillChange` annotation for userland code (see: https://github.com/mongodb/mongo-php-library/pull/855). Both of these were introduced by https://wiki.php.net/rfc/internal_method_return_types and are intended to serve as a bridge in PHP 8.x until PHP 9.0 arrives. In the case of #1245, we needed to use them on our classes that implement core PHP interfaces, since those interfaces now declare a tentative return type.

As far as it being "tentative", I think this has to do with not being able to formally change the return type of interface methods since that would constitute a BC break (see: [Semantic Versioning and Public Interfaces](http://paul-m-jones.com/post/2015/06/02/semantic-versioning-and-public-interfaces/) for an explanation). Instead, code in PHP 8.1+ will need to start declaring a tentative return type or risk a deprecation warning (which is enough to make our own tests fail, which is why we addressed this).

In contrast to all of that, `ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX` is declaring the formal return type. I don't believe you actually need to worry about using this here, as we'll eventually get to it when cleaning up the entire extension in [PHPC-1709](https://jira.mongodb.org/browse/PHPC-1709). However, there's certainly no harm in doing so now if you can get this working.

The more important thing to address with respect to arginfo is the requirements for the ReadPreference parameter, as that's not something we could alter later without introducing a BC break.

Tying this back to the earlier blog post I referenced about semantic versioning and interfaces, we can effectively narrow the return type (e.g. anything -> bool) without causing a BC break, but we _cannot_ narrow the type of an accepted parameter (e.g. anything -> ReadPreference) without causing a BC break.

src/MongoDB/TopologyDescription.c
```diff
@@ -0,0 +1,246 @@
...
+{
+	php_phongo_topologydescription_t* intern;
+	const mongoc_read_prefs_t*        read_preference   = NULL;
+	zval*                             z_read_preference = NULL;
+
+	intern = Z_TOPOLOGYDESCRIPTION_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_START(0, 1)
+	Z_PARAM_OPTIONAL
+	Z_PARAM_ZVAL_EX(z_read_preference, 1, 0)
```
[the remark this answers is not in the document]
tanlisu: Hmm, I was thinking that I should parse the read preference as a zval in order to then get `mongoc_read_prefs_t` from it using `phongo_read_preference_from_zval`. I just updated to use `Z_PARAM_OBJECT_OF_CLASS(z_read_preference, php_phongo_readpreference_ce)` - does this still work with my declaration of `z_read_preference` as a `zval*`?

src/MongoDB/TopologyDescription.c
tanlisu: Ok, that makes sense. I'm just using `ZEND_BEGIN_ARG_INFO_EX` then since I couldn't get `ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX` to work.

tests/topologyDescription/topologyDescription-debug-001.phpt
jmikola: Noted. In this case it would have been helpful to preemptively open a comment thread to communicate your intention (especially since a "todo" comment can't be left here). Allowing the comment to stay unresolved until it's addressed would serve double duty to inform reviewers and also serve as a personal reminder.

src/MongoDB/Monitoring/TopologyChangedEvent.c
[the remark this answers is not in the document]
jmikola: Noted. Searching for "UNKNOWN mach_msg unhandled MACH_SEND_TRAILER option" and "WARNING: unhandled amd64-darwin syscall" (omitting the suffix) turns up various discussion threads so these appear to be common issues on macOS. Unfortunately nothing I can provide any assistance with.

jmikola:
> I think it's caused by the most recent libmongoc submodule update since the test works when I revert the submodule and run it locally.

[PHPC-1538](https://jira.mongodb.org/browse/PHPC-1538) looks like the relevant issue. I discerned that by looking at [recent libmongoc commits](https://github.com/mongodb/mongo-c-driver/commits/master) and the commit message for https://github.com/mongodb/mongo-c-driver/commit/fdc2de4ff0e472a03c6c8a22dccaa1916ef3fd08 stood out. I then traced that through JIRA to the related PHPC issue.

> Note: it still doesn't seem to be working after updating to the most recent commit (mongodb/mongo-c-driver@6abd837)

Since I rebased my PR branch on master, https://github.com/mongodb/mongo-c-driver/commit/fdc2de4ff0e472a03c6c8a22dccaa1916ef3fd08 is included in its history. I'm going to rebase again so that commit is excluded to unblock you here, since that PR appears to need more work (https://github.com/mongodb/mongo-c-driver/pull/785#issuecomment-897123606).

tanlisu:
Got it, just updated again

src/MongoDB/TopologyDescription.c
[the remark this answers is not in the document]
jmikola: `Z_PARAM_OBJECT_OF_CLASS` should still work. The argument is still parsed into a zval, but the macro will additionally ensure it is an object that is an instance of a particular class entry.

You can also search [php/php-src](https://github.com/php/php-src) for other uses of that macro, particularly within the `ext/` directory where other extensions are contained.

src/MongoDB/TopologyDescription.c
```diff
@@ -0,0 +1,248 @@
...
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	RETVAL_STRING(mongoc_topology_description_type(intern->topology_description));
+} /* }}} */
+
+/* {{{ MongoDB\Driver\TopologyDescription function entries */
+/* clang-format off */
+ZEND_BEGIN_ARG_INFO_EX(ai_TopologyDescription_hasReadableServer, 0, 0, 0)
+	ZEND_ARG_OBJ_INFO(0, readPreference, MongoDB\\Driver\\ReadPreference, 1)
```
jmikola, line 108: I _think_ we should specify `0` for the final macro argument here (`allow_null`). I realize this contradicts what I said previously in https://github.com/mongodb/mongo-php-driver/pull/1244#issuecomment-896114319, so please test it locally to be sure. Try to run what you have now in a test that passes `null` for ReadPreference and see if it fails. If it does, that would make for a reasonable "error" test for the method (along with passing some other bad types, like an integer or an empty array). We have a few such tests in the driver to check for type errors on various methods, so you can base it on that.

Looking at what you have in the `hasReadableServer` implementation earlier in this file, I don't actually think `null` will be accepted as it'd fail the `Z_PARAM_OBJECT_OF_CLASS` check. `Z_PARAM_OPTIONAL` is independent of the parameter's _value_ -- it's just controlling whether we can omit the parameter entirely. In the old `zend_parse_paramaters` syntax, `|` and `!` modifiers were used for optional and nullable, respectively. Offhand, I don't know what the modern equivalent of `!` is (I don't believe it's covered in [Registering and using PHP functions](https://www.phpinternalsbook.com/php7/extensions_design/php_functions.html either)).

tanlisu: Just checked and what you said seems right, so I'll add the error test for `hasReadableServer`.

tests/topologyDescription/topologyDescription-debug-001.phpt
```diff
@@ -0,0 +1,46 @@
...
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+object(MongoDB\Driver\TopologyDescription)#%d (%d) {
+  ["servers"]=>
+  array(%d) {
+    %a
+  }
+  ["type"]=>
+  string(%d) "%rSingle|Sharded|ReplicaSetWithPrimary%r"
```
jmikola, line 44: This is technically correct, but I'd suggest wrapping the pattern in parens for readability (also done in other tests).
```suggestion
  string(%d) "%r(Single|Sharded|ReplicaSetWithPrimary)%r"
```

tests/topologyDescription/topologyDescription-getServers-001.phpt
```diff
@@ -0,0 +1,50 @@
...
+        $manager = create_test_manager();
+        $manager->addSubscriber($this);
+        $manager->executeCommand(DATABASE_NAME, new MongoDB\Driver\Command(['ping' => 1]));
+        $manager->removeSubscriber($this);
+
+        return $this->topologyDescription;
+    }
+}
+
+function assertIsServerDescriptionsArray(array $sds) {
```
jmikola, line 30: Suggested rename since this function technically isn't asserting anything. It's just returning whether the parameter is an array of SDs.

```suggestion
function isArrayOfServerDescriptions(array $sds) {
```

tests/topologyDescription/topologyDescription-getServers-001.phpt
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+    return true;
+}
+
+$subscriber = new TopologyDescriptionProvider;
+$topologyDescription = $subscriber->getTopologyDescription();
+$serverDescriptions = $topologyDescription->getServers();
+
+var_dump(assertIsServerDescriptionsArray($serverDescriptions));
```
jmikola, line 43:
```suggestion
var_dump(isArrayOfServerDescriptions($serverDescriptions));
```

jmikola: I think it also makes sense to assert that the array has at least one element in it (using [count](https://www.php.net/count)). Otherwise, it's possible that the array is empty and this still passes.

In practice, we wouldn't expect that to happen since the subscriber captures the final TD; however, the extra assertion will make the test more resilient if that were ever to change.

tests/topologyDescription/topologyDescription-getType-001.phpt
```diff
@@ -0,0 +1,39 @@
...
+
+$subscriber = new TopologyDescriptionProvider;
+$topologyDescription = $subscriber->getTopologyDescription();
+var_dump($topologyDescription->getType());
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+string(%d) "%rSingle|Sharded|ReplicaSetWithPrimary%r"
```
jmikola, line 38:
```suggestion
string(%d) "%r(Single|Sharded|ReplicaSetWithPrimary)%r"
```

jmikola (commented):
A few suggestions and just waiting on a new test: `topologyDescription-hasReadableServer_error-001.phpt`.

jmikola:
@tanlisu Also note that https://github.com/mongodb/mongo-c-driver/pull/785 (for [CDRIVER-3909](https://jira.mongodb.org/browse/CDRIVER-3909)) has been merged. You can revert the change to `.gitmodules` in this PR and update the submodule to https://github.com/mongodb/mongo-c-driver/commit/d2838038aa7d1808b872ea7bd39df33db5bba364 (current HEAD of master). Switching back to the master branch should also mean `make libmongoc-version-current` should detect the correct version as well.

src/MongoDB/TopologyDescription.c
```diff
@@ -0,0 +1,248 @@
...
+
+	for (i = 0; i < n; i++) {
+		zval obj;
+
+		phongo_serverdescription_init(&obj, sds[i]);
+		add_next_index_zval(return_value, &obj);
+	}
+} /* }}} */
+
+/* {{{ proto boolean MongoDB\Driver\TopologyDescription::hasReadableServer([?MongoDB\Driver\ReadPreference $readPreference])
```
jmikola, line 61: `?` can be removed here since `null` is not accepted.

tests/topologyDescription/topologyDescription-hasReadableServer_error-001.phpt
```diff
@@ -0,0 +1,55 @@
+--TEST--
+MongoDB\Driver\TopologyDescription::hasReadableServer() (argument with bad type)
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_live(); ?>
```
jmikola, line 5: This test will need to be split for PHP 7 and 8 since the error messages differ. See `session-startTransaction_error-004.phpt` and `session-startTransaction_error-005.phpt` for prior art. You can lift the same skip logic for PHP versions from those tests.

jmikola: Alternatively, just remove the `echo` and trailing newline from `throws()` output and we can make this portable for both PHP versions.

src/MongoDB/TopologyDescription.c
```diff
@@ -0,0 +1,248 @@
...
+		array_init_size(&servers, n);
+
+		for (i = 0; i < n; i++) {
+			zval obj;
+
+			if (!php_phongo_server_description_to_zval(&obj, sds[i])) {
+				/* Exception already thrown */
+				zval_ptr_dtor(&obj);
+				zval_ptr_dtor(&servers);
+				mongoc_server_descriptions_destroy_all(sds, n);
```
jmikola, line 185: This function leaks because the `sds` aren't freed on the successful code path.

src/MongoDB/TopologyDescription.c
```diff
@@ -0,0 +1,248 @@
...
+{
+	php_phongo_topologydescription_t* intern;
+	mongoc_server_description_t**     sds;
+	size_t                            i, n = 0;
+
+	intern = Z_TOPOLOGYDESCRIPTION_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	sds = mongoc_topology_description_get_servers(intern->topology_description, &n);
```
jmikola, line 49: This appears to leak, but not because of the server descriptions themselves. Those are used to initialize PHP objects and ultimately get freed in ServerDescription's `free_object` handler. Rather, the allocated array that stores the server description pointers is leaking here. If you look at `mongoc_server_descriptions_destroy_all` in libmongoc, you'll see that it frees both the pointers and finally the array.

Two possible solutions here:

 * Use `mongoc_server_description_new_copy` in the `for` loop below to send copies to `phongo_serverdescription_init`, which will be owned by the PHP object. And then call `mongoc_server_descriptions_destroy_all` after the loop.

 * Avoid calling `mongoc_server_descriptions_destroy_all` and making extra copies of the server descriptions and just `bson_free(sds)` after the `for` loop.

I'd lean towards the first even though it's a bit more inefficient, since knowing to free to array alone is kind of assuming libmongoc's internal logic and not abiding its public API.

jmikola (approved):

PR #1245, opened 2021-08-05
PHPC-1858 and PHPC-1933: Return type info for interface implementations in PHP 8.1 

jmikola:
https://jira.mongodb.org/browse/PHPC-1858
https://jira.mongodb.org/browse/PHPC-1933

### DIFF
--- a/phongo_compat.h
+++ b/phongo_compat.h
@@ -218,6 +218,15 @@
 	} while (0)
 #endif
 
+/* Per https://wiki.php.net/rfc/internal_method_return_types, "Non-final
+ * internal method return types - when possible - are declared tentatively in
+ * PHP 8.1, and they will become enforced in PHP 9.0." This can be revisited
+ * when more general typing improvements are made in PHPC-1709. */
+#ifndef ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_TYPE_INFO_EX
+#define ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_TYPE_INFO_EX(name, return_reference, required_num_args, type, allow_null) \
+	ZEND_BEGIN_ARG_INFO_EX(name, 0, return_reference, required_num_args)
+#endif
+
 void      phongo_add_exception_prop(const char* prop, int prop_len, zval* value);
 zend_bool php_phongo_zend_hash_apply_protection_begin(HashTable* ht);
 zend_bool php_phongo_zend_hash_apply_protection_end(HashTable* ht);
--- a/src/BSON/Binary.c
+++ b/src/BSON/Binary.c
@@ -271,6 +271,7 @@ static PHP_METHOD(Binary, unserialize)
 } /* }}} */
 
 /* {{{ MongoDB\BSON\Binary function entries */
+/* clang-format off */
 ZEND_BEGIN_ARG_INFO_EX(ai_Binary___construct, 0, 0, 2)
 	ZEND_ARG_INFO(0, data)
 	ZEND_ARG_INFO(0, type)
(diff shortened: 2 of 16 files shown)


phongo_compat.h
```diff
@@ -218,6 +218,15 @@
...
 #endif
 
+/* Per https://wiki.php.net/rfc/internal_method_return_types, "Non-final
+ * internal method return types - when possible - are declared tentatively in
+ * PHP 8.1, and they will become enforced in PHP 9.0." This can be revisited
+ * when more general typing improvements are made in PHPC-1709. */
+#ifndef ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_TYPE_INFO_EX
+#define ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_TYPE_INFO_EX(name, return_reference, required_num_args, type, allow_null) \
+	ZEND_BEGIN_ARG_INFO_EX(name, 0, return_reference, required_num_args)
+#endif
```
jmikola, line 228: This macro was introduced in https://github.com/php/php-src/commit/532c60cb921757151f4a60b9f56e2805ce23d819 (8.1.0alpha1). For earlier versions where this isn't required, we'll just fall back to standard arginfo.

Actual return type arginfo will be handled by [PHPC-1709](https://jira.mongodb.org/browse/PHPC-1709) down the line.

alcaeus: Ah, this is nice so we don't have to add conditionals all over the place. Thanks!

src/BSON/Binary.c
```diff
@@ -280,6 +280,9 @@ ZEND_BEGIN_ARG_INFO_EX(ai_Binary___set_state, 0, 0, 1)
 	ZEND_ARG_ARRAY_INFO(0, properties, 0)
 ZEND_END_ARG_INFO()
 
+ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_TYPE_INFO_EX(ai_Binary_jsonSerialize, 0, 0, IS_ARRAY, 0)
+ZEND_END_ARG_INFO()
+
```
alcaeus, line 285: Since these are reused, can we declare these `ai_*_jsonSerialize` variables somewhere else and reuse them? The signature of these methods is fixed by the interface, so technically we could have this definition in the same place. Happy to defer this to a follow-up PR if you'd like to get it fixed first.

src/MongoDB/Cursor.c
```diff
@@ -351,6 +351,21 @@ PHP_METHOD(Cursor, rewind)
...
+
+ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_TYPE_INFO_EX(ai_Cursor_next, 0, 0, IS_VOID, 0)
+ZEND_END_ARG_INFO()
+
+ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_TYPE_INFO_EX(ai_Cursor_valid, 0, 0, _IS_BOOL, 0)
+ZEND_END_ARG_INFO()
+
+ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_TYPE_INFO_EX(ai_Cursor_rewind, 0, 0, IS_VOID, 0)
+ZEND_END_ARG_INFO()
+
```
alcaeus, line 368: Same as with `JsonSerializable`, these could be defined in a central place, but for iterator methods I'm not sure they're needed in many other places.

alcaeus (approved):
LGTM!

src/BSON/Binary.c
jmikola: Deferring:

 * https://jira.mongodb.org/browse/PHPC-1938
 * https://jira.mongodb.org/browse/PHPC-1939

src/MongoDB/Cursor.c
jmikola: The Iterator and Countable methods are definitely not reused, so I think this is fine as-is. [PHPC-1938](https://jira.mongodb.org/browse/PHPC-1938) will definitely help to keep this more organized.

tanlisu (approved):
LGTM

PR #1246, opened 2021-08-10
PHPC-1849: Handle deprecation of Serializable in PHP 8.1

jmikola:
https://jira.mongodb.org/browse/PHPC-1849

### DIFF
--- a/php_phongo.h
+++ b/php_phongo.h
@@ -182,26 +182,26 @@ bool php_phongo_manager_unregister(php_phongo_manager_t* manager);
 	} while (0)
 #endif
 
-#define PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_debug, intern, props, size) \
-	do {                                                                   \
-		if (is_debug) {                                                    \
-			ALLOC_HASHTABLE(props);                                        \
-			zend_hash_init((props), (size), NULL, ZVAL_PTR_DTOR, 0);       \
-		} else if ((intern)->properties) {                                 \
-			(props) = (intern)->properties;                                \
-		} else {                                                           \
-			ALLOC_HASHTABLE(props);                                        \
-			zend_hash_init((props), (size), NULL, ZVAL_PTR_DTOR, 0);       \
-			(intern)->properties = (props);                                \
-		}                                                                  \
+#define PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_temp, intern, props, size) \
+	do {                                                                  \
+		if (is_temp) {                                                    \
+			ALLOC_HASHTABLE(props);                                       \
+			zend_hash_init((props), (size), NULL, ZVAL_PTR_DTOR, 0);      \
+		} else if ((intern)->properties) {                                \
+			(props) = (intern)->properties;                               \
+		} else {                                                          \
+			ALLOC_HASHTABLE(props);                                       \
+			zend_hash_init((props), (size), NULL, ZVAL_PTR_DTOR, 0);      \
+			(intern)->properties = (props);                               \
+		}                                                                 \
 	} while (0)
 
-#define PHONGO_GET_PROPERTY_HASH_FREE_PROPS(is_debug, props) \
-	do {                                                     \
-		if (is_debug) {                                      \
-			zend_hash_destroy((props));                      \
-			FREE_HASHTABLE(props);                           \
-		}                                                    \
+#define PHONGO_GET_PROPERTY_HASH_FREE_PROPS(is_temp, props) \
+	do {                                                    \
+		if (is_temp) {                                      \
+			zend_hash_destroy((props));                     \
+			FREE_HASHTABLE(props);                          \
+		}                                                   \
 	} while (0)
 
 #define PHONGO_ZVAL_CLASS_OR_TYPE_NAME(zv) (Z_TYPE(zv) == IS_OBJECT ? ZSTR_VAL(Z_OBJCE(zv)->name) : zend_get_type_by_const(Z_TYPE(zv)))
--- a/src/BSON/Binary.c
+++ b/src/BSON/Binary.c
@@ -69,6 +69,32 @@ static bool php_phongo_binary_init_from_hash(php_phongo_binary_t* intern, HashTa
 	return false;
 } /* }}} */
 
+static HashTable* php_phongo_binary_get_properties_hash(phongo_compat_object_handler_type* object, bool is_temp) /* {{{ */
+{
+	php_phongo_binary_t* intern;
+	HashTable*           props;
+
+	intern = Z_OBJ_BINARY(PHONGO_COMPAT_GET_OBJ(object));
+
+	PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_temp, intern, props, 2);
+
+	if (!intern->data) {
+		return props;
+	}
+
+	{
+		zval data, type;
+
+		ZVAL_STRINGL(&data, intern->data, intern->data_len);
+		zend_hash_str_update(props, "data", sizeof("data") - 1, &data);
+
+		ZVAL_LONG(&type, intern->type);
+		zend_hash_str_update(props, "type", sizeof("type") - 1, &type);
+	}
+
+	return props;
+} /* }}} */
+
 /* {{{ proto void MongoDB\BSON\Binary::__construct(string $data, int $type)
    Construct a new BSON binary type */
 static PHP_METHOD(Binary, __construct)
@@ -270,6 +296,28 @@ static PHP_METHOD(Binary, unserialize)
 	zval_ptr_dtor(&props);
 } /* }}} */
 
+/* {{{ proto array MongoDB\Driver\Binary::__serialize()
+*/
+static PHP_METHOD(Binary, __serialize)
+{
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	RETURN_ARR(php_phongo_binary_get_properties_hash(PHONGO_COMPAT_OBJ_P(getThis()), true));
+} /* }}} */
+
+/* {{{ proto array MongoDB\Driver\Binary::__unserialize()
+*/
+static PHP_METHOD(Binary, __unserialize)
+{
+	zval* data;
+
+	PHONGO_PARSE_PARAMETERS_START(1, 1)
+	Z_PARAM_ARRAY(data)
+	PHONGO_PARSE_PARAMETERS_END();
+
+	php_phongo_binary_init_from_hash(Z_BINARY_OBJ_P(getThis()), Z_ARRVAL_P(data));
+} /* }}} */
+
 /* {{{ MongoDB\BSON\Binary function entries */
 /* clang-format off */
 ZEND_BEGIN_ARG_INFO_EX(ai_Binary___construct, 0, 0, 2)
@@ -279,6 +327,10 @@ ZEND_END_ARG_INFO()
 
 ZEND_BEGIN_ARG_INFO_EX(ai_Binary___set_state, 0, 0, 1)
 	ZEND_ARG_ARRAY_INFO(0, properties, 0)
+ZEND_END_ARG_INFO()
+
+ZEND_BEGIN_ARG_INFO_EX(ai_Binary___unserialize, 0, 0, 1)
+	ZEND_ARG_ARRAY_INFO(0, data, 0)
 ZEND_END_ARG_INFO()
 
 ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_TYPE_INFO_EX(ai_Binary_jsonSerialize, 0, 0, IS_ARRAY, 0)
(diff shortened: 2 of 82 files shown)


php_phongo.h
```diff
@@ -182,26 +182,26 @@ bool php_phongo_manager_unregister(php_phongo_manager_t* manager);
...
-			ALLOC_HASHTABLE(props);                                        \
-			zend_hash_init((props), (size), NULL, ZVAL_PTR_DTOR, 0);       \
-		} else if ((intern)->properties) {                                 \
-			(props) = (intern)->properties;                                \
-		} else {                                                           \
-			ALLOC_HASHTABLE(props);                                        \
-			zend_hash_init((props), (size), NULL, ZVAL_PTR_DTOR, 0);       \
-			(intern)->properties = (props);                                \
-		}                                                                  \
+#define PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_temp, intern, props, size) \
```
jmikola, line 185: Since `__serialize` will also return a newly allocated HashTable, this is no longer just used by the `get_debug_info` handlers. `is_temp` seemed like a more accurate name.

src/BSON/Binary.c
```diff
@@ -270,6 +296,28 @@ static PHP_METHOD(Binary, unserialize)
...
+/* {{{ proto array MongoDB\Driver\Binary::__serialize()
+*/
+static PHP_METHOD(Binary, __serialize)
+{
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	RETURN_ARR(php_phongo_binary_get_properties_hash(PHONGO_COMPAT_OBJ_P(getThis()), true));
+} /* }}} */
+
+/* {{{ proto array MongoDB\Driver\Binary::__unserialize()
```
jmikola, line 308: These protos are incorrect. The methods should have a `void` return annotation and take an `array $data` parameter and . I'll revise in a subsequent commit.

src/BSON/Binary.c
```diff
@@ -377,32 +431,6 @@ static int php_phongo_binary_compare_objects(zval* o1, zval* o2) /* {{{ */
 	return zend_binary_strcmp(intern1->data, intern1->data_len, intern2->data, intern2->data_len);
 } /* }}} */
 
-static HashTable* php_phongo_binary_get_properties_hash(phongo_compat_object_handler_type* object, bool is_debug) /* {{{ */
```
jmikola, line 380: Various `get_properties_hash` functions needed to be moved earlier in the file because they're used by `__serialize` methods. The alternative was to just add forward declarations at the top of the file, but that will ultimately be sorted by [PHPC-1728](https://jira.mongodb.org/browse/PHPC-1728).

src/MongoDB/CursorId.c
```diff
@@ -58,14 +58,14 @@ static bool php_phongo_cursorid_init_from_hash(php_phongo_cursorid_t* intern, Ha
 	return false;
 } /* }}} */
 
-static HashTable* php_phongo_cursorid_get_properties_hash(phongo_compat_object_handler_type* object, bool is_debug) /* {{{ */
+static HashTable* php_phongo_cursorid_get_properties_hash(phongo_compat_object_handler_type* object, bool is_temp, bool is_serialize) /* {{{ */
```
jmikola, line 61: The existing implementation of `Serializable::serialize()` always used `ADD_ASSOC_INT64_AS_STRING`. That didn't seem necessary if the value falls within the range of a 32-bit signed integer. For example, `WriteConcern::serialize()` encodes `wtimeout` as either an integer or string based on its range.

However, `php_phongo_cursorid_init_from_hash` currently only expects a string, so I'll either need to change that to also accept an integer or remove the conditional encoding for `is_serialize` below.

src/MongoDB/CursorId.c
```diff
@@ -74,14 +74,22 @@ static HashTable* php_phongo_cursorid_get_properties_hash(phongo_compat_object_h
 	{
 		zval value;
 
-		if (is_debug) {
+		if (is_serialize) {
+			if (intern->id > INT32_MAX || intern->id < INT32_MIN) {
+				ZVAL_INT64_STRING(&value, intern->id);
+			} else {
+				ZVAL_LONG(&value, intern->id);
```
jmikola, line 81: Per the comment above, this will need to be removed unless `php_phongo_cursorid_init_from_hash` is improved to also accept an integer.

Since this issue wasn't caught by any new tests, I'll be sure to add a regression test before modifying either function.

src/MongoDB/WriteConcern.c
```diff
@@ -298,7 +298,7 @@ static PHP_METHOD(WriteConcern, isDefault)
 	RETURN_BOOL(mongoc_write_concern_is_default(intern->write_concern));
 } /* }}} */
 
-static HashTable* php_phongo_write_concern_get_properties_hash(phongo_compat_object_handler_type* object, bool is_debug, bool is_bson) /* {{{ */
+static HashTable* php_phongo_writeconcern_get_properties_hash(phongo_compat_object_handler_type* object, bool is_temp, bool is_bson, bool is_serialize) /* {{{ */
```
jmikola, line 301: Before introducing `is_serialize`, `WriteConcern::serialize()` would enter the `! is_bson` code path and conditionally encode `wtimeout` based on the system architecture. That actually seemed like a mistake, since it could lead to a 64-bit system producing serialized output that cannot be processed on a 32-bit system (e.g. if `wtimeout` was `INT32_MAX + 1`).

This adds a new code path to ignore the system architecture and instead encode according to the range of the value itself to maximum compatibility. The `init_from_hash` function already supports both formats, so this isn't the same issue that came up with CursorId earlier.

src/MongoDB/WriteConcern.c
```diff
@@ -344,6 +344,12 @@ static HashTable* php_phongo_write_concern_get_properties_hash(phongo_compat_obj
 
 			if (is_bson) {
 				ZVAL_INT64(&z_wtimeout, wtimeout);
+			} else if (is_serialize) {
+				if (wtimeout > INT32_MAX || wtimeout < INT32_MIN) {
+					ZVAL_INT64_STRING(&z_wtimeout, wtimeout);
+				} else {
+					ZVAL_LONG(&z_wtimeout, wtimeout);
+				}
```
jmikola, line 352: I'm not sure if it's already covered in the tests, but I'll double-check that we're testing this code path of encoding `wtimeout` according to its range. I think some of the existing tests might be expecting either a string or integer (with `%r`), which is no longer necessary since the serialization output will now be determined entirely by the value.

src/MongoDB/CursorId.c
jmikola: Decided to keep the old behavior of always serializing "id" as a string. Also added a regression test case to each of `cursorid-serialization-001.phpt` and `cursorid-serialization-002.phpt` for that behavior.

src/MongoDB/CursorId.c
jmikola: Opted to remove conditional encoding for `is_serialize` since that always maintains existing behavior before this PR. See my comment below.

src/MongoDB/WriteConcern.c
jmikola: Confirmed that this _is_ covered by the "64-bit wtimeout is always encoded as as string" test cases in `writeconcern-serialization-001.phpt` and `writeconcern-serialization-002.phpt`.

tests/bson/bson-utcdatetime-serialization_error-004.phpt
```diff
@@ -0,0 +1,23 @@
...
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/basic.inc';
+
+echo throws(function() {
+    unserialize('O:24:"MongoDB\BSON\UTCDateTime":1:{s:12:"milliseconds";s:9:"1234.5678";}');
+}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
+
+/* TODO: Add tests for out-of-range values once CDRIVER-1377 is resolved */
```
jmikola, line 15: This TODO was copied from `bson-utcdatetime-serialization_error-002.phpt`. I created [PHPC-1957](https://jira.mongodb.org/browse/PHPC-1957) to track this.

src/BSON/Binary.c
```diff
@@ -69,6 +69,32 @@ static bool php_phongo_binary_init_from_hash(php_phongo_binary_t* intern, HashTa
...
+
+	{
+		zval data, type;
+
+		ZVAL_STRINGL(&data, intern->data, intern->data_len);
+		zend_hash_str_update(props, "data", sizeof("data") - 1, &data);
+
+		ZVAL_LONG(&type, intern->type);
+		zend_hash_str_update(props, "type", sizeof("type") - 1, &type);
+	}
```
alcaeus, line 93: I've wondered about this before. What's the practical reason for nesting? Is it to only conditionally allocate the two `zval` structs on the stack and avoid doing so if there's no data to handle?

src/BSON/Binary.c
```diff
@@ -281,6 +329,10 @@ ZEND_BEGIN_ARG_INFO_EX(ai_Binary___set_state, 0, 0, 1)
 	ZEND_ARG_ARRAY_INFO(0, properties, 0)
 ZEND_END_ARG_INFO()
 
+ZEND_BEGIN_ARG_INFO_EX(ai_Binary___unserialize, 0, 0, 1)
+	ZEND_ARG_ARRAY_INFO(0, data, 0)
+ZEND_END_ARG_INFO()
```
alcaeus, line 334: I've been wondering about this for a while as well: shouldn't this be in PHP's realm to provide arginfo structs for its own interface? We're duplicating this in a lot of our classes, where using `ai_Serializable___unserialize` would work just fine everywhere else 🤔

// cc @nikic

src/BSON/DBPointer.c
```diff
@@ -68,6 +68,31 @@ static bool php_phongo_dbpointer_init_from_hash(php_phongo_dbpointer_t* intern,
...
+		return props;
+	}
+
+	{
+		zval ref, id;
+
+		ZVAL_STRING(&ref, intern->ref);
+		ZVAL_STRING(&id, intern->id);
+		zend_hash_str_update(props, "ref", sizeof("ref") - 1, &ref);
+		zend_hash_str_update(props, "id", sizeof("id") - 1, &id);
```
alcaeus, line 90: No need to change anything here, but getting to the second class in the code review I thought "I suppose this is now 20 times the same change", and I immediately catch the different code organisation between this and the Binary class up there. No need to change since I'd like to refactor this code in general: macros that call the combination of `ZVAL_*` and `zend_hash_str_update` would help since they could handle the scoping (if it's necessary, see earlier comment).

src/BSON/DBPointer.c
```diff
@@ -187,8 +212,34 @@ static PHP_METHOD(DBPointer, unserialize)
...
+static PHP_METHOD(DBPointer, __unserialize)
+{
+	zval* data;
+
+	PHONGO_PARSE_PARAMETERS_START(1, 1)
+	Z_PARAM_ARRAY(data)
+	PHONGO_PARSE_PARAMETERS_END();
+
+	php_phongo_dbpointer_init_from_hash(Z_DBPOINTER_OBJ_P(getThis()), Z_ARRVAL_P(data));
+} /* }}} */
```
alcaeus, line 235: So, this is another one where I feel PHP could offer a macro:
```c
/* {{{ proto void MongoDB\Driver\DBPointer::__unserialize(array $data)
*/
IMPL_SERIALIZABLE___UNSERIALIZE_START()
	php_phongo_dbpointer_init_from_hash(Z_DBPOINTER_OBJ_P(getThis()), Z_ARRVAL_P(data));
IMPL_SERIALIZABLE___UNSERIALIZE_END() /* }}} */
```

Since the other logic is always the same, this may be an improvement to help reducing the menial task of adding this to multiple classes...

src/MongoDB/WriteConcern.c
alcaeus: For what it's worth, looking at our platform stats, I think we'll stop considering 32-bit systems soon.

tests/bson/bson-binary-serialization_error-004.phpt
```diff
@@ -0,0 +1,33 @@
...
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_php_version('<', '7.4.0'); ?>
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/basic.inc';
+
+echo throws(function() {
+    unserialize('O:19:"MongoDB\BSON\Binary":1:{s:4:"data";s:6:"foobar";}');
+}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
```
alcaeus, line 13: I know many of these files don't do this yet, but we should be using the `::class` constant. No need to change this in these files as they're most likely copypasta from the originals, so I'll see if I can do a nice find/replace to take care of these.

alcaeus (approved):
Changes look good, just a couple of questions on the code to judge potential refactoring effort.

src/BSON/Binary.c
jmikola: These aren't even conditionally allocated. We're just using braces to create a scope and allow us to declare variables inline. Our C compiler options would otherwise flag this as a warning and require that we declare these at the top of the function scope.

src/BSON/Binary.c
jmikola: Interesting point, although I could see this being problematic if PHP happened to change its arginfo between versions. I think this may relate to why interfaces in 8.1 started using _tentative_ return types, to defer the actual changes to PHP 9.0.

Allowing extensions to define their own arginfo presumably lets us use a narrower API if we want, which would still conform with the interface in core.

nikic: Nowadays we auto-generate all arginfo (as well as class declarations in general) anyway, so we don't have any particular interest in making reuse simpler.

src/BSON/DBPointer.c
jmikola: Created [PHPC-1962](https://jira.mongodb.org/browse/PHPC-1962) to track this.

tests/bson/bson-binary-serialization_error-004.phpt
jmikola: It would likely require a `sprintf` since we'd also need to capture the string length in the preceding integer. I'll leave this to you to make a ticket if you'd like to track the work.

jmikola:
Noted one test failure on AppVeyor for 32-bit CursorId `var_dump()` output. Fixed in the most recent push and will wait for green CI before merging.

PR #1248, opened 2021-08-16
PHPC-1538: authSource in URI string no longer requires credentials

jmikola:
https://jira.mongodb.org/browse/PHPC-1538

Unfortunately, the restriction is still in place for authSource in the URI options array (pending PHPC-1950).


tanlisu (approved):
LGTM

PR #1249, opened 2021-08-17
PHPC-1925: Change ServerDescription::getType() to return a string

tanlisu:
https://jira.mongodb.org/browse/PHPC-1925

### DIFF
--- a/src/MongoDB/ServerDescription.c
+++ b/src/MongoDB/ServerDescription.c
@@ -28,6 +28,16 @@
 
 zend_class_entry* php_phongo_serverdescription_ce;
 
+#define PHONGO_SERVER_UNKNOWN "Unknown"
+#define PHONGO_SERVER_STANDALONE "Standalone"
+#define PHONGO_SERVER_MONGOS "Mongos"
+#define PHONGO_SERVER_POSSIBLE_PRIMARY "PossiblePrimary"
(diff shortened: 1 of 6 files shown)


src/MongoDB/ServerDescription.c
```diff
@@ -115,7 +125,7 @@ static PHP_METHOD(ServerDescription, getType)
 
 	PHONGO_PARSE_PARAMETERS_NONE();
 
-	RETVAL_LONG(php_phongo_server_description_type(intern->server_description));
+	RETVAL_STRING(mongoc_server_description_type(intern->server_description));
```
jmikola, line 128: This would be a good opportunity to revise `serverDescription-getType-001.phpt` similar to what you did for the `TopologyDescription::getType()` test.

Additionally, `php_phongo_serverdescription_get_properties_hash` can be changed to output a "type" as a string instead of calling `php_phongo_server_description_type`. Its test can also be revised to use `%r` like the corresponding TopologyDescription tests.

jmikola:
According to CI, `server-getServerDescription-001.phpt` still needs to be updated.

tanlisu:
Yes sorry, I saw but didn't have time to finish yesterday

jmikola:
No worries. I wasn't sure if you caught it since it was after the review was requested.

jmikola (approved):

PR #1251, opened 2021-08-18
PHPC-1893: Implement SDAM event classes

tanlisu:
https://jira.mongodb.org/browse/PHPC-1893


src/phongo_apm.c
```diff
@@ -262,6 +262,35 @@ static void phongo_apm_command_failed(const mongoc_apm_command_failed_t* event)
...
+	/* Return early if there are no APM subscribers to notify */
+	if (zend_hash_num_elements(subscribers) == 0) {
+		goto cleanup;
+	}
+
+	object_init_ex(&z_event, php_phongo_serveropeningevent_ce);
+	p_event = Z_SERVEROPENINGEVENT_OBJ_P(&z_event);
+
+	mongoc_apm_server_opening_get_topology_id(event, &p_event->topology_id);
+	p_event->host = mongoc_apm_server_opening_get_host(event);
```
jmikola, line 284: Note that the documentation for [`mongoc_apm_server_opening_get_host`](http://mongoc.org/libmongoc/1.17.5/mongoc_apm_server_opening_get_host.html) states that the return value is only valid for the lifetime of the `mongoc_apm_server_opening_t` event. Storing the returned pointer directly on the `php_phongo_serveropeningevent_t` struct is likely to lead to a segfault down the line when libmongoc eventually frees this value and PHPC retains a pointer to it.

php_phongo_structs.h
```diff
@@ -284,6 +284,12 @@ typedef struct {
 	zend_object std;
 } php_phongo_commandsucceededevent_t;
 
+typedef struct {
+	bson_oid_t                topology_id;
+	const mongoc_host_list_t* host;
```
jmikola, line 289: Related to my later comment in ServerOpeningEvent.c, [`mongoc_apm_server_opening_get_host`](http://mongoc.org/libmongoc/1.17.5/mongoc_apm_server_opening_get_host.html) instructs us to copy this if we need it to live beyond the lifetime of libmongoc's APM event object (which we do).

`mongoc_apm_server_opening_get_host` also tells us that the return value is not actually a list and it's just using this type to represent a single host (i.e. the `next` pointer should be `NULL`). That means we need only worry about copying a single struct. I'll also note that the [`mongoc_host_list_t`](http://mongoc.org/libmongoc/1.17.5/mongoc_host_list_t.html) is part of libmongoc's public API, unlike some of the other opaque types we've worked with.

I think we have two options for `phongo_apm_server_opening`:

 * Declare this as a single `mongoc_host_list_t` (not a pointer) and use [`memcpy`](https://man7.org/linux/man-pages/man3/memcpy.3.html) to copy the contents of the `mongoc_apm_server_opening_get_host`'s return value (dereferencing the pointer) into this struct.
 * Declare separate fields for the host string and port based on the same declarations in `mongoc_host_list_t`'s definition, and copy those values accordingly. Since the host string has a fixed size (`BSON_HOST_NAME_MAX + 1`), you won't even need to allocate/free it.

I think either approach is fine. The choice you make will also impact the functions in ServerOpeningEvent.c that need to access the host and port, so in that sense it may be easier to use the second style and avoid needing to work with a `mongoc_host_list_t` in that file.

src/MongoDB/Monitoring/ServerOpeningEvent.c
```diff
@@ -0,0 +1,141 @@
...
+
+/* {{{ proto string ServerOpeningEvent::getHost()
+   Returns this event's host */
+static PHP_METHOD(ServerOpeningEvent, getHost)
+{
+	php_phongo_serveropeningevent_t* intern = Z_SERVEROPENINGEVENT_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	RETVAL_STRING((intern->host)->host);
```
jmikola, line 37: Note that the host _string_ is not the only vital information in [`mongoc_host_list_t`](http://mongoc.org/libmongoc/1.17.5/mongoc_host_list_t.html). The port is equally important, which is why we also have a `Server::getPort()` method. I think you'll need to add `getPort()` on this class as well.

I realize `mongoc_host_list_t` also has a `family` field; however, we don't currently make use of that. The libmongoc docs also don't go into much detail about that field (reported in [CDRIVER-4133](https://jira.mongodb.org/browse/CDRIVER-4133)), but I believe it's used to indicate the socket type (e.g. IPv4, IPv6, Unix domain socket). Nothing to worry about in the context of this PR, though. I've also opened [PHPC-1955](https://jira.mongodb.org/browse/PHPC-1955) to track this.

src/MongoDB/Monitoring/ServerOpeningEvent.c
```diff
@@ -0,0 +1,141 @@
...
+{
+	php_phongo_serveropeningevent_t* intern;
+	zval                             retval = ZVAL_STATIC_INIT;
+	char                             topology_id[25];
+
+	intern   = Z_OBJ_SERVEROPENINGEVENT(PHONGO_COMPAT_GET_OBJ(object));
+	*is_temp = 1;
+	array_init_size(&retval, 2);
+
+	ADD_ASSOC_STRING(&retval, "host", (intern->host)->host);
```
jmikola, line 105: Note that this line will change depending on how you restructure `php_phongo_serveropeningevent_t`. At the very least, you'll definitely need to add a `port` field.

tests/apm/monitoring-serverOpening-001.phpt
```diff
@@ -0,0 +1,44 @@
...
+
+    public function serverOpening(MongoDB\Driver\Monitoring\ServerOpeningEvent $event)
+    {
+        if ($this->serverOpened) {
+            return;
+        }
+
+        $this->serverOpened = true;
+        echo "- getHost() returns a string: ", is_string($event->getHost()) ? 'yes' : 'no', "\n";
+        echo "- getTopologyId() returns an ObjectId: ", ($event->getTopologyId() instanceof MongoDB\BSON\ObjectId) ? 'yes' : 'no', "\n";
```
jmikola, line 24: A separate idea to make this test more robust would be to ensure that `ServerOpeningEvent::getTopologyId()` is the same as the topology ID reported for other events. If you'd rather not address it in this PR then please make a ticket in the epic to track that idea. I think it's reasonable to put it off for now since it could be done independently once all of the events are implemented. And when doing so, we'd probably also want to switch over to using `Manager::addSubscriber()` in all of these tests. It doesn't make much difference since this PHP process only constructs a single Manager, but it's technically correct to do so (and serves as a good example for anyone that might reference this test as example code).

tests/apm/monitoring-serverOpening-001.phpt
```diff
@@ -0,0 +1,44 @@
...
+    private $serverOpened = false;
+
+    public function serverOpening(MongoDB\Driver\Monitoring\ServerOpeningEvent $event)
+    {
+        if ($this->serverOpened) {
+            return;
+        }
+
+        $this->serverOpened = true;
+        echo "- getHost() returns a string: ", is_string($event->getHost()) ? 'yes' : 'no', "\n";
```
jmikola, line 23: Reminder to test `getPort()` as well once that's implemented.

php_phongo_structs.h
tanlisu: Got it, I'll do the second approach then

tests/apm/monitoring-serverOpening-001.phpt
tanlisu: Check topology ID: https://jira.mongodb.org/browse/PHPC-1958
Use Manager::addSubscriber: https://jira.mongodb.org/browse/PHPC-1959

tanlisu:
I just realized that the return types in the function header comments of `TopologyChangedEvent::getNewDescription()` and `TopologyChangedEvent::getPreviousDescription()` are wrong: 
https://github.com/mongodb/mongo-php-driver/blob/e3d105a5c3dc9901f3f266246a471c76edc5ad3f/src/MongoDB/Monitoring/TopologyChangedEvent.c#L29
https://github.com/mongodb/mongo-php-driver/blob/e3d105a5c3dc9901f3f266246a471c76edc5ad3f/src/MongoDB/Monitoring/TopologyChangedEvent.c#L43
Should I just fix that here or make a separate PR?

jmikola:
> Should I just fix that here or make a separate PR?

Separate PR makes sense. You can probably just edit it directly through GitHub: https://github.com/mongodb/mongo-php-driver/blob/feature/sdam-monitoring/src/MongoDB/Monitoring/TopologyChangedEvent.c

tests/apm/monitoring-serverClosed-001.phpt
```diff
@@ -0,0 +1,50 @@
...
+$m->executeCommand(DATABASE_NAME, $command);
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+- getHost() returns a string: yes
+- getPort() returns an integer: yes
+- getTopologyId() returns an ObjectId: yes
+===DONE===
```
tanlisu, line 50: The `monitoring-serverClosed-001.phpt` test fails on the GitHub CI (https://github.com/mongodb/mongo-php-driver/pull/1251/checks?check_run_id=3365180493) because I think no `ServerClosedEvent` is emitted from the 'ping' command - how should I make sure that we get a server closed event?

jmikola: Did this test ever pass locally for you, or was CI the first time you observed it in action?

According to the [SDAM Monitoring spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#events), ServerClosedEvent is:

> Published when the server monitor’s connection is closed and the server is shutdown.

You may need to use the `disableClientPersistence` driver option for Manager (third argument) to ensure the connection get closed when the manager object goes out of scope. Without that, the libmongoc client will outlive the Manager object.

While you're doing this, I'd also suggest using `Manager::addSubscriber()` for this test (ahead of [PHPC-1959](https://jira.mongodb.org/browse/PHPC-1959)). Offhand, I'm not sure if adding a subscriber on a specific Manager will add a reference to it, which might prevent it from being garbage collected (and thus the libmongoc client being destroyed due to `disableClientPersistence`). I don't believe that's the case, as `phongo_apm_copy_manager_for_client` may be the only place that is done (and only for the lifetime of the command monitoring events being dispatched). Using `Manager::addSubscriber()` would allow us to confirm that's indeed the case.

tanlisu: Ah ok, the test did pass locally for me but not on the CI.

src/phongo_apm.c
```diff
@@ -262,6 +262,108 @@ static void phongo_apm_command_failed(const mongoc_apm_command_failed_t* event)
...
+	if (zend_hash_num_elements(subscribers) == 0) {
+		goto cleanup;
+	}
+
+	object_init_ex(&z_event, php_phongo_serverclosedevent_ce);
+	p_event = Z_SERVERCLOSEDEVENT_OBJ_P(&z_event);
+
+	host_list = mongoc_apm_server_closed_get_host(event);
+	memset(p_event->host, 0, sizeof(p_event->host));
+	strncpy(p_event->host, host_list->host, sizeof(p_event->host) - 1);
```
tanlisu, line 321: I'm getting a build error on the GitHub CI using `strncpy` (https://github.com/mongodb/mongo-php-driver/pull/1251/checks?check_run_id=3373304261):

> In function ‘strncpy’,
    inlined from ‘phongo_apm_server_closed’ at /home/runner/work/mongo-php-driver/mongo-php-driver/src/phongo_apm.c:321:2:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:106:10: error: ‘__builtin_strncpy’ output may be truncated copying 255 bytes from a string of length 255 [-Werror=stringop-truncation]

jmikola: I believe https://stackoverflow.com/a/50198710 discusses the exact situation. I suggest reading through that question and the selected answer to understand what's going on here.

That said, we can also use [`bson_strncpy`](http://mongoc.org/libbson/current/bson_strncpy.html) here instead. That appears to be what libmongoc uses internally to copy the same field (see `_mongoc_host_list_push`). And `bson_strncpy` has the added benefit of always assigning a null byte to the end of the buffer (something `strncpy` does not do on its own).

tests/apm/monitoring-serverClosed-001.phpt
tanlisu: Hmm I added the `disableClientPersistence` option but the test seems to fail for the same reason as before (https://github.com/mongodb/mongo-php-driver/pull/1251/checks?check_run_id=3374366170).

jmikola:
I can confirm that `monitoring-serverClosed-001.phpt` passes locally with a replica set. I noted that CI is using a standalone 4.4 server. When I run the test with that, it does fail because a ServerClosedEvent is never emitted. I wonder if that may be the result of libmongoc emitting a TopologyClosedEvent (not yet implemented in PHPC) when the entire client shuts down rather than a ServerClosedEvent. I'd suggest looking through libmongoc to see if you can trace how these events are dispatched. Alternatively, ask in the libmongoc channel on Slack to see if @kevinAlbs has some quick insight. If nothing comes of that, I'll take a look later tonight when I get a chance. If the answer ends up being that standalone clusters only issue a TopologyClosedEvent instead of ServerClosedEvent, then perhaps we'll need to restrict this test to a replica set (I assume sharded clusters might behave similarly to standalones in that libmongoc may open mongos connections and just keep them open for the lifetime of the client).

On a separate note, I observed something strange when running the test on a replica set. While trying to diagnose a possible failure, I modified the event subscriber to dump all events:

```php
class MySubscriber implements MongoDB\Driver\Monitoring\SDAMSubscriber
{
    public function serverChanged(MongoDB\Driver\Monitoring\ServerChangedEvent $event) { var_dump($event); }

    public function serverClosed(MongoDB\Driver\Monitoring\ServerClosedEvent $event) { var_dump($event); }

    public function serverOpening(MongoDB\Driver\Monitoring\ServerOpeningEvent $event) { var_dump($event); }
    
    public function topologyChanged(MongoDB\Driver\Monitoring\TopologyChangedEvent $event) { var_dump($event); }

    public function topologyOpening(MongoDB\Driver\Monitoring\TopologyOpeningEvent $event) { var_dump($event); }
}
```

I had a 4.4 replica set running locally and configured `MONGODB_URI` as `mongodb://127.0.0.1:27017/?replicaSet=rs0`. Running the test and dumping all events resulted in some very strange output for the final TopologyChangedEvent event:

```
object(MongoDB\Driver\Monitoring\TopologyChangedEvent)#4 (3) {
  ["topologyId"]=>
  string(24) "611e9dcded56d969d04f6591"
  ["newTopologyDescription"]=>
  array(2) {
    ["servers"]=>
    array(2) {
      ["servers"]=>
      array(0) {
      }
      ["type"]=>
      string(19) "ReplicaSetNoPrimary"
    }
    ["type"]=>
    string(21) "ReplicaSetWithPrimary"
  }
  ["oldTopologyDescription"]=>
  array(2) {
    ["servers"]=>
    array(0) {
    }
    ["type"]=>
    string(19) "ReplicaSetNoPrimary"
  }
}
```

The output for `servers` definitely looks wrong, so please see if you can sort that out.

tanlisu:
> The output for `servers` definitely looks wrong, so please see if you can sort that out.

Should be fixed now - I had some extra lines in`php_phongo_topology_description_to_zval` that shouldn't have been there.

tanlisu:
Notes: Currently the `ServerClosedEvent` and `TopologyClosedEvent` tests fail because those two events aren't being emitted. I also added the `ServerHeartbeatFailedEvent` class, but it's missing the `ServerHeartbeatFailedEvent::getError()` function because I wasn't sure how to deal with the `bson_error_t`. I also haven't looked into how to produce a `ServerHeartbeatFailedEvent` for testing.
The only SDAM events left are `ServerHeartbeatStartedEvent` and `ServerHeartbeatSucceededEvent`.

jmikola:
> I also added the ServerHeartbeatFailedEvent class, but it's missing the ServerHeartbeatFailedEvent::getError() function because I wasn't sure how to deal with the bson_error_t.

`phongo_apm_command_failed` may be a useful reference, since it also has code that creates a PHP exception (stored as `php_phongo_commandfailedevent_t.z_error`) from a `bson_error_t`.

> I also haven't looked into how to produce a ServerHeartbeatFailedEvent for testing.

Per the spec, this is emitted when a `hello` command (or `isMaster` on pre-5.0 servers) fails. You should be able to use the `configureFailPoint` command (also used within PHPC's test suite via the `configureFailPoint()` utility function) to coerce such a failure. [Server Fail Points](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#server-fail-points) in the Unified Test Format spec may be helpful reading there. Additionally, you can grep various spec tests in that repository for a `failPoint` operation to see how other tests configure fail points for different purposes. There are definitely examples of setting fail points on `hello`/`isMaster`.

Since setting the fail point requires connecting to the server, there are at least two ways to go about this:

 * Create a two Managers with `disableClientPersistence` to ensure they use separate libmongoc clients. Use one to configure the fail point and the other to issue a ping command with SDAM monitoring enabled. You may get a heartbeat event when the `hello`/`isMaster` fails during discovery.
 * Create one Manager and configure the fail point. Wait a bit longer than [`heartbeatFrequencyMS`](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#heartbeatfrequencyms), which is configurable as low as 500ms, before issuing a ping command.

jmikola:
Superseded by https://github.com/mongodb/mongo-php-driver/pull/1287.

PR #1252, opened 2021-08-18
Fix message for catching wrong exception in raises()

jmikola:
The original implementation of raises() in c6a793d466ef1dda46ce33bbd5e2d77e914249bb apparently referenced an undeclared $exceptionname variable several times. That was later addressed 804f383aea602c2bf3698169d893fc3cd2f10782 but that commit neglected to fix this reference.

This change avoids a possible message where the actual exception is printed twice:

```
ALMOST: Got Exception - expected Exception
```

Given the [error handler](https://www.php.net/set_error_handler) implementation here, we should always except an [ErrorException](https://www.php.net/manual/en/class.errorexception.php).

### DIFF
--- a/tests/utils/tools.php
+++ b/tests/utils/tools.php
@@ -607,7 +607,7 @@ function raises($function, $type, $infunction = null) {
             }
             printf("OK: Got %s\n", severityToString($e->getSeverity()));
         } else {
-            printf("ALMOST: Got %s - expected %s\n", get_class($e), $exceptionname);
+            printf("ALMOST: Got %s - expected %s\n", get_class($e), ErrorException::class);
         }
         restore_error_handler();
         return $e->getMessage();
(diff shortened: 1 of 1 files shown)


tests/utils/tools.php
```diff
@@ -607,7 +607,7 @@ function raises($function, $type, $infunction = null) {
             }
             printf("OK: Got %s\n", severityToString($e->getSeverity()));
         } else {
-            printf("ALMOST: Got %s - expected %s\n", get_class($e), $exceptionname);
+            printf("ALMOST: Got %s - expected %s\n", get_class($e), ErrorException::class);
```
jmikola, line 610: Actually, this is not a complete fix. This `else` condition is also reached if the severity (e.g. notice, warning) is incorrect. I'll revise that in a bit.

alcaeus: Yes. In this case, I think we can change the `catch` in L593 to only catch `ErrorException`, so we only have to handle the case where we've caught messages with a different severity.

We should also wrap the `restore_error_handler` call in a `finally` block for the `try` to properly restore the error handler regardless of outcome.

jmikola: Added `finally` with `restore_error_handler`. I also improved output for converting severity to a string and refactored both `raises()` and `throws()` to reduce conditional nesting.

alcaeus (approved):

PR #1256, opened 2021-08-24
PHPC-1918: Update Javascript and Symbol BSON corpus tests

jmikola:
https://jira.mongodb.org/browse/PHPC-1918

POC for mongodb/specifications#1056


scripts/convert-bson-corpus-tests.php
```diff
@@ -6,6 +6,7 @@
     'Int64 type: -1' => 'PHP encodes integers as 32-bit if range allows',
     'Int64 type: 0' => 'PHP encodes integers as 32-bit if range allows',
     'Int64 type: 1' => 'PHP encodes integers as 32-bit if range allows',
+    'Javascript Code: Embedded nulls' => 'Embedded null in code string is not supported in libbson (CDRIVER-1879)',
```
jmikola, line 9: This test originally tested the wrong type (string). Now that it correctly uses code w/o scope, we need to add an expected failure just as we had for the code w/ scope variant below (for which the test always used the correct type).

PR #1261, opened 2021-09-21
PHPC-1975: Disable SKIPIF caching for skip_if_not_clean on PHP 8.1

jmikola:
https://jira.mongodb.org/browse/PHPC-1975

### DIFF
--- a/tests/utils/skipif.php
+++ b/tests/utils/skipif.php
@@ -11,6 +11,18 @@ require_once __DIR__ . '/basic.inc';
 require_once __DIR__ . '/tools.php';
 
 /**
+ * Disables SKIPIF caching (PHP 8.1+).
+ */
+function disable_skipif_caching()
+{
+    if (PHP_VERSION_ID < 80100) {
+        return;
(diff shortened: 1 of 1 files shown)


tests/utils/skipif.php
```diff
@@ -10,6 +10,18 @@
 require_once __DIR__ . '/basic.inc';
 require_once __DIR__ . '/tools.php';
 
+/**
+ * Disables SKIPIF caching (PHP 8.1+).
+ */
+function skip_if_nocache()
```
alcaeus, line 16: Nit: this suggests that this method skips tests if there's no cache. disable_skipif_caching seems more appropriate, but I defer to your judgment.

alcaeus (approved):
LGTM - wording can stay if you don't like the alternative.

tests/utils/skipif.php
jmikola: That thought did occur to me when writing this. I agree that `disable_skipif_caching` is a much better name.

PR #1272, opened 2021-11-05
PHPC-1997: Migrate to new libbson atomic API

jmikola:
https://jira.mongodb.org/browse/PHPC-1997


php_phongo.c
```diff
@@ -3789,7 +3789,7 @@ PHP_GSHUTDOWN_FUNCTION(mongodb)
 	 * is the last thread, MSHUTDOWN has been called, persistent clients from
 	 * all threads have been destroyed, and it is now safe to shutdown libmongoc
 	 * and restore libbson's original vtable. */
-	if (bson_atomic_int_add(&phongo_num_threads, -1) == 0) {
+	if (bson_atomic_int32_fetch_sub(&phongo_num_threads, 1, bson_memory_order_seq_cst) - 1 == 0) {
```
jmikola, line 3792: @vector-of-bool: Would you mind sanity-checking these two lines? In the one above, I'm just doing an increment and don't need the return value. Here's we're decrementing and checking the return value, so I believe I need to also subtract from the result since the `__atomic_fetch_OP` built-ins return the original value (in contrast to `__atomic_OP_fetch` variants).

vector-of-bool: This change is correct: The value returned will be the value from _before_ the operation completes.

vector-of-bool (approved):
LGTM

PR #1274, opened 2021-11-10
PHPC-2004 and PHPC-2007: Specify __toString() return type and implement Stringable

jmikola:
https://jira.mongodb.org/browse/PHPC-2004
https://jira.mongodb.org/browse/PHPC-2007

See https://externals.io/message/116308#116318 for context. There is a subtle BC break here, as userland classes implementing our BSON interfaces will now need to explicitly specify a string return type on PHP versions before 8.1 (where it's added automatically). AFAICT, the only way to avoid the BC break entirely would be to continue to omit return type info and let PHP 8.1+ add the return type info and Stringable implementation automatically; however, that's not a permanent solution as PHP 8.2 will start raising warnings for the automatic behavior (https://github.com/php/php-src/commit/86379b6710f972e0d4a11c89ce28d5768d9824d3).

For that reason, I'm inclined to explicitly add return type info and Stringable implementations (for PHP 8+) and force users to confront the BC break now. This is also unlikely to affect many users, since the BSON interfaces are not widely used ([PHPC-640](https://jira.mongodb.org/browse/PHPC-640) is one such example).

### DIFF
--- a/tests/session/session-advanceOperationTime-003.phpt
+++ b/tests/session/session-advanceOperationTime-003.phpt
@@ -21,7 +21,7 @@ class MyTimestamp implements MongoDB\BSON\TimestampInterface
         return 1234;
     }
 
-    public function __toString()
+    public function __toString(): string
     {
         return sprintf('[%d:%d]', $this->getIncrement(), $this->getTimestamp());
     }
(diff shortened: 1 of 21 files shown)


jmikola:
Note: Evergreen failures are due to unexpected SSL cert verification ([PHPC-1981](https://jira.mongodb.org/browse/PHPC-1981))

tests/session/session-advanceOperationTime-003.phpt
```diff
@@ -21,7 +21,7 @@ class MyTimestamp implements MongoDB\BSON\TimestampInterface
         return 1234;
     }
 
-    public function __toString()
+    public function __toString(): string
```
jmikola, line 24: This is an example of the BC break where users will need to add a type hint to any implementations of our BSON interfaces.

PR #1283, opened 2021-12-08
PHPC-2004: Use tentative return types for BSON interface toString methods

jmikola:
https://jira.mongodb.org/browse/PHPC-2004

This should avoid the BC break introduced in 6c82ec8b52fad33c3e9444ae8bd2c27a729c4688. PHP 8.1 will automatically add return type info for Stringable, which means the ReturnTypeWillChange attribute should not be necessary on userland classes.

### DIFF
--- a/src/BSON/BinaryInterface.c
+++ b/src/BSON/BinaryInterface.c
@@ -27,7 +27,7 @@ zend_class_entry* php_phongo_binary_interface_ce;
 
 /* {{{ MongoDB\BSON\BinaryInterface function entries */
 /* clang-format off */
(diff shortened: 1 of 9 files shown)


jmikola:
I also tested this PR with PHP's master branch (8.2-dev) and there seems to be no conflict between tentative return type info for interface `__toString` methods and the Stringable behavior previously discussed on [internals](https://externals.io/message/116308).

Note: Compiling 8.2-dev did require a small addition to `config.m4` to work around a `memrchr` issue: https://github.com/php/php-src/commit/067df263448ee26013cddee1065bc9c1f028bd23#commitcomment-61570954

PR #1292, opened 2022-01-13
PHPC-2023 and PHPC-2030: Allow observation of events dispatched during client destruction

jmikola:
https://jira.mongodb.org/browse/PHPC-2023
https://jira.mongodb.org/browse/PHPC-2030

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -3352,6 +3352,9 @@ void php_phongo_client_reset_once(php_phongo_manager_t* manager, int pid)
 	{
 		if (pclient->client == manager->client) {
 			phongo_pclient_reset_once(pclient, pid);
+
+			/* Request-scoped clients are only used by a single Manager, so we
+			 * can return early after finding a match. */
 			return;
 		}
 	}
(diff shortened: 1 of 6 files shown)


php_phongo.c
```diff
@@ -3369,16 +3372,12 @@ static void php_phongo_pclient_destroy(php_phongo_pclient_t* pclient)
...
-		 * redesigned to not reference a client (see: PHPC-1666).
-		 *
-		 * Note: this is only relevant for request-scoped clients. APM
-		 * subscribers will no longer exist when persistent clients are
-		 * destroyed in GSHUTDOWN. */
-		mongoc_client_set_apm_callbacks(pclient->client, NULL, NULL);
+		/* If we are in request shutdown, disable APM to avoid dispatching more
+		 * events. This means that certain events (e.g. TopologyClosedEvent,
+		 * command monitoring for endSessions) may not be observed. */
+		if (EG(flags) & EG_FLAGS_IN_SHUTDOWN) {
```
jmikola, line 3378: @sgolemon: If I may pick your brain for a moment: Is this something we can reasonably rely on going forward? It's hard to tell what's public and private in the Zend headers, but this seemed pretty accessible.

sgolemon: Per slack. This seems to be the best you'll get for now. :(

Will followup with php-internals about this though.

PR #1317, opened 2022-05-05
PHPC-2083: Allow enums to be instantiated during BSON decoding

jmikola:
https://jira.mongodb.org/browse/PHPC-2083


src/phongo_bson.c
```diff
@@ -1037,6 +1087,62 @@ bool php_phongo_bson_to_zval_ex(const unsigned char* data, int data_len, php_pho
...
+
+				if (!value) {
+					char* path = php_phongo_field_path_as_string(state->field_path);
+					phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "Missing enum value for field path '%s'", path);
+					efree(path);
+
+					goto cleanup;
+				}
+
+				retval = zend_call_method_with_1_params(NULL, state->odm, NULL, "from", NULL, value);
```
jmikola, line 1104: @sgolemon: while the approach below seems like it'd be portable for both pure and backed enums, I wanted to see if invoking [`BackedEnum::from()`](https://www.php.net/manual/en/language.enumerations.backed.php) would work to reduce internal code.

Running this code path in the corresponding test, I see the following error:

```
Fatal error: MyEnum::from(): Return value must be of type static, MyEnum returned in Unknown on line 0
```

I traced this error back to either `zend_verify_return_error` or `zend_verify_internal_return_error` in [`zend_execute.c`](https://github.com/php/php-src/blob/php-8.1.5/Zend/zend_execute.c#L1306), but it's not clear to me why this would trigger.

I'm passing the enum's CE in `zend_call_method_with_1_params`. It wasn't possible to pass a `zend_object*` as the first parameter since enums cannot be instantiated.

I find it odd that I wouldn't be able to invoke these methods from an internal context (presumably I should be able to do everything one can do in userland), so I wonder if I may just be hitting an edge case that may be a bug in the original enum implementation. Any ideas?

sgolemon: The good news is: What you're trying to do SHOULD work.  The bad news is that there seems to be a bug in the engine where we make an assumption during return type checking that we have a relevant call frame to look at when verifying the called scope.  Turns out for Enums, which have an internal implementation of the from/tryFrom methods, we don't actually have a call frame, so the scope is essentially meaningless.

I've got a hack you can use, which violates the call stack by replacing whatever it's current notion of "thisness" with the actual calling scope you want enforced, and the rest "just works".  I wouldn't recommend this in a general use case, but for calling Enum::from() it should be safe enough.

```
ZEND_FUNCTION(smg)
{
    zend_string *classname = zend_string_init("ZendTestStringEnum", strlen("ZendTestStringEnum"), 0);
    zend_class_entry *ce = zend_lookup_class(classname);
    ZEND_ASSERT(ce);

    zval value;
    ZVAL_STRING(&value, "Test1");

    /* Preserve current scope and replace with target scope. */
    zval oldThis = EG(current_execute_data)->This;
    ZVAL_CE(&(EG(current_execute_data)->This), ce); 

    zend_call_method_with_1_params(NULL, ce, NULL, "from", return_value, &value);

    /* Restore current scope */
    EG(current_execute_data)->This = oldThis;

    zend_string_release(classname);
    zval_ptr_dtor(&value);
}
```

The two commented sections being what you wanna pay attention to.

tests/bson/bson-enum-002.phpt
```diff
@@ -0,0 +1,104 @@
...
+
+    $bson = fromPHP($document);
+    echo "Test ", toJSON($bson), "\n";
+    hex_dump($bson);
+    var_dump(toPHP($bson, $typeMap));
+    echo "\n";
+}
+
+?>
+===DONE===
```
jmikola, line 68: When executed with `TEST_PHP_ARGS=-m`, both `bson-enum-002.phpt` and `bson-enum-003.phpt` fail the following assertion:

```
php: /tmp/build_php-8.1.8.FWj/php-8.1.8/Zend/zend_objects_API.c:195: zend_objects_store_del: Assertion `(!(((zend_uintptr_t)((executor_globals.objects_store).object_buckets[handle])) & (1<<0)))' failed.
Aborted

Termsig=6
```

The corresponding `.mem` file reports many leaks (none include `phongo_bson.c` in the stack trace), but I believe those are just because of the abort failure immediately terminating the process without allowing PHP to clean up internally allocated memory.

The assertion failure corresponds to [this line in LXR](https://heap.space/xref/PHP-8.1/Zend/zend_objects_API.c?r=fdc22744#195).

tests/bson/bson-enum-003.phpt
```diff
@@ -0,0 +1,108 @@
...
+
+    public function bsonSerialize(): array
+    {
+        return (array) $this;
+    }
+ 
+    public function bsonUnserialize(array $data)
+    {
+        printf("%s called with: %s\n", __METHOD__, json_encode($data));
+    }
```
alcaeus, line 45: We might want to provide a `PersistableEnum` trait to reduce copy-pasta and make this less error-prone for users:
```php
enum MyBackedEnum: string implements Persistable
{
    use PersistableEnum;

    case foo = 'bar';
}
```

tests/bson/bson-enum-002.phpt
```diff
@@ -0,0 +1,104 @@
...
+BSON encoding: Enums implementing Unserializable
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_php_version('<', '8.1.0'); ?>
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/basic.inc';
+
+enum MyEnum implements MongoDB\BSON\Unserializable
```
alcaeus, line 11: Can you add a test that shows what happens when we provide a typeMap specifying an enum that doesn't implement `Unserializable`? IMO we shouldn't require the interface for enums as the method doesn't need to contain any logic.

alcaeus (commented):
Logic looks good, I do have a couple of questions around making this easier to use. Requiring `Unserializable` isn't really necessary IMO, and a trait to ease implementing enums that implement `Persistable` would reduce the number of foot-guns we provide to users.

src/phongo_bson.c
```diff
@@ -768,6 +768,55 @@ static void php_phongo_handle_field_path_entry_for_compound_type(php_phongo_bson
...
 
+#if PHP_VERSION_ID >= 80100
+/* Resolves an enum class and case name to a zval. On error, an exception will
+ * have been thrown and NULL will be returned.
+ *
+ * This function is modeled after php_var_unserialize_internal in php-src. */
+static zval* resolve_enum_case(zend_class_entry* ce, const char* case_name)
+{
+	zval*                return_value = NULL;
+	zend_string*         c_str;
```
sgolemon, line 779: `c_str` will be uninitialized when `!(ce->ce_flags & ZEND_ACC_ENUM)`

```suggestion
	zend_string*         c_str = NULL;
```

Edit: Looking below, I realize that can't actually happen since you only call this method from with a `ce->ce_flags & ZEND_ACC_ENUM` block, but if you're going to belt and suspender it here, then you might as well do it right. :)

src/phongo_bson.c
```diff
@@ -1035,9 +1126,44 @@ bool php_phongo_bson_to_zval_ex(const unsigned char* data, int data_len, php_pho
...
+				if (pfree) {
+					efree(case_name);
+				}
+
+				if (!enum_case) {
+					/* Exception already thrown */
+					goto cleanup;
+				}
+
+				ZVAL_COPY_VALUE(&obj, enum_case);
```
sgolemon, line 1159: I'm wondering what you expect to happen in the `bsonUnserialize` call for a typical unserializable enum case.  Since these are essentially singleton instances any side effects will be essentially global.

tests/bson/bson-enum-003.phpt
jmikola: I see a few options, which may not be mutually exclusive:

 * Allow an enum to be specified in a `typeMap` without requiring it to implement `Unserializable` (since we know the method is irrelevant). In this case, I suppose we'd still invoke `bsonUnserialize()` if it _is_ available on the enum, just to preserve expected behavior for BSON decoding. Since enforcement of `Unserializable` happens elsewhere, this will require some additional code changes.
 * Create an UnserializableEnum trait
 * Create a PersistableEnum trait

Presumably, the traits would exist in PHPC.

tests/bson/bson-enum-002.phpt
jmikola: I can add that, but the exception is thrown in unrelated code (and I believe it was already tested -- for non enums, of course).

src/phongo_bson.c
jmikola: I mentioned this in one of the test file comments, but there's really no practical purpose for `bsonUnserialize`. Since enums have no state, there's nothing that would be initialized. We're just invoking it because of the existing contract. The example implementation I used in the test file was to log a message that `bsonUnserialize` was called.

I think it sucks that we need to require enums to implement an interface, but I'd be more concerned about implicitly encoding their class name (i.e. pretending they're Persistable w/o the explicit implementation). The trait that @alcaeus suggested is probably the best compromise.

tests/bson/bson-enum-002.phpt
jmikola: Added a test in `bson-enum_error-003.phpt` (Typemap specifies enum that does not implement Unserializable).

tests/bson/bson-enum-003.phpt
jmikola: Added a PersistableEnum trait, which includes a `bsonSerialize()` method that returns the enum object cast as an array and a `bsonUnserialize()` method that does nothing. Presumably, this can be used just fine with an enum implementing any combination of Serializable, Unserializable, and/or Persistable.

If users implement Persistable, the encoded BSON is going to include `__pclass`; otherwise, they'll need to use a typemap and implement Unserializable (as demonstrated in the tests). We'd want to document both approaches, and I suppose the PersistableEnum docs would be a good place to do that.

I don't think there's any reason to create UnserializableEnum, as it'd just be a subset of PersistableEnum with no `bsonSerialize()` method. And if users actually want a different implementation for `bsonSerialize()` (we're well in edge case territory here), they can avoid the trait and just roll their own implementations.

Lastly, I looked into whether we could restrict use of this trait to enum objects (e.g. throw an exception if a non-enum imported the trait) but found no easy way to do so after digging through `zend_compile_use_trait()` in PHP core. I expect this would be possible with some compiler hacking, but I have no idea how to do that and it'd probably be a waste of @sgolemon's time for relatively little benefit.

src/phongo_bson.c
jmikola: FYI: a PersistableEnum trait with a no-op `bsonUnserialize()` method has been created.

tests/bson/bson-enum-002.phpt
jmikola: I spent some time investigating this and realized it was caused by using `ZVAL_COPY_VALUE` instead of `ZVAL_COPY` on the resolved enum case. These are discussed in more detail in [Copying Zvals](https://www.phpinternalsbook.com/php7/zvals/memory_management.html#copying-zvals) in the internals book.

 `ZVAL_COPY_VALUE` was previously used on the return value from `bsonSerialize()`, which made sense there since we didn't need to increment the ref count (the value was being moved). But for enums, this left the original enum case in a bad state, which triggered the assertion failure in `zend_objects_store_del()`. `ZVAL_COPY` correctly increments the case's refcount, which allows all tests to pass w/o leaks.

src/BSON/PersistableEnum.stub.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+/**
+ * @generate-class-entries static
+ * @generate-function-entries
+ */
+
+namespace MongoDB\BSON;
+
+trait PersistableEnum
```
alcaeus, line 10: Seeing these changes makes me appreciate the work people put into gen_stub.php for PHP core - it's just so much easier to define classes now than it was before when everything had to be done manually...

alcaeus (approved):
LGTM!

tests/bson/bson-enum-003.phpt
alcaeus: I agree with only providing `PersistableEnum` - also agree with not restricting the use to traits. No matter how much effort you'll put in it, people will always find a way to shoot themselves in the foot.

PR #1337, opened 2022-07-22
PHPC-1709: Add typing information to arginfo

alcaeus:
PHPC-1708
PHPC-1711
PHPC-2015
PHPC-2115

This PR adds stub files to generate arginfo. As a first proof of concept, this is done for BSON classes, but other classes will follow. Wherever we can't add return types yet (e.g. for interfaces and non-final classes where this constitutes a BC break) we declare a tentative return type to alert people to the upcoming change. In the stub files, this is characterised by the `@tentative-return-type` annotation in the docblock for the corresponding method.

There are a couple of requirements for this, which will be added to the contribution docs later:
- arginfo headers have to be generated using `build/gen_stub.php`. To get this file, `phpize` needs to have been called with PHP 8.0 or later
- Changing stub files entails regenerating arginfo files, as well as removing any `.lo` files for the class in question. I haven't yet managed to define the arginfo files as dependency for the respective class which would enable automatic recompilation.

- [x] Create stub files for BSON classes
- [x] Create stub files for remaining classes
- [x] Update contributing docs
- [x] ~Integrate arginfo generation into build process (if possible)~
- [x] Add checks to build process to ensure stub files are up to date (if the above is not possible)


src/BSON/Binary.c
```diff
@@ -92,7 +93,7 @@ static HashTable* php_phongo_binary_get_properties_hash(phongo_compat_object_han
 
 /* {{{ proto void MongoDB\BSON\Binary::__construct(string $data, int $type)
    Construct a new BSON binary type */
-static PHP_METHOD(Binary, __construct)
+PHP_METHOD(MongoDB_BSON_Binary, __construct)
```
alcaeus, line 96: Since the generated arginfo files declares these methods as non-static, we have to remove this from all classes. @jmikola is there a particular reason these were declared as static?

tests/bson/bson-decimal128_error-001.phpt
```diff
@@ -11,7 +11,7 @@ require_once __DIR__ . '/../utils/basic.inc';
 
 echo throws(function() {
     new MongoDB\BSON\Decimal128([]);
-}, MongoDB\Driver\Exception\InvalidArgumentException::class), "\n";
+}, TypeError::class), "\n";
```
alcaeus, line 14: I'm not entirely sure about these changes - some people may see this as a BC break. I'll have to double-check why the exception class changes for this and ObjectId.

alcaeus:
@jmikola this is ready for a first look before I continue with the rest of these classes. Build failures on PHP 8.1 are due to deprecation warnings since we're introducing tentative return types for BSON interfaces. The affected classes will be updated in the next step once we've confirmed that the general direction is ok.

src/BSON/Binary.c
jmikola: It was just a matter of avoiding namespace pollution when compiling. These functions aren't referenced outside of the corresponding C file.

tests/bson/bson-decimal128_error-001.phpt
jmikola: Does this change apply to _all_ PHP versions? It calls to mind some of the variations between PHP 7.x and 8.0 when we were implementing [PHPC-1631](https://jira.mongodb.org/browse/PHPC-1631).

alcaeus: It affects all PHP versions, since we unconditionally add argument types to methods. I'll take another look at this tomorrow to figure out where this is coming from.

src/BSON/Javascript.stub.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+/** @generate-function-entries */
+
+namespace MongoDB\BSON
+{
+    final class Javascript
+    {
+#if PHP_VERSION_ID >= 80000
+        final public function __construct(string $code, array|object|null $scope = null) {}
```
jmikola, line 10: Can the nullable type hint (`?` prefix) be used here or is `|null` required when using type unions?

src/BSON/RegexInterface.stub.php
```diff
@@ -0,0 +1,18 @@
...
+    interface RegexInterface
+    {
+        /** @tentative-return-type */
+        public function getPattern(): string;
+
+        /** @tentative-return-type */
+        public function getFlags(): string;
+
+        /** @tentative-return-type */
+        public function __toString(): string;
```
jmikola, line 16: Just to confirm: we need to use tentative return types only for the standalone _interfaces_ but not the concrete classes thanks to PHP's internal handling for Stringable implementations. This relates to https://github.com/mongodb/mongo-php-driver/commit/6c82ec8b52fad33c3e9444ae8bd2c27a729c4688.

src/BSON/UTCDateTime.stub.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+/** @generate-function-entries */
+
+namespace MongoDB\BSON
+{
+    final class UTCDateTime
+    {
+#if PHP_VERSION_ID >= 80000
+        public final function __construct(int|string|float|\DateTime|\DateTimeImmutable|null $milliseconds = null) {}
```
jmikola, line 10: I suppose [DateTimeInterface](https://www.php.net/manual/en/class.datetimeinterface.php) isn't feasible to use, since we actually depend on the specific DateTime or DateTimeImmutable classees to access internal fields and don't merely invoke interface methods (i.e. the construct would fail to handle a userland class implementing DateTimeInterface, assuming that's even possible).

src/BSON/DBPointer.stub.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    final class DBPointer
+    {
+        final private function __construct() {}
+
+        final public function __toString(): string {}
+
+        final public function serialize(): string {}
+
+        /** @param string $serialized */
```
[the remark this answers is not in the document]
alcaeus: No changes in the arginfo files, but the `gen_stub.php` script complains if there is no typing information at all, hence the PHPDoc comment.

src/BSON/Binary.stub.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    final class Binary
+    {
+        public final function __construct(string $data, int $type) {}
+
+        final public function getData(): string {}
+
+        final public function getType(): int {}
+
+        public static function __set_state(array $properties): object {}
```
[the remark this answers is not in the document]
alcaeus: Unfortunately, `self` isn't possible:
> $ ./build/gen_stub.php 
In ./src/BSON/Binary.stub.php:
MongoDB\BSON\Binary::__set_state(): The exact class name must be used instead of "self"

However, using `Binary` works, so I'll be changing those.

src/BSON/Binary_arginfo.h
```diff
@@ -0,0 +1,79 @@
...
+
+ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(arginfo_class_MongoDB_BSON_Binary___serialize, 0, 0, IS_ARRAY, 0)
+ZEND_END_ARG_INFO()
+
+#if PHP_VERSION_ID >= 80000
+ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(arginfo_class_MongoDB_BSON_Binary_jsonSerialize, 0, 0, IS_MIXED, 0)
+ZEND_END_ARG_INFO()
+#endif
+
+#if !(PHP_VERSION_ID >= 80000)
```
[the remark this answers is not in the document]
alcaeus: Yeah, I was also irritated by that, but came to realise that I'll just have to live with (in my opinion) sub-optimal generated code.

src/BSON/Javascript.stub.php
alcaeus: PHP requires `|null` for nullable unions. This came up in the discussion for [nullable intersection types](https://wiki.php.net/rfc/nullable_intersection_types), which would essentially be a union of `null` and the intersection type.

src/BSON/RegexInterface.stub.php
alcaeus: That is correct. Strictly speaking, the interface should already extend the `Stringable` interface, which would make this a hard BC break. Adding the `string` return type to `__toString` in the implementation of BSON type interfaces removes this BC breaks, as those classes are all final. We can clean all this up in a 2.0 release where we're free to break BC. Until then, the tentative return type definition will alert implementers of the BSON interfaces that we're adding a return type that they should be adding already.

src/BSON/UTCDateTime.stub.php
alcaeus: The check currently in the constructor currently only accepts `DateTime` or `DateTimeImmutable` instances, but that could be changed. However, we then access the object through `Z_PHPDATE_P`, which in turn calls [`php_date_obj_from_obj`](https://heap.space/xref/PHP-8.1/ext/date/php_date.h?r=c85fcbc2#46). As `DateTimeInterface` technically can't be implemented by user land code and was specifically created to accept both `DateTime` and `DateTimeImmutable` instances (see [documentation](https://www.php.net/manual/en/class.datetimeinterface.php)) we can also try changing it if you think it's safe to do so.

src/BSON/Binary.c
alcaeus: Ack, in that case we'll have to live with the polluted namespace, since we can't change the generated stub files :(

src/BSON/Binary.stub.php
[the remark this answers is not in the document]
jmikola: The prohibition on `self` dates back to https://github.com/php/php-src/commit/80e5ad5a293117552cf1d616b3a7e4b23dfc8fa5 and https://github.com/php/php-src/pull/7215. That PR references https://github.com/php/doc-en/pull/624, although I didn't find any outright discussion about why `self` is prohibited there (just that they opted to use class names instead of `self` for docs).

It looks like `static` might work if you'd rather use that; otherwise, no objections to just hard-coding the specific class name for these methods.

src/BSON/UTCDateTime.stub.php
jmikola: Ah, I forgot that `DateTimeInterface` was only available for internal use. We actually use that in our documentation for [`UTCDateTime::__construct()`](https://www.php.net/manual/en/mongodb-bson-utcdatetime.construct.php) so I think it'd make sense to do so here, too. See: https://github.com/mongodb/mongo-php-driver/blob/1.14.0/src/BSON/UTCDateTime.c#L152-L161

src/BSON/Binary.stub.php
```diff
@@ -0,0 +1,65 @@
...
+  * @generate-class-entries static
+  * @generate-function-entries static
+  */
+
+namespace MongoDB\BSON
+{
+    final class Binary implements BinaryInterface, JsonSerializable, Type, \Serializable
+    {
+        /** @var int */
+        public const TYPE_GENERIC = 0x01;
```
alcaeus, line 13: I've added these in a separate commit since we used to use the libbson enum to define these constants. Since these are protected by the BC promise, I don't mind hardcoding the values here but can revert the change if you'd rather use libbson values.

src/BSON/Binary.c
```diff
@@ -400,17 +357,8 @@ static HashTable* php_phongo_binary_get_properties(phongo_compat_object_handler_
 
 void php_phongo_binary_init_ce(INIT_FUNC_ARGS) /* {{{ */
 {
-	zend_class_entry ce;
-
-	INIT_NS_CLASS_ENTRY(ce, "MongoDB\\BSON", "Binary", php_phongo_binary_me);
-	php_phongo_binary_ce                = zend_register_internal_class(&ce);
+	php_phongo_binary_ce                = register_class_MongoDB_BSON_Binary(php_phongo_binary_interface_ce, php_phongo_json_serializable_ce, php_phongo_type_ce, zend_ce_serializable);
```
alcaeus, line 360: This function is generated by specifying `generate-class-entries` in the stub file. It can't take care of object handlers, but it allows us to rely on the stub files for the class structure (modifiers, extended classes, implemented interfaces, class and property definitions, etc.).

src/BSON/BinaryInterface.stub.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+/**
+  * @generate-class-entries static
+  * @generate-function-entries
```
alcaeus, line 5: Note that for interfaces there is no `static` suffix behind `generate-function-entries`. Specifying that generates an invalid stub file due to a bug in gen_stub.php. I'll look at that separately.

alcaeus (commented):
Added `generate-function-entries` to existing stubs and added stubs for exception classes. See a couple of notes below.

src/BSON/Binary.c
alcaeus: I realised that using `generate-function-entries static` in the stub file fixes this, so we're back to `static PHP_METHOD(...)` :)

src/BSON/Binary.stub.php
jmikola: Ah, I didn't consider that. Since `php_phongo_binary_init_ce` isn't going away, I think keeping the `zend_declare_class_constant_long` lines in there is sensible. This is going to come up with other files as well, and I'd rather depend on the libmongoc constants if possible, since that might also catch a bug on their end.

If not, I'd at least want to have comments in these stub files referring to the libmongoc constant we're hard-coding. For example:

```
public const TYPE_GENERIC = 0x01; /* BSON_SUBTYPE_BINARY */
```

src/BSON/BinaryInterface.stub.php
jmikola: What does `static` do for `generate-class-entries`?

alcaeus: Yes, the tag value is injected into the function definitions: `generate-class-entries` maps to the `register_class_MongoDB_...` function, while `generate-function-entries` maps to the `ZEND_METHOD(MongoDB_..., ...)` lines.

src/BSON/Binary.stub.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+/**
+  * @generate-class-entries static
+  * @generate-function-entries static
+  */
+
+namespace MongoDB\BSON
+{
+    final class Binary implements BinaryInterface, JsonSerializable, Type, \Serializable
```
jmikola, line 10: Should `JsonSerializable` be `\JsonSerializable`? It's not defined in the `MongoDB\BSON` namespace.

----

Since `php_phongo_binary_init_ce` includes:

```
#if PHP_VERSION_ID >= 80000
	zend_class_implements(php_phongo_binary_ce, 1, zend_ce_stringable);
#endif
```

Would it make sense to use `#if` here to define two different class declarations, or perhaps just an extra line between this and the opening `{` to avoid duplication? Something like:

```
final class Binary implements BinaryInterface, JsonSerializable, Type, \Serializable
#if PHP_VERSION_ID >= 80000
    , \Stringable
#endif
{
```

alcaeus: Yes - I'll change this to `\JsonSerializable`. I was confused by the `php_phongo_json_serializable_ce` variable 🤦 

Unfortunately, the `#if` is completely ignored in the class header (both with and without duplication of the class definition in conditionals). We're unfortunately stuck adding this ourselves until we require PHP 8.0, at which point we can add it to the stub files.

jmikola: See: https://github.com/mongodb/mongo-php-driver/blob/1.14.0/php_phongo.c#L183

We may actually want to continue using `php_phongo_json_serializable_ce`, since it's fetched during MINIT to avoid problems on some distributions where the JSON extension was _not_ compiled in to PHP and instead loaded dynamically. That was definitely an issue back in the 5.x releases (see: [PHPC-878](https://jira.mongodb.org/browse/PHPC-878)), and it may still apply to 7.x+. We may need to confirm with @remicollet.

And while I'm looking at the above, I think the manual fetching of `DateTimeImmutable` may be obsolete. I've opened [PHPC-2115](https://jira.mongodb.org/browse/PHPC-2115) to track that.

src/BSON/Binary.stub.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+/**
+  * @generate-class-entries static
+  * @generate-function-entries static
```
remicollet, line 5: Also nice to add   `@generate-legacy-arginfo 80000` to take benefit of compatibility stuff when using 8.2. Example when using SensitiveParameter attributes, see

See https://github.com/pierrejoye/php_zip/blob/master/php81/php_zip.stub.php#L91
And https://github.com/pierrejoye/php_zip/blob/master/php81/php_zip_arginfo.h#L532

src/BSON/Binary.stub.php
remicollet: Try (may requires 8.2 to generate the arginfo)
```
/** @cvalue BSON_SUBTYPE_BINARY */
public const TYPE_GENERIC = UNKNOWN;
```

See https://github.com/php/php-src/commit/1362fefc85ea8da7e32338291fa36182745c9622

src/BSON/Binary.stub.php
remicollet: Json is always build statically in 8.x, so this hack is still useful on 7.x

(btw, only there to avoid "undefined symbol" when load order is not the correct one for linux distribution using `--enable-rtld-now`)

src/BSON/Binary.stub.php
remicollet: Also notice that is some additional compatibility checks are missing, is still possible to add them in gen_stub.php before 8.2.0 GA

src/BSON/Binary.stub.php
alcaeus: Ah, that is nice. I was already using the 8.2 file (the previous version didn't handle constants), so I switched to using `@cvalue`.

src/BSON/Binary.stub.php
remicollet: If you use 70200 it will also generate a `foo_legady_arginfo.h` without type hinting.

alcaeus: `generate-legacy-arginfo 70200` is not supported, so we could only generate legacy arginfo files for PHP 7.0. Since those don't include any argument types at all (which we want since PHP 7.2 allows for this), we'll just do without legacy arginfo for now. We've got everything we wanted covered this way as well:
* argument types (`mixed` and unions only on PHP 8.0+)
* return types for final methods/classes
* tentative return types for non-final methods/classes (PHP 8.1+, no return type info in other versions)

remicollet: generate-legacy-arginfo accepts any PHP_VERISON_ID, "legacy" file is generated as soon as < 80000 ;)

BTW? I really think it should be used, at least for SensitiveParameter.
And hacks for compatibility could probably be proposed in php_src so every ext will be benefit of them.

This is how OpenSource work ;)

alcaeus:
```
$ ./build/gen_stub.php 
In ./src/MongoDB/WriteConcern.stub.php:
Legacy PHP version must be one of: "70000" (PHP 7.0), "80000" (PHP 8.0), "80100" (PHP 8.1), "80200" (PHP 8.2), "70200" provided
```

I'll take a look at adding a compatibility mode for PHP 7.2, but at the moment it doesn't exist. As it is, legacy arginfo isn't useful for us.

src/BSON/UTCDateTime.stub.php
alcaeus: Changed as suggested 👍

tests/exception/bulkwriteexception-getwriteresult-001.phpt
```diff
@@ -1,22 +1,60 @@
...
+    var_dump($e->getWriteResult());
+}
 
 ?>
 ===DONE===
 <?php exit(0); ?>
---EXPECT--
-bool(true)
+--EXPECTF--
+object(MongoDB\Driver\WriteResult)#%d (9) {
```
jmikola, line 26: Note that we typically use `%d` for the number of object properties as well. That keeps the diff smaller if we end up adding additional properties down the line.

Change can be made here and in the other two `object(` lines below.

tests/bson/bson-decimal128_error-001.phpt
jmikola: I'm pretty sure this is the result of parameter types now being validated by arginfo instead of zpp. That's probably something we should document in release notes (you can add `documentation-needed` on the ticket and make a note of this in a comment). I don't think there's any place to mention this in the extension docs, though -- it'd be pedantic to add this to _every_ method's changelog.

jmikola (changes requested):

.github/workflows/arginfo-files.yml
```diff
@@ -0,0 +1,45 @@
+name: "Check generated arginfo files"
+
+on:
+  pull_request:
+    branches:
+      - "v*.*"
+      - "master"
+      - "feature/*"
```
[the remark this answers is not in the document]
alcaeus: This was added in https://github.com/mongodb/mongo-php-driver/commit/69cc2b5c7418ef4bf320a160b3d6b0918ef7102d, and it's a nice way to prevent builds on all branches while retaining the ability to test feature branches in our forks without messing with the workflow files.

.github/workflows/arginfo-files.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+        with:
+          submodules: true
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          extensions: ":mongodb"
```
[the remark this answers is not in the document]
alcaeus: I think this probably was copied from PHPLIB and isn't necessary. Removing it might save us some build time, so I'll go ahead and do that.

config.m4
```diff
@@ -66,7 +66,6 @@ if test "$PHP_MONGODB" != "no"; then
     AX_CHECK_COMPILE_FLAG(-Wenum-compare,                   _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wenum-compare"                 ,, -Werror)
     AX_CHECK_COMPILE_FLAG(-Wempty-body,                     _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wempty-body"                   ,, -Werror)
     AX_CHECK_COMPILE_FLAG(-Wparentheses,                    _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wparentheses"                  ,, -Werror)
-    AX_CHECK_COMPILE_FLAG(-Wdeclaration-after-statement,    _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wdeclaration-after-statement"  ,, -Werror)
```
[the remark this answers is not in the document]
alcaeus: Correct. A quick test with C99 showed some more errors, so when I realised that we add this flag specifically when enabling developer flags I figured the easiest way to fix this was to remove this single flag. We can create a ticket if we want to move to C99, but I figured it probably wasn't as easy as adding `std=C99` and call it a day.

src/MongoDB/ClientEncryption.stub.php
```diff
@@ -0,0 +1,58 @@
...
+ * @generate-function-entries static
+ */
+
+namespace MongoDB\Driver
+{
+    final class ClientEncryption
+    {
+        /**
+          * @var string
+          * @cvalue MONGOC_AEAD_AES_256_CBC_HMAC_SHA_512_DETERMINISTIC
```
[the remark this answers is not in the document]
alcaeus: The stub generator doesn't know about this, and the generated arginfo files are dumb in that regard - not just towards our code but towards everything. The file expects the including code to import all necessary symbols before including the arginfo header file.

php_phongo.c
```diff
@@ -44,7 +44,6 @@ ZEND_TSRMLS_CACHE_DEFINE();
 static int32_t phongo_num_threads = 0;
 
 /* Declare zend_class_entry dependencies, which are initialized in MINIT */
-zend_class_entry* php_phongo_date_immutable_ce;
```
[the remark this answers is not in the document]
alcaeus: Good point, I amended the commit.

src/BSON/MaxKey.stub.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+/**
+  * @generate-class-entries static
+  * @generate-function-entries static
+  */
+
+namespace MongoDB\BSON
```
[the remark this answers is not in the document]
alcaeus: Yes - not sure why I opted for the indented namespace declaration, but it's fixed now.

src/MongoDB/BulkWrite.stub.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace MongoDB\Driver
+{
+    final class BulkWrite implements \Countable
+    {
+        public final function __construct(array $options = []) {}
+
+        public function count(): int {}
+
+#if PHP_VERSION_ID >= 80000
```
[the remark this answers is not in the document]
alcaeus: Changed, I have no preference as to what the generated code looks like :)

php_phongo.h
```diff
@@ -109,6 +109,20 @@ zend_object_handlers* phongo_get_std_object_handlers(void);
 		(intern)->created_by_pid = (int) getpid(); \
 	} while (0)
 
+#define PHONGO_DISABLED_CONSTRUCTOR(classname)                                         \
+	PHP_METHOD(classname, __construct)                                                 \
```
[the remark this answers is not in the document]
alcaeus: Good point - added.

src/MongoDB/ReadPreference.stub.php
```diff
@@ -0,0 +1,119 @@
...
+ * @generate-class-entries static
+ * @generate-function-entries static
+ */
+
+namespace MongoDB\Driver
+{
+    final class ReadPreference implements \MongoDB\BSON\Serializable, \Serializable
+    {
+        /**
+          * @var int
```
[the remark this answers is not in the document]
alcaeus: Fixed while updating the namespace declaration.

src/MongoDB/Server.stub.php
```diff
@@ -0,0 +1,123 @@
...
+        /**
+          * @var int
+          * @cvalue PHONGO_SERVER_LOAD_BALANCER
+          */
+        public const TYPE_LOAD_BALANCER = UNKNOWN;
+
+        final private function __construct() {}
+
+#if PHP_VERSION_ID >= 80000
+        final public function executeBulkWrite(string $namespace, BulkWrite $zwrite, array|WriteConcern|null $options = null): WriteResult {}
```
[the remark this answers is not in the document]
alcaeus: Done 👍

src/MongoDB/ServerApi.stub.php
```diff
@@ -0,0 +1,35 @@
...
+ * @generate-class-entries static
+ * @generate-function-entries static
+ */
+
+namespace MongoDB\Driver
+{
+    final class ServerApi implements \MongoDB\BSON\Serializable, \Serializable
+    {
+        /** @var string */
+        public const V1 = '1';
```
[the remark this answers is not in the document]
alcaeus: There is the `mongoc_server_api_version_t` enum in libmongoc, but that doesn't store the string version. Instead, it uses `mongoc_server_api_version_to_string` to return the version string for the enum, hence the duplicate definition in our `ServerApi` class.

src/MongoDB/Session.h
```diff
@@ -21,6 +21,12 @@
 
 #include <php.h>
 
+#define PHONGO_TRANSACTION_NONE "none"
+#define PHONGO_TRANSACTION_STARTING "starting"
+#define PHONGO_TRANSACTION_IN_PROGRESS "in_progress"
+#define PHONGO_TRANSACTION_COMMITTED "committed"
+#define PHONGO_TRANSACTION_ABORTED "aborted"
```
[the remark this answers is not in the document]
alcaeus: Correct. I can also move them back to Session.c and define them before including the arginfo file if we want to keep these private.

tests/utils/classes.inc
```diff
@@ -38,17 +39,18 @@ class Address implements MongoDB\BSON\Persistable {
     protected $zip;
     protected $country;
 
-    function __construct($zip, $country) {
+    public function __construct($zip, $country) {
         $this->zip = $zip;
         $this->country = $country;
     }
-    function bsonSerialize() {
+    #[\ReturnTypeWillChange]
```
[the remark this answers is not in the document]
alcaeus: That is correct - PHP 8.1 adds the concept of tentative return types to alert users of upcoming return type changes (which have to be added downstream first). There's PHPC-1710 to document these type changes in interfaces, which I'll tackle once this PR is done. This will involve adding language to the interface documentation on php.net to point out this change and how to silence the errors without adding types first. In this case I didn't add a return type since it would be `mixed` which would require a test split due to the wide range of PHP versions we support. Users may be more flexible as to what return types they can add, especially as they can narrow the return type.

tests/topologyDescription/topologyDescription-hasReadableServer_error-001.phpt
```diff
@@ -44,7 +44,6 @@ class TopologyDescriptionProvider implements MongoDB\Driver\Monitoring\SDAMSubsc
 }
 
 $tests = [
-    null,
```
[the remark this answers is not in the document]
alcaeus: I can add language to the documentation in PHPC-1710.

This is one of those special cases that can't be reproduced in PHP userland code as it doesn't know the concept of optional, non-nullable arguments. The only way to prevent users from passing an explicit `null` value would be to make the parameter required, which would be a BC break. I figured if users want to pass `null` there's nothing that would prevent us from handling it the same way as if they didn't pass anything at all.

tests/readConcern/readconcern-ctor_error-002.phpt
```diff
@@ -15,15 +15,15 @@ $tests = [
...
 ?>
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
-OK: Got MongoDB\Driver\Exception\InvalidArgumentException
-%SMongoDB\Driver\ReadConcern::__construct()%sstring, array given
-OK: Got MongoDB\Driver\Exception\InvalidArgumentException
-%SMongoDB\Driver\ReadConcern::__construct()%sstring, %r(object|stdClass)%r given
+OK: Got TypeError
+%SMongoDB\Driver\ReadConcern::__construct()%sstring or null, array given
```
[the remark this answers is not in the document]
alcaeus: That is correct, the error message now gets thrown by the zpp logic.

tests/manager/manager-executeQuery-005.phpt
```diff
@@ -10,7 +10,7 @@ MongoDB\Driver\Manager::executeQuery() with filter and projection
 require_once __DIR__ . "/../utils/basic.inc";
 
 class MyArrayObject extends ArrayObject implements MongoDB\BSON\Unserializable {
-    function bsonUnserialize(array $data) {
+    public function bsonUnserialize(array $data): void {
```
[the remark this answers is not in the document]
alcaeus: I added this on my own since I did a batch find/replace across all test files for this method.

tests/exception/bulkwriteexception-getwriteresult-001.phpt
alcaeus: Good point, done.

tests/bson/bson-decimal128_error-001.phpt
alcaeus: Yes, I believe adding this to each method's changelog would be overkill. Maybe a paragraph in the release notes would be appropriate for this.

src/MongoDB/Manager.stub.php
```diff
@@ -0,0 +1,64 @@
...
+
+        final public function getWriteConcern(): WriteConcern {}
+
+        final public function removeSubscriber(Monitoring\Subscriber $subscriber): void {}
+
+        final public function selectServer(?ReadPreference $readPreference = null): Server {}
+
+        final public function startSession(array $options = []): Session {}
+
+        final public function __wakeup(): void {}
```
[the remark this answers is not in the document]
alcaeus: I did pretty much that, so now:
* all classes that use `PHONGO_CE_DISABLE_SERIALIZATION` also use `PHONGO_DISABLED_WAKEUP` and vice-versa
* all classes that use `PHONGO_DISABLED_WAKEUP` also define the method in their stub file

.github/workflows/arginfo-files.yml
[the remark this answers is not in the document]
jmikola: Oh, I forgot I added that. IIRC, this was back when we had a separate branch for SDAM monitoring.

src/MongoDB/Session.h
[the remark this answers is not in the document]
jmikola: No objections. I'd rather keep all of our includes above any definitions and code, so moving these to `Session.h` makes sense.

src/MongoDB/Cursor.c
```diff
@@ -291,49 +292,8 @@ PHP_METHOD(Cursor, rewind)
...
-	PHP_ME(Cursor, rewind, ai_Cursor_rewind, ZEND_ACC_PUBLIC)
-
-	ZEND_NAMED_ME(__construct, PHP_FN(MongoDB_disabled___construct), ai_Cursor_void, ZEND_ACC_PRIVATE | ZEND_ACC_FINAL)
-	ZEND_NAMED_ME(__wakeup, PHP_FN(MongoDB_disabled___wakeup), ai_Cursor_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
-	PHP_FE_END
-};
-/* clang-format on */
-/* }}} */
+PHONGO_DISABLED_CONSTRUCTOR(MongoDB_Driver_Cursor)
+PHONGO_DISABLED_WAKEUP(MongoDB_Driver_Cursor)
```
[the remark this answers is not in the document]
jmikola: This was addressed in recent commits. Related to: https://github.com/mongodb/mongo-php-driver/pull/1337#discussion_r935259248

src/MongoDB/Manager.c
```diff
@@ -808,85 +811,6 @@ static PHP_METHOD(Manager, startSession)
...
-	PHP_ME(Manager, executeBulkWrite, ai_Manager_executeBulkWrite, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
-	PHP_ME(Manager, getEncryptedFieldsMap, ai_Manager_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
-	PHP_ME(Manager, getReadConcern, ai_Manager_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
-	PHP_ME(Manager, getReadPreference, ai_Manager_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
-	PHP_ME(Manager, getServers, ai_Manager_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
-	PHP_ME(Manager, getWriteConcern, ai_Manager_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
-	PHP_ME(Manager, removeSubscriber, ai_Manager_removeSubscriber, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
-	PHP_ME(Manager, selectServer, ai_Manager_selectServer, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
-	PHP_ME(Manager, startSession, ai_Manager_startSession, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
-	ZEND_NAMED_ME(__wakeup, PHP_FN(MongoDB_disabled___wakeup), ai_Manager_void, ZEND_ACC_PUBLIC | ZEND_ACC_FINAL)
```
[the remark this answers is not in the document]
jmikola: This was addressed in recent commits. Related to: https://github.com/mongodb/mongo-php-driver/pull/1337#discussion_r935259248

jmikola (approved):
GitHub decided to stop rendering the full PR diff, but I reviewed the commits since my last review and it looks like everything has been addressed.

jmikola:
The [1.15.0 release](https://github.com/mongodb/mongo-php-driver/releases/tag/1.15.0) uses [tentative return types](https://wiki.php.net/rfc/internal_method_return_types) for interfaces methods. Our reasoning at the time was that this was a BC-friendly migration path since it'd have no effect on PHP 8.0 and earlier (see: [`phongo_compat.h`](https://github.com/mongodb/mongo-php-driver/blob/1.15.0/src/phongo_compat.h#L267)).

That said, I now realize that users _already_ on PHP 8.1 would have encountered this issue for the first time when upgrading from driver version 1.14.x to 1.15.0. I apologize for any inconvenience this may have caused on your side.

> Should this change be followed with a major release?

In hindsight, it may have been preferable punt on introducing tentative return types altogether. I don't think the change alone would be enough to warrant a bump to 2.0 (there are other larger changes planned, but not yet implemented). Tentative return types perhaps make more sense in PHP core where BC breaks regularly show up in minor releases (e.g. https://www.php.net/manual/en/migration81.incompatible.php), but not PECL extensions which follow external release schedules.

[PHPC-2139](https://jira.mongodb.org/browse/PHPC-2139) is the canonical ticket for BC-breaking typing changes for 2.0. That will be the next opportunity to replace these tentative return types with actual return types.

PR #1339, opened 2022-07-29
PHPC-2093: Key Management API

jmikola:
https://jira.mongodb.org/browse/PHPC-2093


src/MongoDB/ClientEncryption.c
```diff
@@ -70,6 +151,44 @@ static PHP_METHOD(ClientEncryption, createDataKey)
...
+		/* Exception already thrown */
+		goto cleanup;
+	}
+
+	if (!mongoc_client_encryption_delete_key(Z_CLIENTENCRYPTION_OBJ_P(getThis())->client_encryption, &keyid, &reply, &error)) {
+		phongo_throw_exception_from_bson_error_t(&error);
+		goto cleanup;
+	}
+
+	RETVAL_OPTIONAL_BSON_T(reply);
```
alcaeus, line 185: Strictly speaking, this could return `null` if `reply` were to come back empty for some reason, which would then cause a fatal error due to a return type mismatch. In my opinion we should handle that before to ensure we can't run into the `RETVAL_NULL()` case without throwing an exception. Alternatively, we can change the return value to `object|null` if that makes sense.

src/MongoDB/Cursor.c
```diff
@@ -261,7 +262,7 @@ PHP_METHOD(Cursor, rewind)
 
 	PHONGO_PARSE_PARAMETERS_NONE();
 
-	/* If the intern was never advanced (e.g. command intern), do so now */
+	/* If the cursor was never advanced (e.g. command cursor), do so now */
```
alcaeus, line 265: 👍

src/MongoDB/ClientEncryption.c
```diff
@@ -120,7 +120,7 @@ ZEND_BEGIN_ARG_INFO_EX(ai_ClientEncryption_encrypt, 0, 0, 1)
 ZEND_END_ARG_INFO()
 
 ZEND_BEGIN_ARG_INFO_EX(ai_ClientEncryption_decrypt, 0, 0, 1)
-	ZEND_ARG_OBJ_INFO(0, keyVaultClient, MongoDB\\BSON\\BinaryInterface, 0)
+	ZEND_ARG_OBJ_INFO(0, value, MongoDB\\BSON\\BinaryInterface, 0)
```
alcaeus, line 455: Note that this will become obsolete once #1337 is merged.

src/MongoDB/Cursor.c
```diff
@@ -492,6 +493,78 @@ void php_phongo_cursor_init_ce(INIT_FUNC_ARGS) /* {{{ */
 	php_phongo_handler_cursor.offset         = XtOffsetOf(php_phongo_cursor_t, std);
 } /* }}} */
 
+static void phongo_cursor_init(zval* return_value, zval* manager, mongoc_cursor_t* cursor, zval* readPreference, zval* session) /* {{{ */
```
alcaeus, line 496: Noted that this logic was moved from phongo_execute.c

src/MongoDB/ClientEncryption.c
jmikola: Good point. I believe this is the only method where the reply document should never be empty. Other methods allow empty documents when nothing matches.

I'll add a catch for this that throws an UnexpectedValueException before we call `RETVAL_OPTIONAL_BSON_T`.

src/MongoDB/ClientEncryption.c
jmikola: This was removed after rebasing on master.

tests/utils/basic.inc
```diff
@@ -8,5 +8,7 @@ define('DATABASE_NAME', getenv('MONGODB_DATABASE') ?: 'phongo');
 define('COLLECTION_NAME', makeCollectionNameFromFilename($_SERVER['SCRIPT_FILENAME']));
 define('NS', DATABASE_NAME . '.' . COLLECTION_NAME);
 define('SSL_DIR', realpath(getenv('SSL_DIR')));
-define('CSFLE_KEY_VAULT_NS', 'encryption.__keyVault');
+define('CSFLE_KEY_VAULT_DATABASE_NAME', 'encryption');
+define('CSFLE_KEY_VAULT_COLLECTION_NAME', '__keyVault');
+define('CSFLE_KEY_VAULT_NS', CSFLE_KEY_VAULT_DATABASE_NAME . '.' . CSFLE_KEY_VAULT_COLLECTION_NAME);
```
jmikola, line 13: Splitting these was necessary to utilize `skip_if_not_clean` in tests. I originally created a `skip_if_ns_not_clean` method but this seemed more straightforward.

src/MongoDB/ClientEncryption.stub.php
```diff
@@ -50,12 +52,24 @@ final public function decrypt(\MongoDB\BSON\Binary $value): mixed {}
...
 
+    final public function getKey(\MongoDB\BSON\Binary $keyId): ?object {}
+
+    final public function getKeyByAltName(string $keyAltName): ?object {}
+
+    final public function getKeys(): \MongoDB\Driver\Cursor {}
+
+    final public function removeKeyAltName(\MongoDB\BSON\Binary $keyId, string $keyAltName): ?object {}
+
+    final public function rewrapManyDataKey(array|object $filter, array $options = []): object {}
```
alcaeus, line 72: This should be wrapped in a conditional, as union types are not available on PHP < 8.0. That should also deal with the missing `ZEND_ARG_TYPE_MASK` macro.

jmikola: Done. Thanks for pointing this out.

alcaeus (approved):
LGTM!

src/MongoDB/ClientEncryption.c
```diff
@@ -106,6 +237,223 @@ static PHP_METHOD(MongoDB_Driver_ClientEncryption, decrypt)
...
+	bson_error_t error         = { 0 };
+	bson_t       reply         = BSON_INITIALIZER;
+
+	mongoc_client_encryption_rewrap_many_datakey_result_t* result = NULL;
+	const bson_t*                                          bulk_write_result;
+
+	PHONGO_PARSE_PARAMETERS_START(1, 2)
+	PHONGO_PARAM_ARRAY_OR_OBJECT(zfilter)
+	Z_PARAM_OPTIONAL
+	Z_PARAM_ARRAY(options)
```
jmikola, line 405: While this is technically correct, we should change this to also accept null for consistency with older APIs.

Related to https://github.com/mongodb/mongo-php-driver/pull/1346#issuecomment-1216712630

tests/clientEncryption/clientEncryption-rewrapManyDataKey-003.phpt
```diff
@@ -0,0 +1,31 @@
+--TEST--
+MongoDB\Driver\ClientEncryption::rewrapManyDataKey() accepts null for $options
```
jmikola, line 2: FYI: this test covers the changes from https://github.com/mongodb/mongo-php-driver/pull/1347 for [PHPC-2124](https://jira.mongodb.org/browse/PHPC-2124). I don't believe we had existing tests for passing `null`, as we probably would have caught the stub conflict sooner if so.

jmikola:
@alcaeus: Feel free to review my last commit. I'll wait on https://github.com/mongodb/mongo-php-driver/pull/1347 to be merged before merging this.

One question: when merging generated `arginfo.h` files, I assume we always want to regenerate those _after_ merging to ensure the "Stub hash" is consistent. Is that correct?

alcaeus:
> One question: when merging generated `arginfo.h` files, I assume we always want to regenerate those _after_ merging to ensure the "Stub hash" is consistent. Is that correct?

That is correct. Luckily, the stub hash in the arginfo file will conflict if the file was changed in both head and base branch of a PR, requiring a rebase and re-generation of the arginfo file.

alcaeus (approved):

PR #1340, opened 2022-08-01
PHPC-2117: Test on PHP 8.2

alcaeus:
PHPC-2117

### DIFF
--- a/tests/apm/commandFailedEvent-002.phpt
+++ b/tests/apm/commandFailedEvent-002.phpt
@@ -12,6 +12,9 @@ $m = create_test_manager();
 
 class MySubscriber implements MongoDB\Driver\Monitoring\CommandSubscriber
 {
+    private $startRequestId;
+    private $startOperationId;
+
     public function commandStarted( \MongoDB\Driver\Monitoring\CommandStartedEvent $event ): void
     {
         echo "started: ", $event->getCommandName(), "\n";
--- a/tests/bson/bson-binary-set_state-001.phpt
+++ b/tests/bson/bson-binary-set_state-001.phpt
@@ -25,27 +25,27 @@ foreach ($tests as $test) {
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
-MongoDB\BSON\Binary::__set_state(array(
+%r\\?%rMongoDB\BSON\Binary::__set_state(array(
 %w'data' => 'foobar',
 %w'type' => 0,
 ))
 
-MongoDB\BSON\Binary::__set_state(array(
+%r\\?%rMongoDB\BSON\Binary::__set_state(array(
 %w'data' => '',
 %w'type' => 0,
 ))
 
-MongoDB\BSON\Binary::__set_state(array(
+%r\\?%rMongoDB\BSON\Binary::__set_state(array(
 %w'data' => '' . "\0" . 'foo',
 %w'type' => 0,
 ))
 
-MongoDB\BSON\Binary::__set_state(array(
+%r\\?%rMongoDB\BSON\Binary::__set_state(array(
 %w'data' => '>Egè›Ó¤VBfUD' . "\0" . '' . "\0" . '',
 %w'type' => 4,
 ))
 
-MongoDB\BSON\Binary::__set_state(array(
+%r\\?%rMongoDB\BSON\Binary::__set_state(array(
 %w'data' => '8Xö"0¬<‘_0fCÆ?',
 %w'type' => 5,
 ))
--- a/tests/bson/bson-encode-003.phpt
+++ b/tests/bson/bson-encode-003.phpt
@@ -6,6 +6,8 @@ BSON encoding: Encoding objects into BSON representation
 require_once __DIR__ . '/../utils/basic.inc';
 
 class MyClass implements MongoDB\BSON\Persistable {
+    private $props;
+
     #[\ReturnTypeWillChange]
     public function bsonSerialize() {
         return array(
@@ -18,6 +20,8 @@ class MyClass implements MongoDB\BSON\Persistable {
     }
 }
 class MyClass2 implements MongoDB\BSON\Persistable {
+    private $props;
+
     #[\ReturnTypeWillChange]
     public function bsonSerialize() {
         return array(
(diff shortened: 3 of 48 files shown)


tests/apm/commandFailedEvent-002.phpt
```diff
@@ -12,6 +12,9 @@ $m = create_test_manager();
 
 class MySubscriber implements MongoDB\Driver\Monitoring\CommandSubscriber
 {
+    private $startRequestId;
+    private $startOperationId;
```
jmikola, line 16: Related: [PHP 8.2 warns on dynamic property creation](https://wiki.php.net/rfc/deprecate_dynamic_properties)

tests/bson/bson-binary-clone-001.phpt
```diff
@@ -18,7 +18,6 @@ $types = array(
 foreach($types as $type) {
     // Use 16-byte data to satisfy UUID requirements
     $binary = new MongoDB\BSON\Binary('randomBinaryData', $type);
-    $binary->foo = 'bar';
```
jmikola, line 21: This was actually a regression test for [PHPC-1230](https://jira.mongodb.org/browse/PHPC-1230) dating back to https://github.com/mongodb/mongo-php-driver/commit/aed838497bd8f87fb2a5b0cc0fbf0e66a8a6d54d.

If we're removing this across the board, would it make sense to reintroduce a `clone` test for at least one class that runs on PHP versions before 8.2?

tests/bson/bson-binary-set_state-001.phpt
```diff
@@ -25,27 +25,27 @@ foreach ($tests as $test) {
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
-MongoDB\BSON\Binary::__set_state(array(
+%r\\?%rMongoDB\BSON\Binary::__set_state(array(
```
jmikola, line 28: I assume PHP 8.2+ now emits the FQCN with a leading backslash. Is that correct?

tests/bson/bson-encode-003.phpt
```diff
@@ -6,6 +6,8 @@ BSON encoding: Encoding objects into BSON representation
 require_once __DIR__ . '/../utils/basic.inc';
 
 class MyClass implements MongoDB\BSON\Persistable {
+    private $props;
```
jmikola, line 9: image

tests/bson/bson-toPHP-003.phpt
```diff
@@ -9,6 +9,7 @@ class MyClass
 {
 }
 
+#[\AllowDynamicProperties]
```
jmikola, line 12: I suppose there's no way to add this attribute to internally defined classes, right? If so, we could use this for testing preservation of dynamic properties during object cloning in lieu of restricting the test to pre-8.2.

tests/bson/bug1598-002.phpt
```diff
@@ -1,5 +1,8 @@
 --TEST--
 PHPC-1598: BSON type get_gc should delegate to zend_std_get_properties
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_php_version('>=', '8.1.99'); ?>
```
jmikola, line 5: Necessary since "< 8.2.0" would allow the test to run on pre-release versions, correct?

What's the actual reason that we can't run this on 8.2+? Is it due to dynamic creation of the `a` property on each BSON object? I noted from https://wiki.php.net/rfc/deprecate_dynamic_properties that stdClass isn't restricted.

jmikola (approved):
A few comments/questions but I don't think any changes are needed.

jmikola:
Note: the [one test failure on server "latest" appears](https://evergreen.mongodb.com/lobster/evergreen/test/mongo_php_driver_test_mongodb_versions__mongodb_versions~latest_php_edge_versions~latest_stable_os~debian11_test_replicaset_auth_patch_e2f91446bc6bc73ff4149bf352d883f4be6dd641_62e7b22d3e8e8658590c2475_22_08_01_10_59_58/0/a25612ad0d09fd2ca51fb6ac5f40da26/#bookmarks=0%2C11&l=1&shareLine=1) appears unrelated to [PHPC-2075](https://jira.mongodb.org/browse/PHPC-2075) and may warrant a new `failing-on-waterfall` issue.

tests/bson/bson-toPHP-003.phpt
alcaeus: It can be added to internally defined classes using `zend_add_class_attribute`, although I'd much rather take advantage of this new functionality and disallow dynamic attributes on our BSON classes.

tests/bson/bug1598-002.phpt
alcaeus: > Necessary since "< 8.2.0" would allow the test to run on pre-release versions, correct?

I remember prior art for doing this, but I can also change it to `< 8.2.0`.

> What's the actual reason that we can't run this on 8.2+? Is it due to dynamic creation of the `a` property on each BSON object? I noted from https://wiki.php.net/rfc/deprecate_dynamic_properties that stdClass isn't restricted.

That is correct - starting with PHP 8.2 creating dynamic properties on classes emits a deprecation notice, with the functionality being removed in PHP 9.

tests/bson/bson-binary-set_state-001.phpt
alcaeus: Yes, that's correct. I opted for the `EXPECTF` regex over separate tests for this change, despite the rough look of the regular expression.

tests/bson/bson-binary-clone-001.phpt
alcaeus: That's a good point - I'll duplicate the tests and have them create dynamic properties on PHP before 8.2. 👍

alcaeus: Duplicated each test for PHP 8.2 and restored the originals 👍

tests/bson/bug1598-002.phpt
jmikola: > I remember prior art for doing this, but I can also change it to < 8.2.0.

Possible misunderstanding here, but wouldn't `skip_if_php_version('>', '8.2.0');` cause test failures since the test would then be allowed to run on 8.2 pre-release versions? I think it's correct as-is, but was just asking for confirmation.

tests/bson/bson-binary-clone-001.phpt
jmikola: If this test is now restricted to PHP < 8.2, shouldn't we add back the dynamic property testing? This might have been an oversight.

Also, quoting [`version_compare`](https://www.php.net/version_compare):

> Note that pre-release versions, such as 5.3.0-dev, are considered lower than their final release counterparts (like 5.3.0).

So I think we do want to use `<?php skip_if_php_version('>=', '8.1.99'); ?>` as discussed in https://github.com/mongodb/mongo-php-driver/pull/1340#discussion_r934725681

PR #1346, opened 2022-08-16
PHPC-2111: Update ZPP info to forbid explicit null for optional args

alcaeus:
PHPC-2111, also PHPC-2015

While working on the documentation, I noticed that a number of optional arguments were changed to also allow explicit `null` values. This PR undoes this, cleaning up the methods in the process. In some cases this also led to changes in the stub files, but in many cases the stub file already didn't declare the argument nullable, so allowing `null` while parsing parameters was definitely wrong.

### DIFF
--- a/src/BSON/functions.c
+++ b/src/BSON/functions.c
@@ -60,7 +60,7 @@ PHP_FUNCTION(MongoDB_BSON_toPHP)
 	PHONGO_PARSE_PARAMETERS_START(1, 2)
 	Z_PARAM_STRING(data, data_len)
 	Z_PARAM_OPTIONAL
-	Z_PARAM_ARRAY_OR_NULL(typemap)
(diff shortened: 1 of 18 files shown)


jmikola:
@alcaeus and I spoke about this PR and agreed it should be closed. The issue is that we've historically accepted `null` for these optional parameters, which was accepted by our original arg strings in `zend_parse_parameters()` (before we switched to macros). The PHP.net docs never discussed this, and did not reveal that null was accepted.

We agreed the most consistent course of action is to update the docs to note that `null` is accepted, and leave the stubs as-is. In the future, if those stubs end up being used to generate PHP.net docs, we won't have any unexpected changes.

This also means the code changes in this PR to _not_ allow null are unnecessary, hence its closure.

PR #1349, opened 2022-08-23
PHPC-2128: Create stub file for functions

alcaeus:
PHPC-2128

This adds the last stub file needed to no longer have to write arginfo definitions ourselves. However, there's a small catch to it: due to the way gen_stub.php generates the code, the function entries are not prefixed with their namespace (unlike for methods where they are prefixed by the class name, which includes the namespace). This essentially means that as soon was we want to have a function with one name in two separate namespaces, the autogenerated file will no longer work, as shown [in this gist](https://gist.github.com/alcaeus/0fdbaacab8427e95dee4e36b14db7dc5).

I'm fine with this for now and am working to improve gen_stub upstream to fix this issue; however we can also defer this until the upstream script is fixed.


alcaeus:
The naming issue upstream is fixed by https://github.com/php/php-src/pull/9406.

php_phongo.c
```diff
@@ -483,41 +482,6 @@ PHP_FUNCTION(MongoDB_disabled___wakeup) /* {{{ */
 /* }}} */
 
 /* {{{ Extension functions, module dependencies, and module entry */
-ZEND_BEGIN_ARG_INFO_EX(ai_bson_fromPHP, 0, 0, 1)
```
jmikola, line 486: Should you remove "Extension functions" from the comment above?

src/functions.stub.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+/**
+  * @generate-function-entries
```
jmikola, line 4: The asterisk here and on the line below are indented an extra space. Looks like this was done in several other stub files, so you may want to just correct it all now.

Alternatively, this can probably just be a single line like some of the `@param` lines below, since you're not specifying `@generate-class-entries static` here.

jmikola (approved):
Minor suggestions but LGTM.

Is the plan to merge this now and then regenerate after https://github.com/php/php-src/pull/9406 is merged? If so, do you want to create a separate tracking ticket for that?

php_phongo.c
alcaeus: Good point, done.

src/functions.stub.php
alcaeus: Ah, yeah. I fixed those for method comments in the original PR, but forgot the file docblock. I fixed them 👍

alcaeus:
> Is the plan to merge this now and then regenerate after [php/php-src#9406](https://github.com/php/php-src/pull/9406) is merged? If so, do you want to create a separate tracking ticket for that?

If the PR to php-src is merged today, I'd make the changes to avoid circling back to this, but if the PR is delayed I'd merge this down and create a separate ticket for making the changes.

alcaeus:
Merging now since the PR might take a few more days to be merged. For reference, changes to the stub generation script don't automatically cause a regeneration of the arginfo files, as the stub file hash stored in the arginfo file is not dependent on the generation script. Changes to the script will only take effect when the stub file changes, which protects us from these BC breaks happening randomly.

PR #1350, opened 2022-08-26
PHPC-2130: Remove proto comments in favour of stubs

alcaeus:
PHPC-2130

~I've left the folding comments in place for the time being, but depending on the tooling used we may be better off dropping those as well.~

I've also removed the unnecessary emacs folding markers (see second commit)


jmikola (approved):
Admittedly not going to spend my time examining the removed comments in each file. Find letting you merge this once CI is green.

PR #1351, opened 2022-08-30
PHPC-2103: Report loaded crypt_shared version in phpinfo()

jmikola:
https://jira.mongodb.org/browse/PHPC-2103
https://jira.mongodb.org/browse/PHPC-2132

Note: this was rebased after closing https://github.com/mongodb/mongo-php-driver/pull/1352 for [PHPC-2131](https://jira.mongodb.org/browse/PHPC-2131) and merging v1.14 up to master.


jmikola:
Note: the one [task failure](https://spruce.mongodb.com/task/mongo_php_driver_test_mongodb_versions__mongodb_versions~latest_php_edge_versions~latest_stable_os~debian11_test_replicaset_auth_patch_dfa4c20f8b41128d4acd64dc51d8316aeb928695_630e6bc057e85a55b7e01b77_22_08_30_19_57_54/tests?execution=0&sortBy=STATUS&sortDir=ASC) is related to [PHPC-2121](https://jira.mongodb.org/browse/PHPC-2121).

src/LIBMONGOC_VERSION_CURRENT
```diff
@@ -1 +1 @@
-1.22.1
+1.11.1-20220829+git623d659f00
```
alcaeus, line 1: This looks wrong. I don't recall how I used to fix it - the problem either appeared when checking out a specific commit hash or when checking out the master branch. Either way, the version should definitely not be 1.11 ;)

src/phongo_client.c
```diff
@@ -51,6 +51,47 @@ typedef struct {
...
+}
+
+/* Returns the version of the crypt_shared library, or NULL if it's unavailable.
+ * Querying the version requires a mongoc_client_t pointer. Since the shared
+ * library can only be loaded once, any client will return the same result so we
+ * consult the first persistent or request-scoped client we can find.
+ *
+ * Note: this may incorrectly return NULL if crypt_shared was loaded through a
+ * mongoc_client_t since destroyed (e.g. single requested-scoped client);
+ * however, that's the best can do with libmongoc's API. */
```
alcaeus, line 77: If I read this correctly, then we wouldn't report a version if calling something like `php --ri mongodb`, since that never instantiates a client - is that correct? Is there a way to work around this limitation?

tests/manager/manager-ctor-auto_encryption-error-004.phpt
```diff
@@ -25,14 +25,17 @@ foreach ($tests as $autoEncryptionOptions) {
...
 ===DONE===
 <?php exit(0); ?>
---EXPECT--
+--EXPECTF--
 OK: Got MongoDB\Driver\Exception\EncryptionException
 A crypt_shared override path was specified [/not/found], but we failed to open a dynamic library at that location
 
 OK: Got MongoDB\Driver\Exception\EncryptionException
-Option 'cryptSharedLibRequired' is 'true', but failed to load the crypt_shared runtime libary
+Option 'cryptSharedLibRequired' is 'true', but failed to load the crypt_shared runtime lib%r(r?)%rary
```
alcaeus, line 38: Nice way to work around the typo being fixed eventually :)

alcaeus (changes requested):
Version in the libmongoc version file looks off, LGTM otherwise.

src/LIBMONGOC_VERSION_CURRENT
jmikola: This has long been tracked in [CDRIVER-3315](https://jira.mongodb.org/browse/CDRIVER-3315). When you've updated this file in the past, did you manually change `1.11.1` to a particular version?

I've generally ignored this and haven't pushed for it to be fixed upstream since our releases always use actual libmongoc tags.

src/phongo_client.c
jmikola: It goes beyond that. `crypt_shared` is only loaded dynamically as needed, so if the Manager objects are not constructed with `autoEncryption` driver options that would necessitate loading the library, `_mongoc_crypt_get_crypt_shared_version` will return NULL.

We could theoretically try to get around that by constructing an extra client (without `cryptSharedLibRequired` to avoid errors), but there's no guarantee that it would pick up the same library or even a library at all since we don't know what path a user might specify in `cryptSharedLibPath`. We would be limited to searching the system path.

src/LIBMONGOC_VERSION_CURRENT
alcaeus: I think I cheated and updated it manually. Either way, fine to have it as is 👍

src/phongo_client.c
alcaeus: I see. No worries then, and "unknown" is vague enough to cover those cases.

alcaeus (approved):
LGTM

PR #1357, opened 2022-09-07
PHPC-1944: Add final modifier to __set_state methods

levon80999:
https://jira.mongodb.org/browse/PHPC-1944

### DIFF
--- a/src/BSON/Binary.stub.php
+++ b/src/BSON/Binary.stub.php
@@ -69,7 +69,7 @@ final class Binary implements BinaryInterface, \JsonSerializable, Type, \Seriali
 
     final public function getType(): int {}
 
-    public static function __set_state(array $properties): Binary {}
+    final public static function __set_state(array $properties): Binary {}
 
     final public function __toString(): string {}
 
--- a/src/BSON/ObjectId.stub.php
+++ b/src/BSON/ObjectId.stub.php
@@ -15,7 +15,7 @@ final class ObjectId implements ObjectIdInterface, \JsonSerializable, Type, \Ser
 
     public final function __toString(): string {}
 
-    public static function __set_state(array $properties): ObjectId {}
+    final public static function __set_state(array $properties): ObjectId {}
 
     final public function serialize(): string {}
 
--- a/src/BSON/Binary_arginfo.h
+++ b/src/BSON/Binary_arginfo.h
@@ -1,5 +1,5 @@
 /* This is a generated file, edit the .stub.php file instead.
- * Stub hash: fac819b0ca7564fd16a3a7f972a5b3cbee650c04 */
+ * Stub hash: 7b9fd21a206d97f242415b2b2f0fc1ca847524ca */
 
 ZEND_BEGIN_ARG_INFO_EX(arginfo_class_MongoDB_BSON_Binary___construct, 0, 0, 2)
 	ZEND_ARG_TYPE_INFO(0, data, IS_STRING, 0)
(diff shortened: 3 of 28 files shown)


src/BSON/ObjectId.stub.php
```diff
@@ -15,7 +15,7 @@ public final function getTimestamp(): int {}
 
     public final function __toString(): string {}
 
-    public static function __set_state(array $properties): ObjectId {}
+    public final static function __set_state(array $properties): ObjectId {}
```
jmikola, line 18: For consistency, please place the `final` keyword _before_ the method visibility (e.g. `final public static`). I realize many of the stubs are inconsistent about this. I've opened [PHPC-2136](https://jira.mongodb.org/browse/PHPC-2136) to address that.

jmikola (changes requested):
`configure~` should not have been included in this PR. Please remove that file.

Additionally, this PR only updates `ObjectId::__set_state()`. There are many more `__set_state()` methods in the extension that also need to be updated.

src/BSON/ObjectId.stub.php
levon80999: I've removed configure~ and add final to other __set_state() methods and by the way moving final keyword before the method visibility where I noted it but still there are many places where order is not correct

levon80999: Also after generation stubs I noticed that order for key words does not any affect to generated files.

this is the code fragment that add arg infos
`private function getFlagsAsArginfoString(): string
    {
        $flags = "ZEND_ACC_PUBLIC";
        if ($this->flags & Class_::MODIFIER_PROTECTED) {
            $flags = "ZEND_ACC_PROTECTED";
        } elseif ($this->flags & Class_::MODIFIER_PRIVATE) {
            $flags = "ZEND_ACC_PRIVATE";
        }

        if ($this->flags & Class_::MODIFIER_STATIC) {
            $flags .= "|ZEND_ACC_STATIC";
        }

        if ($this->flags & Class_::MODIFIER_FINAL) {
            $flags .= "|ZEND_ACC_FINAL";
        }

        if ($this->flags & Class_::MODIFIER_ABSTRACT) {
            $flags .= "|ZEND_ACC_ABSTRACT";
        }

        if ($this->isDeprecated) {
            $flags .= "|ZEND_ACC_DEPRECATED";
        }

        return $flags;
    }`

jmikola: > there are many places where order is not correct

Indeed, this is why I opened [PHPC-2136](https://jira.mongodb.org/browse/PHPC-2136). That can be addressed independently from this PR.

The order of modifiers has no impact on the generated C code or evaluation by PHP. I just want this addressed so the prototypes throughout our stubs are consistent.

levon80999: If I understand correct there is no more changes. Can we merge this PR and continue work on PHPC-2136 after next triage

src/BSON/Binary.stub.php
```diff
@@ -63,13 +63,13 @@ final class Binary implements BinaryInterface, \JsonSerializable, Type, \Seriali
      */
     public const TYPE_USER_DEFINED = UNKNOWN;
 
-    public final function __construct(string $data, int $type) {}
+    final public function __construct(string $data, int $type) {}
```
jmikola, line 66: This PR should only change `__set_state` methods. Please leave these other corrections to [PHPC-2136](https://jira.mongodb.org/browse/PHPC-2136) (as mentioned in https://github.com/mongodb/mongo-php-driver/pull/1357#discussion_r967541061) so we don't end up with unrelated changes in the final commit.

jmikola (changes requested):

src/BSON/Binary.stub.php
levon80999: Done, please review again

levon80999: There is one CI fail I will fix it and request review again

levon80999: Done

jmikola (approved):

PR #1367, opened 2022-10-05
PHPC-2141: Emit deprecation notice for WriteResult getters with unacknowledged writes

alcaeus:
PHPC-2141

With this change, we're deprecating calling the various `get<field>Count` methods on unacknowledged write results. In 2.0, this will throw an exception, as is already done in the PHP library. For fields missing in the response, we no longer return `null`, but rather `0` to make the method signature more compatible. In 2.0, the return value of these methods will change from `int|null` to `int`.

### DIFF
--- a/src/MongoDB/WriteResult.c
+++ b/src/MongoDB/WriteResult.c
@@ -31,10 +31,17 @@
 #include "MongoDB/WriteError.h"
 #include "WriteResult_arginfo.h"
 
+#define PHONGO_WRITERESULT_CHECK_ACKNOWLEDGED(method)                                                                                                                                        \
+	if (!mongoc_write_concern_is_acknowledged(intern->write_concern)) {                                                                                                                      \
+		php_error_docref(NULL, E_DEPRECATED, "Calling MongoDB\\Driver\\WriteResult::" method "() for an unacknowledged write is deprecated and will throw an exception in ext-mongodb 2.0"); \
+		RETURN_NULL();                                                                                                                                                                       \
+	}
+
 #define PHONGO_WRITERESULT_RETURN_LONG_FROM_BSON_INT32(iter, bson, key)                \
 	if (bson_iter_init_find((iter), (bson), (key)) && BSON_ITER_HOLDS_INT32((iter))) { \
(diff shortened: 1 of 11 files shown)


tests/writeResult/writeresult-getdeletedcount-002.phpt
```diff
@@ -24,6 +24,7 @@ var_dump($result->getDeletedCount());
 ?>
 ===DONE===
 <?php exit(0); ?>
---EXPECT--
+--EXPECTF--
+Deprecated: MongoDB\Driver\WriteResult::getDeletedCount(): Calling MongoDB\Driver\WriteResult::getDeletedCount() for an unacknowledged write is deprecated and will throw an exception in %s
```
jmikola, line 28: This uses `%s` prematurely but still passes. Contrast with the `getInsertedCount()` test.

jmikola (approved):
LGTM with the trailing period removed and updated expectations.

src/MongoDB/WriteResult.c
```diff
@@ -31,10 +31,17 @@
 #include "MongoDB/WriteError.h"
 #include "WriteResult_arginfo.h"
 
+#define PHONGO_WRITERESULT_CHECK_ACKNOWLEDGED(method)                                                                                                                                         \
+	if (!mongoc_write_concern_is_acknowledged(intern->write_concern)) {                                                                                                                       \
+		php_error_docref(NULL, E_DEPRECATED, "Calling MongoDB\\Driver\\WriteResult::" method "() for an unacknowledged write is deprecated and will throw an exception in ext-mongodb 2.0."); \
```
jmikola, line 36: I'd suggest omitting the trailing period here. IIRC, we only do that when a message includes multiple sentences.

tests/writeResult/writeresult-getdeletedcount-002.phpt
alcaeus: Find/replace was a little too eager here. Thanks!

src/MongoDB/WriteResult.c
alcaeus: Removed the trailing period, which hopefully also fixes tests.

PR #1373, opened 2022-10-19
Bump required libmongoc version

alcaeus:
This updates our evergreen config and build scripts to require libmongoc 1.23.0. #1351 introduced the usage of `mongoc_client_get_crypt_shared_version` and updated the bundled libmongoc to a 1.23 version, but the other scripts were not updated. This caused build failures in the `lowest-supported` libmongoc axis due to it still attempting to build with 1.22.1. See the [patch build](https://spruce.mongodb.com/version/634fe0bca4cf47674d8d5a03/tasks).

Note that building against libmongoc master also fails at the moment, but this is unrelated and probably can't be fixed until we update the build file list to account for new files in libmongoc.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1154,9 +1154,9 @@ axes:
     display_name: libmongoc version
     values:
       - id: "lowest-supported"
-        display_name: "1.22.1"
+        display_name: "1.23.0"
         variables:
-          LIBMONGOC_VERSION: "1.22.1"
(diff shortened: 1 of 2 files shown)


jmikola:
This should be addressed in the merge commit for v1.14 into master: https://github.com/mongodb/mongo-php-driver/commit/dadb93fb3420bdc02a3960e1e8460a95a20594dc

PR #1374, opened 2022-10-20
PHPC-2157: Upgrade libmongoc to 1.22.2

jmikola:
https://jira.mongodb.org/browse/PHPC-2157


alcaeus (approved):

jmikola:
Forgot to update the `pkg-config` requirements and Evergreen matrix. Fixed in https://github.com/mongodb/mongo-php-driver/commit/e853c4c0e65847248685a74ea3956222dd69412c.

alcaeus:
> Forgot to update the `pkg-config` requirements and Evergreen matrix. Fixed in [e853c4c](https://github.com/mongodb/mongo-php-driver/commit/e853c4c0e65847248685a74ea3956222dd69412c).

I was on the fence about this - technically the driver would work fine with a lower patch version of libmongoc - similar to how we don't bump the patch version of the `ext-mongodb` dependency in PHPLIB. With new API it's a different story as there will be "undefined symbol" errors, but in these cases we could get away without it.

Either way, updating Evergreen is probably best to ensure builds are stable 👍

jmikola:
You're correct that a libmongoc patch release wouldn't introduce new symbols and thus PHPC would still compile without error. Since we don't currently test with system libraries, all that's required is the submodule bump and updating the Evergreen matrix (for a successful CI build); however, any attempt to use an earlier version of a system _would_ result in CI failures if we started testing with those.

[Update package dependencies](https://github.com/mongodb/mongo-php-driver/blob/master/CONTRIBUTING.md#update-package-dependencies) in the contributing docs don't refer to patch bumps directly, but I'm on board with keeping the `pkg-config` requirement in step with our submodules. This ensures that no user should ever be running a combination of the driver and libmongoc (or libmongocrypt) that fails on something we expect was already fixed.

This might also be relevant to @remicollet if he uses this for his own package requirements.

> similar to how we don't bump the patch version of the ext-mongodb dependency in PHPLIB

It's rare, but we have bumped `ext-mongodb` to a patch version in the past when it was required for testing. See: https://github.com/mongodb/mongo-php-library/commit/10b578ffd5ce67e558e92ce14fb7a799820c12d0#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34L15

remicollet:
IMHO dependency should be on API/ABI, so not on the exact version (patch level) so 1.22

jmikola:
> IMHO dependency should be on API/ABI, so not on the exact version (patch level) so 1.22

I understand the reason behind that, but complicates things when we have tickets like [PHPC-2141](https://jira.mongodb.org/browse/PHPC-2141) where we need to guarantee a certain patch level to ensure that a bug has been fixed (either for PHPC directly or downstream in the library).

remicollet:
Relying on version (patch level) for security purposes seems a bad idea, most enterprise distributions are used to backport critical fixes (mostly security) to old versions.

jmikola:
> Relying on version (patch level) for security purposes seems a bad idea, most enterprise distributions are used to backport critical fixes (mostly security) to old versions.

In this case, libmongocrypt fixed the bug in a patch release (1.5.2). We haven't released PHPC 1.15.0 yet, but it will require libmongocrypt 1.5+ for API and I can think of no reason we'd want to enable someone to build against a version before 1.5.2 if we know they're vulnerable.

I might be misunderstanding something, but I don't follow how backports are related to this the bug was introduced in 1.5.0 and fixed in 1.5.2.

In PHP userland, I think it's commonplace to require patch releases in `composer.json` (e.g. `^1.2.3`) to ensure dependencies with known security vulnerabilities are never utilized.

----

I could understand if we were only utilizing in libmongoc as a shared library, but if the most common usage of the driver is with bundled submodules (via PECL installs). From the POV of a user building against system libs, what benefit is there to allow compiling a new release of the driver with outdated libraries? It makes it more likely they'll end up with an untested driver, which could potentially cause more harm than good.

In this particular case, the pkg-config bump also ensures that users are aware that they need to upgrade libmongoc to pull in the bug fixes for Alpine Linux builds and connecting to Atlas Data Lake, the latter of which _was_ a backport from libmongoc's 1.23.x branch.

IMO, the main benefit of shared library installs is that users are free to user _newer_ versions of libmongoc and libmongocrypt if they wish -- before we get a chance to bump submodules and publish a PHP driver release.

I would never suggest this approach for general dependencies (e.g. OpenSSL, compression libs), but I think libmongoc and libmongocrypt are a special case for usability. That's the entire reason we have this submodule approach (for convenient PECL installs), despite all of the maintenance headache it creates needing to keep Autotools in sync with CMake.

PR #1380, opened 2022-11-11
PHPC-2143: Build Windows DLLs with GitHub Actions

jmikola:
https://jira.mongodb.org/browse/PHPC-2143

### DIFF
--- a/.github/workflows/windows-release-build.yml
+++ b/.github/workflows/windows-release-build.yml
@@ -1,4 +1,4 @@
-name: "Releases"
+name: "Windows Release Build"
 
 on:
   release:
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         # Note: keep this in sync with the Windows matrix in tests.yml
-        php: [ "7.2", "8.2" ] # [ "7.2", "7.3", "8.0", "8.1", "8.2" ]
+        php: [ "7.2", "7.3", "8.0", "8.1", "8.2" ]
         arch: [ x64, x86 ]
         ts: [ ts, nts ]
     runs-on: ubuntu-latest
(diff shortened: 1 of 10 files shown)


jmikola:
One more test after cleaning up the workflow files (e.g. removing temporary skips) and changing the release workflow script to use the release tag instead of grepping `phongo_version.h`:

 * [1.15.0alpha3 release](https://github.com/jmikola/mongo-php-driver/releases/tag/1.15.0alpha3)
 * [Original CI build](https://github.com/jmikola/mongo-php-driver/actions/runs/3513019998)
 * [Release workflow](https://github.com/jmikola/mongo-php-driver/actions/runs/3513138040]

Note how the attachments are named "1.15.0alpha3" instead of "1.15.0-dev" as in the previous release.

If we end up doing full release automation, grepping `phongo_version.h` would make sense as a verification check. For this purpose, I don't think it was necessary and the workflow can be made simpler without it.

alcaeus (commented):
Changes LGTM - I didn't expect the publishing of artifacts to look this simple, this is definitely a first step toward some more release automation!

Should the CONTRIBUTING docs be updated to reflect changes necessary for this automation to work, e.g. creating the tag and only publishing a release once the tag build has completed?

jmikola:
> Should the CONTRIBUTING docs be updated to reflect changes necessary for this automation to work, e.g. creating the tag and only publishing a release once the tag build has completed?

Done in the latest commits.

alcaeus (approved):
Great work, excited to see this in action! 🎉

.github/workflows/windows-release-build.yml
```diff
@@ -0,0 +1,37 @@
...
+  release:
+    types: [ published ]
+
+jobs:
+  windows-release-build:
+    strategy:
+      fail-fast: false
+      matrix:
+        # Note: keep this in sync with the Windows matrix in tests.yml
+        php: [ "7.2", "7.3", "8.0", "8.1", "8.2" ]
```
jmikola, line 13: @mickverm: This would explain the missing PHP 7.4 builds.

@alcaeus: Can you make a ticket for this? We're missing 7.4 both here and in `tests.yml`.

I don't think it's possible to modify the action configs and re-run for 1.15.0, since the tag includes the config file omitting 7.4. Short of creating a 1.15.1 release, perhaps you can manually create 7.4 archives on your own fork (e.g. rewrite git history to create the 1.15.0 tag anew) and then attach those to the [1.15.0 release](https://github.com/mongodb/mongo-php-driver/releases/tag/1.15.0).

alcaeus: Created PHPC-2185 to track this. I'll tag a 1.15.1 release soon to also include a fix for PHPC-2184, which will then provide 7.4 builds for the 1.15 branch.

alcaeus:
@mickverm 1.15.1 has just been released, including DLLs for PHP 7.4. You may find them attached to the release: https://github.com/mongodb/mongo-php-driver/releases/tag/1.15.1

PR #1383, opened 2022-11-15
PHPC-2168: Use consistent int types for APM fields and snprintf

jmikola:
https://jira.mongodb.org/browse/PHPC-2168

The buffer for snprintf is slightly larger than necessary (21 bytes), but this is consistent with the buffer size we use elsewhere.


alcaeus (approved):
LGTM. I was surprised to see that the duration is stored as int64_t instead of uint64_t, but no reason to deviate from the base types 👍

PR #1384, opened 2022-11-23
Revise release instructions

jmikola:
Push "Package X.Y.Z" and "Back to -dev" commits separately to ensure Windows build artifacts are created for the tag.

Add instructions for creating a maintenance branch after releasing a new minor version.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -281,8 +281,18 @@ generation script.
 
 ## Releasing
 
-The follow steps outline the release process for a maintenance branch (e.g.
-releasing the `vX.Y` branch as X.Y.Z).
+The follow steps outline the release process for both new minor versions (e.g.
+releasing the `master` branch as X.Y.0) and patch versions (e.g. releasing the
+`vX.Y` branch as X.Y.Z).
+
+The command examples below assume that the canonical "mongodb" repository has
+the remote name "mongodb". You may need to adjust these commands if you've given
+the remote another name (e.g. "upstream"). The "origin" remote name was not used
+as it likely refers to your personal fork.
+
+It helps to keep your own fork in sync with the "mongodb" repository (i.e. any
+branches and tags on the main repository should also exist in your fork). This
+is left as an exercise to the reader.
 
 ### Transition JIRA issues and version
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -281,8 +281,18 @@ generation script.
...
-The follow steps outline the release process for a maintenance branch (e.g.
-releasing the `vX.Y` branch as X.Y.Z).
+The follow steps outline the release process for both new minor versions (e.g.
+releasing the `master` branch as X.Y.0) and patch versions (e.g. releasing the
+`vX.Y` branch as X.Y.Z).
+
+The command examples below assume that the canonical "mongodb" repository has
+the remote name "mongodb". You may need to adjust these commands if you've given
+the remote another name (e.g. "upstream"). The "origin" remote name was not used
+as it likely refers to your personal fork.
```
jmikola, line 291: I imagine most development environments are clones of a personal fork, so it's unlikely that "origin" (the default) would be correct. In any event, specifying a remote name in the examples adds clarity.

CONTRIBUTING.md
```diff
@@ -281,8 +281,18 @@ generation script.
...
+
+It helps to keep your own fork in sync with the "mongodb" repository (i.e. any
+branches and tags on the main repository should also exist in your fork). This
+is left as an exercise to the reader.
```
jmikola, line 295: @alcaeus: I dunno if I've shared this with you, but this is what I use locally:

```sh
#!/bin/sh

current=$(git rev-parse --abbrev-ref HEAD)

git fetch mongodb
git fetch origin

for branch in master v1.14 v1.15; do
  git checkout $branch && git pull mongodb $branch && git push origin $branch
done

git checkout $current
```

Dunno if you think it's worth documenting elsewhere.

CONTRIBUTING.md
```diff
@@ -352,46 +362,47 @@ After copying release notes, use `make package` to create the package file (e.g.
...
 
 ```
 $ git add phongo_version.h
 $ git commit -m "Package X.Y.Z"
+$ git push mongodb
 ```
 
-### Tag release
+> **Note:** Pushing this commit independently from the subsequent "Back to -dev"
+> commit will ensure that Windows build artifacts are created for the release.
```
jmikola, line 371: GitHub Actions are smart enough to not build individual commits if multiple are pushed at once. The original example we had where "Package X.Y.Z" _and_ "Back to -dev" were pushed simultaneously would have meant we never generate build artifacts for the actual tag.

This is clunky and definitely something we could improve down the line if we stop depending on build artifacts from a previous CI run and instead build DLLs on-demand for releases. We could also avoid debug builds and PDB file generation for typical CI runs (I imagine that's a bit slower than without).

CONTRIBUTING.md
```diff
@@ -400,44 +406,72 @@ $ git commit -m "Back to -dev" phongo_version.h
...
+$ git pull mongodb master
 $ git merge vX.Y --strategy=ours
-$ git push
+$ git push mongodb
 ```
 
 The `--strategy=ours` option ensures that all changes from the merged commits
-will be ignored.
+will be ignored. This is OK because previously ensured that the `master` branch
+was up-to-date with all code changes in this maintenance branch before tagging.
```
jmikola, line 474: Thought it'd be helpful to remind readers why this is OK. Technically, we're just ignoring the `phongo_version.h` changes here.

CONTRIBUTING.md
```diff
@@ -281,8 +281,18 @@ generation script.
 
 ## Releasing
 
-The follow steps outline the release process for a maintenance branch (e.g.
-releasing the `vX.Y` branch as X.Y.Z).
+The follow steps outline the release process for both new minor versions (e.g.
```
alcaeus, line 284:
```suggestion
The following steps outline the release process for both new minor versions (e.g.
```

CONTRIBUTING.md
```diff
@@ -312,18 +322,13 @@ Update the version and stability constants in `phongo_version.h`. This should
...
-#define PHP_MONGODB_VERSION "1.1.8"
-#define PHP_MONGODB_STABILITY "stable"
-#define PHP_MONGODB_VERSION_DESC 1,1,8,1
+```diff
+-#define PHP_MONGODB_VERSION "1.1.8-dev"
+-#define PHP_MONGODB_STABILITY "devel"
+-#define PHP_MONGODB_VERSION_DESC 1,1,8,0
++#define PHP_MONGODB_VERSION "1.1.8"
++#define PHP_MONGODB_STABILITY "stable"
++#define PHP_MONGODB_VERSION_DESC 1,1,8,1
```
alcaeus, line 331: It may be worth mentioning that for going from `x.y.z-dev` to `x.y.z` the last digit in this number needs to be increased by 1, not reset to `0`. I've been confused by this before, especially during my first releases.

CONTRIBUTING.md
alcaeus: There is the option of pushing them at the same time, then pushing the tag separately and ensuring the tag is built. The workflow syntax allows you to specify tag filters to build on a push: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches-and-tags.

If we're updating the process down the line, we could change the process to build tags and create a draft release for them, attaching build artifacts in the process. We could then revise the release notes separately. For now I don't think we need to change this, as I'm sure we'll be working on automating releases (including PECL packages) in the near future.

alcaeus (approved):
LGTM with a few optional suggestions.

CONTRIBUTING.md
jmikola: Isn't that stated in the preceding paragraph?

> This should entail removing the version's "-dev" suffix, changing the stability to "stable", and _increasing the last digit for `PHP_MONGO_VERSION_DESC`_:

CONTRIBUTING.md
jmikola: Noted. I didn't consider that since we currently only have branch filters in place.

If we revisit release automation down the line I'd definitely suggest writing up a design doc first so we can plan out exactly what we want to implement.

CONTRIBUTING.md
alcaeus: I may need new glasses. Thanks for pointing this out!

CONTRIBUTING.md
alcaeus: I agree, that change exceeds the scope of getting this to work. I suggest putting release automation on our quarterly plan and designing this ahead of time 👍

PR #1395, opened 2023-01-04
PHPC-2180: Allow applications to register a log message subscriber

jmikola:
https://jira.mongodb.org/browse/PHPC-2180
https://jira.mongodb.org/browse/PHPC-2289


src/functions.stub.php
```diff
@@ -26,6 +26,16 @@ function toPHP(string $bson, ?array $typemap = null) {}
     function toRelaxedExtendedJSON(string $bson): string {}
 }
 
+namespace MongoDB\Driver\Logging {
+    function addLogger(Logger $logger): void {}
+
+    function levelToString(int $level): string {}
+
+    function log(int $level, string $message): void {}
```
alcaeus, line 34: Why do we expose a `log` function? Is this to allow PHPLIB to add its own log messages to the registered logger? If so, I think we should mark this as `@internal` and document it as such.

GromNaN: I see it's used in tests.
I'd rather transfer PHPC log messages to a PSR logger in PHPLIB than having PHPLIB sending message to PHPC logger. It allows more flexibility like adding the `$context` with the current client.

src/MongoDB/Logging/functions.c
```diff
@@ -0,0 +1,85 @@
...
+	PHONGO_PARSE_PARAMETERS_END();
+
+	/* TODO: throw if level is invalid */
+
+	if (strlen(message) != message_len) {
+		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "Log messages cannot contain null bytes. Unexpected null byte after \"%s\".", message);
+		return;
+	}
+
+	mongoc_log(level, PHONGO_LOG_DOMAIN, "%s", message);
```
GromNaN, line 59: Not yet a C expert like you, but I think the `mongoc_log` function accepts extra arguments. That looks like the [PSR-3 log functions](https://www.php-fig.org/psr/psr-3/#3-psrlogloggerinterface) that accept a `$context` that is injected into the message.

https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-log.c#L115C1-L115C1

I would be nice if `MongoDB\Driver\Logging\Logger::log` could receive this raw `format` and `args`. Log systems that accept objects are becoming increasingly common (I'm thinking of kibana, which lets you filter on log fields).

Also, would it be possible to pass the manager (or a Client identifier) that triggered the log message?

jmikola: > I think the mongoc_log function accepts extra arguments.

It does, but the variable args are only used for the format string (see: [Logging](
https://mongoc.org/libmongoc/current/logging.html) in the libmongoc docs). I'd rather not expose that in in PHPC when userland can filter messages through [`sprintf()`](https://www.php.net/sprintf) and get much better error reporting.

If you were referring to `gLogData` in the C source, that's an arbitrary pointer provided when registering a handler, which libmongoc then passes to the handler function with each invocation. That wouldn't be suitable for 

> Log systems that accept objects are becoming increasingly common

I think that's what [structured logging](https://github.com/mongodb/specifications/blob/master/source/logging/logging.rst#structured-versus-unstructured-logging) in the driver spec refers to, but it's not supported in libmongoc. And I don't think we need it ourselves, since PHPC and PHPLIB will likely only use logging for messages mandated by our own specifications.

> Also, would it be possible to pass the manager (or a Client identifier) that triggered the log message?

Logging in libmongoc is fundamentally disconnected from `mongoc_client_t` and only exists at a global level. Some drivers include a topology ID (from SDAM) in logged messages to address that; however, libmongoc does not (yet) do so.

src/functions.stub.php
jmikola: > Is this to allow PHPLIB to add its own log messages to the registered logger?

Correct. I suppose it _could_ be used by an application using PHPC directly to inject messages into libmongoc's log output (e.g. `mongodb.debug`), but there's no mandate for PHPC to actually provide such an API to users.

> I'd rather transfer PHPC log messages to a PSR logger in PHPLIB than having PHPLIB sending message to PHPC logger. It allows more flexibility like adding the $context with the current client.

My initial thought was that we'd need to route log messages through a single place in order to test spec compliance, and I suppose that's still true, but there's also no reason that "single place" cannot be PHPLIB instead of PHPC.

I'll consider removing this before merging, but let's leave it in place pending the PHPLIB POC.

src/MongoDB/Logging/Logger.stub.php
```diff
@@ -0,0 +1,55 @@
...
+
+namespace MongoDB\Driver\Logging;
+
+interface Logger
+{
+    /**
+     * @var int
+     * @cvalue MONGOC_LOG_LEVEL_ERROR
+     */
+    public const LEVEL_ERROR = UNKNOWN;
```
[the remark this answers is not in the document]
jmikola: > a provision to ensure that the value defined here matches the one extracted from the constant found in `@cvalue`

I think we discussed this over video. Can you remind me what the "provision" is? Does `gen_stub.php` generate code that asserts consistency between the `@cvalue` and whatever we hard-code here?

src/functions.stub.php
```diff
@@ -26,6 +26,16 @@ function toPHP(string $bson, ?array $typemap = null) {}
     function toRelaxedExtendedJSON(string $bson): string {}
 }
 
+namespace MongoDB\Driver\Logging {
+    function addLogger(Logger $logger): void {}
```
jmikola, line 30: Realistically, this is also going to need an optional integer that specifies the minimum logging level so that we can default that to `MONGOC_LOG_LEVEL_DEBUG` and suppress trace-level logs by default. This will also allow us to preserve the optimization whereby PHPC disables trace logging unless absolutely necessary (previously only if `mongodb.debug` was active).

src/phongo_ini.c
```diff
@@ -14,96 +14,82 @@
...
+	int          fd = -1;
+	char*        prefix;
+	int          len;
+	zend_string* filename = NULL;
+	FILE*        stream   = NULL;
 
-	bson_gettimeofday(&tv);
-	t  = tv.tv_sec;
-	tu = tv.tv_usec;
+	/* TODO: consider using time() retval directly instead of time_t struct */
```
jmikola, line 37: I do intend to come back to this before merging. I just want to double-check with the C team that I'm not inviting compatibility issues by moving away from `bson_gettimeofday()`. One would _hope_ that [time(2)](https://man7.org/linux/man-pages/man2/time.2.html) is old enough to rely on across platforms.

src/phongo_ini.c
```diff
@@ -14,96 +14,82 @@
...
+	time(&t);
+	len = spprintf(&prefix, 0, "PHONGO-%ld", t);
 
-	dt = php_format_date((char*) ZEND_STRL("Y-m-d\\TH:i:s"), t, 0);
+	/* TODO: Refactor this to use fdopen (see: PHPC-2181) */
```
jmikola, line 41: This was discovered while working on this PR in any earlier phase and I created [PHPC-2181](https://jira.mongodb.org/browse/PHPC-2181) to defer this task since it's a bit of a rabbit hole and we've been using `php_open_temporary_fd()` for a very long time already.

src/phongo_log.c
```diff
@@ -0,0 +1,193 @@
...
+	}
+}
+
+/* Sets or unsets our libmongoc handler according to whether logging is enabled
+ * (i.e. there is a stream or the loggers HashTable is not empty). This should
+ * be called each time after updating the stream or logger HashTable. */
+static void phongo_log_sync_handler(void)
+{
+	if (MONGODB_G(debug_fd) || (MONGODB_G(loggers) && zend_hash_num_elements(MONGODB_G(loggers)) > 0)) {
+		mongoc_log_trace_enable();
```
jmikola, line 102: This is the optimization I was referring to above. It's responsible for setting a `gLogTrace` global that libmongoc uses to early-return in `mongoc_log()`.

Note: this is different than the optimization proposed in [CDRIVER-4697](https://jira.mongodb.org/browse/CDRIVER-4697), although related.

src/MongoDB/Logging/Logger.stub.php
```diff
@@ -0,0 +1,55 @@
...
+     */
+    public const LEVEL_DEBUG = UNKNOWN;
+
+    /**
+     * @var int
+     * @cvalue MONGOC_LOG_LEVEL_TRACE
+     */
+    public const LEVEL_TRACE = UNKNOWN;
+
+    public function log(int $level, string $domain, string $message): void;
```
jmikola, line 54: I'm not entirely happy with this API, as I think most would expect the `log()` method of a `Logger` interface to be used to _write_ log messages, not receive them.

And then I considered renaming this class to `LogSubscriber` and found myself asking why it's in a separate namespace to begin with.

What do you think about the following:

 * Rename `MongoDB\Logging\Logger` to `MongoDB\Monitoring\LogSubscriber` (leave its `log()` method as-is).
 * `MongoDB\Monitoring\LogSubscriber` extends the Subscriber interface, like the APM and SDAM listeners.
 * Allow a log listener to be registered through the existing `addSubscriber()` methods. LogSubscriber would then get treated differently from the other two subscriber interfaces.
 * We could choose to prohibit passing a LogSubscriber to `Manager::addSubscriber()` and force users to add them with `MongoDB\Driver\Monitoring\addSubscriber()` because libmongoc only has a global logger. If that ever changes and we can implement client-level logging atop libmongoc, we can then allow `Manager::addSubscriber()` to take a LogSubscriber. Basically, this would avoid a potential BC issue.

Logging and monitoring are already intertwined (the APM spec was renamed to [Command Logging and Monitoring](https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/command-logging-and-monitoring.rst), so I don't think this would be too strange a proposal. It would also let us avoid expanding our API/docs footprint.

src/MongoDB/Logging/Logger.stub.php
```diff
@@ -0,0 +1,55 @@
...
+     * @var int
+     * @cvalue MONGOC_LOG_LEVEL_WARNING
+     */
+    public const LEVEL_WARNING = UNKNOWN;
+
+    /**
+     * @var int
+     * @cvalue MONGOC_LOG_LEVEL_MESSAGE
+     */
+    public const LEVEL_MESSAGE = UNKNOWN;
```
jmikola, line 34: Are we OK with adding these constants on the Logger interface, or would it make sense to push these to their own interface similar to [Psr\Log\LogLevel](https://www.php-fig.org/psr/psr-3/#5-psrlogloglevel)?

I could go either way, but lean towards the side of consolidating our API/docs footprint (i.e. less classes).

src/MongoDB/Logging/Logger.stub.php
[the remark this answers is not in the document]
jmikola: @alcaeus: :point_up:

src/functions.stub.php
jmikola: Note: I've removed `levelToString()` but kept this for now until I have a suitable workaround in place within PHPLIB. I think that might require a separate method on the PsrLogAdapter class (in https://github.com/mongodb/mongo-php-library/pull/1158). We can't use the name `log()` there, as it'd conflict with the `MongoDB\Driver\Logging\Logger::log()` method for _receiving_ a log message; however, it would also be `@internal` to PHPLIB so we could really pick any name we like for it (e.g. `writeLog()`).

The additional benefit of that approach is that we could use custom domains within PHPLIB instead of requiring PHPC to fill in its own "php" domain.

src/MongoDB/Logging/functions.c
```diff
@@ -0,0 +1,82 @@
...
+#include "bson/bson.h"
+#include "mongoc/mongoc.h"
+
+#include <php.h>
+
+#include "php_phongo.h"
+#include "phongo_error.h"
+#include "phongo_log.h"
+
+#define PHONGO_LOG_DOMAIN "php"
```
jmikola, line 26: Note: if the `log()` function is relocated to PHPLIB this can be removed. See https://github.com/mongodb/mongo-php-driver/pull/1395#discussion_r1324787456 for more context.

src/MongoDB/Logging/functions.c
jmikola: @GromNaN: OK to resolve this?

GromNaN: Yes, thanks.

src/MongoDB/Logging/Logger.stub.php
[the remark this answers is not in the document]
alcaeus: Sorry for missing this. This is best explained with a diff. In this case, I added values for `TYPE_GENERIC` and `TYPE_FUNCTION`, with the latter being wrong on purpose. The stub and generated arginfo changes are as follows:
```diff
diff --git a/src/BSON/Binary.stub.php b/src/BSON/Binary.stub.php
index 6e67401f..978d7ab4 100644
--- a/src/BSON/Binary.stub.php
+++ b/src/BSON/Binary.stub.php
@@ -13,13 +13,13 @@ final class Binary implements BinaryInterface, \JsonSerializable, Type, \Seriali
      * @var int
      * @cvalue BSON_SUBTYPE_BINARY
      */
-    public const TYPE_GENERIC = UNKNOWN;
+    public const TYPE_GENERIC = 0x00;
 
     /**
      * @var int
      * @cvalue BSON_SUBTYPE_FUNCTION
      */
-    public const TYPE_FUNCTION = UNKNOWN;
+    public const TYPE_FUNCTION = 0x02; // intentionally wrong
 
     /**
      * @var int
diff --git a/src/BSON/Binary_arginfo.h b/src/BSON/Binary_arginfo.h
index 5808000b..a6688841 100644
--- a/src/BSON/Binary_arginfo.h
+++ b/src/BSON/Binary_arginfo.h
@@ -1,5 +1,5 @@
 /* This is a generated file, edit the .stub.php file instead.
- * Stub hash: 3ecb4933da21ef2b58ca95d74e2aac8d26ca6dc6 */
+ * Stub hash: b2a2108c037f062e9df833f6250c557ab22cd5a7 */
 
 ZEND_BEGIN_ARG_INFO_EX(arginfo_class_MongoDB_BSON_Binary___construct, 0, 0, 1)
        ZEND_ARG_TYPE_INFO(0, data, IS_STRING, 0)
@@ -110,12 +110,14 @@ static zend_class_entry *register_class_MongoDB_BSON_Binary(zend_class_entry *cl
        zend_string *const_TYPE_GENERIC_name = zend_string_init_interned("TYPE_GENERIC", sizeof("TYPE_GENERIC") - 1, 1);
        zend_declare_class_constant_ex(class_entry, const_TYPE_GENERIC_name, &const_TYPE_GENERIC_value, ZEND_ACC_PUBLIC, NULL);
        zend_string_release(const_TYPE_GENERIC_name);
+       ZEND_ASSERT(BSON_SUBTYPE_BINARY == 0);
 
        zval const_TYPE_FUNCTION_value;
        ZVAL_LONG(&const_TYPE_FUNCTION_value, BSON_SUBTYPE_FUNCTION);
        zend_string *const_TYPE_FUNCTION_name = zend_string_init_interned("TYPE_FUNCTION", sizeof("TYPE_FUNCTION") - 1, 1);
        zend_declare_class_constant_ex(class_entry, const_TYPE_FUNCTION_name, &const_TYPE_FUNCTION_value, ZEND_ACC_PUBLIC, NULL);
        zend_string_release(const_TYPE_FUNCTION_name);
+       ZEND_ASSERT(BSON_SUBTYPE_FUNCTION == 2);
 
        zval const_TYPE_OLD_BINARY_value;
        ZVAL_LONG(&const_TYPE_OLD_BINARY_value, BSON_SUBTYPE_BINARY_DEPRECATED);
```

On a debug build of PHP, `ZEND_ASSERT` defers to `assert`, resulting in the following output when running tests:
> Assertion failed: (BSON_SUBTYPE_FUNCTION == 2), function register_class_MongoDB_BSON_Binary, file Binary_arginfo.h, line 120.

src/MongoDB/Logging/Logger.stub.php
alcaeus: Since we have decided to not have a `$minLogLevel`, I think the last reason to not make it a `LogSubscriber` went out the door. IMO this sounds like a good plan.

src/MongoDB/Logging/Logger.stub.php
alcaeus: I think these can stay in as constants, to be deprecated when we bump the PHP version enough to create a proper `enum` for the log levels.

src/functions.stub.php
alcaeus: `writeLog` sounds like a good idea, and it'd make more sense to have this in PHPLIB than opening up the entire libmongoc logging infrastructure.

jmikola: https://github.com/mongodb/mongo-php-library/pull/1158 now has an internal, static `writeLog()` method that can be used by PHPLIB. It has the same API as PHPC's `Logger::log()` method, so we can specify different domains as needed. It currently uses libmongoc domains, but I suppose we can always change the internal API to use [spec-derived levels](https://github.com/mongodb/specifications/blob/master/source/logging/logging.rst#log-severity-levels) if needed (or PSR levels directly).

src/MongoDB/Logging/Logger.stub.php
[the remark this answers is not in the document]
jmikola: I'll create a PHPC ticket to replace this across all of PHPC.

src/functions.stub.php
jmikola: No longer needed because we decided to suppress trace-level logs from registered Logger instances.

src/MongoDB/Logging/Logger.stub.php
jmikola: SGTM. I'll be very pleased with that API.

And we'll have `Manager::addSubscriber()` throw for now, instead of introducing an intermediary interface to distinguish global and Manager subscribers (see [Slack thread](https://mongodb.slack.com/archives/C026UA4GF0B/p1694619954582389?thread_ts=1694618364.328659&cid=C026UA4GF0B)).

src/MongoDB/Logging/Logger.stub.php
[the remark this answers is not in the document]
jmikola: Created https://jira.mongodb.org/browse/PHPC-2292 but addressing this for LogSubscriber immediately in this PR.

src/functions.stub.php
jmikola: I decided to leave an internal `MongoDB\Driver\Monitoring\mongoc_log()` function in place for internal testing, but we won't document it.

I did add `@internal` in the stub, which admittedly has no tangible effect. I considered going a step further an adding preprocessor directives to only define the function under certain circumstances, but I realized it would only complicate things for Windows since we create a single build for testing _and_ releases. I've opened PHPC-2293 to revisit that at a later point (if we care).

src/MongoDB/Logging/functions.c
jmikola: This was removed because the `mongoc_log()` internal function takes a `domain` variable.

src/MongoDB/Logging/Logger.stub.php
jmikola: This is now implemented, and I also added tests for registering a single object that is both an APM _and_ log subscriber.

src/phongo_ini.c
jmikola: Replaced this with `time(NULL)`, which is already used in PHP core: https://github.com/php/php-src/blob/php-8.2.10/ext/date/php_date.c#L70

tests/logging/logging-addSubscriber-005.phpt
```diff
@@ -0,0 +1,51 @@
+--TEST--
+MongoDB\Driver\Monitoring\addSubscriber(): Adding a combined CommandSubscriber and LogSubscriber
```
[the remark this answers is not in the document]
jmikola: Good question. This seems very much related to [PHPC-2289](https://jira.mongodb.org/browse/PHPC-2289), but I had yet to consider objects implementing both interfaces when I opened that issue. The test you're commenting on here came later.

This is rare enough that I'd be in favor of failing outright. The whole point of the exception is to inform the user that something is unsupported. I don't expect anyone to actually implement both interfaces like this, and merely created the test to cover the hypothetical edge case and document our behavior.

I'll leave a comment on the issue above to clarify that we should throw for unions.

tests/logging/logging-addSubscriber-004.phpt
```diff
@@ -0,0 +1,38 @@
...
+    public function log(int $level, string $domain, string $message): void
+    {
+        printf("%d: %s: %s\n", $level, $domain, $message);
+    }
+}
+
+$logger = new MyLogger;
+addSubscriber($logger);
+
+ini_set('mongodb.debug', 'stdout');
```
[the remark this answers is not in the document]
jmikola: It's possible, but would require an additional wildcard to match all of the log output when the extension is initialized. Manually toggling `mongodb.debug` within the test lets me focus on just the `mongoc_log()` calls below.

alcaeus (approved):

GromNaN (approved):

PR #1408, opened 2023-04-05
PHPC-2200: Add MongoDB rapid releases to Evergreen matrix

jmikola:
https://jira.mongodb.org/browse/PHPC-2200

Related to https://github.com/mongodb/mongo-php-library/pull/1056

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1240,24 +1240,30 @@ buildvariants:
   tasks:
     - name: "test-atlas"
 
-# OCSP is available from MongoDB 4.4+ (Debian 10 has MongoDB 4.4+)
 - matrix_name: "test-ocsp"
-  matrix_spec: { "os": "debian10", "mongodb-versions": ["4.4", "5.0", "6.0"], "php-edge-versions": "latest-stable" }
+  matrix_spec: { "os": "debian10", "mongodb-versions": "*", "php-edge-versions": "latest-stable" }
   display_name: "OCSP tests - ${mongodb-versions}"
+  exclude_spec:
+    # OCSP is available from MongoDB 4.4+ (Debian 10 has MongoDB 4.4+)
+    - { "os": "debian10", "mongodb-versions": ["3.6", "4.0", "4.2"], "php-edge-versions": "latest-stable" }
   tasks:
     - name: ".ocsp"
 
-# Stable API is available from MongoDB 5.0+
 - matrix_name: "test-requireApiVersion"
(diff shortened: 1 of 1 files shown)


.evergreen/config.yml
```diff
@@ -1035,43 +1035,47 @@ axes:
...
 
   - id: mongodb-edge-versions
     display_name: MongoDB Version
     values:
       - id: "latest-stable"
-        display_name: "MongoDB 5.0"
+        display_name: "MongoDB 6.0"
         variables:
-          VERSION: "5.0"
+          MONGODB_VERSION: "6.0"
```
jmikola, line 1074: Thoughts on using "rapid" here?

See also: https://github.com/mongodb/mongo-php-library/pull/1056#discussion_r1158056586

.evergreen/config.yml
```diff
@@ -1236,24 +1240,30 @@ buildvariants:
...
     - name: "test-requireApiVersion"
 
-# Load balancer is available from MongoDB 5.0+
 - matrix_name: "test-loadBalanced"
-  matrix_spec: { "os": "debian11", "mongodb-versions": ["5.0", "6.0"], "php-edge-versions": "latest-stable" }
+  matrix_spec: { "os": "debian11", "mongodb-versions": "*", "php-edge-versions": "latest-stable" }
   display_name: "Load balanced - ${mongodb-versions}"
+  exclude_spec:
+    # Load balancer is available from MongoDB 5.0+
+    - { "os": "debian11", "mongodb-versions": ["3.6", "4.0", "4.2", "4.4"], "php-edge-versions": "latest-stable" }
```
jmikola, line 1266: The `test-csfle-skip_crypt_shared` matrix below currently use 6.0 exclusively. Let me know if you think it makes sense to test on 6.0+ using an `exclude_spec` similar to what we have here.

See also: https://github.com/mongodb/mongo-php-library/pull/1056#discussion_r1158070106

.evergreen/config.yml
alcaeus: SGTM 👍

.evergreen/config.yml
alcaeus: SGTM 👍

.evergreen/config.yml
jmikola: Quoting @JamesKovacs from Slack:

> When M.0.0 goes GA, rapid points to (M-1).3.x for a quarter, which isn't useful since no one can run that quarterly release

With that in mind, I think keeping 6.0 makes sense here and we can just bump to 7.0 when that gets released.

PR #1409, opened 2023-04-06
PHPC-2197: Support queryable encryption range indexes

jmikola:
https://jira.mongodb.org/browse/PHPC-2197

### DIFF
--- a/src/MongoDB/ClientEncryption.c
+++ b/src/MongoDB/ClientEncryption.c
@@ -1012,6 +1012,10 @@ static void phongo_clientencryption_encrypt(php_phongo_clientencryption_t* clien
 
 	php_phongo_zval_to_bson_value(zvalue, PHONGO_BSON_NONE, &value);
 
+	if (EG(exception)) {
+		goto cleanup;
+	}
+
 	opts = phongo_clientencryption_encrypt_opts_from_zval(options);
 
 	if (!opts) {
@@ -1084,6 +1088,10 @@ static void phongo_clientencryption_decrypt(php_phongo_clientencryption_t* clien
 	bson_error_t error      = { 0 };
 
 	php_phongo_zval_to_bson_value(zciphertext, PHONGO_BSON_NONE, &ciphertext);
+
+	if (EG(exception)) {
+		goto cleanup;
+	}
 
 	if (!mongoc_client_encryption_decrypt(clientencryption->client_encryption, &ciphertext, &value, &error)) {
 		phongo_throw_exception_from_bson_error_t(&error);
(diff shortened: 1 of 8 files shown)


tests/clientEncryption/clientEncryption-encryptExpression-001.phpt
```diff
@@ -0,0 +1,54 @@
...
+    ],
+];
+
+$encryptedExpr = $clientEncryption->encryptExpression($expr, $encryptOpts);
+
+var_dump($encryptedExpr->{'$and'}[0]->encryptedInt->{'$gte'} instanceof MongoDB\BSON\Binary);
+var_dump($encryptedExpr->{'$and'}[0]->encryptedInt->{'$gte'}->getType() === MongoDB\BSON\Binary::TYPE_ENCRYPTED);
+
+var_dump($encryptedExpr->{'$and'}[1]->encryptedInt->{'$lte'} instanceof MongoDB\BSON\Binary);
+var_dump($encryptedExpr->{'$and'}[1]->encryptedInt->{'$lte'}->getType() === MongoDB\BSON\Binary::TYPE_ENCRYPTED);
```
jmikola, line 44: Note: dumping the entire `$encryptedExpr` object and asserting the contents of both Binary objects with `%a` (see: [EXPECTF](https://qa.php.net/phpt_details.php#expectf_section)) was unreliable for the same reasons as Session assertions. See https://github.com/mongodb/mongo-php-driver/commit/2104d6dff3a81fd79b00d00ce03eaf272501a05f for more context.

src/MongoDB/ClientEncryption.c
```diff
@@ -943,6 +1089,10 @@ static void phongo_clientencryption_decrypt(php_phongo_clientencryption_t* clien
 
 	php_phongo_zval_to_bson_value(zciphertext, PHONGO_BSON_NONE, &ciphertext);
 
+	if (EG(exception)) {
+		goto cleanup;
```
jmikola, line 1093: This is added for consistency but we don't actually test for this since `decrypt()` expects a Binary and that is always going to successfully encode to BSON.

This is also mentioned in the commit message.

GromNaN: You mean it's inaccessible code, intentionally added? That seems strange to me. A comment in the code would be more appropriate.

jmikola: See https://github.com/mongodb/mongo-php-driver/pull/1412#discussion_r1206321486 for a recent discussion about this very issue in @alcaeus' PR.

I think ideally we'd use assertions here, as is sometimes done in libmongoc; however, there's no prior art in PHPC for doing so.

I created [PHPC-2235](https://jira.mongodb.org/browse/PHPC-2235) to track this, but I'd prefer to leave this in place for now since it'd be easily detected by tooling should we end up integrating that.

GromNaN (approved):

PR #1410, opened 2023-04-11
Add missing void arg to fix MacOS builds

alcaeus:
Upgraded to MacOS 13.3 recently and building the driver exposed this failure. Not sure why this didn't appear earlier as `strict-prototypes` was part of our developer build flags for a long time already.

```
src/phongo_client.c:78:44: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
const char* php_phongo_crypt_shared_version()
```

### DIFF
--- a/src/phongo_client.c
+++ b/src/phongo_client.c
@@ -75,7 +75,7 @@ static const mongoc_client_t* get_first_pclient_client(HashTable* ht)
  * Note: this may incorrectly return NULL if crypt_shared was loaded through a
  * mongoc_client_t since destroyed (e.g. single requested-scoped client);
  * however, that's the best can do with libmongoc's API. */
-const char* php_phongo_crypt_shared_version()
+const char* php_phongo_crypt_shared_version(void)
 {
 	const mongoc_client_t* client = NULL;
 


src/phongo_client.c
```diff
@@ -75,7 +75,7 @@ static const mongoc_client_t* get_first_pclient_client(HashTable* ht)
  * Note: this may incorrectly return NULL if crypt_shared was loaded through a
  * mongoc_client_t since destroyed (e.g. single requested-scoped client);
  * however, that's the best can do with libmongoc's API. */
-const char* php_phongo_crypt_shared_version()
+const char* php_phongo_crypt_shared_version(void)
```
jmikola, line 78: Very interesting considering I addressed this exact issue in [PHPC-2166](https://jira.mongodb.org/browse/PHPC-2166) (and related libmongoc/libmongocrypt tickets). I'm surprised that this was overlooked. Please double-check that it's the last occurrence before merging. Also, feel free to reference PHPC-2166 in the commit message to link it up (we don't need to change the fixVersion).

alcaeus: I checked, and there are none left in PHPC :)

PR #1412, opened 2023-04-21
PHPC-2206: Implement BSON\Value class

alcaeus:
PHPC-2206

This PR introduces the `MongoDB\BSON\Value` class as a wrapper for `bson_value_t`. The `Document`, `PackedArray`, and `Iterator` classes will be updated to return value instances instead of native values. This allows for a better handling of BSON types whenever raw BSON is consumed. For example, by using the `Value` class, users are able to differentiate between a 32-bit integers and 64-bit integers, even when the value does not fall in the 64-bit range.

The class contains a generic `getValue` method that will return the value contained in the `Value` instance. For documents and arrays, this returns `Document` and `PackedArray` instances respectively. For BSON types, it returns the corresponding `MongoDB\BSON\Type` instance, and for primitives it will return the PHP value.

In addition to the `getValue` method, it comes with a set of `is*` and `get*` methods for each BSON type. For added convenience, this includes accessors for integers (if one does not need to differentiate between 32-bit and 64-bit values), as well as numbers in general (if one does not need to differentiate between integers and doubles).

The PR also extracts some previously static factories from `phongo_bson.c` to the various BSON source files, as they are now used in multiple places. This will eventually be extracted into its own pull request. For the time being, the PR does not change the encoding and decoding of `bson_value_t`. As is, a `bson_value_t` is converted to a `zval` by adding it to an empty document, converting the document and extracting the relevant key. When converting a `zval` to `bson_value_t` this is handled the same way. Ideally, we'd skip this round-tripping and work with the `bson_value_t` directly. I'd like to complete this in a separate pull request as this is a larger effort if we want to avoid code duplication.

TODO:
- [x] Return `Value` instances when accessing document offsets
- [x] Return `Value` instances when accessing array offsets
- [x] Return `Value` instances in BSON iterator
- [x] Refactor encoding and decoding of `bson_value_t` instances.


src/BSON/Value.c
```diff
@@ -0,0 +1,631 @@
...
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	intern = Z_VALUE_OBJ_P(getThis());
+
+	switch (intern->value.value_type) {
+		case BSON_TYPE_INT32:
+			RETURN_LONG(intern->value.value.v_int32);
+
+		case BSON_TYPE_INT64:
+			RETURN_INT64(intern->value.value.v_int64);
```
alcaeus, line 309: This will always an `int` on 64-bit platforms and only use an `Int64` instance on 32-bit platforms if the value exceeds the 64-bit range. For the time being I consider this the better way, as `$value->getType()` and `$value->isInt64()` are available to check if the data received was a 64-bit integer.

src/BSON/Value.c
```diff
@@ -0,0 +1,631 @@
...
+			intern->value.value_type   = BSON_TYPE_BOOL;
+			intern->value.value.v_bool = true;
+			break;
+
+		case IS_LONG:
+			lvalue = Z_LVAL_P(data);
+
+			if (lvalue > INT32_MAX || lvalue < INT32_MIN) {
+				intern->value.value_type    = BSON_TYPE_INT64;
+				intern->value.value.v_int64 = lvalue;
```
alcaeus, line 117: This factory currently lacks a way to explicitly assign an int64 type for a 32-bit integer. A workaround would be to pass an `Int64` instance, but we may want to refactor this to allow for easier creation of 64-bit BSON integers.

src/BSON/Value.c
```diff
@@ -0,0 +1,631 @@
...
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	intern = Z_VALUE_OBJ_P(getThis());
+
+	PHONGO_VALUE_CHECK_TYPE(intern->value, BSON_TYPE_NULL);
+
+	RETURN_NULL();
+}
+
+static PHP_METHOD(MongoDB_BSON_Value, getNumber)
```
alcaeus, line 381: I'm not sure if `getNumber` makes a lot of sense in PHP, but I figured I'd add it anyways for convenience.

src/BSON/Value.c
```diff
@@ -0,0 +1,631 @@
...
+
+static PHP_METHOD(MongoDB_BSON_Value, isCode)
+{
+	php_phongo_value_t* intern;
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	intern = Z_VALUE_OBJ_P(getThis());
+
+	RETURN_BOOL(intern->value.value_type == BSON_TYPE_CODE || intern->value.value_type == BSON_TYPE_CODEWSCOPE);
```
alcaeus, line 523: This is the only case (apart from the `int` and `number` convenience helpers) where we return the same class instance for two different BSON types, hence the explicit declaration instead of using `PHONGO_VALUE_TYPE_ISSER`.

src/BSON/Value.stub.php
```diff
@@ -0,0 +1,214 @@
...
+
+namespace MongoDB\BSON;
+
+final class Value
+{
+    /**
+     * @var int
+     * @cvalue BSON_TYPE_EOD
+     */
+    public const TYPE_EOD = 0x00;
```
alcaeus, line 16: This is copy-pasta and included for the sake of completeness, but we may want to exclude the `EOD` type as it's irrelevant for users.

src/BSON/Value.stub.php
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * @var int
+     * @cvalue BSON_TYPE_DOUBLE
+     */
+    public const TYPE_DOUBLE = 0x01;
```
alcaeus, line 22: A note on the values: in other stub files we've used `UNKNOWN` as value whenever we use `@cvalue` to fetch the value from a C definition. Since I was lazy and copied over the declaration from bson-types.h I was able to include the values. There is no downside to this TBH: the generated arginfo has a `ZEND_ASSERT` call to ensure that the value contained in the `@cvalue` member is the same as defined here. The advantage is that when we start using gen_stub.php to generate class synopsis the documentation will contain the correct value for the constant.

src/BSON/Value.stub.php
```diff
@@ -0,0 +1,214 @@
...
+
+#if PHP_VERSION_ID >= 80000
+    final public function getValue(): mixed {}
+#else
+    /** @return mixed */
+    final public function getValue() {}
+#endif
+
+#if PHP_VERSION_ID >= 80000
+    final static public function fromPHP(mixed $value): Value {}
```
alcaeus, line 156: For now, this is the only factory for a BSON value instance. We could also have factories for each type, which would also allow to explicitly create an int64 value from the 32-bit range.

src/phongo_bson.h
```diff
@@ -115,9 +115,6 @@ bool php_phongo_bson_value_to_zval(const bson_value_t* value, zval* zv);
 bool php_phongo_bson_typemap_to_state(zval* typemap, php_phongo_bson_typemap* map);
 void php_phongo_bson_typemap_dtor(php_phongo_bson_typemap* map);
 
-void php_phongo_bson_new_timestamp_from_increment_and_timestamp(zval* object, uint32_t increment, uint32_t timestamp);
-void php_phongo_bson_new_int64(zval* object, int64_t integer);
```
alcaeus, line 119: These were moved to separate header files and renamed in the case of timestamp.

src/phongo_compat.h
```diff
@@ -152,6 +153,13 @@
 	} else {                                            \
 		ZVAL_LONG((_zv), (_value));                     \
 	}
+#define RETURN_INT64(_value)                               \
```
alcaeus, line 156: Adding a new compatibility macro felt nicer than using `ZVAL_INT64` on `return_value`. This also made me realise that we don't test any non-Windows 32-bit platforms, but I don't know if we even have 32-bit platforms on evergreen anymore. I'm checking with the build team regarding this.

tests/bson/bson-value-fromPHP-001.phpt
```diff
@@ -0,0 +1,181 @@
+--TEST--
+MongoDB\BSON\Value::fromPHP()
```
alcaeus, line 2: I've only created a single test for rapid prototyping. This will be split up into multiple tests in the future.

src/BSON/Value.stub.php
```diff
@@ -0,0 +1,214 @@
+<?php
+
+/**
+ * @generate-class-entries static
+ * @generate-function-entries static
+ */
+
+namespace MongoDB\BSON;
+
+final class Value
```
alcaeus, line 10: This class doesn't yet implement any interfaces, as I haven't thought about whether users should be able to serialise/unserialise these instances. Feedback is welcome.

src/BSON/Value.c
```diff
@@ -0,0 +1,631 @@
...
+static HashTable* php_phongo_value_get_properties_hash(phongo_compat_object_handler_type* object, bool is_temp)
+{
+	php_phongo_value_t* intern;
+	HashTable*          props;
+
+	intern = Z_OBJ_VALUE(PHONGO_COMPAT_GET_OBJ(object));
+
+	PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_temp, intern, props, 1);
+
+	// TODO: Print value (somehow)
```
alcaeus, line 69: I was thinking of including the type here, and we may just want to include the result of `bson_value_to_zval` here instead of displaying raw data.

src/BSON/Value.c
```diff
@@ -0,0 +1,631 @@
...
+	zend_long           lvalue;
+
+	PHONGO_PARSE_PARAMETERS_START(1, 1)
+	Z_PARAM_ZVAL(data);
+	PHONGO_PARSE_PARAMETERS_END();
+
+	object_init_ex(&zv, php_phongo_value_ce);
+	intern = Z_VALUE_OBJ_P(&zv);
+
+	switch (Z_TYPE_P(data)) {
```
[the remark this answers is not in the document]
alcaeus: Sorry. I plan to refactor the `zval_to_bson` function to create the `bson_value_t` directly instead of wrapping it in a document and passing it to the other function.

src/BSON/Value.stub.php
```diff
@@ -0,0 +1,214 @@
...
+    final public function isDBPointer(): bool {}
+    final public function isDecimal128(): bool {}
+    final public function isDocument(): bool {}
+    final public function isDouble(): bool {}
+    final public function isInt(): bool {}
+    final public function isInt32(): bool {}
+    final public function isInt64(): bool {}
+    final public function isMaxKey(): bool {}
+    final public function isMinKey(): bool {}
+    final public function isNumber(): bool {}
```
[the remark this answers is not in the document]
alcaeus: I agree with `isNumber` not really making any sense. I'm dropping it as Decimal128 doesn't behave like a number in PHP and should be included for consistency.

src/BSON/Value.c
```diff
@@ -0,0 +1,631 @@
...
+#include "BSON/Int64.h"
+#include "BSON/Javascript.h"
+#include "BSON/ObjectId.h"
+#include "BSON/PackedArray.h"
+#include "BSON/Regex.h"
+#include "BSON/Symbol.h"
+#include "BSON/Timestamp.h"
+#include "BSON/UTCDateTime.h"
+
+#define PHONGO_VALUE_TYPE_ISSER(name, type)                              \
```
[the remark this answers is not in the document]
alcaeus: > Can we use literally anything other than "isser" that won't make me think of [Die Antwoord](https://en.wikipedia.org/wiki/Die_Antwoord) every time I read this?

LOL...I've renamed it to `PHONGO_VALUE_IS_TYPE`.


> Also, I'd suggest defining this right above the macro usage and then undefining it immediately after. Prior art:

Makes sense 👍

src/MongoDB/Monitoring/CommandFailedEvent.c
```diff
@@ -23,6 +23,7 @@
 #include "phongo_bson_encode.h"
 #include "phongo_error.h"
 
+#include "BSON/Int64.h"
```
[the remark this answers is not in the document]
alcaeus: Not quite. `php_phongo_bson_new_int64` was previously declared in `phongo_bson.h`. This file is included in `phongo_structs.h`, which again is included in `php_phongo.h`, which is included in almost all of our files. With the function being moved to Int64.h, the `_INT64` macros require the inclusion of this file. What makes the problem worse is that the `_INT64` macros pass through to the `_LONG` macros on 64-bit machines and only make use of `php_phongo_bson_new_int64` on 32-bit machines. This makes this hard to spot during development, as tooling such as CLion marks the include as unnecessary.

Thinking about this some more, it would make sense to add `Int64.h` to `phongo_compat.h` as that's where usage comes from. That's probably the better alternative.

src/BSON/ObjectId.h
```diff
@@ -21,6 +21,6 @@
 
 #include <php.h>
 
-void phongo_objectid_init(zval* return_value, const bson_oid_t* oid);
+void php_phongo_bson_new_objectid(zval* return_value, const bson_oid_t* oid);
```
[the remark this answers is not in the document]
alcaeus: I've refactored the factories to return a bool and updated all instances to handle the return value. The only place I wasn't able to reliably handle were the compatibility macros for int64, as the macro itself can't decide what to do in case construction of the Int64 instance goes wrong. Since the factory currently always returns true, I'm ignoring that for the time being.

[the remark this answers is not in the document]
jmikola: > Since the factory currently always returns true, I'm ignoring that for the time being.

Please leave a comment in or around the macro if you haven't already.

src/BSON/Value.c
```diff
@@ -0,0 +1,631 @@
...
+			break;
+
+		case IS_ARRAY:
+		case IS_OBJECT:
+			php_phongo_zval_to_bson_value(data, PHONGO_BSON_NONE, &intern->value);
+			break;
+
+		default:
+			// Unsupported
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Received unsupported type %s for initialiser", zend_zval_type_name(data));
```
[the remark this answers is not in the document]
jmikola: > we're complaining about a PHP type that isn't supported as BSON

Oh, totally missed that when reading this earlier. SGTM, and glad you caught the extra deref bug.

src/BSON/Value.c
alcaeus: If I understand correctly, even after opening up the constructor we'd always return values as an `int` on 64-bit platforms, but allow users to create `Int64` instances if they want to force a `$numberLong` BSON type? That would indeed solve the problem.

src/BSON/Value.c
alcaeus: I believe your comment refers to encoding a `Javascript` instance as BSON? I didn't touch that part yet (instead relying on `zval_to_bson` to encode objects), but I agree that we should ensure to not use `CODEWSCOPE` if the `Javascript` instance does not contain a scope.

src/BSON/Value.stub.php
alcaeus: I'll leave this for later, as I'm not entirely sure what the BSON representation of a single value should be, as BSON always needs a root-level document.

src/BSON/Value.c
alcaeus: Reporting the BSON type along with the zeal makes sense, I'll implement that 👍

src/BSON/Value.c
```diff
@@ -0,0 +1,631 @@
...
+			php_phongo_bson_new_javascript(return_value, intern->value.value.v_code.code, intern->value.value.v_code.code_len, NULL);
+			return;
+
+		case BSON_TYPE_CODEWSCOPE:
+			scope = bson_new_from_data(intern->value.value.v_codewscope.scope_data, intern->value.value.v_codewscope.scope_len);
+			php_phongo_bson_new_javascript(return_value, intern->value.value.v_codewscope.code, intern->value.value.v_codewscope.code_len, scope);
+			return;
+
+		default:
+			phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "Unexpected BSON type %d", intern->value.value_type);
```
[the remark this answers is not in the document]
alcaeus: I took the liberty of defining a macro that throws the correct exception.

src/BSON/Value.c
jmikola: Opening the Int64 constructor in [PHPC-1949](https://jira.mongodb.org/browse/PHPC-1949) will just make it easier for users to create int64 BSON types. It does nothing for converting BSON to PHP.

With that in mind, I was suggesting `Value::fromPHP()` accept an Int64 type to force creation of a int64 BSON type.

src/BSON/Value.stub.php
```diff
@@ -0,0 +1,206 @@
...
+    final public function getDocument(): Document {}
+    final public function getDouble(): float {}
+    final public function getInt(): int {}
+    final public function getInt32(): int {}
+#if PHP_VERSION_ID >= 80000
+    final public function getInt64(): int|Int64 {}
+#else
+    /** @return int|Int64 */
+    final public function getInt64() {}
+#endif
```
jmikola, line 172: https://github.com/mongodb/mongo-php-driver/pull/1412#discussion_r1176193974 just made me think bank to this.

Is `getInt(): int` and `getInt32(): int` redundant? Or will they have different logic internally where `getInt()` warns on truncation and `getInt32()` does not?

I have more thoughts on `getInt64(): int|Int64`. If the method behaves just like `getInt()` except that it returns an Int64 instead of a truncation warning, then it's only relevant to 32-bit platforms. Such a method would always return `int` on 64-bit platforms and it'd effectively be an alias of `getInt()`.

Does it make more sense to have `getInt64()` always return Int64? There's a need on both 32-bit and 64-bit platforms to maintain type fidelity for 64-bit integers with values within the range of signed 32-bit integers. I do realize a user could just keep the Value as-is and pass it along for BSON encoding, so maybe this _isn't_ important.

Thoughts?

src/BSON/Value.c
```diff
@@ -53,15 +54,23 @@ static HashTable* php_phongo_value_get_properties_hash(phongo_compat_object_hand
...
+
+		zend_hash_str_update(props, "type", sizeof("type") - 1, &type);
+	}
+
+	{
+		zval value;
+
+		php_phongo_bson_value_to_zval(&intern->value, &value);
+
+		zend_hash_str_update(props, "value", sizeof("value") - 1, &value);
```
jmikola, line 72: Note that the `get_properties_hash` function typically corresponds to `init_from_hash`, and both are used for `__set_state` and PHP serialization in addition to debug output. That could be problematic here:

 * There will need to be an inverse of `php_phongo_bson_type_to_string`
 * Relying on `php_phongo_bson_value_to_zval` means we might have trouble round-tripping BSON int64 (regardless of system architecture).

src/BSON/Value.stub.php
alcaeus: All of the `get*` methods check the internal type and throw if that isn't the case. For example, calling `getInt64` on an instance storing an int32 will throw:
> Uncaught MongoDB\Driver\Exception\UnexpectedValueException: Unexpected BSON type 0x10

This is similar to `bson_iter_t`, which also refuses to coerce in its getters (e.g. `bson_iter_int64`), but includes specific methods to coerce types (e.g. `bson_iter_as_int64`). While the iterator class returns `0` (or nothing in case of `bson_iter_document`) when types do not match, we explicitly throw, which I believe is fine considering the API allows for checking types first.

This brings me to `getInt`, `getInt32`, and `getInt64`. Let's take the following code:
```php
$int32 = MongoDB\BSON\Value::fromPHP(2);
var_dump($int32->isInt());
var_dump($int32->isInt32());
var_dump($int32->isInt64());

var_dump($int32->getValue());
var_dump($int32->getInt());
var_dump($int32->getInt32());
var_dump($int32->getInt64());
```

The output of this is as follows:
```
bool(true)
bool(true)
bool(false)
int(2)
int(2)
int(2)
Uncaught MongoDB\Driver\Exception\UnexpectedValueException: Unexpected BSON type 0x10
```

To summarise, the consumer should use:
* `getValue` when the type does not matter at all and they are only interested in the value
* `get*` (e.g. `getInt32` or `getDocument`) when they want to ensure a specific BSON type; the corresponding `is` method should be used to avoid exceptions
* `getInt` when they want to use an integer but don't care whether it's 32-bit or 64-bit.

`getInt` is a special case here, but it behaves different from `getInt32` and `getInt64` as it only throws if the type is not int32 or int64, whereas the type-specific getters only check for the specific type they are designed for. I hope this clears up the confusion between `getInt` and `getInt32`.

---

As for `getInt64` always returning an `Int64` instance, this would definitely be an option. If we decide that `getInt64` should always return an object, I'd like to ensure that PHPC-1949 (opening the constructor of `Int64`) is implemented and that operations on an `Int64` work the same way they do with scalar integers. That would allow us to immediately define the return type of `getInt64` as `Int64` instead of the current `int | Int64`, avoiding an unnecessary change down the line.

The only question that remains then is how `getInt` should behave when returning the value of an int64 `bson_value_t` on 64-bit platforms: should it return an `int` for convenience or should it return an `Int64` for consistence.

src/BSON/Value.c
alcaeus: Serialisation is something I've left out as per the original comment in the stub class. While I think we can avoid thinking about BSON serialisation, regular PHP serialisation is something we should implement. The only issue here is int64 with 32-bit values, which might be solved if `phongo_bson_value_to_zval` always returns an `Int64` instance for the type.

That said, I'll work on `init_from_hash` next and make sure that serialisation of types works.

src/BSON/Value.c
```diff
@@ -0,0 +1,631 @@
...
+
+	new_intern = Z_OBJ_VALUE(new_object);
+	zend_objects_clone_members(&new_intern->std, &intern->std);
+
+	bson_value_copy(&intern->value, &new_intern->value);
+
+	return new_object;
+}
+
+// TODO: can we even compare values?
```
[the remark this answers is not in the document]
alcaeus: I'm considering implementing a comparison handler that is able to indicate equality, i.e. returns `0` if type and value match, and returns `1` if they don't match. The use case here would be to allow for easier comparison in tests (e.g. for `__clone` or `__serialize`) and potentially for users, as checking equality/inequality is a common use case. I'll note here (and later in the docs) that using `<`, `>`, `<=`, `>=`, and `<=>` would not produce consistent results due to us always returning `1` for unequal objects.

A workaround for the latter would be to order `bson_value_t` by type first, and by value (using `memcmp`) second, i.e.:
```php
return (value1.type <=> value2.type) ?: (value1.value <=> value2.value);
```

While this doesn't handle comparison of equal values of different types (e.g. `int32(2) == int64(2)`), it would provide a more consistent behaviour than the first suggestion by ensuring that `(value1 > value2) === (value2 < value1)`. This would not be the case if we always returned `1` on inequality, as that would mean that `value1 > value2` and `value2 > value1` would both be true at the same time.

src/BSON/Value.stub.php
jmikola: Oh, I thought we were talking about PHP serialization, not our BSON interfaces.

The advantage of using a BSON envelope for PHP serialization is that we can leverage existing APIs and BSON validation when unserializing. Not constructing a `bson_value_t` directly from, say, a base64 encoded string would also avoid potential attack vectors to create invalid BSON.

src/BSON/Value.c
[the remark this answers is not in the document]
jmikola: Oh, I forgot `bson_compare()` was so simplistic. Would you mind moving the above comments to [PHPC-2209](https://jira.mongodb.org/browse/PHPC-2209) so we can continue the conversation there?

src/BSON/Value.stub.php
jmikola: Would `getValue()` only return an Int64 iff we're on a 32-bit platform _and_ the value is beyond the 32-bit range? If it's meant to be a convenience function, I assume we'd want to return a scalar int whenever possible (e.g. `int64(0)` would not yield an Int64 object and folks can use `getInt64()` if they care about that).

> As for getInt64 always returning an Int64 instance, this would definitely be an option

I'm in favor of this, provided we can ensure Int64 objects can be cast to PHP integers. I don't believe that's tested/implemented now, but I created [PHPC-2213](https://jira.mongodb.org/browse/PHPC-2213) to track that.

> operations on an Int64 work the same way they do with scalar integers

Created [PHPC-2212](https://jira.mongodb.org/browse/PHPC-2212) to track that.

> how `getInt()` should behave

I'd need to give this more thought before we commit, but if `getValue()` is already a convenience method I'm not sure what the real use case for `getInt()` is.

 * If it only returns Int64 when absolutely necessary (i.e. values in 64-bit range on a 32-bit platform) then it's basically `getValue()` with additional type checks.
 * If it always returns a PHP integer and warns on truncation then it's probably not usable on 32-bit platforms (warnings could be unpredictable). On a 64-bit platform, it's still basically `getValue()` with type checking, expect you get the benefit of an `int` return type hint (is that terribly useful for static analysis?).
 * If it always returns BSON int32 as a PHP int and BSON int64 as an Int64 object, then it seems like a more pure alias of the `getInt32()` and `getInt64()` methods. Combined with casting support mentioned above, x64 users could do `(int) $value->getInt()`, but then this is basically no different than the preceding bullet.

How do you see this method being used?

src/BSON/Value.c
jmikola: If we're dealing with PHP serialization and `var_export()`, I think it'd be sensible to output a base64 binary string (i.e. raw BSON in an envelope). I don't see any reason to rely on zval conversion, and it'd hopefully make the `init_from_hash` function much easier to implement. There, we probably wouldn't even need to output the type code at all.

This would mean the `get_debug_info` handler would need separate logic, assuming you'd like `var_dump()` to display a user-friendly value. But there, we wouldn't need to care about the int64 conversion since we'd still have the BSON type to report alongside it.

src/BSON/Value.c
jmikola: I think the impetus for that last comment was realizing that `getCode()` was essentially an alias like `getInt()`. I was considering how users might want to determine the original BSON type given just the returned Javascript object. AFAIK, the only way to do so is checking whether `getScope()` is `null`.

I don't think there's anything to do here other than make sure we document how `getCode()` works by allowing both types.

src/BSON/Value.c
alcaeus: I agree. At the moment, `get_debug_info` also displays a `Document` or `PackedArray` instance, which isn't helpful to the user. In those cases, I'd rather use `phongo_bson_value_to_zval_legacy` to convert documents and arrays to their PHP representations. It will mean that I have to either add a flag to `phongo_bson_value_to_zval` (to indicate whether documents and arrays should be returned as BSON types) or find a better name than `_legacy`, as the latter indicates that the method will be dropped eventually.

I'll implement serialisation and then see what can be refactored. 👍

src/BSON/Value.c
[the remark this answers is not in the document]
alcaeus: I expanded the description of PHPC-2209 to capture this.

src/BSON/Value.stub.php
alcaeus: `isInt()` and `getInt()` have been removed. I didn't create a ticket yet as I agree with your point that it's just `getValue()` with added type checks, so there is no reason to have this method.

For integers, this leaves us with:
* `getInt32()`, which will always return an int value but throw if the value is an int64
* `getInt64()`, which will always return an `Int64` instance but throw if the value is an int32 (i.e., not `{$numberLong: ... }` in extended JSON). This is true on all platforms and for all values, so even int64 values that fall in the 32-bit range are returned as `Int64` instances.
* `getValue()`, which will return whatever the corresponding type specific getter would return.

src/BSON/Value.c
alcaeus: Serialisation rules have been updated. For `Document`, `PackedArray`, and `Value`, the debug representation now includes raw data as well as the PHP representation. The type that is returned is a formatted string for convenience. Take the following example:
```
var_dump(Value::fromPHP([0, 1, 2]);
// Output
object(MongoDB\BSON\Value)#1 (2) {
  ["type"]=>
  string(5) "array"
  ["value"]=>
  object(MongoDB\BSON\PackedArray)#1 (2) {
    ["data"]=>
    string(36) "GgAAABAwAAEAAAAQMQACAAAAEDIAAwAAAAA="
    ["value"]=>
    array(3) {
      [0]=>
      int(1)
      [1]=>
      int(2)
      [2]=>
      int(3)
    }
  }
}
```

When serialising a `Document` or `PackedArray` instance, only the raw data is serialised. When serialising a `Value` instance, the result includes the type as well as the correct BSON instance. Take an Int64 for example:
```
var_dump(serialize(Value::fromPHP(new Int64(2))));
// Output
string(110) "O:18:"MongoDB\BSON\Value":2:{s:4:"type";i:18;s:5:"value";O:18:"MongoDB\BSON\Int64":1:{s:7:"integer";s:1:"2";}}"
```

The same is true for var_export:
```
var_export(Value::fromPHP(new Int64(2)), true);
// Output
\MongoDB\BSON\Value::__set_state(array(
   'type' => 18,
   'value' => \MongoDB\BSON\Int64::__set_state(array(
     'integer' => '2',
  )),
))
```

src/BSON/Value.stub.php
alcaeus: Resolving this comment as `Value` implements `\Serializable` and serialisation has been added.

src/MongoDB/ClientEncryption.c
```diff
@@ -941,14 +942,15 @@ static void phongo_clientencryption_decrypt(php_phongo_clientencryption_t* clien
...
-	php_phongo_zval_to_bson_value(zciphertext, PHONGO_BSON_NONE, &ciphertext);
+	phongo_zval_to_bson_value(zciphertext, &ciphertext);
 
 	if (!mongoc_client_encryption_decrypt(clientencryption->client_encryption, &ciphertext, &value, &error)) {
 		phongo_throw_exception_from_bson_error_t(&error);
 		goto cleanup;
 	}
 
-	if (!php_phongo_bson_value_to_zval(&value, zvalue)) {
+	/** Use the legacy decoder to return PHP types instead of BSON types */
```
jmikola, line 952: Correct me if I'm mistaken, but this preserves BC and will only differ from the non-legacy function in that Document and PackedArray instances will _not_ be returned. The subtle implication there is that there's no way to obtain a Value instance.

If so, we might want to considering adding an alternative `decrypt()` function that returns a Value directly. Feel free to create a PHPC issue for that if you agree.

src/MongoDB/Monitoring/CommandFailedEvent.c
```diff
@@ -249,7 +249,11 @@ static HashTable* php_phongo_commandfailedevent_get_debug_info(phongo_compat_obj
 	if (intern->has_service_id) {
 		zval service_id;
 
-		phongo_objectid_init(&service_id, &intern->service_id);
+		if (!phongo_objectid_new(&service_id, &intern->service_id)) {
+			/* Exception should already have been thrown */
+			goto done;
```
jmikola, line 254: I noticed you didn't check the return value of `phongo_objectid_new` earlier in the file. Is there any reason to do so here? Looking at the implementation it always returns `true` so I think we can forgo the conditional.

Likewise for other calls to `phongo_objectid_new` and any other similar "new" methods.

src/phongo_bson.c
```diff
@@ -982,34 +896,140 @@ bool php_phongo_bson_data_to_zval(const unsigned char* data, int data_len, zval*
...
+		case BSON_TYPE_UNDEFINED:
+			return object_init_ex(zv, php_phongo_undefined_ce) == SUCCESS;
+
+		case BSON_TYPE_ARRAY:
+			if (!bson_init_static(&bson, value->value.v_doc.data, value->value.v_doc.data_len)) {
+				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "Invalid BSON received for BSON_TYPE_ARRAY");
+				return false;
+			}
+
+			return phongo_packedarray_new(zv, &bson, true);
```
jmikola, line 1019: Ah, this answers my earlier question about whether the BSON data was ever copied for `phongo_packedarray_new()` and `phongo_document_new()`. Disregard my earlier question (it's long gone in some collapsed file by this point of the review).

src/BSON/Value.c
```diff
@@ -0,0 +1,665 @@
...
+	smart_str            buf = { 0 };
+
+	intern = Z_VALUE_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	array_init_size(&retval, 2);
+	add_assoc_long(&retval, "type", intern->value.value_type);
+
+	phongo_bson_value_to_zval(&intern->value, &value);
```
jmikola, line 518: Should you handle a `false` return value and `goto` a done/cleanup label?

src/phongo_bson.c
```diff
@@ -982,34 +896,140 @@ bool php_phongo_bson_data_to_zval(const unsigned char* data, int data_len, zval*
 	return retval;
 }
 
-/* Converts a BSON value to a ZVAL. */
-bool php_phongo_bson_value_to_zval(const bson_value_t* value, zval* zv)
+/** Converts a BSON value to a ZVAL, returning BSON objects and arrays as
+ * standard PHP types instead of Document or PackedArray instances */
+bool phongo_bson_value_to_zval_legacy(const bson_value_t* value, zval* zv)
```
jmikola, line 901: If you expect that a `false` return value will be accompanied by an exception, please note that in the comment.

src/phongo_bson.c
```diff
@@ -982,34 +896,140 @@ bool php_phongo_bson_data_to_zval(const unsigned char* data, int data_len, zval*
...
+		zval_ptr_dtor(&state.zchild);
+		return retval;
 	}
 
-	retval = true;
+	return phongo_bson_value_to_zval(value, zv);
+}
+
+/* Converts a BSON value to a ZVAL. */
+bool phongo_bson_value_to_zval(const bson_value_t* value, zval* zv)
```
jmikola, line 936: If you expect that a `false` return value will be accompanied by an exception, please note that in the comment.

src/phongo_bson.c
```diff
@@ -982,34 +896,140 @@ bool php_phongo_bson_data_to_zval(const unsigned char* data, int data_len, zval*
...
+				return false;
+			}
+
+			return phongo_document_new(zv, &bson, true);
 	}
 
-cleanup:
-	zval_ptr_dtor(&state.zchild);
-	return retval;
+	/* Throw exception? */
```
jmikola, line 1030: I think so, yes. This seems no different than other default cases where you throw for an unexpected type.

Consider handling this in a `default:` within the above `switch`. That's what `php_phongo_bson_iter_to_zval` did:

```c
default:
    phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "Detected unsupported BSON type %d at offset %d", iter->type, iter->off);
    return false;
```

src/phongo_bson_encode.c
```diff
@@ -611,3 +608,67 @@ void php_phongo_zval_to_bson_value(zval* data, php_phongo_bson_flags_t flags, bs
 	bson_destroy(&bson);
 	zval_ptr_dtor(&data_object);
 }
+
+/* Converts the argument to a bson_value_t. If the object is an instance of
+ * MongoDB\BSON\Serializable, the return value of bsonSerialize() will be
+ * used. It is the caller's responsibility to call bson_value_destroy. */
+bool phongo_zval_to_bson_value(zval* data, bson_value_t* value)
```
jmikola, line 615: Note that a false return value will be accompanied by an exception.

src/phongo_compat.h
```diff
@@ -117,43 +121,57 @@
 		}                \
 	}
 
-#define ZVAL_INT64_OBJ(_zv, _value) php_phongo_bson_new_int64((_zv), (_value))
+#define ZVAL_INT64_OBJ(_zv, _value) phongo_int64_new((_zv), (_value))
+#define RETURN_INT64_OBJ(_value)              \
```
jmikola, line 125: Per my earlier comment this might also be unused.

src/phongo_compat.h
```diff
@@ -117,43 +121,57 @@
...
 		add_assoc_long((_zv), (_key), (_value));        \
 	}
 #define ZVAL_INT64(_zv, _value)                         \
 	if ((_value) > INT32_MAX || (_value) < INT32_MIN) { \
-		php_phongo_bson_new_int64((_zv), (_value));     \
+		phongo_int64_new((_zv), (_value));              \
 	} else {                                            \
 		ZVAL_LONG((_zv), (_value));                     \
 	}
+#define RETURN_INT64(_value)                            \
```
jmikola, line 168: Is this used anywhere?

tests/bson/bson-value-var_export-001.phpt
```diff
@@ -0,0 +1,235 @@
...
+    $bsonValue = MongoDB\BSON\Value::fromPHP($value);
+    echo var_export($bsonValue, true), "\n";
+}
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+Testing null:
+%r\\?%rMongoDB\BSON\Value::__set_state(array(
```
jmikola, line 48: Is `Value::__set_state()` tested? We have tests for other `__set_state()` implementations.

I think various other Value methods might be missing tests as well, but I'm fine to put those off towards the end of the PR until all code changes are addressed.

src/BSON/Value.c
```diff
@@ -0,0 +1,665 @@
...
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	intern = Z_VALUE_OBJ_P(getThis());
+
+	PHONGO_VALUE_CHECK_TYPE(intern->value, BSON_TYPE_ARRAY);
+
+	bson = bson_new_from_data(intern->value.value.v_doc.data, intern->value.value.v_doc.data_len);
+
+	phongo_packedarray_new(return_value, bson, false);
```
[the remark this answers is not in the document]
alcaeus: `phongo_bson_value_to_zval` uses `bson_init_static` to initialise a BSON object, then passes it to `phongo_packedarray_new` with `copy=true`. This could probably be refactored to use `bson_new_from_data` and use `copy=false`, in which case we should document that ownership of the BSON structure will move to the newly created instance.
I can create a ticket for refactoring this.

src/phongo_compat.h
```diff
@@ -117,43 +121,57 @@
 		}                \
 	}
 
-#define ZVAL_INT64_OBJ(_zv, _value) php_phongo_bson_new_int64((_zv), (_value))
+#define ZVAL_INT64_OBJ(_zv, _value) phongo_int64_new((_zv), (_value))
```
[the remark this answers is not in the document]
alcaeus: This is used in `phongo_bson_value_to_zval` only.

src/BSON/Value.c
```diff
@@ -0,0 +1,665 @@
...
+{
+	php_phongo_value_t* intern;
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	intern = Z_VALUE_OBJ_P(getThis());
+
+	PHONGO_VALUE_CHECK_TYPE(intern->value, BSON_TYPE_INT64);
+
+	RETURN_INT64_OBJ(intern->value.value.v_int64);
```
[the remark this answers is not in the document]
alcaeus: I mainly created this for consistency with `RETURN_INT64`, and there were previously more usages of this. We could also use this in the `RETURN_INT64` macro for 32-bit platforms, as that initialises an int64 object if the value falls out of the 32-bit range. I'll defer to your judgment on this.

src/BSON/Value.stub.php
```diff
@@ -0,0 +1,210 @@
...
+    final public function isMinKey(): bool {}
+    final public function isNull(): bool {}
+    final public function isObjectId(): bool {}
+    final public function isRegex(): bool {}
+    final public function isSymbol(): bool {}
+    final public function isTimestamp(): bool {}
+    final public function isUndefined(): bool {}
+    final public function isUtf8(): bool {}
+
+    final public static function __set_state(array $properties): Regex {}
```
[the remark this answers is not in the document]
alcaeus: It did not create a test failure, as per your last comment in this review: I forgot to add tests creating a `Value` instance from a `var_export` result. I'll add proper tests.

src/BSON/Value.stub.php
```diff
@@ -0,0 +1,210 @@
...
+#else
+    /** @return mixed */
+    final public function getValue() {}
+#endif
+
+#if PHP_VERSION_ID >= 80000
+    final static public function fromPHP(mixed $value): Value {}
+#else
+    /** @param mixed $value */
+    final static public function fromPHP($value): Value {}
```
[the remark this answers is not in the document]
alcaeus: That is correct: https://github.com/php/php-src/blob/fbe6696d49980ec52018fb4281798fb873bf0a54/build/gen_stub.php#L221. `static` might be allowed, but then that isn't supported until PHP 8.0.

tests/bson/bson-document-clone-001.phpt
```diff
@@ -23,5 +23,22 @@ bool(false)
 object(MongoDB\BSON\Document)#%d (%d) {
   ["data"]=>
   string(48) "JAAAAARmb28AGgAAABAwAAEAAAAQMQACAAAAEDIAAwAAAAAA"
+  ["value"]=>
+  object(stdClass)#%d (%d) {
+    ["foo"]=>
+    object(MongoDB\BSON\PackedArray)#%d (%d) {
+      ["data"]=>
+      string(36) "GgAAABAwAAEAAAAQMQACAAAAEDIAAwAAAAA="
```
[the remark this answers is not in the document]
alcaeus: That is the downside of showing raw and formatted values, but I'm hopeful that this won't become an issue. 🤞

src/MongoDB/Monitoring/CommandFailedEvent.c
jmikola: SGTM. I overlooked that the other instances were at the end of a function where the check would have made no difference.

tests/bson/bson-value-var_export-001.phpt
alcaeus: Tests for `__set_state` have been added. While there are no explicit test files for other methods, the main constructor tests also calls the `is<Type>` and `get<Type>` methods for each value that we're testing.

src/BSON/Value.stub.php
```diff
@@ -0,0 +1,210 @@
...
+    final public function isTimestamp(): bool {}
+    final public function isUndefined(): bool {}
+    final public function isUtf8(): bool {}
+
+    final public static function __set_state(array $properties): Value {}
+
+    final public function serialize(): string {}
+
+    /** @param string $serialized */
+    final public function unserialize($serialized): void {}
```
jmikola, line 205: I realize this is consistent with every other stub, but why do we have an `@param` annotation here?

The interface has a `string` type since https://github.com/php/php-src/commit/273731fb762d661fb40015afd851ab340cf7a429#diff-c8adc7ea754be4705d22956277447f2a8cdcd19e3185caa1513391d864a82384R44

It had no type hint in 7.4: https://github.com/php/php-src/blob/dca8d5565b947270a29f232bc54efd9df3b92b94/Zend/zend_interfaces.c#L572-L580

Given that, should we be using `#if PHP_VERSION_ID >= 80000` to conditionally declare a type hint? If so, can you create a ticket for that?

src/MongoDB/ClientEncryption.c
jmikola: :point_up: Outstanding question

src/BSON/Value.c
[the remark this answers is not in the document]
jmikola: Sorry for not following up here. A separate ticket SGTM to we don't lose track of this.

src/phongo_compat.h
```diff
@@ -22,6 +22,10 @@
 #include <Zend/zend_string.h>
 #include <Zend/zend_portability.h>
 
+/* Include stdbool.h as it might not have been implicitly loaded yet */
+#include <stdbool.h>
```
jmikola, line 26: Is this still relevant? I don't see `bool` used anywhere in this file. Just `zend_bool`.

src/BSON/Value.c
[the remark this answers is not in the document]
jmikola: `RETURN_INT64` seems to be gone. I forget if that was in a previous version of this PR or something else.

In any event, I see you started using `phongo_int64_new()` in the other Int64 macros. Feel free to resolve this after reading.

src/MongoDB/ClientEncryption.c
```diff
@@ -921,7 +921,8 @@ static void phongo_clientencryption_encrypt(php_phongo_clientencryption_t* clien
 		goto cleanup;
 	}
 
-	if (!php_phongo_bson_value_to_zval(&ciphertext, zciphertext)) {
+	/** Use the legacy decoder to return PHP types instead of BSON types */
+	if (!phongo_bson_value_to_zval_legacy(&ciphertext, zciphertext)) {
```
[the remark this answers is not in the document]
jmikola: :point_up: Outstanding question

src/MongoDB/Session.c
```diff
@@ -643,7 +644,11 @@ static HashTable* php_phongo_session_get_debug_info(phongo_compat_object_handler
 		if (timestamp && increment) {
 			zval ztimestamp;
 
-			php_phongo_bson_new_timestamp_from_increment_and_timestamp(&ztimestamp, increment, timestamp);
+			if (!phongo_timestamp_new(&ztimestamp, increment, timestamp)) {
```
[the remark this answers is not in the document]
jmikola: I think we previously agreed to do error checking regardless of the potential to return `false`. If you concur, feel free to resolve this.

src/phongo_bson.c
```diff
@@ -332,24 +334,15 @@ static bool php_phongo_bson_visit_undefined(const bson_iter_t* iter, const char*
...
-}
-
 static bool php_phongo_bson_visit_oid(const bson_iter_t* iter ARG_UNUSED, const char* key, const bson_oid_t* v_oid, void* data)
 {
 	zval*                  retval = PHONGO_BSON_STATE_ZCHILD(data);
 	php_phongo_bson_state* state  = (php_phongo_bson_state*) data;
 	zval                   zchild;
 
-	php_phongo_objectid_new_from_oid(&zchild, v_oid);
+	if (!phongo_objectid_new(&zchild, v_oid)) {
```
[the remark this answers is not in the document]
jmikola: Per my earlier comment, I think we agreed to leave the error checking in place so nothing to change here. I see you did add comments about assuming a `true` return from `phongo_int64_new()` in the Int64 macros.

Feel free to resolve after reading.

src/phongo_bson.c
```diff
@@ -982,34 +896,147 @@ bool php_phongo_bson_data_to_zval(const unsigned char* data, int data_len, zval*
 	return retval;
 }
 
-/* Converts a BSON value to a ZVAL. */
-bool php_phongo_bson_value_to_zval(const bson_value_t* value, zval* zv)
+/** Converts a BSON value to a zval, returning BSON objects and arrays as
```
jmikola, line 899: We don't use doc blocks in PHPC.

```suggestion
/* Converts a BSON value to a zval, returning BSON objects and arrays as
```

jmikola (approved):
Just a few questions and some things to follow up on, but I think we're all set here.

Enjoy your customary Gordon Ramsey GIF:

.

src/BSON/Value.stub.php
alcaeus: This is due to the stub tool (gen_stub.php), which complains about parameters without types. Each parameter either needs to have a declared type or a corresponding `@param` annotation. The same goes for return types, where a return type or `@return` annotation are necessary.

src/phongo_compat.h
alcaeus: You are right - this was left over from a previous version. I've removed it.

src/MongoDB/ClientEncryption.c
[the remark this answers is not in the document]
alcaeus: This is correct - I've changed this to use `phongo_bson_value_to_zval`.

src/MongoDB/ClientEncryption.c
alcaeus: Yes, if we were to use `phongo_bson_value_to_zval` here, any encrypted document or array would suddenly no longer be returned as PHP value but rather as a `PackedArray` or `Document` instance.  To avoid the BC break, I'm using the "legacy" function here.

I created PHPC-2240 to handle this in preparation for the future where we no longer return arrays and `stdClass` instances in PHPC.

src/MongoDB/Session.c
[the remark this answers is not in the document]
alcaeus: That is correct - these functions may decide to throw at some point, so I opted for consistency with other `phongo_<bsonType>_new` functions here.

src/phongo_bson.c
alcaeus: Good point, especially considering that C doesn't have the concept of docblocks at all.

That said, there are a number of other instances where we use docblocks which we may want to clean up at some point. I've created PHPC-2241 to track cleaning those up.

src/phongo_compat.h
alcaeus: Correction: this is still necessary, at least on some versions of PHP. With PHP 8 or newer, `zend_bool` was changed to use a standard `bool` instead of a `unsigned char` (see https://github.com/php/php-src/commit/4b77a158ef2850582aeb4834c588aba49942776c). This adds an include for stdbool.h in zend_types.h, which we include in phongo_compat.h.

We can either leave the include in place here (to ensure we have stdbool.h available when needed on PHP < 8) or include it explicitly in every file that makes use of a `bool` type. I'm fine leaving it in phongo_compat.h for the time being.

alcaeus:
Ignoring Windows test failures for the time being. There were successful Windows builds on PHP 7.2 (which is usually the most critical version). That said, we'll have to revisit these tests as they are quite fragile (usually due to the same timeout when starting MongoDB).

src/phongo_compat.h
jmikola: If https://github.com/php/php-src/commit/4b77a158ef2850582aeb4834c588aba49942776c is the first place `zend_bool` is defined, I don't understand how `stdbool.h` would not already be included by that point. We're already including `Zend/zend_types.h` at the top of this file.

I remember encountering this before but I never fully wrapped my head around it.

OK to leave as-is but we can consider opening a ticket to get to the bottom of it later.

alcaeus: For PHP versions before 8.0 `zend_types.h` does not include `stdbool.h`, triggering the error there. Given that, I think including it in phongo_compat.h makes sense.

PR #1413, opened 2023-04-21
Bump cmb69/setup-php-sdk to 0.7

jmikola:
This pulls cmb69/setup-php-sdk@26487290af8a51fa8ce948f9c4ee18e2fa54529e, which migrates deprecated set-output commands to use $GITHUB_OUTPUT

----

Per [GitHub Actions: Deprecating save-state and set-output commands](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/):

> Starting 1st June 2023 workflows using save-state or set-output commands via stdout will fail with an error.

We may want to backport this to older branches to avoid any trouble if we need to create a patch release down the line.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -109,7 +109,7 @@ jobs:
 
       - name: Setup PHP SDK
         id: setup-php
-        uses: cmb69/setup-php-sdk@v0.6
+        uses: cmb69/setup-php-sdk@v0.7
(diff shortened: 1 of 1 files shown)


alcaeus (approved):

jmikola:
Rebased to v1.15. Will come back to this next week and merge once CI looks good.

PR #1415, opened 2023-04-26
PHPC-2210 and PHPC-2211: Fix direct BSON copy and toJson memory leak

jmikola:
https://jira.mongodb.org/browse/PHPC-2210
https://jira.mongodb.org/browse/PHPC-2211

### DIFF
--- a/src/phongo_bson_encode.c
+++ b/src/phongo_bson_encode.c
@@ -383,13 +383,13 @@ try_again:
 
 /* This is based on bson_copy_to_excluding_noinit() and is necessary because
  * bson_copy_to() cannot be used with a bson_t allocated with bson_new(). */
-static void phongo_bson_copy_to_noinit (const bson_t *src, bson_t *dst)
+static void phongo_bson_copy_to_noinit(const bson_t* src, bson_t* dst)
 {
 	bson_iter_t iter;
 
-	if (bson_iter_init (&iter, src)) {
-		while (bson_iter_next (&iter)) {
-			if (!bson_append_iter (dst, NULL, 0, &iter)) {
+	if (bson_iter_init(&iter, src)) {
+		while (bson_iter_next(&iter)) {
+			if (!bson_append_iter(dst, NULL, 0, &iter)) {
 				/* This should not be able to happen since we are copying from
 				 * within a valid bson_t. */
 				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "Error copying \"%s\" field from source document", bson_iter_key(&iter));
--- a/src/phongo_bson_encode.c
+++ b/src/phongo_bson_encode.c
@@ -381,6 +381,24 @@ try_again:
 	}
 }
 
+/* This is based on bson_copy_to_excluding_noinit() and is necessary because
+ * bson_copy_to() cannot be used with a bson_t allocated with bson_new(). */
+static void phongo_bson_copy_to_noinit (const bson_t *src, bson_t *dst)
+{
+	bson_iter_t iter;
+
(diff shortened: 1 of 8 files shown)


tests/bulk/bulkwrite-insert-002.phpt
```diff
@@ -0,0 +1,29 @@
...
+<?php
+
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = create_test_manager();
+
+$document = MongoDB\BSON\Document::fromJSON('{ "_id": 2 }');
+
+$bulk = new MongoDB\Driver\BulkWrite();
+$insertedId = $bulk->insert($document);
```
jmikola, line 17: While implementing this test, I realized there's a separate bug where `_id` fields are not generated for Document instances. I opened [PHPC-2216](https://jira.mongodb.org/browse/PHPC-2216) to track that.

src/phongo_bson_encode.c
```diff
@@ -403,17 +421,15 @@ static void php_phongo_zval_to_bson_internal(zval* data, php_phongo_field_path*
 			if (instanceof_function(Z_OBJCE_P(data), php_phongo_document_ce)) {
 				php_phongo_document_t* intern = Z_DOCUMENT_OBJ_P(data);
 
-				bson_destroy(bson);
-				bson_copy_to(intern->bson, bson);
+				phongo_bson_copy_to_noinit(intern->bson, bson);
```
jmikola, line 424: Note that only `php_phongo_zval_to_bson_internal` was affected.

While `php_phongo_bson_append_object` also handles direct copying of Document and PackedArray data, it does so via `bson_append_document` and `bson_append_array()` functions, respectively. There was no problem there.

alcaeus: Ah. I assumed that I could destroy the existing `bson_t` instance and have `bson_copy_to` handle the rest. If I understand this correctly, the issue is that `bson_destroy(bson)` frees the memory `bson` is pointing at. `bson_copy_to` doesn't catch this as it asserts on `dst`, which is still a valid `bson_t*` (i.e. not `NULL`) but pointing at memory no longer owned. Classic use of a footgun provided by C I have to admit, and a great catch on your part!

tests/query/query-ctor-007.phpt
```diff
@@ -0,0 +1,54 @@
...
+$bulk->insert(['_id' => 1]);
+$bulk->insert(['_id' => 2]);
+$manager->executeBulkWrite(NS, $bulk);
+
+$filter = MongoDB\BSON\Document::fromJSON('{ "_id": { "$gt": 1 } }');
+$query = new MongoDB\Driver\Query($filter);
+
+var_dump($query);
+
+$cursor = $manager->executeQuery(NS, $query);
```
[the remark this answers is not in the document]
alcaeus: I agree - I believe the tests you added cover every aspect of potential document usage. 👍

alcaeus (approved):
LGTM. I've pushed a commit with clang-format to fix code style issues stemming from copying the logic over from libbson.

src/phongo_bson_encode.c
jmikola: That's part of the problem, but removing `bson_destroy()` also doesn't help us. The crux of the matter is that `bson_copy_to()` isn't designed to work on an allocated `bson_t` struct, which always exist in an initialized state.

I spoke to @kevinAlbs about this and we think it's technically possible to alter `bson_copy_to()` to support both, but it doesn't seem necessary at this time -- and a local copy method in PHPC works just fine for our needs.

PR #1416, opened 2023-04-27
PHPC-2216: Generate _id field when inserting Document or PackedArray instances

alcaeus:
PHPC-2216

Test is named to not conflict with the new test added in #1415.

### DIFF
--- a/tests/bulk/bulkwrite-insert-003.phpt
+++ b/tests/bulk/bulkwrite-insert-003.phpt
@@ -1,0 +1,76 @@
+--TEST--
+MongoDB\Driver\BulkWrite::insert() appends ID when inserting Document instances
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_live(); ?>
+<?php skip_if_not_clean(); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = create_test_manager();
+
+$document = MongoDB\BSON\Document::fromPHP(['foo' => 'bar']);
+$array = MongoDB\BSON\PackedArray::fromPHP(['bar']);
+
+$bulk = new MongoDB\Driver\BulkWrite();
+
+$insertedDocumentId = $bulk->insert($document);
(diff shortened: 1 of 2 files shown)


tests/bulk/bulkwrite-insert-003.phpt
```diff
@@ -0,0 +1,76 @@
...
+
+$manager = create_test_manager();
+
+$document = MongoDB\BSON\Document::fromPHP(['foo' => 'bar']);
+$array = MongoDB\BSON\PackedArray::fromPHP(['bar']);
+
+$bulk = new MongoDB\Driver\BulkWrite();
+
+$insertedDocumentId = $bulk->insert($document);
+$insertedArrayId = $bulk->insert($array);
```
jmikola, line 19: Is there a use case for allowing `php_phongo_zval_to_bson_internal` to encode a PackedArray as a root document? I'm not sure if the code path is needed by anything in PHPC, or if you just added it because an array is technically just a `bson_t`.

Would it make sense to throw instead?

jmikola (commented):
Code changes LGTM, but I do have a question about PackedArray.

tests/bulk/bulkwrite-insert-003.phpt
alcaeus: `php_phongo_zval_to_bson_internal` (and `php_phongo_zval_to_bson`) is used in several places, and we can't make an assumption that whatever we're converting will be a root document to be inserted to the database. If we want to prevent this, the correct course of action would be to catch this in the various `BulkWrite` operations, preventing the usage as a document in `insert`. `BulkWrite::update` will be a bit more complicated: while the replacement document shouldn't be a document directly, the `PackedArray` instance could also hold an aggregation pipeline to be used for the update, in which case we'd want to convert and send it to the server as is.

Long story short, I don't think this is something we should worry about. The main reason for the separation was to ignore the stored keys when iterating a `PackedArray` instance (as opposed to `Document` where they are relevant).

alcaeus: Created PHPC-2219 to track this.

PR #1417, opened 2023-05-03
PHPC-2214: Int64 improvements

alcaeus:
PHPC-2214

This PR makes `Int64` objects more usable by allowing the construction of said objects on both 32 and 64 bit platforms, as well as getting it to (almost) behave like a number.

The first part (PHPC-1949) allows creating objects by making the constructor public. The constructor argument can be an int or string (to allow using 64-bit values on 32-bit platforms).

The second part concerns casting (PHPC-2213). With `Int64` already being a string able object, the cast_object helper now takes care of casting to int and bool types. The cast also supports the `_IS_NUMBER` pseudo-type available in PHP 7.3 and later. When casting an `Int64` instance containing a value that exceeds the `INT32` range to an int on 32-bit platforms, a warning is emitted to let the user know that the value is truncated.
Note that the cast helper explicitly does not support casting to a float. Depending on the precision, the value would still get truncated. With truncation being rather unpredictable in this case, I decided it was better to let PHP handle those shenanigans.

Last but not least, the `Int64` class now overloads certain operators (PHPC-2212). For the time being, the following operators are supported: `+`, `-`, `*`, `/`, `%`, `<<`, `>>`, and `**`. Operator overloading only supports operations with an int or `Int64` value. Other values and other operations are not implemented and will result in the default handling for PHP. For example, creating an `Int64` and adding it to a float results in PHP trying to cast the int64 to a number, then adding the float to it. While the cast to a number results in an int, the addition of a float value makes the result a float:
```
$int64 = new Int64(10);
var_dump($int64 + 3.14); // Will result in double(13.14)
```

In other cases, the result will be an `Int64` instance, even if the value is within the 32-bit range:
```
$int64 = new Int64(10);
var_dump($int64 * 2); // Will result in Int64(20)
```


src/BSON/Int64.c
```diff
@@ -244,6 +263,100 @@ static int php_phongo_int64_compare_objects(zval* o1, zval* o2)
...
+		case ZEND_SL:
+			ZVAL_INT64_OBJ(result, value1 << value2);
+			return SUCCESS;
+
+		case ZEND_SR:
+			ZVAL_INT64_OBJ(result, value1 >> value2);
+			return SUCCESS;
+
+		case ZEND_POW:
+			ZVAL_INT64_OBJ(result, pow(value1, value2));
```
jmikola, line 352: Note that `pow()` returns a `float`, so this can easily result in silent truncation based on `ZVAL_INT64_OBJ` calling `php_phongo_bson_new_int64` and C implicitly casting the rturn value to an `int64_t`.

You can contrast this with [`pow_function` in `zend_operators.c`](https://github.com/php/php-src/blob/php-8.2.5/Zend/zend_operators.c#L1288), which has much more complex handling to decide if the result can be represented as an integer or must utilize a float.

If we intentionally omit this opcode here, we could rely on `pow_function` to convert the Int64 to a numeric type. That would make `$int64a ** $int64b` would be more likely to return a PHP float, but it'd be more predictable and consistent with how `**` might work with regular integers.

Whatever we decide, we should ensure there are tests for the overflow conditions. Also, some tests for a negative `value2` operand. Currently, I think a negative `value2` would always produce `Int64(0)` as a result.

tests/bson/bson-int64-operation-001.phpt
```diff
@@ -0,0 +1,72 @@
+--TEST--
+MongoDB\BSON\Int64 operations
```
jmikola, line 2: Note my previous comments on the operation handler. This file will likely get much longer if you cover the overflow/underflow cases, so you may want to split it up by operation or groups of operations.

tests/bson/bson-int64-cast_error-001.phpt
```diff
@@ -0,0 +1,23 @@
...
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/basic.inc';
+
+$int64 = new MongoDB\BSON\Int64(2);
+
+echo raises(function () use ($int64) {
+    var_dump((float) $int64);
+}, E_WARNING), "\n";
```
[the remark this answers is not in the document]
alcaeus: Added and dropped the other file.

tests/bson/bson-int64-operation_error-002.phpt
```diff
@@ -0,0 +1,31 @@
...
+
+// Unsupported type. This yields a warning about a non-numeric value
+echo raises(function() use ($value) {
+    var_dump($value + 'foo');
+}, E_WARNING), "\n";
+
+// Division by zero
+echo throws(function() use ($value) {
+    var_dump($value / 0);
+}, DivisionByZeroError::class), "\n";
```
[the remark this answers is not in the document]
alcaeus: I'd ignore the weird behaviour in PHP 7.x and always throw - since there is no well-defined result I'd rather not introduce such behaviour.

src/BSON/Int64.c
```diff
@@ -244,6 +263,100 @@ static int php_phongo_int64_compare_objects(zval* o1, zval* o2)
...
+		value2 = Z_LVAL_P(op2);
+	} else if (Z_TYPE_P(op2) == IS_OBJECT && Z_OBJCE_P(op2) == php_phongo_int64_ce) {
+		value2 = Z_INT64_OBJ_P(op2)->integer;
+	} else {
+		return FAILURE;
+	}
+
+	switch (opcode) {
+		case ZEND_ADD:
+			ZVAL_INT64_OBJ(result, value1 + value2);
```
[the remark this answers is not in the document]
alcaeus: This has taken me a lot longer than I'd like to admit, but I finally got somewhere. Here's what's going on:
* For `ZEND_ADD` and `ZEND_SUB` I extracted the method used in [`fast_long_add_function`](https://github.com/php/php-src/blob/7304b56f11771ba4ed51c0053c47b94c1b9bb9a3/Zend/zend_operators.h#L699..L700) to check for a potential overflow and return the result as a double.
* For `ZEND_MUL`, I duplicated the behaviour used in [`ZEND_SIGNED_MULTIPLY_LONG`](https://github.com/php/php-src/blob/master/Zend/zend_multiply.h#L112..L115) which compares the int64 result to the result of the same operation in a `long double`, which has enough precision to handle the int64 range.
* For `ZEND_DIV`, the only way it can overflow is the case of `INT64_MIN / -1`. I explicitly covered that case and also found an issue where the result was always treated as an int even if it should've been a float. This is now properly covered by tests.
* `ZEND_SL` does not produce overflows ([see 3v4l.org](https://3v4l.org/uksb3)), and I added a check that `INT64_MAX << 1` does indeed not overflow
* `ZEND_SR` cannot overflow or underflow by definition as it shifts everything to the right by 1 bit
* `ZEND_POW` was fixed accordingly.

src/BSON/Int64.c
```diff
@@ -244,6 +263,100 @@ static int php_phongo_int64_compare_objects(zval* o1, zval* o2)
...
+			if (value2 == 0) {
+				zend_throw_exception(zend_ce_division_by_zero_error, "Division by zero", 0);
+				return FAILURE;
+			}
+
+			ZVAL_INT64_OBJ(result, value1 / value2);
+			return SUCCESS;
+
+		case ZEND_MOD:
+			ZVAL_INT64_OBJ(result, value1 % value2);
```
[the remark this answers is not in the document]
alcaeus: I was so happy I got to check against a division by 0 that I forgot to add this one. This is now fixed and a test added to the corresponding `_error` test.

tests/bson/bson-int64-get_properties-001.phpt
```diff
@@ -4,13 +4,13 @@ MongoDB\BSON\Int64 get_properties handler (get_object_vars)
...
-    unserialize('C:18:"MongoDB\BSON\Int64":48:{a:1:{s:7:"integer";s:20:"-9223372036854775808";}}'),
-    unserialize('C:18:"MongoDB\BSON\Int64":28:{a:1:{s:7:"integer";s:1:"0";}}'),
+    '9223372036854775807',
+    '-9223372036854775808',
+    0,
 ];
 
 foreach ($tests as $test) {
-    var_dump(get_object_vars($test));
+    var_dump(get_object_vars(new MongoDB\BSON\Int64($test)));
```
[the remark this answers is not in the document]
alcaeus: I originally did this for simplicity, but creating them outside of the loop makes potential exceptions more useful by referring directly to the item where creation failed, instead of pointing to the loop and leaving us to figure out which one failed. Updated.

tests/bson/bson-int64-ctor-002.phpt
```diff
@@ -0,0 +1,18 @@
+--TEST--
+MongoDB\BSON\Int64 constructor (64-bit int value)
+--SKIPIF--
+<?php if (8 !== PHP_INT_SIZE) { die('skip Only for 64-bit platform'); } ?>
+--FILE--
+<?php
+
+var_dump(new MongoDB\BSON\Int64(9223372036854775807));
```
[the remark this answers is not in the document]
alcaeus: Added.

tests/bson/bson-int64-cast_error-002.phpt
```diff
@@ -0,0 +1,23 @@
+--TEST--
+MongoDB\BSON\Int64 casting errors (PHP < 8)
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_php_version('>=', '8.0'); ?>
```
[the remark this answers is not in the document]
alcaeus: I believe so - given that there won't be any 7.x releases anymore this is likely irrelevant, but we might want to keep this in mind for future versions. That said, the file is gone now.

tests/bson/bson-int64-cast-002.phpt
```diff
@@ -0,0 +1,26 @@
+--TEST--
+MongoDB\BSON\Int64 casting (64-bit platform)
+--SKIPIF--
+<?php if (8 !== PHP_INT_SIZE) { die('skip Only for 64-bit platform'); } ?>
+--FILE--
+<?php
+
+$int64 = new MongoDB\BSON\Int64(9223372036854775807);
```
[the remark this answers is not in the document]
alcaeus: That's some interesting behaviour around `PHP_INT_MIN` - I did not know that. I added it to the test.

tests/bson/bson-int64-cast-003.phpt
```diff
@@ -0,0 +1,30 @@
+--TEST--
+MongoDB\BSON\Int64 casting (32-bit platform)
+--SKIPIF--
+<?php if (4 !== PHP_INT_SIZE) { die('skip Only for 32-bit platform'); } ?>
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/basic.inc';
+
+$int64 = new MongoDB\BSON\Int64('9223372036854775807');
```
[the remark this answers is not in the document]
alcaeus: Added

tests/bson/bson-int64-operation-001.phpt
alcaeus: I split this test into several tests for various operations (arithmetic, exponentiation, bitwise operators, others).

tests/bson/bson-int64-ctor-001.phpt
```diff
@@ -0,0 +1,32 @@
+--TEST--
+MongoDB\BSON\Int64 constructor
+--FILE--
+<?php
+
+$tests = [
+    2,
+    '2',
+    '9223372036854775807',
```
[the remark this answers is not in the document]
alcaeus: Added tests for invalid values. Also added `0x2` for good measure.

src/BSON/Int64.c
```diff
@@ -244,6 +263,100 @@ static int php_phongo_int64_compare_objects(zval* o1, zval* o2)
...
+
+		case ZEND_SR:
+			ZVAL_INT64_OBJ(result, value1 >> value2);
+			return SUCCESS;
+
+		case ZEND_POW:
+			ZVAL_INT64_OBJ(result, pow(value1, value2));
+			return SUCCESS;
+
+		default:
```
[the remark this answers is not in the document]
alcaeus: * `ZEND_BOOL_NOT` and `ZEND_BOOL_XOR` worked through implicitly casting, and I added tests for that.
* Increment/decrement is converted to `ZEND_ADD`/`ZEND_SUB`, but tests were added
* Bitwise operations were added
* `fast_long_add_function` (and all other built-in functions) can't be used due to them using `zend_long`, which on 32-bit platforms is `long` instead of `long long`. If we ever drop 32-bit support we can revisit this, but for now we can't leverage them.

src/BSON/Int64.c
```diff
@@ -244,6 +263,229 @@ static int php_phongo_int64_compare_objects(zval* o1, zval* o2)
...
+	PHONGO_GET_INT64(value1, op1);
+
+	switch (opcode) {
+		case ZEND_ADD:
+			PHONGO_GET_INT64(value2, op2);
+
+			lresult = value1 + value2;
+
+			// Check for overflow by comparing the sign bits of the operands and the result
+			if ((value1 & INT64_MIN) == (value2 & INT64_MIN) && (value1 & INT64_MIN) != (lresult & INT64_MIN)) {
```
jmikola, line 356: I assume this was adapted from the [`#else` block of `fast_long_add_function()`](https://github.com/php/php-src/blob/php-8.2.6/Zend/zend_operators.h#L691-L703). If so, can you leave a comment for future reference?

I was initially confused as to why you were using `INT64_MIN` here. Defining an `INT64_SIGN_MASK` (or similar) constant from `INT64_MIN` may make this and the `ZEND_SUB` case below more readable.

Side note: this isn't a critical code path, but I found it interesting that php-src performs the addition twice. I think it makes sense to do it once like you've done here with `lresult`, but I'm curious if there's a particular reason php-src did not. :shrug: 

----

And if you'd like to add the relative conditions to the comment (like I suggested for `ZEND_SUB` below):

 * Adding a positive to a positive yields a negative
 * Adding a negative to a negative (i.e. subtraction) yields a positive

src/BSON/Int64.c
```diff
@@ -244,6 +263,229 @@ static int php_phongo_int64_compare_objects(zval* o1, zval* o2)
...
+			}
+
+			return SUCCESS;
+
+		case ZEND_SUB:
+			PHONGO_GET_INT64(value2, op2);
+
+			lresult = value1 - value2;
+
+			// Check for overflow by comparing the sign bits of the operands and the result
```
jmikola, line 369: If you add a reference to `fast_long_add_function()` above, you can also reference `fast_long_sub_function()` here. I see the conditional logic is slightly different.

Just working through this truth table in my head, the two overflow cases (i.e. use a double) are:

 * Subtracting a negative from a positive (i.e. addition) yields a negative
 * Subtracting a positive from a negative yields a positive

That seems short enough to capture in a comment to help future readers.

src/BSON/Int64.c
```diff
@@ -244,6 +263,229 @@ static int php_phongo_int64_compare_objects(zval* o1, zval* o2)
...
+
+		case ZEND_MUL:
+			PHONGO_GET_INT64(value2, op2);
+
+			{
+				int64_t     lres  = value1 * value2;
+				long double dres  = (long double) value1 * (long double) value2;
+				long double delta = (long double) lres - dres;
+
+				// Guard against overflow, return double if so
```
jmikola, line 386: I assume this is based on [`ZEND_SIGNED_MULTIPLY_LONG`](https://github.com/php/php-src/blob/php-8.2.6/Zend/zend_multiply.h#L155) if you'd like to add a reference.

src/BSON/Int64.c
```diff
@@ -244,6 +263,229 @@ static int php_phongo_int64_compare_objects(zval* o1, zval* o2)
...
+
+		case ZEND_DIV:
+			PHONGO_GET_INT64(value2, op2);
+			if (value2 == 0) {
+				zend_throw_exception(zend_ce_division_by_zero_error, "Division by zero", 0);
+				return FAILURE;
+			}
+
+			// INT64_MIN / 1 exceeds the int64 range, return double in that case.
+			// Same goes for any other division that would not yield an integer.
```
jmikola, line 404: Consider a reference to [`div_function_base`](https://github.com/php/php-src/blob/php-8.2.6/Zend/zend_operators.c#L1324).

tests/bson/bson-int64-ctor-002.phpt
```diff
@@ -0,0 +1,23 @@
...
+var_dump(new MongoDB\BSON\Int64(~9223372036854775807));
```
jmikola, line 9:
```suggestion
/* Note: -9223372036854775808 is parsed as a double, since PHP applies the
 * negation operator after parsing the integer, which exceeds INT64_MAX.
 * Therefore, we can either negate INT64_MAX or use PHP_INT_MIN. */
var_dump(new MongoDB\BSON\Int64(~9223372036854775807));
```

I won't push for including this every time we use `~9223372036854775807` in a test, but having it documented at least once is a good idea. The ctor test seemed like the most recognizable place to do so.

tests/bson/bson-int64-cast_error-001.phpt
```diff
@@ -0,0 +1,20 @@
...
+echo raises(function () use ($int64) {
+    var_dump((float) $int64);
+}, E_NOTICE | E_WARNING), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+OK: Got %r(E_NOTICE|E_WARNING)%r
+Object of class MongoDB\BSON\Int64 could not be converted to float
```
jmikola, line 19: Forgive me if I'm beating a dead horse, but does it still make sense to prohibit an explicit float case knowing that we can so easily return a float from the Int64 arithmetic operations?

tests/bson/bson-int64-ctor_error-001.phpt
```diff
@@ -0,0 +1,36 @@
+--TEST--
+MongoDB\BSON\Int64 constructor with invalid data
+--FILE--
+<?php
+
+require_once __DIR__ . "/../utils/basic.inc";
+
+$tests = [
+    '0x2', // Hex data
```
jmikola, line 9: This makes me wonder how we handle other alternative [integer literals](https://www.php.net/manual/en/language.types.integer.php):

```
<?php
$a = 1234; // decimal number
$a = 0123; // octal number (equivalent to 83 decimal)
$a = 0o123; // octal number (as of PHP 8.1.0)
$a = 0x1A; // hexadecimal number (equivalent to 26 decimal)
$a = 0b11111111; // binary number (equivalent to 255 decimal)
$a = 1_234_567; // decimal number (as of PHP 7.4.0)
```

I won't push for everything, but adding one more test for an octal with leading zeroes seems worthwhile, even if it doesn't result in an failure and ends up in the normal ctor test. It'd be good to have that behavior documented in a test.

tests/bson/bson-int64-operation-001.phpt
```diff
@@ -0,0 +1,64 @@
...
+  string(1) "5"
+}
+object(MongoDB\BSON\Int64)#%d (%d) {
+  ["integer"]=>
+  string(1) "1"
+}
+float(9.22%dE+18)
+float(-9.22%dE+18)
+float(1.844%dE+19)
+float(4.61%dE+18)
```
jmikola, line 57: Is PHP consistent about using "E" for scientific notation? Unrelated to PHP, I know we had an issue in BSON tests where "E" or "e" might be used on different platforms.

The [grammar definition](https://www.php.net/manual/en/language.types.float.php) in PHP.net shows `[eE]`, so I might suggest using an inline regex pattern such as `%r(e|E)%r` or `%r[eE]%r`.

This applies to anywhere else you might be expecting scientific notation in EXPECTF.

tests/bson/bson-int64-operation-003.phpt
```diff
@@ -0,0 +1,43 @@
+--TEST--
+MongoDB\BSON\Int64 operations: increment/decrement
+--FILE--
+<?php
+
+$value = new MongoDB\BSON\Int64(10);
+
+var_dump($value++); // This will print 11 instead of 10 due to PHP not implementing operator overloading for increment
```
jmikola, line 8: Consider a comment above all of this explaining that inc/dec operations _do not_ modify the object itself, just so we have that documented somewhere.

I remember we talked about that behavior, but I was reminded when reading the test wondering why `$value` always seems to reset to 10 for each operation.

tests/bson/bson-int64-serialization-001.phpt
```diff
@@ -7,14 +7,14 @@ MongoDB\BSON\Int64 serialization (Serializable interface)
...
-    unserialize('C:18:"MongoDB\BSON\Int64":28:{a:1:{s:7:"integer";s:1:"0";}}'),
+    '9223372036854775807',
+    '-9223372036854775808',
+    0,
 ];
 
 foreach ($tests as $test) {
-    var_dump($test);
-    var_dump($s = serialize($test));
+    var_dump($int64 = new MongoDB\BSON\Int64($test));
```
jmikola, line 16: Consider constructing the Int64 object outside of the loop to be consistent with other tests such as `bson-int64-get_properties-001.phpt`.

Related: https://github.com/mongodb/mongo-php-driver/pull/1417#discussion_r1190723745

tests/bson/bson-int64-serialization-002.phpt
```diff
@@ -7,14 +7,14 @@ MongoDB\BSON\Int64 serialization (__serialize and __unserialize)
...
-    unserialize('O:18:"MongoDB\BSON\Int64":1:{s:7:"integer";s:1:"0";}'),
+    '9223372036854775807',
+    '-9223372036854775808',
+    0,
 ];
 
 foreach ($tests as $test) {
-    var_dump($test);
-    var_dump($s = serialize($test));
+    var_dump($int64 = new MongoDB\BSON\Int64($test));
```
jmikola, line 16: Consider constructing outside of the loop, per my previous comment.

src/BSON/Int64.c
alcaeus: Added references here and for other arithmetic operations.

tests/bson/bson-int64-ctor-002.phpt
alcaeus: Changed this to use PHP_INT_MIN as we're on a 64-bit platform - that leaves no instances of negating `INT64_MAX`.

tests/bson/bson-int64-ctor_error-001.phpt
alcaeus: The alternative notations work fine when used in integer literals as the parser handles this and passes us an integer like any other. However, these values can't be used in strings, so it's not possible to pass octal numbers from the 64-bit range on 32-bit platforms. I don't think this is a particularly popular use case, but it may be worth documenting nonetheless.

tests/bson/bson-int64-operation-001.phpt
alcaeus: Updated.

tests/bson/bson-int64-serialization-001.phpt
alcaeus: Updated here and below.

jmikola (approved):
Just some suggestions that you can address on your own. Thanks for all the work on this!

jmikola (approved):

tests/bson/bson-int64-operation-003.phpt
jmikola: Quick follow-up, as I just saw [this reply in room11](https://chat.stackoverflow.com/transcript/message/56345198#56345198) to your question about inc/dec operations on objects:

> PHP allows you to overload operators in any way you want, thus you're responsible to make a new clone for the result - hopefully Int64 is already immutable. I've learned how operator overloading is working from [gmp implementation](https://github.com/php/php-src/blob/8d0345d94ef98bfe29df264b338aeef16ddefda9/ext/gmp/gmp.c#L410C20-L427), as you can see there - if result and op1 are the same, gmp makes clone of op1.

It looks like the `do_operation` handler always initializes a new `result` zval, but I can't be sure. Can you add a test or two to assert that Int64 objects are immutable?

I suppose that could be accomplished with something like `$value !== $value++`.

PR #1418, opened 2023-05-04
PHPC-2220: Return int64 values for server connection ID

alcaeus:
PHPC-2220
Also fixes PHPC-1941 (using OP_MSG to authenticate if the server supports it) by using the new libmongoc version.

Due to BC constraints, the API of `getServerConnectionId` in the command events can't be changed and we're forced to return an int value here. For that reason, the methods now raise an `E_WARNING` if a 64-bit value is truncated on a 32-bit platform. This is in line with previous behaviour, where the value was also truncated. Values within the 32-bit range and all values on 64-bit platforms will be returned as integers.


jmikola (approved):
I'm very much OK with the truncation warning, as we're doing that elsewhere in a few places. IMO this isn't an essential API on the same level of needing to preserve BSON types in user data.

PR #1421, opened 2023-05-12
PHPC-2222: Ensure packed arrays return PackedArray instances when converted to BSON

alcaeus:
PHPC-2222

While PHPC-2219 may make changes to what can be handled as a root document, serialising a `PackedArray` instance through toPHP with a type map that specifies a BSON root type should always yield a `PackedArray` instance, no matter how absurd the use case may seem.

### DIFF
--- a/src/phongo_bson.c
+++ b/src/phongo_bson.c
@@ -1042,10 +1042,10 @@ bool php_phongo_bson_to_zval_ex(const bson_t* b, php_phongo_bson_state* state)
 
 		if (state->is_visiting_array) {
 			object_init_ex(&obj, php_phongo_packedarray_ce);
-			bson = &Z_DOCUMENT_OBJ_P(&obj)->bson;
+			bson = &Z_PACKEDARRAY_OBJ_P(&obj)->bson;
 		} else {
 			object_init_ex(&obj, php_phongo_document_ce);
-			bson = &Z_PACKEDARRAY_OBJ_P(&obj)->bson;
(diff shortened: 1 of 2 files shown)


src/phongo_bson.c
```diff
@@ -1035,15 +1035,20 @@ bool php_phongo_bson_to_zval_ex(const bson_t* b, php_phongo_bson_state* state)
...
+		zval     obj;
+		bson_t** bson;
 
-		object_init_ex(&obj, php_phongo_document_ce);
+		if (state->is_visiting_array) {
+			object_init_ex(&obj, php_phongo_packedarray_ce);
+			bson = &Z_DOCUMENT_OBJ_P(&obj)->bson;
+		} else {
+			object_init_ex(&obj, php_phongo_document_ce);
+			bson = &Z_PACKEDARRAY_OBJ_P(&obj)->bson;
```
jmikola, line 1048:
```suggestion
			bson = &Z_PACKEDARRAY_OBJ_P(&obj)->bson;
		} else {
			object_init_ex(&obj, php_phongo_document_ce);
			bson = &Z_DOCUMENT_OBJ_P(&obj)->bson;
```

This bug could easily go unnoticed because the structs have identical memory layouts.

jmikola (changes requested):

src/phongo_bson.c
alcaeus: Oof, nice catch!

PR #1424, opened 2023-05-16
Show how to verify installation after building from source

GromNaN:
For contributing, I had to search for the `mongodb.so` file and find a way to validate the extension was correctly compiled.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,13 @@ phpize > /dev/null && \
 make clean > /dev/null && make all > /dev/null && make install
 ```
 
+To verify that the installation was successful, run the following command, which
+will report `phpinfo()` output for the extension:
+
+```
+$ php --ri mongodb
+```
+
 ## Generating arginfo from stub files
 
 Arginfo structures are generated from stub files using the `gen_stub.php`


CONTRIBUTING.md
```diff
@@ -21,6 +21,20 @@ phpize > /dev/null && \
...
+
+```
+php -d extension=$(pwd)/.libs/mongodb.so -r "echo (new ReflectionExtension('mongodb'))->getVersion();"
+```
+
+For larger tests, configure the extension in your `php.ini` file.
+
+```ini
+extension="<replace-with-absolute-project-path>/mongo-php-driver/.libs/mongodb.so"
+```
```
alcaeus, line 35: What's the advantage of this over `make install`?

GromNaN: Nothing. I missed that part of the command when running step by step.

GromNaN:
Updated this PR to add a manual check in the doc.

CONTRIBUTING.md
```diff
@@ -21,6 +21,14 @@ phpize > /dev/null && \
 make clean > /dev/null && make all > /dev/null && make install
 ```
 
+To verify that the installation was successful, run the following command which indicates the version of the installed
+extension. It must end with `-dev` suffix.
+
+```
+php -r "echo (new ReflectionExtension('mongodb'))->getVersion();"
```
alcaeus, line 28: It might also be sensible to point out `php --ri mongodb`, which will print the MongoDB extension portion of phpinfo, including the build configuration (e.g. libmongoc/libmongocrypt versions being used).

GromNaN: That's much better. 
```
php --ri mongodb

mongodb

MongoDB support => enabled
MongoDB extension version => 1.16.0-dev
MongoDB extension stability => devel
libbson bundled version => 1.24.0-20230504+git89717a3beb
libmongoc bundled version => 1.24.0-20230504+git89717a3beb
libmongoc SSL => enabled
libmongoc SSL library => Secure Transport
libmongoc crypto => enabled
libmongoc crypto library => Common Crypto
libmongoc SASL => enabled
libmongoc ICU => disabled
libmongoc compression snappy => disabled
libmongocrypt bundled version => 1.8.0-alpha0
libmongocrypt crypto library => Common Crypto
crypt_shared library version => unknown

Directive => Local Value => Master Value
```

CONTRIBUTING.md
```diff
@@ -21,6 +21,14 @@ phpize > /dev/null && \
 make clean > /dev/null && make all > /dev/null && make install
 ```
 
+To verify that the installation was successful, run the following command which indicates the version of the installed
+extension. It must end with `-dev` suffix.
+
+```
+php --ri mongodb
```
jmikola, line 28:
```suggestion
$ php --ri mongodb
```

For consistency with earlier shell snippets.

CONTRIBUTING.md
```diff
@@ -21,6 +21,14 @@ phpize > /dev/null && \
 make clean > /dev/null && make all > /dev/null && make install
 ```
 
+To verify that the installation was successful, run the following command which indicates the version of the installed
+extension. It must end with `-dev` suffix.
```
jmikola, line 25:
```suggestion
To verify that the installation was successful, run the following command, which
will report `phpinfo()` output for the extension:
```

A few changes here:

 * Applying 80-character wrap for consistency with other lines
 * Added comma before "which" (see: [grammar rules](https://www.grammarly.com/blog/comma-before-which/))
 * Referred to `phpinfo()` since that's directly what `--ri` reports. We can optionally added "(including its version)" but I don't think it's necessary for this audience.
 * Even before the previous change, I was going to suggest removing the bit about the `-dev` suffix as these are general rules for building from source. Even though the shell example above starts with a fresh git clone, these are the same instructions one might follow when building a tag from source. This is basically a local copy of [Building the MongoDB PHP Driver from source ](https://www.php.net/manual/en/mongodb.installation.manual.php) in the PHP.net docs.

CONTRIBUTING.md
```diff
@@ -21,6 +21,14 @@ phpize > /dev/null && \
...
+
+```
+php --ri mongodb
+```
+
+
```
jmikola, line 31:
```suggestion
```

Not sure if the extra line was intentional, but other sections use a single break before the next title. I'm open to changing this but if so we should just do it across the entire file (same PR or a different is OK by me).

jmikola (approved):

PR #1429, opened 2023-05-29
Increase default timeout for starting MongoDB service on Windows

jmikola:
This is based on suggestions from actions/runner-images#5949

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -133,6 +133,7 @@ jobs:
 
       - name: Start MongoDB
         run: |
+          reg add HKLM\System\CurrentControlSet\Control /v ServicesPipeTimeout /t REG_DWORD /d 300000 /f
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -133,6 +133,7 @@ jobs:
 
       - name: Start MongoDB
         run: |
+          reg add HKLM\System\CurrentControlSet\Control /v ServicesPipeTimeout /t REG_DWORD /d 300000 /f
```
jmikola, line 136: This did not have the desired effect, as the service still timed out: https://github.com/mongodb/mongo-php-driver/actions/runs/5109526170/jobs/9184410368?pr=1429#step:8:1

I've opened [PHPC-2233](https://jira.mongodb.org/browse/PHPC-2233) to consider a different approach.

PR #1432, opened 2023-06-13
PHPC-2133: Fix building with SRV support on FreeBSD

jmikola:
https://jira.mongodb.org/browse/PHPC-2133

Verified locally with an Atlas SRV connection string. I installed FreeBSD 13.2 in a virtual machine and compiled the driver for PHP 8.0 both with and without this patch.

I _think_ the patch might also apply to OpenBSD, where resolve functions exist in libc, but I'm not confident enough to enable that and would rather wait for an actual bug report to come our way before I expend the effort to test it myself.


alcaeus (approved):
LGTM. We can ignore Windows test failures in this case.

PR #1435, opened 2023-06-16
PHPC-2242, PHPC-2243, PHPC-2244: BSON class improvements

alcaeus:
PHPC-2242
PHPC-2243
PHPC-2244

This PR cleans up the BSON class handling a bit. While working with the changes committed in #1412, I noticed that wrapping BSON values in `MongoDB\BSON\Value` instances was cumbersome and no longer necessary after changing the way int64 values are handled in raw BSON. To summarise, we decided that `MongoDB\BSON\Value::getValue()` would always return an `Int64` instance, even when the value would fit within a platform-sized integer. That removed the main incentive for the `MongoDB\BSON\Value` class, as we originally planned to use it to differentiate between 32-bit and 64-bit integers when the value was within the 32-bit range.

With that in mind, this PR undoes some of the changes introduced in the pull request. The main change is that `Document::get`, `PackedArray::get`, and `Iterator::current` now return `Int64` instances whenever a BSON long is encountered. For one, the raw BSON classes are designed to interact with BSON structures, so returning BSON values that can't be represented as PHP scalars as BSON class instances makes sense. Furthermore, with the recent changes to the `Int64` class (see #1417), instances of that class can be used like numbers in arithmetic operations and cast to integer if a user wishes to do so.
This change also applies to `Document::toPHP` and `PackedArray::toPHP`, which return BSON long values as `Int64` instances. The already existing `MongoDB\BSON\toPHP` function retains its current behaviour, as do any other instances of where we convert BSON data to PHP. The new behaviour only applies when interacting with the new BSON `Document` and `PackedArray` classes.

Last but not least, the `MongoDB\BSON\Value` class was dropped without a replacement. The improvements and refactoring made during the creation of the class remain in place.


tests/bson/bson-document-get-001.phpt
```diff
@@ -5,9 +5,14 @@ MongoDB\BSON\Document::get() key access
...
 
-$document = MongoDB\BSON\Document::fromPHP(['foo' => 'bar', 'bar' => 'baz']);
+$document = MongoDB\BSON\Document::fromPHP([
+    'foo' => 'bar',
+    'bar' => 'baz',
+    'int64' => new MongoDB\BSON\Int64(123),
+]);
 var_dump($document->get('foo'));
 var_dump($document->get('bar'));
+var_dump($document->get('int64'));
```
jmikola, line 15: Noted that you've added tests for always returning an Int64.

tests/bson/bson-toPHP-017.phpt
```diff
@@ -0,0 +1,23 @@
+--TEST--
+MongoDB\BSON\toPHP(): int64 values are returned as int scalars
```
jmikola, line 2: I'll note that 64-bit values on a 32-bit platform are an edge case where Int64 objects would still be returned, but I don't think we need to demonstrate that in a test. If you want you can add a note above `'int64' => new MongoDB\BSON\Int64(123),` explaining that you're using a 32-bit value to simplify testing.

jmikola (approved):
Minor suggestion but LGTM.

PR #1436, opened 2023-06-20
Add script to generate function map for static analysis tools

alcaeus:
The newly added script generates function maps in the format that PHPStan and Psalm use. The function map is stored in `scripts/functionmap.php`. The script is invoked via a make target, ensuring that the extension is built to generate the correct map:
```bash
$ make generate-function-map
```

The script generates maps for all functions, classes, and interfaces in the `MongoDB\BSON` and `MongoDB\Driver` namespaces.


scripts/create-functionmap.php
```diff
@@ -0,0 +1,95 @@
...
+$members = array_merge($classes, $interfaces);
+sort($members);
+
+$skippedMethods = ['__set_state', '__wakeup', '__serialize', '__unserialize'];
+
+foreach ($members as $member) {
+    $reflectionClass = new ReflectionClass($member);
+
+    foreach ($reflectionClass->getMethods(ReflectionMethod::IS_PUBLIC) as $method) {
+        if ($method->getDeclaringClass() != $reflectionClass && $method->getName() != '__toString') {
```
jmikola, line 62: AFAICT, this is skipping inherited methods that are _not_ `__toString()`. What is this covering?

I think our exception classes are the only cases of inheritance, and I'm not sure why `__toString()` would be relevant there.

scripts/create-functionmap.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+}
+
+$map = var_export($callmap, true);
+
+// Format output
+
+$replacements = [
+    '#array \(#' => '[',     // Short array syntax (opening)
+    '#\)#' => ']',           // Short array syntax (closing)
```
jmikola, line 81: Is it problematic that this could match any closing parenthesis?

scripts/create-functionmap.php
```diff
@@ -0,0 +1,95 @@
...
+    '#0 => #' => '',         // Numeric index for return type
+    '#\\\\{2}#' => '\\',     // Double backslash
+    '# => $\s+#m' => ' => ', // Newlines at start of array
+    '#\s+\[$\s+#m' => ' [',  // Newlines at start of array
+    '#,$\s+#m' => ', ',      // Newline after array elements
+    '#, \],\s+#' => "],\n",  // Newline after array elements
+    '#^\s*\'#m' => '    \'', // Fix indentation
+];
+
+echo preg_replace(
```
jmikola, line 91: I trust the replacements above are correct, but it might be more maintainable to use templates or `printf()` to generate the output. Iteration would be required for the class methods or functions, and then again for the parameters following the return type.

I'll defer to you, though.

scripts/create-functionmap.php
```diff
@@ -0,0 +1,95 @@
...
+        if (str_starts_with($name, $namespace) || str_starts_with($name, strtolower($namespace))) {
+            return true;
+        }
+    }
+
+    return false;
+};
+
+$getFunctionCallMapEntry = function(ReflectionFunctionAbstract $function): array {
+    $returnType = match(true) {
```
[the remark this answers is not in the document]
alcaeus: > This may be one of the first scripts to require PHP 8+.

Almost: gen_stub.php also requires the latest PHP version by default. Either way, I'll add a note about PHP 8 in a comment above, but as you already suggested the `match` statement will fail to parse.

scripts/create-functionmap.php
```diff
@@ -0,0 +1,95 @@
...
+        $function->hasReturnType() => (string) $function->getReturnType(),
+        $function->hasTentativeReturnType() => (string) $function->getTentativeReturnType(),
```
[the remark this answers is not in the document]
alcaeus: The `getReturnType` and `getTentativeReturnType` methods return a `ReflectionType` instance, not a string, hence the cast here.

scripts/create-functionmap.php
```diff
@@ -0,0 +1,95 @@
...
+    foreach ($function->getParameters() as $parameter) {
+        $parameterKey = $parameter->getName();
+        if ($parameter->isOptional()) {
+            $parameterKey .= '=';
+        }
+
+        $parameterType = (string) $parameter->getType();
+        if ($function->getName() === 'unserialize' && $parameter->getName() === 'serialized') {
+            $parameterType = 'string';
+        }
```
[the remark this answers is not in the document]
alcaeus: I took a trip through the PHP source code to figure out what's going on.

In PHP < 8.0, the parameter had no type and was called `$serialized`. In PHP 8.0, the parameter name was changed to `$data`, and a string type added. Since the PHP 7 interface has no type, we can't add our own type in the method signature directly, which is why there's a `@param` annotation. This in turn causes our definition to be without a parameter type, which is why an empty type is returned.

There are two options for this: we can add a conditional method signature for PHP >= 8, which fixes the parameter name and adds the `string` type. As these maps are generated on PHP 8 only, this would ensure correct signature generation without a workaround. The other option would be to add a comment explaining this and doing away with it once we drop the `Serializable` interface. `Serializable` is only used on PHP < 7.4, but one could argue that dropping the interface in a minor release could be a BC break. With all of this in mind, adding conditional method signatures seems like a sensible choice and doesn't cause any follow-up work outside of dropping the interface whenever we're ready to do so.

alcaeus:
I made some more changes. Previously, the script always generated the function map for the installed extension version. There is now a make target called `generate-function-map`, which invokes the script the same way `run-tests.php` is invoked. This ensures that the extension is built before invoking the script with the latest extension version loaded. Additionally, the script now places a `functionmap.php` file in the scripts directory. This file is not committed to git, but it allows reformatting the code a little easier before copying it over to the files for PHPStan and Psalm.

scripts/create-functionmap.php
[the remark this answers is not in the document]
jmikola: > The other option would be to add a comment explaining this and doing away with it once we drop the Serializable interface. Serializable is only used on PHP < 7.4, but one could argue that dropping the interface in a minor release could be a BC break

I agree that it'd be a BC break to remove Serializable. I'm already tracking that in [PHPC-2248] with a 2.0 fixVersion.

> With all of this in mind, adding conditional method signatures seems like a sensible choice and doesn't cause any follow-up work outside of dropping the interface whenever we're ready to do so.

Switching to a conditional declaration sounds good, especially if that would simplify this script. I think that deserves its own PHPC ticket for visibility. If you want to merge this as-is you can just make a note to come back and clean up this workaround later.

scripts/create-functionmap.php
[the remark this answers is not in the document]
jmikola: Oh, I totally missed that. Thanks.

scripts/generate-functionmap.php
```diff
@@ -0,0 +1,139 @@
...
+if (PHP_VERSION_ID < 80000) {
+    echo 'This script requires PHP 8.0 or higher';
+    exit(1);
+}
+
+$filename = __DIR__ . '/functionmap.php';
+(new FunctionMapGenerator)->createFunctionMap($filename);
+printf("Created call map in %s\n", $filename);
+
+class FunctionMapGenerator
```
jmikola, line 12: I didn't realize PHP allowed a class to be instantiated _before_ it was declared. Was that always the case?

Makefile.frag
```diff
@@ -66,3 +66,25 @@ libmongoc-version-current:
 
 libmongocrypt-version-current:
 	cd src/libmongocrypt/ && python etc/calc_release_version.py > ../LIBMONGOCRYPT_VERSION_CURRENT
+
+generate-function-map: all
```
jmikola, line 70: I didn't think of this before, but I think this warrants a new section in CONTRIBUTING.md.

Also, should this be a [phony target](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) since the name is not an actual filename? I think we may be missing that for a few other things as well, like `libmongoc-version-current` and `libmongocrypt-version-current`.

scripts/generate-functionmap.php
```diff
@@ -0,0 +1,139 @@
...
+                $parameterType = 'string';
+            }
+
+            $functionMapEntry[$parameterKey] = $parameterType;
+        }
+
+        return $functionMapEntry;
+    }
+
+    private function removeDoubleBackslash(string $string): string {
```
jmikola, line 136: I know we're not enforcing coding standards here, but should the function braces be on their own line?

scripts/generate-functionmap.php
```diff
@@ -0,0 +1,139 @@
...
+        $functionMapEntry = [$returnType];
+
+        foreach ($function->getParameters() as $parameter) {
+            $parameterKey = $parameter->getName();
+            if ($parameter->isOptional()) {
+                $parameterKey .= '=';
+            }
+
+            $parameterType = (string) $parameter->getType();
+            if ($function->getName() === 'unserialize' && $parameter->getName() === 'serialized') {
```
jmikola, line 126: Noted that this is sticking around. If you intend to add a conditional declaration in a separate PHPC ticket we can leave this as-is and then come back and remove this later.

jmikola (approved):
Some small suggestions but LGTM with some contributing docs added.

scripts/generate-functionmap.php
alcaeus: I'm not sure when that was added - that was the reason I initially had everything differently. Once I realised this works I decided to put all the logic in a class.

Makefile.frag
alcaeus: Added paragraph in contribution docs and updated the list of phone targets.

scripts/generate-functionmap.php
alcaeus: I missed this when refactoring from anonymous functions to a class. Fixed now.

scripts/generate-functionmap.php
alcaeus: This has been removed now that we have conditional declarations for the method in question.

scripts/create-functionmap.php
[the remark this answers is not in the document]
alcaeus: Updated to contain reworked method signatures for `Serializable::unserialize`.

scripts/generate-functionmap.php
GromNaN: It seems that it works since PHP 4, except if the class declaration is in a condition.
https://3v4l.org/NW6YX

PR #1438, opened 2023-06-21
PHPC-2236: Allow compression libraries to be configured at build time

jmikola:
https://jira.mongodb.org/browse/PHPC-2236

This introduces several configure options:

```
  --with-mongodb-snappy=[auto/yes/no]
  --with-mongodb-zlib=[auto/yes/system/bundled/no]
  --with-mongodb-zstd=[auto/yes/no]
```

The "auto", "yes", and "no" options behave like existing options: "auto" attempts to use a library but allows it to not be found; "yes" requires a library and errors if it's not found; "no" disables a library.

The "system" and "bundled" options for zlib function like "yes", but for using the system library or bundled zlib sources in libmongoc, respectively. Since bundled sources are always available, configuring zlib with "auto", "yes", or "bundled" should never fail.

### DIFF
--- a/scripts/autotools/libmongoc/CheckCompression.m4
+++ b/scripts/autotools/libmongoc/CheckCompression.m4
@@ -1,72 +1,111 @@
+PHP_ARG_WITH([mongodb-snappy],
+             [whether to enable Snappy for compression],
+             [AS_HELP_STRING([--with-mongodb-snappy=@<:@auto/yes/no@:>@],
+                             [MongoDB: Enable Snappy for compression [default=auto]])],
+             [auto],
+             [no])
+
+PHP_ARG_WITH([mongodb-zlib],
+             [whether to enable zlib for compression],
+             [AS_HELP_STRING([--with-mongodb-zlib=@<:@auto/yes/system/bundled/no@:>@],
+                             [MongoDB: Enable zlib for compression [default=auto]])],
+             [auto],
+             [no])
+
+PHP_ARG_WITH([mongodb-zstd],
+             [whether to enable zstd for compression],
+             [AS_HELP_STRING([--with-mongodb-zstd=@<:@auto/yes/no@:>@],
+                             [MongoDB: Enable zstd for compression [default=auto]])],
+             [auto],
+             [no])
+
 found_snappy="no"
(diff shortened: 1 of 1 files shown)


scripts/autotools/libmongoc/CheckCompression.m4
```diff
@@ -1,72 +1,111 @@
...
+                    [have_snappy_headers=yes],
+                    [have_snappy_headers=no])
+
+    if test "$have_snappy_lib" = "yes" -a "$have_snappy_headers" = "yes"; then
+      PHP_ADD_LIBRARY([snappy],,[MONGODB_SHARED_LIBADD])
+      found_snappy="yes"
+    fi
+  ])
+
+  if test "$PHP_MONGODB_SNAPPY" = "yes" -a "$found_snappy" = "no"; then
```
jmikola, line 48: I have system libraries for all three compression libraries. Rather than remove them, I tested this by flipping this (and related conditionals) to `"$found_snappy" = "yes"` and checking for an error when the system library was picked up.

I also tested "system" and "bundled" interactions for zlib.

When system libraries are found, you'll see the following configure output:

```
checking for PHP_MONGODB_SNAPPY... yes
checking for PHP_MONGODB_ZLIB... yes
checking for PHP_MONGODB_ZSTD... yes

...

MONGODB_SHARED_LIBADD                            : -lsnappy -lz -lzstd
```

Using bundled zlib doesn't result in any configure output, so to test that you'll want to check the definitions in the generated `mongoc-config.h` output file (e.g. `MONGOC_ENABLE_COMPRESSION_X`).

Compiling the driver and checking `phpinfo()` output also works.

alcaeus (approved):
LGTM. Seeing these new configure options made me think that we may want to mention those in [the documentation](https://www.php.net/manual/en/mongodb.installation.manual.php). There are no other configure options mentioned there right now, so I defer to you whether we want to have these options documented.

jmikola:
> There are no other configure options mentioned there right now, so I defer to you whether we want to have these options documented.

Made a note of that in [PHPC-1017](https://jira.mongodb.org/browse/PHPC-1017).

PR #1442, opened 2023-06-27
PHPLIB-1098: Update evergreen build hosts

alcaeus:
PHPLIB-1098

This PR changes the evergreen configuration to only build on supported hosts.
This includes:
- Drop testing on Ubuntu 18.04
- Drop testing on RHEL 7.0
- Add Ubuntu 22.04 and 20.04
- Add RHEL 9.0 and 7.6

RHEL 7.1 and 7.2 are kept to keep testing on the Z and Power architectures. Once the toolchain is available, RHEL 8.0 will take their place.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1124,27 +1124,52 @@ axes:
   - id: os
     display_name: OS
     values:
+      # Debian
       - id: debian11
         display_name: "Debian 11"
-        run_on: debian11
+        run_on: debian11-small
       - id: debian10
         display_name: "Debian 10"
-        run_on: debian10
+        run_on: debian10-small
       - id: debian92
         display_name: "Debian 9.2"
-        run_on: debian92
-      - id: rhel70
-        display_name: "RHEL 7.0"
-        run_on: rhel70
+        run_on: debian92-small
+
+      # RHEL
+      - id: rhel90
+        display_name: "RHEL 9.0"
+        run_on: rhel90-small
+      # TODO: RHEL 8.x is not working yet
+      # - id: rhel82-arm64
+      #   display_name: "RHEL 8.2 ARM64"
+      #   run_on: rhel82-arm64-small
+      # - id: rhel80
+      #   display_name: "RHEL 8.0"
+      #   run_on: rhel80-small
+      - id: rhel76
+        display_name: "RHEL 7.6"
+        run_on: rhel76-small
       - id: rhel71-power8
         display_name: "RHEL 7.1 Power 8"
         run_on: rhel71-power8-build
       - id: rhel72-zseries
         display_name: "RHEL 7.2 zSeries"
         run_on: rhel72-zseries-build
-      - id: ubuntu1804-arm64
-        display_name: "Ubuntu 18.04 ARM64"
-        run_on: ubuntu1804-arm64-test
+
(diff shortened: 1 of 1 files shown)


.evergreen/config.yml
```diff
@@ -1184,15 +1215,29 @@ axes:
...
+    - debian10
+    - rhel90
+    # TODO: RHEL 8.x is not working yet
+    # - rhel82-arm64
+    # - rhel80
+    - rhel76
+    - ubuntu2204-arm64
+    - ubuntu2204
+    - ubuntu2004-arm64
+    - ubuntu2004
```
jmikola, line 1232: Is it redundant to test PHP 8.1+ on both ubuntu2204 and ubuntu2004 (x64 and arm64)?

To make more efficient use of Evergreen, perhaps we can split this around the version boundary, similar to what we do with debian92 and debian11 in the `test-mongodb-versions` matrix.

.evergreen/config.yml
```diff
@@ -1124,27 +1124,58 @@ axes:
...
-      - id: ubuntu1804-arm64
-        display_name: "Ubuntu 18.04 ARM64"
-        run_on: ubuntu1804-arm64-test
+
+      # Ubuntu LTS
+      - id: ubuntu2204
+        display_name: "Ubuntu 22.04 x64"
+        run_on: ubuntu2204-small
+        variables:
+          SKIP_LEGACY_SHELL: "true"
```
jmikola, line 1166: Is `SKIP_LEGACY_SHELL` here in any way related to https://github.com/mongodb/mongo-php-library/pull/1035 and [PHPLIB-1028](https://jira.mongodb.org/browse/PHPLIB-1028)?

I only found `SKIP_LEGACY_SHELL` referenced in [`download-mongodb.sh`](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/2608a86a58be31670e4c7b5aef26d3db44dd3df8/.evergreen/download-mongodb.sh#L714). In that script, it is only used in a test for `-z "${SKIP_LEGACY_SHELL:-}"` (i.e. branch happens when the variable is zero-length or undefined). So in that case, defining this string as _anything_ is considered a "true" value to avoid the conditional download of the legacy shell. That seems a bit odd to me, but I suppose it's just because we tend to do explicit checks for "yes" values in our own scripts.

Why do we only do this for the Ubuntu 22.04 platforms? If the legacy shell isn't necessary at all, perhaps we should be setting `SKIP_LEGACY_SHELL` for all environments.

I see `download-mongodb.sh` is only called in our config for the `getdata` task (only executed manually for debugging). I suppose the only other place it is called is via the `bootstrap mongo-orchestration` task through `run-orchestration.sh`.

.evergreen/config.yml
alcaeus: Since this involves different versions of OpenSSL I found it prudent to test the full range. Also, only testing the old versions would have us continuously update the list of excluded PHP versions on Ubuntu 20.04 as new versions are added. I'm not that concerned about the load on Evergreen on checks running on our branches, but I think it would make sense to limit the tasks we run when testing pull request. I don't think we need to test the full complement of PHP versions on all operating systems for that.

Potentially, we can re-address this when we refactor our build matrices in PHPC-2201.

jmikola (approved):

PR #1447, opened 2023-07-18
PHPC-2262: Replace ICU dep with vendored utf8proc lib

jmikola:
https://jira.mongodb.org/browse/PHPC-2262

PHPC-2262: Replace ICU dep with vendored utf8proc lib
    
This introduces a --with-mongodb-utf8proc configure flag, which accepts "bundled" (default) and "system". Windows will always use bundled sources.
    
CheckICU.m4 and the --with-mongodb-icu configure flag have been removed.
    
Bumps libmongoc to 1.25-dev. This also adds support for a new MONGOC_ENABLE_SRV config constant, which is currently set based on detection in CheckResolv.m4. We can consider exposing that as a configure option down the line in PHPC-2256.


scripts/autotools/libmongoc/CheckUtf8Proc.m4
```diff
@@ -0,0 +1,40 @@
...
+                      [have_utf8proc_lib="no"])
+
+    AC_CHECK_HEADER([utf8proc.h],
+                    [have_utf8proc_headers=yes],
+                    [have_utf8proc_headers=no])
+
+    if test "$have_utf8proc_lib" = "yes" -a "$have_utf8proc_headers" = "yes"; then
+      PHP_ADD_LIBRARY([utf8proc],,[MONGODB_SHARED_LIBADD])
+      found_utf8proc="yes"
+    fi
```
jmikola, line 29: Note: I tested both of these code paths locally.

```
$ pkg-config  libutf8proc --libs
-lutf8proc
```

```
pkg-config  libutf8proc --cflags
-DUTF8PROC_EXPORTS
```

Configure only shows `-lutf8proc`, but I do see `-DUTF8PROC_EXPORTS` added to `$PHP_MONGODB_BUNDLED_CFLAGS` when using pkg-config. I'm not sure if it's required for the second branch. /cc @joshbsiegel

alcaeus (approved):

config.w32
```diff
@@ -86,7 +86,7 @@ if (PHP_MONGODB != "no") {
   ADD_EXTENSION_DEP("mongodb", "openssl", false);
 
   var PHP_MONGODB_CFLAGS="\
-    /D BSON_COMPILATION /D MONGOC_COMPILATION \
+    /D BSON_COMPILATION /D MONGOC_COMPILATION /D UTF8PROC_STATIC \
```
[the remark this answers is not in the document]
alcaeus: Sounds good - I'm not sure whether we should rely on that library being available locally for people that download our DLLs, so compiling with bundled utf8proc seems like the logical choice.

scripts/autotools/libmongoc/CheckUtf8Proc.m4
jmikola: I don't think `UTF8PROC_EXPORTS` is relevant, as it's only referenced in the following conditional for Windows:

https://github.com/JuliaStrings/utf8proc/blob/1cb28a66ca79a0845e99433fd1056257456cef8b/utf8proc.h#L123-L137

joshbsiegel: I agree. I didn't have to do anything with `UTF8PROC_EXPORTS` when making changes in the C driver.

PR #1449, opened 2023-07-31
PHPC-1000: Allow bsonSerialize to return BSON structure

alcaeus:
PHPC-1000

This PR allows implementations of `MongoDB\BSON\Serializable` to return instances of `MongoDB\BSON\Document` and `MongoDB\BSON\PackedArray`. This allows for an easier implementation of the `Serializable` in PHPLIB classes.

The first commit fixes the tentative return type for `Serializable::bsonSerialize`: while the original type was `array|object`, the code actually rejected any object that was not an instance of `stdClass`. I've also tested the impact on this change: on PHP < 8.1 there is no change as tentative return types are only reported in PHP 8.1+. In those versions, users that have a wider return type (e.g. the original `array|object` type or `object` in general), will see a deprecation message that the return types should be compatible. There is no hard break until we make the type non-tentative in PHPC 2.0, so I thought I'd preempt this and fix the return type. This then allows us to specify a return type of `array|stdClass|Document|PackedArray` in the last commit, which once again is consistent with the values the serializer accepts as return value.

### DIFF
--- a/src/BSON/Serializable.stub.php
+++ b/src/BSON/Serializable.stub.php
@@ -11,9 +11,9 @@ interface Serializable extends Type
 {
 #if PHP_VERSION_ID >= 80000
     /** @tentative-return-type */
-    public function bsonSerialize(): array|object;
+    public function bsonSerialize(): array|\stdClass;
 #else
-    /** @return array|object */
+    /** @return array|\stdClass */
     public function bsonSerialize();
 #endif
 }
--- a/src/BSON/Serializable.stub.php
+++ b/src/BSON/Serializable.stub.php
@@ -11,9 +11,9 @@ interface Serializable extends Type
 {
 #if PHP_VERSION_ID >= 80000
     /** @tentative-return-type */
-    public function bsonSerialize(): array|\stdClass;
+    public function bsonSerialize(): array|\stdClass|Document|PackedArray;
 #else
-    /** @return array|\stdClass */
+    /** @return array|\stdClass|Document|PackedArray */
     public function bsonSerialize();
 #endif
 }
--- a/src/MongoDB/ReadConcern.stub.php
+++ b/src/MongoDB/ReadConcern.stub.php
@@ -47,12 +47,7 @@ final class ReadConcern implements \MongoDB\BSON\Serializable, \Serializable
 
     final public static function __set_state(array $properties): ReadConcern {}
 
-#if PHP_VERSION_ID >= 80000
-    final public function bsonSerialize(): array|\stdClass {}
-#else
-    /** @return array|\stdClass */
-    final public function bsonSerialize() {}
-#endif
+    final public function bsonSerialize(): \stdClass {}
 
     final public function serialize(): string {}
 
--- a/src/MongoDB/ReadConcern.stub.php
+++ b/src/MongoDB/ReadConcern.stub.php
@@ -48,9 +48,9 @@ final class ReadConcern implements \MongoDB\BSON\Serializable, \Serializable
     final public static function __set_state(array $properties): ReadConcern {}
 
 #if PHP_VERSION_ID >= 80000
-    final public function bsonSerialize(): array|object {}
+    final public function bsonSerialize(): array|\stdClass {}
 #else
-    /** @return array|object */
+    /** @return array|\stdClass */
     final public function bsonSerialize() {}
 #endif
 
--- a/src/phongo_bson_encode.c
+++ b/src/phongo_bson_encode.c
@@ -104,14 +104,16 @@ static void php_phongo_bson_append_object(bson_t* bson, php_phongo_field_path* f
 	}
 
 	if (Z_TYPE_P(object) == IS_OBJECT && instanceof_function(Z_OBJCE_P(object), php_phongo_type_ce)) {
-		if (Z_TYPE_P(object) == IS_OBJECT && (instanceof_function(Z_OBJCE_P(object), php_phongo_document_ce) || instanceof_function(Z_OBJCE_P(object), php_phongo_packedarray_ce))) {
-			if (instanceof_function(Z_OBJCE_P(object), php_phongo_document_ce)) {
-				php_phongo_document_t* intern = Z_DOCUMENT_OBJ_P(object);
-				bson_append_document(bson, key, key_len, intern->bson);
-			} else {
-				php_phongo_packedarray_t* intern = Z_PACKEDARRAY_OBJ_P(object);
-				bson_append_array(bson, key, key_len, intern->bson);
-			}
+		if (instanceof_function(Z_OBJCE_P(object), php_phongo_document_ce)) {
+			php_phongo_document_t* intern = Z_DOCUMENT_OBJ_P(object);
+			bson_append_document(bson, key, key_len, intern->bson);
+
+			return;
+		}
+
+		if (instanceof_function(Z_OBJCE_P(object), php_phongo_packedarray_ce)) {
+			php_phongo_packedarray_t* intern = Z_PACKEDARRAY_OBJ_P(object);
+			bson_append_array(bson, key, key_len, intern->bson);
 
 			return;
 		}
--- a/src/phongo_bson_encode.c
+++ b/src/phongo_bson_encode.c
@@ -58,6 +58,10 @@ static int php_phongo_is_array_or_document(zval* val)
 	int        count;
 
 	if (Z_TYPE_P(val) != IS_ARRAY) {
+		if (Z_TYPE_P(val) == IS_OBJECT && instanceof_function(Z_OBJCE_P(val), php_phongo_packedarray_ce)) {
+			return IS_ARRAY;
+		}
+
 		return IS_OBJECT;
 	}
 
@@ -121,7 +125,6 @@ static void php_phongo_bson_append_object(bson_t* bson, php_phongo_field_path* f
 		if (instanceof_function(Z_OBJCE_P(object), php_phongo_serializable_ce)) {
 			zval   obj_data;
 			bson_t child;
-			bool   is_array;
 
 			zend_call_method_with_0_params(PHONGO_COMPAT_OBJ_P(object), NULL, NULL, BSON_SERIALIZE_FUNC_NAME, &obj_data);
 
@@ -146,11 +149,9 @@ static void php_phongo_bson_append_object(bson_t* bson, php_phongo_field_path* f
 				return;
 			}
 
-			is_array = php_phongo_is_array_or_document(&obj_data) == IS_ARRAY || (Z_TYPE(obj_data) == IS_OBJECT && instanceof_function(Z_OBJCE(obj_data), php_phongo_packedarray_ce));
-
 			/* Persistable objects must always be serialized as BSON documents;
 			 * otherwise, infer based on bsonSerialize()'s return value. */
-			if (instanceof_function(Z_OBJCE_P(object), php_phongo_persistable_ce) || !is_array) {
+			if (instanceof_function(Z_OBJCE_P(object), php_phongo_persistable_ce) || php_phongo_is_array_or_document(&obj_data) != IS_ARRAY) {
 				bson_append_document_begin(bson, key, key_len, &child);
 				if (instanceof_function(Z_OBJCE_P(object), php_phongo_persistable_ce)) {
 					bson_append_binary(&child, PHONGO_ODM_FIELD_NAME, -1, 0x80, (const uint8_t*) Z_OBJCE_P(object)->name->val, Z_OBJCE_P(object)->name->len);
--- a/src/phongo_bson_encode.c
+++ b/src/phongo_bson_encode.c
@@ -131,6 +131,7 @@ static void php_phongo_bson_append_object(bson_t* bson, php_phongo_field_path* f
 				return;
 			}
 
+			// TODO PHP_VERSION_ID < 80000: obsolete once the tentative return type of bsonSerialize() is enforced
 			if (
 				Z_TYPE(obj_data) != IS_ARRAY && !(Z_TYPE(obj_data) == IS_OBJECT && (instanceof_function(Z_OBJCE(obj_data), zend_standard_class_def) || instanceof_function(Z_OBJCE(obj_data), php_phongo_document_ce) || instanceof_function(Z_OBJCE(obj_data), php_phongo_packedarray_ce)))) {
 				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE,
--- a/src/phongo_bson_encode.c
+++ b/src/phongo_bson_encode.c
@@ -119,6 +119,7 @@ static void php_phongo_bson_append_object(bson_t* bson, php_phongo_field_path* f
 		if (instanceof_function(Z_OBJCE_P(object), php_phongo_serializable_ce)) {
 			zval   obj_data;
 			bson_t child;
+			bool   is_array;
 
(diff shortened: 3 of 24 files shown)


tests/bson/bson-fromPHP-001.phpt
```diff
@@ -21,18 +21,29 @@ class MyDocument implements MongoDB\BSON\Serializable
...
+    }
+}
+
 $tests = array(
     array(1, 2, 3),
     array('foo' => 'bar'),
     (object) array(1, 2, 3),
     (object) array('foo' => 'bar'),
+    MongoDB\BSON\PackedArray::fromPHP([1, 2, 3]),
+    MongoDB\BSON\Document::fromPHP(['foo' => 'bar']),
```
jmikola, line 38: Does this test answer the earlier question I asked about `Persistable::bsonSerialize()` returning a PackedArray? Based on the expected output it looks like PackedArrays are silently encoded as documents and an ODM field appended (not desirable, IMO).

src/phongo_bson_encode.c
```diff
@@ -128,20 +129,25 @@ static void php_phongo_bson_append_object(bson_t* bson, php_phongo_field_path* f
 				return;
 			}
 
-			if (Z_TYPE(obj_data) != IS_ARRAY && !(Z_TYPE(obj_data) == IS_OBJECT && instanceof_function(Z_OBJCE(obj_data), zend_standard_class_def))) {
+			if (
+				Z_TYPE(obj_data) != IS_ARRAY && !(Z_TYPE(obj_data) == IS_OBJECT && (instanceof_function(Z_OBJCE(obj_data), zend_standard_class_def) || instanceof_function(Z_OBJCE(obj_data), php_phongo_document_ce) || instanceof_function(Z_OBJCE(obj_data), php_phongo_packedarray_ce)))) {
```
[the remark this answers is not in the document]
alcaeus: That is correct. I've added a todo for this.

src/phongo_compat.h
```diff
@@ -333,6 +333,11 @@ static inline zend_bool zend_ini_parse_bool(zend_string* str)
 	ZEND_BEGIN_ARG_INFO_EX(name, 0, return_reference, required_num_args)
 #endif
 
+#ifndef ZEND_BEGIN_ARG_WITH_TENTATIVE_RETURN_OBJ_TYPE_MASK_EX
```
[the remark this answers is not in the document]
alcaeus: There's a comment in L320:
> The ZEND_BEGIN_ARG_WITH_TENTATIVE_* set of macros was introduced in PHP 8.1

I've moved the the declaration of `ZEND_BEGIN_ARG_WITH_TENTATIVE_TYPE_INFO_EX` into that block as well and grouped the declarations together.

src/BSON/Serializable.stub.php
```diff
@@ -11,9 +11,9 @@ interface Serializable extends Type
 {
 #if PHP_VERSION_ID >= 80000
     /** @tentative-return-type */
-    public function bsonSerialize(): array|object;
+    public function bsonSerialize(): array|\stdClass|Document|PackedArray;
```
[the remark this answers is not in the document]
alcaeus: Updated the ticket title to indicate we're adding support for returning BSON structures.

src/MongoDB/ReadConcern.stub.php
```diff
@@ -48,9 +48,9 @@ final public function isDefault(): bool {}
     final public static function __set_state(array $properties): ReadConcern {}
 
 #if PHP_VERSION_ID >= 80000
-    final public function bsonSerialize(): array|object {}
+    final public function bsonSerialize(): array|\stdClass {}
```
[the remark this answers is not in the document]
alcaeus: Yes, these will always return an `stdClass` instance and I've changed the return type to reflect this.

src/phongo_bson_encode.c
```diff
@@ -103,22 +103,23 @@ static void php_phongo_bson_append_object(bson_t* bson, php_phongo_field_path* f
...
+				bson_append_document(bson, key, key_len, intern->bson);
+			} else {
+				php_phongo_packedarray_t* intern = Z_PACKEDARRAY_OBJ_P(object);
+				bson_append_array(bson, key, key_len, intern->bson);
+			}
 
-		return;
-	}
+			return;
+		}
```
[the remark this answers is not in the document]
alcaeus: Good point, done as suggested.

src/phongo_bson_encode.c
```diff
@@ -128,20 +129,25 @@ static void php_phongo_bson_append_object(bson_t* bson, php_phongo_field_path* f
...
 
 				return;
 			}
 
+			is_array = php_phongo_is_array_or_document(&obj_data) == IS_ARRAY || (Z_TYPE(obj_data) == IS_OBJECT && instanceof_function(Z_OBJCE(obj_data), php_phongo_packedarray_ce));
+
 			/* Persistable objects must always be serialized as BSON documents;
 			 * otherwise, infer based on bsonSerialize()'s return value. */
-			if (instanceof_function(Z_OBJCE_P(object), php_phongo_persistable_ce) || php_phongo_is_array_or_document(&obj_data) == IS_OBJECT) {
+			if (instanceof_function(Z_OBJCE_P(object), php_phongo_persistable_ce) || !is_array) {
```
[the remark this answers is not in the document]
alcaeus: `is_array` is taken for `obj_data`, which is returned from the `bsonSerialize` method, while the `Persistable` check is for `object`. This has now been changed: we serialise as an object if the object is an instance of `Persistable` (regardless of whether the value returned is an array), or if the result is not an array.

src/phongo_bson_encode.c
```diff
@@ -451,26 +457,35 @@ static void php_phongo_zval_to_bson_internal(zval* data, php_phongo_field_path*
...
 				}
 
+				if (instanceof_function(Z_OBJCE_P(data), php_phongo_persistable_ce)) {
+					bson_append_binary(bson, PHONGO_ODM_FIELD_NAME, -1, 0x80, (const uint8_t*) Z_OBJCE_P(data)->name->val, Z_OBJCE_P(data)->name->len);
+					/* Ensure that we ignore an existing key with the same name
+					 * if one exists in the bsonSerialize() return value. */
+					skip_odm_field = true;
+				}
+
+				// If bsonSerialize() returns a BSON document or packedArray instance, recurse to copy data over directly
```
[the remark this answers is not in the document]
alcaeus: This is now explicitly forbidden through the return type of `Persistable::bsonSerialize` and checked in the code.

src/phongo_compat.h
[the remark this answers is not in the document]
jmikola: Ah, I was wondering why a few of these were missing the leading comment -- but only decided to comment on the one you were addressing here. Thanks for tidying this up!

tests/bson/bson-fromPHP-001.phpt
jmikola: Noted that you added exceptions below:

> Expected MyPersistableDocument::bsonSerialize() to return an array, stdClass, or MongoDB\BSON\Document, MongoDB\BSON\PackedArray given

I'll resolve this.

src/phongo_compat.h
```diff
@@ -285,15 +285,6 @@ static inline zend_bool zend_ini_parse_bool(zend_string* str)
 	Z_PARAM_ARRAY_OR_OBJECT(dest)
 #endif
 
-/* Per https://wiki.php.net/rfc/internal_method_return_types, "Non-final
- * internal method return types - when possible - are declared tentatively in
- * PHP 8.1, and they will become enforced in PHP 9.0." This can be revisited
- * when more general typing improvements are made in PHPC-1709. */
```
jmikola, line 291: [PHPC-1709](https://jira.mongodb.org/browse/PHPC-1709) has long since been resolved. Glad to see this removed :)

src/phongo_bson_encode.c
```diff
@@ -58,6 +58,10 @@ static int php_phongo_is_array_or_document(zval* val)
 	int        count;
 
 	if (Z_TYPE_P(val) != IS_ARRAY) {
+		if (Z_TYPE_P(val) == IS_OBJECT && instanceof_function(Z_OBJCE_P(val), php_phongo_packedarray_ce)) {
+			return IS_ARRAY;
+		}
+
```
jmikola, line 64: While we're here, I think there's some redundant code later in this function if you'd like to address it now.

```c
if (count > 0) {
    ...
} else {
    return Z_TYPE_P(val);
}

return IS_ARRAY;
```

After this conditional for `(Z_TYPE_P(val) != IS_ARRAY)`, we can rely on the final `return IS_ARRAY` unless iteration of a non-empty array indicates that we need to return `IS_OBJECT`. The `else` branch above is redundant.

This can also wait until we refactor this function in [PHPC-2217](https://jira.mongodb.org/browse/PHPC-2217), so I've left a comment there as well.

src/phongo_bson_encode.c
```diff
@@ -141,7 +164,7 @@ static void php_phongo_bson_append_object(bson_t* bson, php_phongo_field_path* f
 
 			/* Persistable objects must always be serialized as BSON documents;
 			 * otherwise, infer based on bsonSerialize()'s return value. */
-			if (instanceof_function(Z_OBJCE_P(object), php_phongo_persistable_ce) || php_phongo_is_array_or_document(&obj_data) == IS_OBJECT) {
+			if (instanceof_function(Z_OBJCE_P(object), php_phongo_persistable_ce) || php_phongo_is_array_or_document(&obj_data) != IS_ARRAY) {
```
jmikola, line 167: I think it'd also help for `php_phongo_is_array_or_document` to assert that only an object or array is passed in. We don't really have to worry about this thanks to the explicit checks on `obj_data` above, but it'd be good for peace of mind. I'll make a note of this in [PHPC-2217](https://jira.mongodb.org/browse/PHPC-2217) as well.

src/phongo_bson_encode.c
```diff
@@ -451,26 +474,47 @@ static void php_phongo_zval_to_bson_internal(zval* data, php_phongo_field_path*
...
+
 				if (Z_TYPE(obj_data) != IS_ARRAY && !(Z_TYPE(obj_data) == IS_OBJECT && instanceof_function(Z_OBJCE(obj_data), zend_standard_class_def))) {
 					phongo_throw_exception(
 						PHONGO_ERROR_UNEXPECTED_VALUE,
-						"Expected %s::%s() to return an array or stdClass, %s given",
+						"Expected %s::%s() to return an array, stdClass, %s, or %s, %s given",
 						ZSTR_VAL(Z_OBJCE_P(data)->name),
 						BSON_SERIALIZE_FUNC_NAME,
+						ZSTR_VAL(php_phongo_document_ce->name),
+						ZSTR_VAL(php_phongo_packedarray_ce->name),
```
jmikola, line 510: This logic all looks correct, although it's a bit more convoluted than the type checking we have in `php_phongo_bson_append_object`. This function might also benefit from a comment like:

```
// TODO PHP_VERSION_ID < 80000: obsolete once the tentative return type of bsonSerialize() is enforced
```

For now, I'd just suggest adding a comment. I think we can leave the code as-is since it hopefully won't stick around much longer (or at least we're unlikely to revisit these functions anytime soon).

If I was implementing this from scratch, I'd probably create a utility function just to check the return type of `bsonSerialize()` and pass in the original object's `zend_class_entry*` to control the logic enforced (really just between persistable or serializable). Something like:

```c
/* Checks the return type of a bsonSerialize() method. Returns true on
 * success; otherwise, throws an exception and returns false.
 *
 * TODO: obsolete once PHP 8.0+ is required and tentative return type
 * is enforced. */
static inline bool phongo_check_bson_serialize_return_type(zval *retval, zend_class_entry* ce)
{
    if (ce instanceof persistable) {
        // check types, throw and return false if needed

        return true;
    }

    if (ce instanceof serializable) {       
        // check types, throw and return false if needed

        return true;
    }

    /* consider throwing PHONGO_ERROR_UNEXPECTED_VALUE since we shouldn't
     * get here, but if you want to just return true here you can remove
     * the true returns above and use if/else instead.
     */ 
}
```

jmikola (approved):
Some small suggestions to clean things up (and some can wait until [PHPC-2217](https://jira.mongodb.org/browse/PHPC-2217)).

LGTM.

src/phongo_bson_encode.c
```diff
@@ -322,7 +345,7 @@ static void php_phongo_bson_append(bson_t* bson, php_phongo_field_path* field_pa
 			break;
 
 		case IS_ARRAY:
-			if (php_phongo_is_array_or_document(entry) == IS_ARRAY) {
+			if (Z_TYPE_P(entry) == IS_ARRAY && php_phongo_is_array_or_document(entry) == IS_ARRAY) {
```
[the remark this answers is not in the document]
alcaeus: Yes. In German we say "doppelt hält besser". I've removed it nonetheless. 🤦‍♂️

src/phongo_bson_encode.c
alcaeus: Left this to be handled in PHPC-2217.

src/phongo_bson_encode.c
alcaeus: Added the helper as suggested, that does indeed clean this up a bit.

src/phongo_bson_encode.c
```diff
@@ -128,20 +177,15 @@ static void php_phongo_bson_append_object(bson_t* bson, php_phongo_field_path* f
...
 			}
 
-			if (Z_TYPE(obj_data) != IS_ARRAY && !(Z_TYPE(obj_data) == IS_OBJECT && instanceof_function(Z_OBJCE(obj_data), zend_standard_class_def))) {
-				phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE,
-									   "Expected %s::%s() to return an array or stdClass, %s given",
-									   ZSTR_VAL(Z_OBJCE_P(object)->name),
-									   BSON_SERIALIZE_FUNC_NAME,
-									   PHONGO_ZVAL_CLASS_OR_TYPE_NAME(obj_data));
+			if (!phongo_check_bson_serialize_return_type(&obj_data, Z_OBJCE_P(object))) {
+				// Exception already thrown
```
jmikola, line 181: @alcaeus: I see this was already merged, but note that `phongo_check_bson_serialize_return_type()` can technically return `false` without throwing an exception if the object is neither Serializable nor Persistable. Perhaps impossible in practice, but I wanted to point it out since this seems like undefined behavior.

src/phongo_bson_encode.c
```diff
@@ -86,6 +90,49 @@ static int php_phongo_is_array_or_document(zval* val)
...
+								   ZSTR_VAL(php_phongo_document_ce->name),
+								   ZSTR_VAL(php_phongo_packedarray_ce->name),
+								   PHONGO_ZVAL_CLASS_OR_TYPE_NAME(*retval));
+			return false;
+		}
+
+		return true;
+	}
+
+	return false;
```
jmikola, line 133: Quoting https://github.com/mongodb/mongo-php-driver/pull/1449#discussion_r1280813934:

> Cconsider throwing PHONGO_ERROR_UNEXPECTED_VALUE since we shouldn't get here, but if you want to just return true here you can remove the true returns above and use if/else instead.

If we return false, we should absolutely throw to follow conventions. Otherwise, just return true and NOP.

jmikola (commented):
@alcaeus: I think this needs a quick follow-up commit.

src/phongo_bson_encode.c
alcaeus: Good point - taken care of in https://github.com/mongodb/mongo-php-driver/pull/1453

PR #1450, opened 2023-07-31
PHPC-2230: Drop support for PHP 7.2 and 7.3

alcaeus:
PHPC-2230

This drops support for PHP 7.2 and 7.3, along with a bunch of compatibility code.

I've also deleted all tests regarding serialisation on PHP < 7.4 and updated the `SKIPIF` blocks on those tests that affect PHP 7.4+.


jmikola (commented):
Noted that you're leaving the Serializable implementations in place for now (per [PHPC-2248](https://jira.mongodb.org/browse/PHPC-2248)).

I recall there being some EXPECTF patterns that varied by PHP version and decided to search for `%r` in the tests. That likely doesn't catch everything, but it did turn up quite a bit:

 * `%r(argument|parameter)%r` and `%r(E_NOTICE|E_WARNING)%r` date back to https://github.com/mongodb/mongo-php-driver/commit/60febd0f28c3d347b5de8f09cad37574873a68c3 and might still be relevant.
 * `%r(object|stdClass)%r` is pretty recent (https://github.com/mongodb/mongo-php-driver/commit/6887226d34566cd772bbacb34e10fc7bd4000f57) and is probably still relevant.
 * `%r[eE]%r` was recently introduced in the Int64 tests. I'm not sure if it's was needed for a particular PHP version or platform (e.g. Windows).
 * `%r(null|NULL)%r` dates back to PHP 5 and can be removed (see: https://github.com/mongodb/mongo-php-driver/commit/651257063d3642aa518e6e27ec85ac0697704e67)
 * ``int%S` and `bool%S` addresses a difference between PHP 7.2 and 7.3 (https://github.com/mongodb/mongo-php-driver/commit/98e583b738a45ddaf0e64a6913ca126162de9381). There may be additional instances of `%r(int|integer)%r` or `%r(bool|boolean)%r`.
 * `%r(double|float)%` addresses a difference between PHP 5 and 7 (see: https://github.com/mongodb/mongo-php-driver/pull/467#discussion_r89211603).
 * `%r(0|"0")%r` was related to PHP 7.2 (see: https://github.com/mongodb/mongo-php-driver/commit/95c467856460a58f856c66a3feefb5f841f4619b)
 * `%w` in `set_state` tests dates back to HHVM (see: https://github.com/mongodb/mongo-php-driver/commit/c2fe3b2089b4299d3a3f723aa70588f76de74c02)

This PR would probably be a good time to address any obsolete patterns in the above list.

alcaeus:
I've cleaned up the following:
* `null|NULL` was normalised to `null`
* `int%S` and `bool%S` were normalised to `int` and `bool`
* `double|float` was normalised to `float`
* The `0` vs. `"0"` keys in objects were normalised to `"0"`
* `%w` in `set_state` tests were removed. There is some fancy behaviour going on with a space at the end of the key line when the value is an object, which I've matched using `%w` to avoid editors automatically removing the trailing whitespace and thus breaking the test.

I've left the others in place for the time being and will check those separately and add comments where possible to indicate when they can be removed.

tests/bson/bug0974-001.phpt
```diff
@@ -31,7 +31,7 @@ object(stdClass)#%d (%d) {
   }
 }
 object(stdClass)#%d (%d) {
-  [%r(0|"0")%r]=>
+  ["0"]=>
```
jmikola, line 34: There are some more occurrences below: `%r(1|"1")%r`. Sorry for not specifying in my original comment.

tests/bson/bug0974-001.phpt
```diff
@@ -44,7 +44,7 @@ object(stdClass)#%d (%d) {
   }
 }
 object(stdClass)#3 (2) {
-  [%r(0|"0")%r]=>
+  ["0"]=>
```
jmikola, line 47: There are more occurrences above and below to replace in this file. Best to search for `")%r]=>` throughout the test files to catch all of these.

tests/bson/bug0974-001.phpt
```diff
@@ -44,7 +44,7 @@ object(stdClass)#%d (%d) {
   }
 }
 object(stdClass)#3 (2) {
```
jmikola, line 46: While you're in this file, `#3 (2)` above should probably be `#%d (%d)`. The number in parentheses is not so important and we can assert the number of fields, but we definitely don't want to expect object sequences.

If you want to take a pass through all the tests looking for `)#\d` that might turn up more things we missed.

tests/bson/bson-toPHP-004.phpt
```diff
@@ -553,7 +553,7 @@ object(stdClass)#%d (1) {
 object(stdClass)#%d (1) {
   ["x"]=>
   object(stdClass)#%d (1) {
-    [%r(0|"0")%r]=>
+    ["0"]=>
```
jmikola, line 556: I think you got everything in this file (they were all `%r(0|"0")%r` AFAICT), but searching the project for `")%r]` will catch everything.

tests/bson/bson-javascript-set_state-001.phpt
```diff
@@ -31,39 +31,39 @@ echo "\n\n";
...
 ))
 
 %r\\?%rMongoDB\BSON\Javascript::__set_state(array(
-%w'code' => 'function foo() { return id; }',
-%w'scope' => 
+   'code' => 'function foo() { return id; }',
+   'scope' =>%w
   %Sarray(
-  %w'id' => 
+     'id' =>%w
```
jmikola, line 57: Was `%w` actually necessary here (and above) _after_ the `=>` for portability across PHP versions?

tests/bson/bson-javascript-set_state-001.phpt
```diff
@@ -31,39 +31,39 @@ echo "\n\n";
...
 ))
 
 %r\\?%rMongoDB\BSON\Javascript::__set_state(array(
-%w'code' => 'function foo(bar) { return bar; }',
-%w'scope' => 
-  %Sarray(
-  %S),
+   'code' => 'function foo(bar) { return bar; }',
+   'scope' =>%w
+  %r\(object\)? %rarray(
```
jmikola, line 41: Why did you add `%r\(object\)? %r` here but not below, where `%Sarray` remains?

tests/bson/bug0974-001.phpt
alcaeus: Removed a number of occurrences across tests.

tests/bson/bson-javascript-set_state-001.phpt
alcaeus: `var_export` has some strange formatting when a property contains an object or array: in that case, one space of the indentation is actually printed at the end of the previous line. As mentioned in my comment above, I've replaced those with `%w` to avoid editors from automatically trimming whitespace at the end and to make it more visible. I found some other occurrences, so those have been adjusted as well.

tests/bson/bson-javascript-set_state-001.phpt
alcaeus: I missed that one. I replaced other occurrences of `%Sarray` with this as it's a more precise match.

tests/bson/bug0974-001.phpt
alcaeus: Looked for any non-zero keys and replaced those as well.

tests/readPreference/readpreference-set_state-001.phpt
```diff
@@ -48,31 +48,31 @@ foreach ($tests as $fields) {
...
        'dc' => 'sf',
        'use' => 'reporting',
-    %S),
-    2 => 
-    %Sarray(
-    %S),
+    %r\)?%r),
+    2 =>%w
+    %r\(object\)? %rarray(
+    %r\)?%r),
```
jmikola, line 75: Why would we expect _another_ optional `)` character here? I understand an optional `(object)` cast in the output, but there is already an expected `)` to balance `array(` on the previous line. And there are other closing parens _below_ that balance out the `array(` instances up top.

I don't think this has anything to do with your previous related comments:

 * https://github.com/mongodb/mongo-php-driver/pull/1450#discussion_r1281459424
 * https://github.com/mongodb/mongo-php-driver/pull/1450#discussion_r1281459748

I see you have `%r\(object\)? %r` above, and the closing paren there is optional. Does that mean that PHP sometimes emits `(object array(...))`? That doesn't make sense as it wouldn't be valid syntax.

In fact, the `%r\(object\)? %r` pattern would only match `(object ` or `(object) `. In that case, it seems like `(object)` is _always_ present (it's the only valid syntax). So perhaps we should expect `(object)` for PHP 7.4+.

I tested this in https://3v4l.org/MaXQ3 and `(object)` is always included in PHP 7.3+. Previous versions would emit `stdClass::__set_state(array(...))`, which explains the extra closing paren, which is no longer applicable.

And PHP 8.2+ differs slightly by emitting a backslash before class names to make them fully-qualified. This is already accounted for with the `%r\\?%r` pattern earlier in the `EXPECTF` block.

----

Note: this applies to other `EXPECTF` sections with `__set_state()` output.

alcaeus: Good observation, I believe I originally added this in error and then it spread via the usual way of copy pasta.

jmikola (approved):

PR #1453, opened 2023-08-04
PHPC-1000: Add missing exception

alcaeus:
Follow-up to PHPC-1000 to take care of https://github.com/mongodb/mongo-php-driver/pull/1449#pullrequestreview-1561537284

### DIFF
--- a/src/phongo_bson_encode.c
+++ b/src/phongo_bson_encode.c
@@ -130,6 +130,10 @@ static inline bool phongo_check_bson_serialize_return_type(zval* retval, zend_cl
 		return true;
 	}
 
+	phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE,
+						   "Expected to receive instance of %s, %s given",
+						   ZSTR_VAL(php_phongo_serializable_ce->name),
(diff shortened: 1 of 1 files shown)


src/phongo_bson_encode.c
```diff
@@ -130,6 +130,10 @@ static inline bool phongo_check_bson_serialize_return_type(zval* retval, zend_cl
 		return true;
 	}
 
+	phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE,
+						   "Expected to receive instance of %s, %s given",
```
jmikola, line 134: Indentation looks way off here. Did you run this through `clang-format`?

alcaeus: This was changed by CLion when I moved code around. `clang-format` doesn't care about these indentations, but we typically only indent an additional level for multi-line method calls. I've updated this and two other instances in this file.

jmikola (approved):

alcaeus:
> Change LGTM but check the formatting before merging.

Turns out I was wrong before: the change I pushed broke clang-format, so I reverted it.

PR #1454, opened 2023-08-04
PHPC-2269: Add internal iterator logic for MongoDB\BSON\Iterator

alcaeus:
PHPC-2269

Adding to v1.17 only as we'd be backporting a number of compat macros from PHP itself, in addition to `ZVAL_OBJ_COPY`. Waiting on #1450 to be merged so I don't have to bother fixing the PHP 7.2 build.

### DIFF
--- a/src/BSON/Iterator.c
+++ b/src/BSON/Iterator.c
@@ -92,6 +92,67 @@ static void php_phongo_iterator_build_current(php_phongo_iterator_t* intern)
 	phongo_bson_value_to_zval(bson_iter_value(&intern->iter), &intern->current);
 }
 
+static zval* php_phongo_iterator_get_current(php_phongo_iterator_t* intern)
+{
+	if (!intern->valid) {
+		phongo_throw_exception(PHONGO_ERROR_LOGIC, "Cannot call current() on an exhausted iterator");
+		return NULL;
+	}
+
+	if (Z_ISUNDEF(intern->current)) {
+		php_phongo_iterator_build_current(intern);
+	}
+
+	return &intern->current;
+}
+
+static void php_phongo_iterator_next(php_phongo_iterator_t* intern)
+{
+	intern->valid = bson_iter_next(&intern->iter);
+	intern->key++;
+	php_phongo_iterator_free_current(intern);
+}
+
+static bool php_phongo_iterator_key(php_phongo_iterator_t* intern, zval* key)
+{
+	const char* ckey;
+
+	if (!intern->valid) {
+		phongo_throw_exception(PHONGO_ERROR_LOGIC, "Cannot call key() on an exhausted iterator");
+		return false;
+	}
+
+	if (intern->is_array) {
+		ZVAL_LONG(key, intern->key);
+
+		return true;
+	}
+
+	ckey = bson_iter_key(&intern->iter);
+	if (!bson_utf8_validate(ckey, strlen(ckey), false)) {
+		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "Detected corrupt BSON data at offset %d", intern->iter.off);
+		return false;
+	}
+
+	ZVAL_STRING(key, ckey);
+
+	return true;
+}
+
+static void php_phongo_iterator_rewind(php_phongo_iterator_t* intern)
+{
+	/* Don't re-initialise iterator if we're still on the first item */
+	if (intern->key == 0) {
+		return;
+	}
+
+	php_phongo_iterator_free_current(intern);
+
+	bson_iter_init(&intern->iter, php_phongo_iterator_get_bson_from_zval(&intern->bson));
+	intern->key   = 0;
+	intern->valid = bson_iter_next(&intern->iter);
+}
+
 static HashTable* php_phongo_iterator_get_properties_hash(phongo_compat_object_handler_type* object, bool is_temp)
 {
 	php_phongo_iterator_t* intern;
@@ -113,48 +174,33 @@ PHONGO_DISABLED_WAKEUP(MongoDB_BSON_Iterator)
 static PHP_METHOD(MongoDB_BSON_Iterator, current)
 {
 	php_phongo_iterator_t* intern = Z_ITERATOR_OBJ_P(getThis());
+	zval*                  data;
 
 	PHONGO_PARSE_PARAMETERS_NONE();
 
-	if (!intern->valid) {
-		phongo_throw_exception(PHONGO_ERROR_LOGIC, "Cannot call current() on an exhausted iterator");
-		return;
+	data = php_phongo_iterator_get_current(intern);
+	if (!data) {
+		// Exception already thrown
+		RETURN_NULL();
 	}
 
-	if (Z_ISUNDEF(intern->current)) {
-		php_phongo_iterator_build_current(intern);
-	}
-
-	if (Z_ISUNDEF(intern->current)) {
+	if (Z_ISUNDEF_P(data)) {
 		RETURN_NULL();
 	} else {
-		ZVAL_COPY_DEREF(return_value, &intern->current);
+		ZVAL_COPY_DEREF(return_value, data);
 	}
 }
 
 static PHP_METHOD(MongoDB_BSON_Iterator, key)
 {
-	const char*            key;
 	php_phongo_iterator_t* intern = Z_ITERATOR_OBJ_P(getThis());
 
 	PHONGO_PARSE_PARAMETERS_NONE();
 
-	if (!intern->valid) {
-		phongo_throw_exception(PHONGO_ERROR_LOGIC, "Cannot call key() on an exhausted iterator");
+	if (!php_phongo_iterator_key(intern, return_value)) {
+		// Exception already thrown
 		return;
 	}
-
-	if (intern->is_array) {
-		RETURN_LONG(intern->key);
-	}
-
-	key = bson_iter_key(&intern->iter);
-	if (!bson_utf8_validate(key, strlen(key), false)) {
-		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "Detected corrupt BSON data at offset %d", intern->iter.off);
-		return;
-	}
-
-	RETURN_STRING(key);
 }
 
 static PHP_METHOD(MongoDB_BSON_Iterator, next)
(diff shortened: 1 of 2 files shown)


src/BSON/Iterator.c
```diff
@@ -263,10 +298,92 @@ static HashTable* php_phongo_iterator_get_properties(phongo_compat_object_handle
...
+static void php_phongo_iterator_it_dtor(zend_object_iterator* iter)
+{
+	zval_ptr_dtor(&iter->data);
+}
+
+static int php_phongo_iterator_it_valid(zend_object_iterator* iter)
+{
+	php_phongo_iterator_t* intern = Z_ITERATOR_OBJ_P(&iter->data);
+
+	return intern->valid ? SUCCESS : FAILURE;
```
alcaeus, line 311: I'm once again reminded how dumb PHP is for having `SUCCESS` defined as `0`, when bools work exactly the other way around. Took me a moment to figure out why as I was initially returning `intern->valid` directly.

src/BSON/Iterator.c
```diff
@@ -263,10 +298,92 @@ static HashTable* php_phongo_iterator_get_properties(phongo_compat_object_handle
 	return php_phongo_iterator_get_properties_hash(object, false);
 }
 
+/* Iterator handlers */
+static void php_phongo_iterator_it_dtor(zend_object_iterator* iter)
```
alcaeus, line 302: This entire logic was shamelessly lifted from [spl_array.c](https://github.com/php/php-src/blob/633e7455e55cb734acea13a14d73b8e976ad2158/ext/spl/spl_array.c#L958..L1048)

src/BSON/Iterator.c
```diff
@@ -263,10 +298,92 @@ static HashTable* php_phongo_iterator_get_properties(phongo_compat_object_handle
...
+	php_phongo_iterator_t* intern = Z_ITERATOR_OBJ_P(&iter->data);
+
+	return intern->valid ? SUCCESS : FAILURE;
+}
+
+static zval* php_phongo_iterator_it_get_current_data(zend_object_iterator* iter)
+{
+	php_phongo_iterator_t* intern = Z_ITERATOR_OBJ_P(&iter->data);
+
+	return php_phongo_iterator_get_current(intern);
```
alcaeus, line 318: Unlike the `current()` method itself, there's no need to copy/deref the zval we're returning here. Another notable difference to `current()` is that we're leaving an `IS_UNDEF` zval as-is, whereas `current` returns `null` in those cases. I don't think it makes a significant difference, and I'm not sure why `Cursor::current()` (which inspired the BSON iterator) returns `null` for an undef zval. I know there are some weird shenanigans going on when returning a `IS_UNDEF`, but I don't think this will ever matter in practice, as having that value returned indicates that something went seriously wrong during the iteration and there should've been an exception anyways.

src/BSON/Iterator.c
```diff
@@ -263,10 +298,92 @@ static HashTable* php_phongo_iterator_get_properties(phongo_compat_object_handle
...
+};
+
+static zend_object_iterator* php_phongo_iterator_get_iterator(zend_class_entry* ce, zval* object, int by_ref)
+{
+	zend_object_iterator* iterator;
+
+	if (by_ref) {
+		phongo_throw_exception(PHONGO_ERROR_LOGIC, "An iterator cannot be used with foreach by reference");
+		return NULL;
+	}
```
alcaeus, line 370: There is precedent for this in PHP itself, although they emit an `E_ERROR` in such cases. I opted to throw instead as it's more consistent how we expose errors.

We can't support iteration by reference as the BSON structures themselves are read-only, so assigning to the decoded value makes no sense.

src/BSON/Iterator.c
```diff
@@ -263,10 +298,92 @@ static HashTable* php_phongo_iterator_get_properties(phongo_compat_object_handle
...
+}
+
+static const zend_object_iterator_funcs php_phongo_iterator_it_funcs = {
+	php_phongo_iterator_it_dtor,
+	php_phongo_iterator_it_valid,
+	php_phongo_iterator_it_get_current_data,
+	php_phongo_iterator_it_get_current_key,
+	php_phongo_iterator_it_move_forward,
+	php_phongo_iterator_it_rewind,
+	NULL,
```
alcaeus, line 359: This is the function pointer for `invalidate_current`, which isn't used. Quoting from [phpinternalsbook](https://www.phpinternalsbook.com/php5/classes_objects/iterators.html):
> The only handler that has no correspondence in userland is `invalidate_current`, which can be used to destroy the current key/value. The handler is largely unused though, in particular `foreach` won’t even call it.

We could point this to `php_phongo_iterator_free_current`, although I'm not sure what the benefits of that would be.

alcaeus:
Here are some results from the performance tests done using the tests added in mongodb/mongo-php-library#1141.

Results in PHP 8.2-debug:
```
$ ./vendor/bin/phpbench run benchmark/BSON/DocumentBench.php --config=phpbench.ci.json --filter="benchIteration$" --report=aggregate --ref=original
PHPBench (1.2.14) running benchmarks... #standwithukraine
with configuration file: phpbench.ci.json
with PHP version 8.2.8, xdebug ✔, opcache ❌
comparing [actual vs. original]

\MongoDB\Benchmark\BSON\DocumentBench

    benchIteration..........................I2 - [Mo23.925ms vs. Mo114.955ms] -79.19% (±0.19%)

Subjects: 1, Assertions: 0, Failures: 0, Errors: 0
+---------------+----------------+-----+------+-----+---------------+------------------+----------------+
| benchmark     | subject        | set | revs | its | mem_peak      | mode             | rstdev         |
+---------------+----------------+-----+------+-----+---------------+------------------+----------------+
| DocumentBench | benchIteration |     | 10   | 3   | 6.498mb 0.00% | 23.925ms -79.19% | ±0.19% -75.23% |
+---------------+----------------+-----+------+-----+---------------+------------------+----------------+
```

This shows an improvement of 79% with this new logic. Using a non-debug version of PHP however yields wildly different results:
```
$ ./vendor/bin/phpbench run benchmark/BSON/DocumentBench.php --config=phpbench.ci.json --filter="benchIteration$" --report=aggregate --ref=original
PHPBench (1.2.14) running benchmarks... #standwithukraine
with configuration file: phpbench.ci.json
with PHP version 8.2.8, xdebug ✔, opcache ❌
comparing [actual vs. original]

\MongoDB\Benchmark\BSON\DocumentBench

    benchIteration..........................I2 - [Mo75.472ms vs. Mo76.018ms] -0.72% (±0.37%)

Subjects: 1, Assertions: 0, Failures: 0, Errors: 0
+---------------+----------------+-----+------+-----+---------------+-----------------+----------------+
| benchmark     | subject        | set | revs | its | mem_peak      | mode            | rstdev         |
+---------------+----------------+-----+------+-----+---------------+-----------------+----------------+
| DocumentBench | benchIteration |     | 10   | 3   | 6.319mb 0.00% | 75.472ms -0.72% | ±0.37% -20.52% |
+---------------+----------------+-----+------+-----+---------------+-----------------+----------------+
```

Here, there is no improvement whatsoever, but interestingly enough the test is slower than in the debug version, which has me somewhat confused.

The extension was compiled the same way for both runs, always including the developer flags (`--enable-mongodb-developer-flags`).

alcaeus:
Here are the latest results comparing iteration in non-debug PHP:
```
benchIteratorToArray....................I0 - [Mo25.636ms vs. Mo77.726ms] -67.02% (±0.00%)
```

These are the results I was looking for - turns out `make install` does not respect a custom `extension_dir` (which I'll have to investigate separately) and I wasn't testing the improvements at all.

src/BSON/Iterator.c
jmikola: I've been bitten by this before, but I've made peace with assuming PHP's behavior is inspired by process exit codes.

src/BSON/Iterator.c
jmikola: `php_phongo_iterator_free_current` is already called from the necessary places.

I'm fine with leaving `invalidate_current` unset, unless there's prior art for other iterator implementations using it.

src/phongo_compat.h
```diff
@@ -303,4 +303,42 @@ zend_bool zend_array_is_list(zend_array* array);
...
+		(dtor),                                                                                                                 \
+			(valid),                                                                                                            \
+			(get_current_data),                                                                                                 \
+			(get_current_key),                                                                                                  \
+			(move_forward),                                                                                                     \
+			(rewind),                                                                                                           \
+			(invalidate_current),                                                                                               \
+			(get_gc),                                                                                                           \
+	}
+#else
```
jmikola, line 319:
```suggestion
#else /* PHP_VERSION_ID < 80000 */
```

This is used in some other longer `#if` blocks elsewhere in the file.

Note that we don't have prior art for this with `#if(n)def`, so I didn't suggest it below.

src/phongo_compat.h
```diff
@@ -303,4 +303,42 @@ zend_bool zend_array_is_list(zend_array* array);
...
+	{                                                                                                                           \
+		(dtor),                                                                                                                 \
+			(valid),                                                                                                            \
+			(get_current_data),                                                                                                 \
+			(get_current_key),                                                                                                  \
+			(move_forward),                                                                                                     \
+			(rewind),                                                                                                           \
+			(invalidate_current),                                                                                               \
+	}
+#endif
```
jmikola, line 330:
```suggestion
#endif /* PHP_VERSION_ID >= 80000 */
```

src/BSON/Iterator.c
jmikola: Does "this entire logic" refer to _all_ of the iterator functions in this file (not just `dtor` here)?

jmikola (approved):
Some small comment suggestions but implementation LGTM.

src/BSON/Iterator.c
```diff
@@ -263,10 +298,93 @@ static HashTable* php_phongo_iterator_get_properties(phongo_compat_object_handle
...
+	iterator->funcs = &php_phongo_iterator_it_funcs;
+
+	return iterator;
+}
+
 void php_phongo_iterator_init_ce(INIT_FUNC_ARGS)
 {
 	php_phongo_iterator_ce                = register_class_MongoDB_BSON_Iterator(zend_ce_iterator);
 	php_phongo_iterator_ce->create_object = php_phongo_iterator_create_object;
+	php_phongo_iterator_ce->get_iterator  = php_phongo_iterator_get_iterator;
```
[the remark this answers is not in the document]
alcaeus: Created PHPC-2278.

src/BSON/Iterator.c
alcaeus: Added the comment as suggested.

src/BSON/Iterator.c
alcaeus: Yes - all of the `_it_*` methods were heavily inspired by PHP's standard implementation.

PR #1460, opened 2023-08-23
PHPC-2233: Split Windows Builds

alcaeus:
PHPC-2233

This pull request adds a new workflow to run tests on Windows, splitting the process into two jobs: one to build the driver and store the built DLL, another to run tests. The previous job sometimes caused issues, as errors while running tests (typically during the setup phase of tests) would cause other builds to be aborted, potentially stopping the compilation before all artifacts were stored. Build artifacts are shared between the builds using the `cache` action to avoid having to unpack and move around build files. For the release builds (and consumption of "nightly" builds), artifacts are created and stored on GitHub for three days as was previously done.

To make the release process easier and less error-prone, the workflow also reacts to push tags. This ensures that the windows-release-build workflow will find build artifacts for a tag even when the tagged commit has not been pushed individually. When pushing a tag, tests are not run as they are not necessary.


.github/workflows/windows/prepare-build/action.yml
```diff
@@ -0,0 +1,53 @@
...
+
+    - name: configure
+      run: configure --enable-mongodb --with-mongodb-sasl=yes --with-mongodb-client-side-encryption=yes --enable-debug-pack --with-prefix=${{ steps.setup-php.outputs.prefix }}
+      shell: cmd
+
+    - name: Get build directory
+      id: get-build-dir
+      shell: cmd
+      # The last echo command to set the output variable intentionally omits a space between the environment variable
+      # and the output redirector to avoid a trailing space in the generated output variable. Do not add a space.
```
jmikola, line 47: Good comment.

.github/workflows/windows/prepare-build/action.yml
```diff
@@ -0,0 +1,53 @@
...
+      run: |
+        cp .github/workflows/get-build-dir.bat .
+        for /F "usebackq tokens=*" %%i in (`get-build-dir.bat`) do set BUILD_DIR=%%i
+        echo BUILD_DIR=%BUILD_DIR%
+        @chcp 65001>nul
+        echo build_dir=%BUILD_DIR%>> %GITHUB_OUTPUT%
```
jmikola, line 53: Was this script adapted from another action, or did you come up with this yourself?

.github/workflows/windows-tests.yml
```diff
@@ -0,0 +1,132 @@
...
+      - "*"
+    branches:
+      - "v*.*"
+      - "master"
+      - "feature/*"
+
+jobs:
+  build:
+    name: "Build Windows DLLs"
+    runs-on: windows-2022
```
jmikola, line 20: Is there any way to enforce that this is standardized between both jobs?

[jobs.<job_id>.runs-on](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on) suggests that this can be a variable.

Not sure this is needed since we only have two jobs, but wanted to ask anyway.

jmikola (approved):

.github/workflows/windows/prepare-build/action.yml
alcaeus: The first two lines were present previously in the steps that publish the DLL as build artifact. The second lines took me a while to figure out, as the syntax to set an output variable for a step depends on the shell being used. [This StackOverflow answer](https://stackoverflow.com/questions/74443940/value-not-set-using-github-output/74444094#74444094) ultimately showed me the correct syntax for this to work in cmd.

.github/workflows/windows-tests.yml
alcaeus: Per the [Workflow syntax documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions), the only other place where to put such values is `env`. However, values set there can't be arrays, so even though we could reference an environment variable in the matrix definition, we can't set the list of PHP versions in an environment variable to reuse it in both matrixes.

PR #1463, opened 2023-09-05
PHPC-2280: Prefer OpenSSL on MacOS over Secure Transport

alcaeus:
PHPC-2280

On MacOS, the driver is compiled using Secure Transport when the `--with-mongodb-ssl` switch is set to `auto`. While this works fine in a CLI context it causes segmentation faults when using the driver in php-fpm, which usually is compiled using OpenSSL. To work around this issue, we now prefer OpenSSL on all platforms, and only check for Secure Transport on MacOS if no OpenSSL installation was found. To continue using Secure Transport on MacOS, users should specify `--with-mongodb-ssl=darwin`.

### DIFF
--- a/scripts/autotools/libmongoc/CheckSSL.m4
+++ b/scripts/autotools/libmongoc/CheckSSL.m4
@@ -18,22 +18,6 @@ AS_IF([test "$PHP_MONGODB_SSL" = "yes"],[
   PHP_MONGODB_SSL="auto"
 ])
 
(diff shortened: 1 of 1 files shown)


jmikola (approved):

tgalopin:
Hi @alcaeus, thanks for this PR!

I am experiencing this specific problem & saw the issue linked. I was wondering whether you plan to release this fix in the near future or if I should consider compiling the extension manually instead?

Thanks!

jmikola:
@tgalopin: The 1.17 release will likely ship sometime in November, following libmongoc 1.25.

If you're presently on 1.16, manually compiling a stable release with `--with-mongodb-ssl=openssl` would be advisable until then.

PR #1464, opened 2023-09-06
PHPLIB-2284: Force regeneration of arginfo files in workflow

alcaeus:
PHPLIB-2284

With this change, arginfo files will be regenerated using the latest version of the script. This helps identify changes to arginfo files as a result of the generation script changing, even when the underlying stub file hasn't changed.

### DIFF
--- a/.github/workflows/arginfo-files.yml
+++ b/.github/workflows/arginfo-files.yml
@@ -38,7 +38,7 @@ jobs:
         run: phpize
 
       - name: "Rebuild arginfo files from stubs"
-        run: "php ./build/gen_stub.php"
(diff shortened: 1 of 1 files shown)


.github/workflows/arginfo-files.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         run: phpize
 
       - name: "Rebuild arginfo files from stubs"
-        run: "php ./build/gen_stub.php"
+        run: "php ./build/gen_stub.php --force-regeneration"
```
jmikola, line 41: image

jmikola (approved):

PR #1465, opened 2023-09-06
PHPC-2282: Remove trace-level logs during BSON encoding

alcaeus:
PHPC-2282


jmikola (approved):

jmikola (commented):
I noted your comments in Slack that these trace statements had minimal impact on benchmarks. Still, glad to have them removed from both our code and emitted log output.

PR #1467, opened 2023-09-12
Enable dependabot for github-actions

GromNaN:
Following @alcaeus demand: https://github.com/mongodb/laravel-mongodb/pull/2612#pullrequestreview-1619728135. This will open PR when new versions of GitHub actions are available.

https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
jmikola, line 3: What does this accomplish for PHPC? Our only direct dependencies are the libmongoc and libmongocrypt submodules, and I doubt it's going to help us write Autoconf files :)

Would this be better suited to PHPLIB? I see that Composer is handled within [dependabot/dependabot-core](https://github.com/dependabot/dependabot-core).

alcaeus: @jmikola FWIW, Dependabot could open pull requests for updates to submodules, but it currently can't handle tags and can only update to the latest commit.

I'm fine with having automatic updates to GitHub Actions dependencies so we don't have to manually keep track of them.

jmikola: Oh, my mistake for not realizing what `package-ecosystem: "github-actions"` did. SGTM.

jmikola (approved):

PR #1472, opened 2023-09-20
PHPC-2245, PHPC-2246: Add array and property accessors to BSON types

alcaeus:
PHPC-2245, PHPC-2246

This adds array access for `MongoDB\BSON\PackedArray` instances, as well as array and property access to `MongoDB\BSON\Document` instances.

Note that the stubs currently don't implement the `ArrayAccess` interface, as the internal functionality does not require it and it would lead to us having to implement the various `offset...` methods defined in the interface. If implementing the interface is preferable to communicate the fact that such array access is possible, we can do so. The same applies to magic methods for property access in `MongoDB\BSON\Document`.

### DIFF
--- a/src/BSON/Document.c
+++ b/src/BSON/Document.c
@@ -165,28 +165,41 @@ static PHP_METHOD(MongoDB_BSON_Document, fromPHP)
 	RETURN_ZVAL(&zv, 1, 1);
 }
 
+static bool php_phongo_document_get(php_phongo_document_t* intern, char* key, size_t key_len, zval* return_value)
+{
+	bson_iter_t iter;
+
+	if (!bson_iter_init(&iter, intern->bson)) {
+		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator.");
+		return false;
+	}
+
+	if (!bson_iter_find_w_len(&iter, key, key_len)) {
+		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not find key \"%s\" in BSON data", key);
+		return false;
+	}
+
+	phongo_bson_value_to_zval(bson_iter_value(&iter), return_value);
+
+	return true;
+}
+
 static PHP_METHOD(MongoDB_BSON_Document, get)
 {
 	php_phongo_document_t* intern;
 	char*                  key;
 	size_t                 key_len;
-	bson_iter_t            iter;
 
 	PHONGO_PARSE_PARAMETERS_START(1, 1)
 	Z_PARAM_STRING(key, key_len)
 	PHONGO_PARSE_PARAMETERS_END();
 
 	intern = Z_DOCUMENT_OBJ_P(getThis());
-	if (!bson_iter_init(&iter, intern->bson)) {
-		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator.");
-	}
 
-	if (!bson_iter_find_w_len(&iter, key, key_len)) {
-		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not find key \"%s\" in BSON data", key);
+	if (!php_phongo_document_get(intern, key, key_len, return_value)) {
+		// Exception already thrown
 		RETURN_NULL();
 	}
-
-	phongo_bson_value_to_zval(bson_iter_value(&iter), return_value);
 }
 
 static PHP_METHOD(MongoDB_BSON_Document, getIterator)
@@ -196,23 +209,31 @@ static PHP_METHOD(MongoDB_BSON_Document, getIterator)
 	phongo_iterator_init(return_value, getThis());
 }
 
+static bool php_phongo_document_has(php_phongo_document_t* intern, char* key, size_t key_len)
+{
+	bson_iter_t iter;
+
+	if (!bson_iter_init(&iter, intern->bson)) {
+		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator.");
+		return false;
+	}
+
+	return bson_iter_find_w_len(&iter, key, key_len);
+}
+
 static PHP_METHOD(MongoDB_BSON_Document, has)
 {
 	php_phongo_document_t* intern;
 	char*                  key;
 	size_t                 key_len;
-	bson_iter_t            iter;
 
 	PHONGO_PARSE_PARAMETERS_START(1, 1)
 	Z_PARAM_STRING(key, key_len)
 	PHONGO_PARSE_PARAMETERS_END();
 
 	intern = Z_DOCUMENT_OBJ_P(getThis());
-	if (!bson_iter_init(&iter, intern->bson)) {
-		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator.");
-	}
 
-	RETURN_BOOL(bson_iter_find_w_len(&iter, key, key_len));
+	RETURN_BOOL(php_phongo_document_has(intern, key, key_len));
 }
 
 static PHP_METHOD(MongoDB_BSON_Document, toCanonicalExtendedJSON)
@@ -267,6 +288,54 @@ static PHP_METHOD(MongoDB_BSON_Document, toPHP)
 	php_phongo_bson_typemap_dtor(&state.map);
 
 	RETURN_ZVAL(&state.zchild, 0, 1);
+}
+
+static PHP_METHOD(MongoDB_BSON_Document, offsetExists)
+{
+	php_phongo_document_t* intern;
+	zval*                  offset;
+
+	PHONGO_PARSE_PARAMETERS_START(1, 1)
+	Z_PARAM_ZVAL(offset)
+	PHONGO_PARSE_PARAMETERS_END();
+
+	intern = Z_DOCUMENT_OBJ_P(getThis());
+
+	if (Z_TYPE_P(offset) != IS_STRING) {
+		RETURN_FALSE;
+	}
+
+	RETURN_BOOL(php_phongo_document_has(intern, Z_STRVAL_P(offset), Z_STRLEN_P(offset)));
+}
+
+static PHP_METHOD(MongoDB_BSON_Document, offsetGet)
+{
+	php_phongo_document_t* intern;
+	zval*                  offset;
+
+	PHONGO_PARSE_PARAMETERS_START(1, 1)
+	Z_PARAM_ZVAL(offset)
+	PHONGO_PARSE_PARAMETERS_END();
+
+	intern = Z_DOCUMENT_OBJ_P(getThis());
+
+	if (Z_TYPE_P(offset) != IS_STRING) {
+		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not find key of type \"%s\" in BSON data", PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(offset));
+		return;
+	}
+
+	// May throw, in which case we do nothing
+	php_phongo_document_get(intern, Z_STRVAL_P(offset), Z_STRLEN_P(offset), return_value);
+}
+
+static PHP_METHOD(MongoDB_BSON_Document, offsetSet)
+{
+	phongo_throw_exception(PHONGO_ERROR_LOGIC, "Cannot write to %s property", ZSTR_VAL(php_phongo_document_ce->name));
+}
+
+static PHP_METHOD(MongoDB_BSON_Document, offsetUnset)
+{
+	phongo_throw_exception(PHONGO_ERROR_LOGIC, "Cannot unset %s property", ZSTR_VAL(php_phongo_document_ce->name));
 }
 
 static PHP_METHOD(MongoDB_BSON_Document, __toString)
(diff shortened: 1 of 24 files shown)


src/BSON/PackedArray.c
```diff
@@ -401,6 +421,48 @@ static HashTable* php_phongo_packedarray_get_properties(phongo_compat_object_han
 	return php_phongo_packedarray_get_properties_hash(object, false, 1);
 }
 
+zval* php_phongo_packedarray_read_dimension(phongo_compat_object_handler_type* object, zval* offset, int type, zval* rv)
```
jmikola, line 424: > If implementing the interface is preferable to communicate the fact that such array access is possible, we can do so.

I think that's the right call. [ArrayObject](https://www.php.net/arrayobject) does the same in [spl_array.c](https://github.com/php/php-src/blob/php-8.2.10/ext/spl/spl_array.c#L673).

My thought is that some userland code might use an `array|ArrayAccess` type hint, which would otherwise prohibit our objects.

src/BSON/PackedArray.c
```diff
@@ -401,6 +421,48 @@ static HashTable* php_phongo_packedarray_get_properties(phongo_compat_object_han
...
+{
+	php_phongo_packedarray_t* intern;
+
+	intern = Z_OBJ_PACKEDARRAY(PHONGO_COMPAT_GET_OBJ(object));
+
+	if (Z_TYPE_P(offset) != IS_LONG) {
```
jmikola, line 430: Do we rely on PHP to convert numeric strings to integers?

I noticed the one error test only uses `'foo'`, so it might be helpful to add some tests with numeric strings and some other types (e.g. bool, float) to see what happens.

tests/bson/bson-packedarray-array-access_error-001.phpt
```diff
@@ -0,0 +1,35 @@
+--TEST--
+MongoDB\BSON\PackedArray array access does not allow writing
```
jmikola, line 2: This also includes the test for validating the offset type. Considering splitting that into a separate file.

This overlaps with my previous comment about testing errors for undefined offsets (with the correct type).

tests/bson/bson-document-array-access-001.phpt
```diff
@@ -0,0 +1,39 @@
...
+var_dump(isset($document['foo']));
+var_dump(isset($document['int64']));
+var_dump(isset($document['baz']));
+
+var_dump($document['foo']);
+var_dump($document['int64']);
+
+echo throws(function() use ($document) {
+    var_dump($document['baz']);
+}, MongoDB\Driver\Exception\RuntimeException::class), "\n";
```
jmikola, line 23: Should this be reserved for an error test? I suppose you can use a new file since it'd warrant a unique description.

src/BSON/Document.c
```diff
@@ -165,28 +165,41 @@ static PHP_METHOD(MongoDB_BSON_Document, fromPHP)
 	RETURN_ZVAL(&zv, 1, 1);
 }
 
+static bool php_phongo_document_get(php_phongo_document_t* intern, char* key, size_t key_len, zval* return_value)
+{
+	bson_iter_t iter;
+
+	if (!bson_iter_init(&iter, intern->bson)) {
+		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator.");
```
jmikola, line 173:
```suggestion
		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator");
```

We don't need periods for single sentence exception messages. Feel free to apply this everywhere.

src/BSON/Document.c
```diff
@@ -196,23 +209,30 @@ static PHP_METHOD(MongoDB_BSON_Document, getIterator)
 	phongo_iterator_init(return_value, getThis());
 }
 
+static bool php_phongo_document_has(php_phongo_document_t* intern, char* key, size_t key_len)
+{
+	bson_iter_t iter;
+
+	if (!bson_iter_init(&iter, intern->bson)) {
+		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator.");
```
jmikola, line 217:
```suggestion
		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator.");
		return false;
```

Otherwise you're accessing uninitialized memory with the subsequent `bson_iter_find_w_len()` call. Interesting that this didn't fail horribly in some test, but I suppose you may have lucked out and hit a guard check in [`_bson_iter_next_internal()`](https://github.com/mongodb/mongo-c-driver/blob/1.24.4/src/libbson/src/bson/bson-iter.c#L529).

src/BSON/Document.c
```diff
@@ -471,6 +491,91 @@ static HashTable* php_phongo_document_get_properties(phongo_compat_object_handle
...
+	char*                  key;
+	size_t                 key_len;
+
+	intern = Z_OBJ_DOCUMENT(PHONGO_COMPAT_GET_OBJ(object));
+
+	PHONGO_COMPAT_PROPERTY_ACCESSOR_NAME_TO_STRING(member, key, key_len);
+
+	if (!php_phongo_document_get(intern, key, key_len, rv)) {
+		// Exception already thrown
+		return &EG(uninitialized_zval);
```
jmikola, line 506: Is this expected by the object handler instead of returning `NULL`? I assume you borrowed this from ArrayObject, so feel free to confirm and resolve on your own.

src/BSON/Document.c
```diff
@@ -471,6 +491,91 @@ static HashTable* php_phongo_document_get_properties(phongo_compat_object_handle
...
+	phongo_throw_exception(PHONGO_ERROR_LOGIC, "Cannot write to %s property", ZSTR_VAL(php_phongo_document_ce->name));
+}
+
+int php_phongo_document_has_dimension(phongo_compat_object_handler_type* object, zval* member, int check_empty)
+{
+	php_phongo_document_t* intern;
+
+	intern = Z_OBJ_DOCUMENT(PHONGO_COMPAT_GET_OBJ(object));
+
+	if (Z_TYPE_P(member) != IS_STRING) {
```
jmikola, line 566: I didn't see any error tests using unsupported types for the dimension and property accessors. I'm specifically curious about passing an integer, and whether that would be converted to a string or preserved via `zval* member`.

Either way, consider adding some type tests like you have for PackedArray (and split those into their own file).

I don't think this is an issue for the property accessors, since `phongo_compat_property_accessor_name_type` is defined as `zend_string` for PHP 8.0+, but perhaps there's an edge case on PHP 7.4 where it's just a `zval`? I expect that'd probably be a syntax error long before any object handler gets invoked, though -- in which case there's nothing for us to test there.

src/BSON/PackedArray.c
```diff
@@ -147,22 +160,29 @@ static PHP_METHOD(MongoDB_BSON_PackedArray, getIterator)
 	phongo_iterator_init(return_value, getThis());
 }
 
+static bool php_phongo_packedarray_has(php_phongo_packedarray_t* intern, zend_long index)
+{
+	bson_iter_t iter;
+
+	if (!bson_iter_init(&iter, intern->bson)) {
+		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator.");
```
jmikola, line 168:
```suggestion
		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator.");
		return false;
```

Same issue as `php_phongo_document_has` above.

src/BSON/PackedArray.c
```diff
@@ -401,6 +421,48 @@ static HashTable* php_phongo_packedarray_get_properties(phongo_compat_object_han
...
+}
+
+int php_phongo_packedarray_has_dimension(phongo_compat_object_handler_type* object, zval* member, int check_empty)
+{
+	php_phongo_packedarray_t* intern;
+
+	intern = Z_OBJ_PACKEDARRAY(PHONGO_COMPAT_GET_OBJ(object));
+
+	if (Z_TYPE_P(member) != IS_LONG) {
+		return false;
```
jmikola, line 455: I assume it's intentional to not throw for an invalid type, as you do in `get`. Would doing so mess up `isset` and the like?

src/BSON/PackedArray.c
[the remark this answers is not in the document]
alcaeus: Added for both `Document` and `PackedArray`.

tests/bson/bson-packedarray-array-access_error-001.phpt
alcaeus: Split up into several tests.

src/BSON/Document.c
alcaeus: Removed in both `Document` and `PackedArray` classes.

src/BSON/Document.c
alcaeus: Ah, yes. Thanks for pointing this out.

src/BSON/Document.c
alcaeus: Yes, copied it over. If it seems more consistent I'm also happy to return `NULL`.

src/BSON/PackedArray.c
alcaeus: Not really, I just figured that we can be more lenient in `isset` checks since I'd assume people use those methods to guard access to the fields in the BSON structure, in which case returning `false` instead of throwing makes more sense.

src/phongo_compat.h
```diff
@@ -70,12 +70,24 @@
 #define phongo_compat_object_handler_type zend_object
 #define PHONGO_COMPAT_GET_OBJ(val) val
 #define PHONGO_COMPAT_SET_COMPARE_OBJECTS_HANDLER(type) php_phongo_handler_##type.compare = php_phongo_##type##_compare_objects;
+#define phongo_compat_property_accessor_name_type zend_string
+#define PHONGO_COMPAT_PROPERTY_ACCESSOR_NAME_TO_STRING(value, key, len) \
+	do {                                                                \
+		(key) = ZSTR_VAL((value));                                      \
+		(len) = ZSTR_LEN((value));                                      \
```
GromNaN, line 77: Why is there floating backslashes ate the end of the line?

GromNaN (commented):
I wonder how the properties will be resolved with [`twig_get_attribute`](https://github.com/twigphp/Twig/blob/3ae6fb8723776f4fc9d8ab490fcabcd3173da8e3/src/Extension/CoreExtension.php#L1471), to see if it optimizes anything or if there are optimizations to be made.

alcaeus:
> I wonder how the properties will be resolved with [`twig_get_attribute`](https://github.com/twigphp/Twig/blob/3ae6fb8723776f4fc9d8ab490fcabcd3173da8e3/src/Extension/CoreExtension.php#L1471) and `symfony/property-accessor`, to see if it optimizes anything or if there are optimizations to be made.

I'd have to check a profile, but IIRC the internal methods are prioritised. The BSON `Iterator` class is similar, as it has internal iterator handlers and extends the `Iterator` interface, but the latter's methods are not used during iteration as PHP relies on the internal iterator handlers to do the job.

src/phongo_compat.h
alcaeus: Macros in C are only single-line, so if you want to add a multi-line macro you need to escape the line break with a backslash. Our coding standard then enforces them to be aligned at the end of the longest line.

tests/bson/bson-packedarray-array-access_error-003.phpt
```diff
@@ -0,0 +1,38 @@
...
+echo throws(function() use ($array) {
+    $array[false];
+}, MongoDB\Driver\Exception\RuntimeException::class), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+OK: Got MongoDB\Driver\Exception\RuntimeException
+Could not find index of type "string" in BSON data
```
GromNaN, line 31: "data" is too generic, one could think it's a Document.
```suggestion
Could not find index of type "string" in BSON array
```

src/phongo_compat.h
GromNaN: So this is it. Thanks, I've just learned about this setting, which I've never needed before.

src/BSON/Document.c
jmikola: No, I think `&EG(uninitialized_zval)` is the right call here. I don't know enough about this API to say whether `NULL` would be equivalent.

jmikola (approved):
I agree with @GromNaN's suggestion to change the "BSON data" exception messages. As an alternative, consider replacing that with the class name instead of "BSON array" and "BSON document".

LGTM otherwise.

src/BSON/PackedArray.c
```diff
@@ -117,27 +117,40 @@ static bool seek_iter_to_index(bson_iter_t* iter, zend_long index)
...
+{
+	bson_iter_t iter;
+
+	if (!bson_iter_init(&iter, intern->bson)) {
+		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator");
+		return false;
+	}
+
+	if (!seek_iter_to_index(&iter, index)) {
+		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not find index \"%d\" in BSON data", index);
```
GromNaN, line 130: BSON array here, isn't it?

alcaeus: 🤦‍♂️ Fixed.

GromNaN (approved):
LGTM

PR #1473, opened 2023-09-22
PHPC-2261: Revamp Evergreen config

alcaeus:
This PR takes care of PHPC-2260 and PHPC-2261. The following changes have been made, while others are still pending:
- [x] Extract build tasks from testing to cache build results
- [x] Make test tasks depend on build tasks
- [x] Remove matrix build variants in favour of separate ones
- [x] Migrate existing special tests (OCSP, etc.) to new format
- [x] Automate generation of build tasks and potentially variants
- [x] Document structure of evergreen files and how to add/remove tasks/variants in future

The last task is still left to do. However, since `generate.tasks` does not show generated tasks when configuring a patch build, it is not suitable for us. However, repetitive files (e.g. the version-dependent tasks) will be generated automatically via a script to avoid repetition and errors.

In addition to the changes above, the following semi-related changes have been made:
- After compilation, a smoketest is run to ensure the library can be loaded. This prevents running a large number of tasks that won't complete successfully
- Add a make target that runs tests without compiling everything. This was done to avoid having to recursively replace directory names in all build output files. We may opt to invoke run-tests.php directly.
- Extracted configuration to multiple files to make files easier to understand and use
- Removed unused functions/tasks (there may be more left to remove)

For now, tests are only run on Debian systems:
* With PHP 8.2, we test all topologies on all MongoDB versions. This uses Debian 11 and Debian 9.2 to cover MongoDB 3.6 up to the latest build
* Older PHP versions only test MongoDB 7.0 in a replica set topology with authentication enabled.
* Other environments are not tested.

We'll have to discuss what tests we want present, and which ones we want to run for pull requests and commits. We may want to include more tasks but not run them by default to allow for testing special scenarios, although those can be added when necessary.


.evergreen/compile-unix.sh
```diff
@@ -1,6 +1,24 @@
-#!/bin/sh
+#!/bin/bash
 set -o errexit  # Exit the script with error if any of the commands fail
 
+# Find PHP binary path for the requested version
```
alcaeus, line 4: This was moved over so we can run compile for multiple PHP versions on the same box if we wanted to.

.evergreen/config/test-variants.yml
```diff
@@ -0,0 +1,126 @@
...
+    tasks:
+      - ".standalone .local !.3.6 !.4.0 !.4.2 !.4.4 !.5.0"
+      - ".replicaset .local !.3.6 !.4.0 !.4.2 !.4.4 !.5.0"
+      - ".sharded .local !.3.6 !.4.0 !.4.2 !.4.4 !.5.0"
+      - "test-atlas-connectivity"
+      - ".ocsp !.4.4"
+    display_tasks:
+      - name: "test-ocsp-latest"
+        execution_tasks:
+          - ".ocsp .latest"
```
alcaeus, line 24: [Display tasks](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files/#display-tasks) group a number of tasks in the "configure patch" view. While this limits us to have to select/unselect all OCSP tests at once without having a choice of enabling only some of them, I believe this is fine. Each OCSP test is an individual task due to the way these are structured.

.evergreen/config/test/ocsp-7.0.yml
```diff
@@ -0,0 +1,389 @@
+tasks:
+  - name: "ocsp-test_1-rsa-delegate-7.0"
```
alcaeus, line 2: The OCSP tasks may be candidates for using a task group. This would allow us to only fetch the build once, install the correct MongoDB version, then run all tasks sequentially on the same build host. This would further cut down resource usage. However, we'll want to make some improvements to mongodb-orchestration to make sure we can set up and tear down OCSP responders properly. We might also have to split up the mongo-orchestration files to allow for separate functions for downloading MongoDB, starting/stopping orchestration, and starting/stopping a deployment.

.evergreen/config.yml
```diff
@@ -1,1303 +1,61 @@
...
-- matrix_name: "test-csfle-skip_crypt_shared"
-  matrix_spec: { "os": "debian11", "mongodb-versions": "*", "php-edge-versions": "latest-stable" }
-  display_name: "CSFLE skip_crypt_shared - ${mongodb-versions}"
-  exclude_spec:
-    # CSFLE crypt_shared is available from MongoDB 6.0+
-    - { "os": "debian11", "mongodb-versions": ["3.6", "4.0", "4.2", "4.4", "5.0"], "php-edge-versions": "latest-stable" }
-  tasks:
-    - name: "test-skip_crypt_shared"
+include:
+  - filename: .evergreen/config/functions.yml
```
alcaeus, line 34: Unfortunately, `include` is only allowed in top-level files, so we can't nest our includes (and include the various test files in `test-tasks.yml`).

Makefile.frag
```diff
@@ -88,3 +88,27 @@ generate-function-map: all
 	else \
 		echo "ERROR: Cannot generate function maps without CLI sapi."; \
 	fi
+
+test-no-build:
```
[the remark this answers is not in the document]
jmikola: > Alternatively, we can also extract the call to run-tests.php

I was originally going to suggest replacing this Makefile target with additional logic in `.evergreen/run-tests.sh`, but seeing how messy this is (and integrated with the rest of the Makefile) I think this is the right call. 

> at that point we may as well consider using a different test runner entirely, such as PHPUnit (which can run phpt files as well).

I think this is totally worth exploring. Would you mind creating a ticket so we can revisit the idea?

I would never consider using PHPUnit with PHPC for local development (since it'd likely complicate things like backtraces and leak detection), but it may actually make sense for CI assuming it handles all of the same syntax of `run-tests.php` and generates comparable test suite output. And we'd also be able to test it alongside `run-tests.php` (e.g. introduce PHPUnit in just an extra task to start). I assume it will also enable better integration with Evergreen, such that we could start populating the "Tests" tab like we do for PHPLIB.

.evergreen/compile-unix.sh
```diff
@@ -1,6 +1,24 @@
...
+if [ -z "$PHP_PATH" ]; then
+  if [ -d "/opt/php/${PHP_VERSION}-64bit/bin" ]; then
+     PHP_PATH="/opt/php/${PHP_VERSION}-64bit/bin"
```
jmikola, line 7: [build-linux-toolchain.sh](https://github.com/10gen/mongo-php-toolchain/blob/master/build-linux-toolchain.sh#L39) defaults to "64bit" but I see that "32bit" is also supported. Do any environments actually use the `FORCE_ARCH` parameter to opt into 32-bit? I traced this back as far as [.evergreen/config.yml](https://github.com/10gen/mongo-php-toolchain/blob/master/.evergreen/config.yml#L42), but can't tell where `${arch}` may come from.

.evergreen/config.yml
```diff
@@ -1,1303 +1,86 @@
...
-      type: test
-      params:
-        working_dir: "src"
-        script: |
-          ${PREPARE_SHELL}
-          [ -f ${PROJECT_DIRECTORY}/${file} ] && sh ${PROJECT_DIRECTORY}/${file} || echo "${PROJECT_DIRECTORY}/${file} not available, skipping"
+# Protect ourselves against rogue test case that runs forever
+# Good rule of thumb: the average length a task takes, times 5, which roughly accounts for variable system performance
+# for various build variants
+exec_timeout_secs: 1800
```
jmikola, line 12: If this is still 6 minutes, does that line up with "the average length a task takes, times 5"? Do our tasks roughly take a bit over a minute to execute?

.evergreen/config.yml
```diff
@@ -1,1303 +1,86 @@
...
-        script: |
-          # Only run if a load balancer was started
-          if [ -n "${SINGLE_MONGOS_LB_URI}" ]; then
-            ${DRIVERS_TOOLS}/.evergreen/run-load-balancer.sh stop
-          fi
-
-pre:
+# These pre and post rules apply to all tasks not part of a task group, which should only ever be tests against local
+# MongoDB instances. All other tasks that require special preparation should be run from within a task group
+pre: &test_setup
```
jmikola, line 16: What references `*test_setup` and `*test_teardown`?

.evergreen/config.yml
```diff
@@ -1,1303 +1,86 @@
...
-post:
-  - func: "upload working dir"
+  - func: "fetch build"
+post: &test_teardown
   - func: "upload mo artifacts"
   - func: "upload test results"
-  - func: "stop load balancer"
   - func: "stop mongo-orchestration"
+  - func: "stop load balancer"
+  - func: "upload working dir"
```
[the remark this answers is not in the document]
jmikola: I'd be in favor of removing this. I've never made use of this, and can't imagine what we'd need it for.

.evergreen/config.yml
```diff
@@ -1,1303 +1,86 @@
...
-  display_name: "CSFLE skip_crypt_shared - ${mongodb-versions}"
-  exclude_spec:
-    # CSFLE crypt_shared is available from MongoDB 6.0+
-    - { "os": "debian11", "mongodb-versions": ["3.6", "4.0", "4.2", "4.4", "5.0"], "php-edge-versions": "latest-stable" }
-  tasks:
-    - name: "test-skip_crypt_shared"
+# Define aliases for patch builds and PR builds. These only apply if no aliases are defined in project settings
+github_pr_aliases:
+  # Run all build tasks, except on Power8 and zSeries due to low number of build hardware
+  - variant: "^build-[^-]+($|-arm64)"
```
jmikola, line 36: This appears to conditionally match "build-" followed by some string of characters omitting a hyphen, and then an optional "-arm64" suffix.

Would it make more sense to explicitly prohibit "-power8" or "-zseries" as a suffix?

.evergreen/config/build/_template-build-libmongoc.yml
```diff
@@ -0,0 +1,26 @@
+tasks:
+  # TODO: Enable the following two builds once libmongoc 1.25.0 is released
```
jmikola, line 2: Should this comment also refer back to `name: "build-php-libmongoc"` in `build-task-groups.yml`?

And I think there's a related area near the bottom of `test-variants.yml`.

.evergreen/config/build-task-groups.yml
```diff
@@ -0,0 +1,45 @@
...
+    max_hosts: 2
+    setup_task: *build_setup
+    setup_task_can_fail_task: true
+    setup_task_timeout_secs: 1800
+    teardown_task: *build_teardown
+    tasks:
+      - ".build !.php7.4 !.php8.0"
+
+  - name: "build-php-libmongoc"
+    # Keep this in sync with the actual number of libmongoc builds (typically 3)
```
jmikola, line 38: Would it make sense to explicitly refer to `_template-build-libmongoc.yml` here?

.evergreen/config/build/build-libmongoc-8.2.yml
```diff
@@ -0,0 +1,27 @@
+# This file is generated automatically - please edit the corresponding template file!
```
jmikola, line 1: The comment is sensible, but what do you think about also nesting these generated files under a sub-directory? That's done in libmongoc (see: [.evergreen/generated_configs](https://github.com/mongodb/mongo-c-driver/tree/master/.evergreen/generated_configs)).

I think this makes sense if we never expect to work with the generated files directly, in which case we can just stuff them away. That's in contrast to the current approach where you're primarily grouping by "build" and "test" types.

Consider the following directory structure:

```
.evergreen/config/
├─ build-task-groups.yml
├─ build-variants.yml
├─ functions.yml
├─ generate-config.php
├─ test-tasks.yml
├─ test-variants.yml
├─ generated/
│  ├─ build-libmongoc-8.2.yml
│  ├─ build-php-7.4.yml
│  ├─ test-load-balanced-5.0.yml
│  ├─ ...
├─ templates/
│  ├─ build-libmongoc-template.yml
│  ├─ build-php-template.yml
│  ├─ test-load-balanced-template.yml
│  ├─ ...
```

Alternatively, we could avoid a `templates/` directory and just keep the `_template-foo.yml` files directly within `.evergreen/config/`, which would then be a central directory of files we might need to modify over time.

I also added a "test-" prefix on generated test configs above to segregate them from build files, but I suppose that's not really important if we don't expect naming conflicts.

You could also replace "build-" and "test-" prefixes with `build/` and `test/` subdirectories within `generated/`; however, I don't think there's strong justification for that if we'd generally have no reason to venture into `generated/` anyway.

I don't mean to bikeshed here, but more thinking about how we can avoid having to look at files we'd have no business touching.

.evergreen/config/functions.yml
```diff
@@ -0,0 +1,355 @@
...
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        local_file: mongodb-start-logs.tar.gz
+        remote_file: ${UPLOAD_BUCKET}/${build_variant}/${revision}/${version_id}/${build_id}/logs/${task_id}-${execution}-mongodb-start-logs.tar.gz
+        bucket: mciuploads
+        permissions: public-read
+        content_type: ${content_type|application/x-gzip}
+        display_name: "mongodb-start-logs.tar.gz"
+
+  "upload working dir":
```
jmikola, line 199: Noted that you proposed deleting this in https://github.com/mongodb/mongo-php-driver/pull/1473/files#r1338184023.

Are there other irrelevant functions? Perhaps `"upload mo artifacts"`? I assume that has to do with server logs, but I can't remember ever needing that.

.evergreen/config/functions.yml
```diff
@@ -0,0 +1,355 @@
...
+  "fix absolute paths":
+    - command: shell.exec
+      params:
+        script: |
+          ${PREPARE_SHELL}
+          for filename in $(find ${DRIVERS_TOOLS} -name \*.json); do
+            perl -p -i -e "s|ABSOLUTE_PATH_REPLACEMENT_TOKEN|${DRIVERS_TOOLS}|g" $filename
+          done
+
+  "windows fix":
```
jmikola, line 297: This looks like a candidate for removal since we don't test on Windows within Evergreen. This is currently called from `pre:` and `build_setup:` (used by `setup_task` definitions within tasks).

.evergreen/config/functions.yml
```diff
@@ -0,0 +1,355 @@
...
+    - command: shell.exec
+      params:
+        script: |
+          ${PREPARE_SHELL}
+          for i in $(find ${DRIVERS_TOOLS}/.evergreen ${PROJECT_DIRECTORY}/.evergreen -name \*.sh); do
+            cat $i | tr -d '\r' > $i.new
+            mv $i.new $i
+          done
+
+  "make files executable":
```
jmikola, line 307: Is this necessary? I would assume any scripts in drivers-evergreen-tools and our own repository would have the correct permissions. And if not, that seems like something we might want to be alerted to so we can fix it properly.

.evergreen/config/functions.yml
```diff
@@ -0,0 +1,355 @@
...
+          done
+
+  "init test-results":
+    - command: shell.exec
+      params:
+        script: |
+          ${PREPARE_SHELL}
+          echo '{"results": [{ "status": "FAIL", "test_file": "Build", "log_raw": "No test-results.json found was created"  } ]}' > ${PROJECT_DIRECTORY}/test-results.json
+
+  "install dependencies":
```
jmikola, line 323: Is this only necessary to install `awscli` on Ubuntu?

See: [.evergreen/install-dependencies.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/install-dependencies.sh)

.evergreen/config/generate-config.php
```diff
@@ -0,0 +1,85 @@
+#!/bin/env php
+<?php
+
+// Supported PHP versions. Add new versions to the beginning of the list
+$supportedPhpVersions = ['8.2', '8.1', '8.0', '7.4'];
```
jmikola, line 5: What do you think about splitting this across lines to make diffs more readable? Likewise for `$supportedMongoDBVersions` below.

.evergreen/config/generate-config.php
```diff
@@ -0,0 +1,85 @@
...
+// Supported MongoDB versions. Add new versions after "rapid"
+$supportedMongoDBVersions = [
+    'latest', 'rapid',
+    '7.0', '6.0', '5.0',
+    '4.4', '4.2', '4.0',
+    '3.6',
+];
+
+// Only test the latest PHP version for libmongoc
+$libmongocBuildPhpVersions = array_slice($supportedPhpVersions, 0, 1);
```
jmikola, line 16: Is this the only reason you wrote "Add new versions to the beginning of the list" above? If so, I'd consider replacing this with:

```suggestion
$libmongocBuildPhpVersions = [ max($supportedPhpVersions) ];
```

.evergreen/config/generate-config.php
```diff
@@ -0,0 +1,85 @@
...
+
+function generateConfigs(
+    string $directory,
+    string $replacementName,
+    string $templateFile,
+    string $outputFormat,
+    array $versions,
+): array {
+    $template = file_get_contents(__DIR__ . '/' . $directory . '/' . $templateFile);
+    $header = '# This file is generated automatically - please edit the corresponding template file!';
```
jmikola, line 70: Consider embedding the template file name in this message, since you have it above. The short filename would probably suffice.

.evergreen/config/test-tasks.yml
```diff
@@ -0,0 +1,28 @@
...
+    tags: ["standalone", "local", "4.0", "storage-engines"]
+    commands:
+      - func: "bootstrap mongo-orchestration"
+        vars:
+          TOPOLOGY: "server"
+          MONGODB_VERSION: "4.0"
+          STORAGE_ENGINE: "inmemory"
+      - func: "run tests"
+  - name: "test-mongodb-4.0-standalone-mmapv1"
+    tags: ["standalone", "local", "4.0", "storage-engines"]
```
jmikola, line 21: I didn't see this in `architecture.md`, but I think it'd be helpful to maintain a glossary of the possible tags we might see within the configs. Offhand, I'm not sure how "local" is used.

.evergreen/config/test-variants.yml
jmikola: SGTM. I think there's also room for improvement here. If we look at libmongoc, I believe they extracted most of their OCSP logic into [run-ocsp-test.sh](https://github.com/mongodb/mongo-c-driver/blob/master/.evergreen/scripts/run-ocsp-test.sh), which makes the Evergreen configuration more readable. That might allow us to do reduce the complexity of `_template-ocsp.yml` and bring it more in line with other templates where we'd only selecting server versions.

.evergreen/config/test-variants.yml
```diff
@@ -0,0 +1,134 @@
...
+      FETCH_BUILD_VARIANT: "build-debian11"
+      FETCH_BUILD_TASK: "build-php-7.4"
+    depends_on:
+      - variant: "build-debian11"
+        name: "build-php-7.4"
+    tasks:
+      - ".replicaset .local .7.0 .auth"
+
+  # Variants with different libmongoc
+  # TODO: Enable the following two builds once libmongoc 1.25.0 is released
```
jmikola, line 98: This might be something else to reference from `_template-build-libmongoc.yml`.

Since I expect these bits will be frequently modified, it may be worth calling out in `architecture.md`, or the higher-level `contributing.md` file (assuming it doesn't already talk about enabling and disabling libmongoc versions for Evergreen).

.evergreen/config/test/_template-load-balanced.yml
```diff
@@ -0,0 +1,15 @@
...
+        vars:
+          TOPOLOGY: "sharded_cluster"
+          MONGODB_VERSION: "%mongodbVersion%"
+          LOAD_BALANCER: "true"
+          SSL: "yes"
+      - func: "start load balancer"
+      - func: "run tests"
+        vars:
+          MONGODB_URI: "${SINGLE_MONGOS_LB_URI}"
+          SSL: "yes"
```
jmikola, line 15: Side note: I have long disliked how `SSL` uses "yes|no" and `AUTH` uses "auth|noauth"; however, I this is is largely influenced by [run-orchestration.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/run-orchestration.sh#L5), so changing it for our own `run-tests.sh` script would just invite inconsistency without our own Evergreen configs.

.evergreen/config/test/_template-local.yml
```diff
@@ -0,0 +1,56 @@
+tasks:
+  - name: "test-mongodb-%mongodbVersion%-standalone-noauth-nossl"
+    tags: ["standalone", "local", "%mongodbVersion%"]
+    commands:
+      - func: "bootstrap mongo-orchestration"
+        vars:
+          TOPOLOGY: "server"
+          MONGODB_VERSION: "%mongodbVersion%"
+      - func: "run tests"
+  - name: "test-mongodb-%mongodbVersion%-standalone-auth"
```
jmikola, line 10:
```suggestion

  - name: "test-mongodb-%mongodbVersion%-standalone-auth"
```

Consider adding newlines between task names. It's done below, but inconsistently.

.evergreen/config/test/_template-ocsp.yml
```diff
@@ -0,0 +1,389 @@
...
+      - func: "bootstrap mongo-orchestration"
+        vars:
+          MONGODB_VERSION: "%mongodbVersion%"
+          TOPOLOGY: "server"
+          ORCHESTRATION_FILE: "rsa-basic-tls-ocsp-mustStaple.json"
+      - func: "run tests"
+        vars:
+          TESTS: "tests/ocsp-success.phpt"
+          APPEND_URI: "/?tls=true&tlsCAFile=${DRIVERS_TOOLS}/.evergreen/ocsp/rsa/ca.pem"
+  - name: "ocsp-test_1-ecdsa-delegate-%mongodbVersion%"
```
jmikola, line 19:
```suggestion

  - name: "ocsp-test_1-ecdsa-delegate-%mongodbVersion%"
```

Likewise, newlines would help with readability.

.evergreen/config/test/_template-require-api-version.yml
```diff
@@ -0,0 +1,13 @@
+tasks:
+  - name: "test-requireApiVersion-%mongodbVersion%"
+    tags: [ "standalone", "local", "%mongodbVersion%", "versioned_api" ]
```
jmikola, line 3: Do I assume correctly that we never use PHP versions for tags? So any "X.Y" strings (and "latest" or "rapid") would only correspond to server versions?

I suppose there's no reason to use a prefix here, as tags are always treated as strings and we're careful to quote these in YAML.

Either way, this is something we could document in a glossary of tags within `architecture.md`.

.evergreen/config/test/_template-skip-crypt-shared.yml
```diff
@@ -0,0 +1,11 @@
+tasks:
+  - name: "test-skip_crypt_shared-%mongodbVersion%"
+    tags: [ "replicaset", "local", "%mongodbVersion%", "skip_crypt_shared" ]
+    commands:
+      - func: "bootstrap mongo-orchestration"
+        vars:
+          TOPOLOGY: "replica_set"
```
jmikola, line 7: Did you want to add auth here, so that replica set with auth was the standard test environment when we're only testing on a single topology? Or is that unnecessary?

.gitattributes
```diff
@@ -0,0 +1,5 @@
+.evergreen/config/build/*.yml linguist-generated=true
```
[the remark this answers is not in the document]
jmikola: Don't forget to update this if we end up reorganizing the generated files.

.evergreen/config.yml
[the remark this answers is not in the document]
alcaeus: Removed 👍

.evergreen/compile-unix.sh
alcaeus: IIRC we no longer have any 32-bit systems that are supported in Evergreen, so this may be completely obsolete at this point.

.evergreen/config.yml
alcaeus: Unfortunately, golang regular expressions do not support negative lookahead, which is why I had to opt for matching the end-of-string character or `-arm64`, which is the only other suffix we use.

.evergreen/compile-unix.sh
jmikola: In that case, feel free to make a ticket to clean that up.

.evergreen/config.yml
alcaeus: The 1800 seconds are 30 minutes, meaning that we'll assume a roughly 6-minute task runtime. I'll double-check to make sure the value is still accurate.

.evergreen/config.yml
alcaeus: In this file, nothing. Off the top of my head I'm not sure whether these could be used in included files as well, but since they're not used in any case I've removed the tags.

.evergreen/config/build/_template-build-libmongoc.yml
alcaeus: Added comments here and in other files to cross-reference usage.

.evergreen/config/build-task-groups.yml
alcaeus: Added

.evergreen/config/functions.yml
alcaeus: Removed a number of other unnecessary functions, including "upload mo artifacts".

.evergreen/config/functions.yml
alcaeus: Removed

.evergreen/config/functions.yml
alcaeus: Removed

.evergreen/config/functions.yml
alcaeus: At the moment, yes. There's some more logic in there for PHPLIB, but in PHPC it doesn't contain anything else.

.evergreen/config/build/build-libmongoc-8.2.yml
alcaeus: Split up into `generated` and `templates` as suggested, but kept the `build` and `test` folders for better distinction.

.evergreen/config/test/_template-require-api-version.yml
alcaeus: Added a list of tags to architecture.md, which should clear this up.

.evergreen/config/test/_template-local.yml
alcaeus: Added a number of newlines in templates and regenerated files. FWIW, I grouped tasks using the same topology together, but it's making the file harder to read. I'll consider using comments if we feel the need to group tasks.

.evergreen/config/test-variants.yml
alcaeus: Added a note in the contribution docs since that explains the libmongoc update process already.

Makefile.frag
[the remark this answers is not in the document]
jmikola: @alcaeus: :point_up:

.evergreen/architecture.md
```diff
@@ -0,0 +1,64 @@
...
+## Task Tags
+
+Tasks are tagged to allow for a better selection in build variants. The following tags are used for build tasks:
+- `build`: All build tasks are tagged with `build`.
+- `build-libmongoc`: These build tasks build with different libmongoc versions.
+- `php<version>`: These tags allow selection based on PHP version, e.g. `php8.2`.
+
+Test tasks use the following tags:
+- `local`: All tasks that run a local MongoDB cluster for testing.
+- `<version>`: These tags allow selection based on MongoDB version, e.g. `6.4`.
```
jmikola, line 56:
```suggestion
- `<version>`: These tags allow selection based on MongoDB version, e.g. `7.0`.
```

Just replacing this with a real server version.

.evergreen/architecture.md
```diff
@@ -0,0 +1,64 @@
...
+- `standalone`, `replicaset`, `sharded`: These tags allow selection based on the MongoDB topology.
+- `loadbalanced`: Allows for selecting tests using a load balancer
```
jmikola, line 58: I noticed that `templates/test/load-balanced.yml` mixes this tag with "sharded". Is that necessary since LB should already imply a sharded topology? Would it make sense to consider this a topology and fold it into the previous line?

.evergreen/architecture.md
```diff
@@ -0,0 +1,64 @@
...
+- `ocsp`: Used for all OCSP tasks
+- `versioned_api`: Used for tests that use the stable API
+- `skip_crypt_shared`: These tasks skip installing the shared library to test with libmongocrypt
+- `nodb`: These tasks do not rely on a local database
+- `atlas`: These tasks work on a MongoDB Atlas cluster
+- `storage-engines`: Tag used for tasks that test the `inmemory` and `mmapv1` storage engines
```
jmikola, line 64: Noted that "serverless" is omitted here. I expect it _will_ be necessary in [PHPLIB-1260](https://jira.mongodb.org/browse/PHPLIB-1260), though.

.evergreen/config.yml
jmikola: @alcaeus: :point_up:

jmikola (commented):
LGTM but there are one or two outstanding questions to follow up on.

Makefile.frag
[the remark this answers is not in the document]
alcaeus: Created PHPC-2306 to evaluate PHPUnit as a test runner for PHPC. I also created PHPC-2307 to track attaching test results separately.

.evergreen/architecture.md
alcaeus: That is correct, `serverless` is used in PHPLIB: https://github.com/mongodb/mongo-php-library/pull/1179/files#diff-741faccf19ca11887ca9ced0106a760c83076a32444223e3e20da68e53bc720eR10

.evergreen/config.yml
alcaeus: Runtime for build tasks is ~2 minutes, while for test tasks it varies between 2 and 5 minutes. Given that this config setting kills tasks that are running too long I don't see an issue with using 30 minutes, but I'm open to reducing it to something more realistic, e.g. 10 minutes.

.evergreen/architecture.md
alcaeus: Changed.

.evergreen/architecture.md
alcaeus: Removed the `sharded` tag from load balanced tests, effectively treating `loadbalanced` like its own topology.

jmikola (approved):

PR #1479, opened 2023-10-11
PHPC-2274 and PHPC-2308: Endian detection for KMS sources

jmikola:
https://jira.mongodb.org/browse/PHPC-2274
https://jira.mongodb.org/browse/PHPC-2308


alcaeus (changes requested):
Compiling on MacOS results in the following error:
> In file included from mongo-php-driver/src/libmongocrypt/src/mongocrypt-buffer.c:18:
src/libmongocrypt/src/mongocrypt-endian-private.h:85:2: error: "The endianness of target architecture is unknown."
#error "The endianness of target architecture is unknown."
 ^

Looking at the corresponding compiler call, `MONGOCRYPT_BIG_ENDIAN` is indeed not present in the call. After looking through the autotools config, I believe this is due to libmongocrypt/Endian.m4 being included in line 375, which correctly sets `PHP_MONGODB_LIBMONGOCRYPT_CFLAGS`. However, line 486 in config.m4 re-initialises that same variable, removing previously added flags.

This used to work because the flags were added to `PHP_MONGODB_BUNDLED_CFLAGS`, which is prepended to `PHP_MONGODB_LIBMONGOCRYPT_CFLAGS` later down the road. FWIW, the endianness detection for libmongoc still writes flags to that variable.

My suggestion would be to keep the flags in the right place to avoid conflicts between bundled libraries, but to initialise the variables before including any autotools configs and then always append values. This would mainly affect the assignment in line 486 of config.m4 where existing flags are reset.

alcaeus (approved):

PR #1480, opened 2023-10-13
PHPC-2219: Prohibit serializing PackedArray as root documents

jmikola:
https://jira.mongodb.org/browse/PHPC-2219

This adds logic to php_phongo_zval_to_bson_internal() to prohibit serializing PackedArray instances as a root document. Since this function is also used in specific cases to encode a BSON array, a PHONGO_BSON_ALLOW_ROOT_ARRAY flag is introduced to relax the restriction.

Includes a separate fix for `php_phongo_field_path_write_item_at_current_level`, which I came across while investigating this.


src/phongo_bson.c
```diff
@@ -162,6 +163,13 @@ void php_phongo_field_path_write_item_at_current_level(php_phongo_field_path* fi
 	php_phongo_field_path_ensure_allocation(field_path, field_path->size);
 
 	if (field_path->owns_elements) {
+		/* Note: owns_elements is only used for field paths parsed from a type
+		 * map, so it's unlikely that an element would already have been
+		 * allocated here. This is in contrast to BSON encoding/decoding, which
+		 * frequently updates the field path and does not own elements. */
+		if (UNEXPECTED(field_path->elements[field_path->size])) {
```
jmikola, line 170: This relies on the PHP macro from [zend_portability.h](https://github.com/php/php-src/blob/php-8.2.11/Zend/zend_portability.h#L362). Alternatively, I could have used [`BSON_UNLIKELY()`](https://github.com/mongodb/mongo-c-driver/blob/1.24.4/src/libbson/src/bson/bson-macros.h#L288).

src/phongo_bson_encode.c
```diff
@@ -466,6 +466,15 @@ static void php_phongo_zval_to_bson_internal(zval* data, php_phongo_field_path*
 			}
 
 			if (instanceof_function(Z_OBJCE_P(data), php_phongo_packedarray_ce)) {
+				/* If we are at the root-level, PackedArray instances should be
+				 * prohibited unless PHONGO_BSON_ALLOW_ROOT_ARRAY is set. */
+				bool is_root_level = (field_path->size == 0);
```
jmikola, line 471: I broke this out into its own variable for readability, since `(field_path->size == 0)` might be intuitive. Not concerned with overhead here, as it seems like something a compiler should easily optimize out.

Note: I didn't bother to share this variable with the enum and Type checks that follow later in the `switch` statement, since the comment there still applies:

> For the error handling that follows, we can safely assume that we are at the root level, since php_phongo_bson_append_object would have already been called for a non-root level.

src/phongo_bson_encode.c
```diff
@@ -466,6 +466,15 @@ static void php_phongo_zval_to_bson_internal(zval* data, php_phongo_field_path*
...
+
+				if (is_root_level && !(flags & PHONGO_BSON_ALLOW_ROOT_ARRAY)) {
+					phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "%s cannot be serialized as a root document", ZSTR_VAL(Z_OBJCE_P(data)->name));
```
jmikola, line 474: I used UnexpectedValueException here for consistency with other BSON encoding exceptions. That is also the only exception we document in [`MongoDB\BSON\fromPHP()`](https://www.php.net/manual/en/function.mongodb.bson-fromphp.php).

It does make the tests a bit odd, since those mostly expect InvalidArgumentException for option validation errors; however, this isn't an error most users should ever encounter anyway.

Lastly, I used "root document" here but left the original error messages for enums and Type instances as "root element". I don't think we need to change those, but I can if you like. Referring to a "document" here seemed more impactful since it contrasts with PackedArray.

src/phongo_bson_encode.h
```diff
@@ -22,9 +22,10 @@
 #include <php.h>
 
 typedef enum {
-	PHONGO_BSON_NONE      = 0x00,
-	PHONGO_BSON_ADD_ID    = 0x01,
-	PHONGO_BSON_RETURN_ID = 0x02
+	PHONGO_BSON_NONE             = 0,
+	PHONGO_BSON_ADD_ID           = (1 << 0),
+	PHONGO_BSON_RETURN_ID        = (1 << 1),
+	PHONGO_BSON_ALLOW_ROOT_ARRAY = (1 << 2)
```
jmikola, line 28: The bit-shift notation here is consistent with various enum flags throughout libmongoc.

tests/bson/bson-fromPHP-001.phpt
```diff
@@ -44,20 +45,28 @@ echo "Testing top-level objects\n";
 foreach ($tests as $test) {
     try {
         echo toJson(fromPHP(new MyDocument($test))), "\n";
+    } catch (Exception $e) {
+        printf("%s: %s\n", get_class($e), $e->getMessage());
+    }
```
jmikola, line 50: Since separate exceptions are thrown here, it was necessary to wrap each `fromPHP()` call in its own try/catch.

While working on this PR, I had briefly used InvalidArgumentException for the PackedArray error, which is why I ended up changing the catch to a generic Exception and printing the class. That's no longer necessary but I think we can keep it as-is since it will lead to more helpful test failure diffs should the exception types ever change.

tests/bulk/bulkwrite-insert-003.phpt
```diff
@@ -11,30 +11,25 @@ require_once __DIR__ . "/../utils/basic.inc";
 $manager = create_test_manager();
 
 $document = MongoDB\BSON\Document::fromPHP(['foo' => 'bar']);
-$array = MongoDB\BSON\PackedArray::fromPHP(['bar']);
 
 $bulk = new MongoDB\Driver\BulkWrite();
 
-$insertedDocumentId = $bulk->insert($document);
-$insertedArrayId = $bulk->insert($array);
```
jmikola, line 19: The PackedArray insertion was entirely removed from this test since it's no longer allowed.

tests/bulk/bulkwrite-insert_error-004.phpt
```diff
@@ -0,0 +1,20 @@
+--TEST--
+MongoDB\Driver\BulkWrite::insert() prohibits PackedArray for document
```
jmikola, line 2: This test covers the case that was removed from `bulkwrite-insert-003.phpt`.

tests/bulk/bulkwrite-update-006.phpt
```diff
@@ -0,0 +1,64 @@
+--TEST--
+MongoDB\Driver\BulkWrite::update() PackedArray for update pipeline
```
jmikola, line 2: This and the arrayFilters test that follows were ported from existing `bulkwrite-update` tests.

tests/readPreference/readpreference-set_state_error-001.phpt
```diff
@@ -7,42 +7,46 @@ require_once __DIR__ . '/../utils/basic.inc';
 
 echo throws(function() {
     MongoDB\Driver\ReadPreference::__set_state(['mode' => 'furthest']);
-}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
+}, MongoDB\Driver\Exception\InvalidArgumentException::class), "\n";
```
jmikola, line 10: Did a search/replace while I was in this file to ensure it was consistent with itself.

tests/bson/bson-javascript_error-004.phpt
```diff
@@ -0,0 +1,18 @@
+--TEST--
+MongoDB\BSON\Javascript::__construct() does not allow PackedArray for scope
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/basic.inc';
+
+echo throws(function() {
+    new MongoDB\BSON\Javascript('function() { return 1; }', MongoDB\BSON\PackedArray::fromPHP([]));
```
jmikola, line 9: On a separate note, I noticed that many of the Javascript tests use `function foo()` as their code string. I don't think the function name is relevant or even correct, given how the server would invoke this, but it also doesn't affect our tests.

That may be worth cleaning up in a separate commit, though, just so we don't have bad example code laying around in the test suite.

alcaeus (approved):
LGTM! Thank you for the comprehensive tests here!

src/phongo_bson.c
alcaeus: SGTM - I believe in the context of a PHP extension `UNEXPECTED` makes more sense than `BSON_UNLIKELY`.

src/phongo_bson_encode.h
alcaeus: Thanks for making the change. This is much easier to read for me.

tests/bson/bson-javascript_error-004.phpt
alcaeus: I agree. Feel free to create a low-hanging-fruits ticket if you don't want to make the change in this pull request.

jmikola: Added a separate commit to fix this and will merge once CI is green.

PR #1490, opened 2023-11-15
Drop ICU information

remicollet:
As utf8proc is now used instead


remicollet:
Having, in phpinfo

`libmongoc ICU => disabled`

may be confusing

jmikola:
@remicollet: Thanks for reporting this. I amended it in https://github.com/mongodb/mongo-php-driver/pull/1491.

PR #1492, opened 2023-11-21
PHPC-2325: Don't throw exception when using coalesce operator in BSON classes

alcaeus:
PHPC-2325

@GromNaN pointed this out today, and while investigating whether this was a PHP issue or our issue I already stumbled over the solution. This PR changes the `read_property` and `read_dimension` object handlers to not throw an exception when it was invoked with `BP_VAR_IS` for `type`. This type is used when `zend_read_property` is invoked with the `silent` parameter set to `true` and in various places throughout `zend_vm_execute.h`.

Through trial and error I discovered that when using the magic `__isset` method in a PHP class, the coalesce operator will use the result of that method call to decide whether to invoke `__get`. If the `__isset` method returns `false`, `__get` is not invoked and the fallback value is used. If `__isset` returns `true`, `__get` is invoked. See [this 3v4l](https://3v4l.org/0WiiV) for a demonstration where `__get` throws.
For built-in classes however, it seems that the `has_property` and `has_dimension` object handlers are not called. Instead, the `read_property` and `read_dimension` handlers are called with a `BP_VAR_IS` value for `type`, and the handler itself is responsible for deciding not to throw.

Note that this only affects property/dimension access, and invoking the `get()` methods will not change. The coalesce operator does not guard against exceptions, it only guards against `null` values, which include undefined properties/dimensions.


tests/bson/bson-document-property-access-002.phpt
```diff
@@ -0,0 +1,19 @@
...
+
+require_once __DIR__ . '/../utils/basic.inc';
+
+$document = MongoDB\BSON\Document::fromPHP([]);
+
+var_dump(isset($document->foo));
+var_dump($document->foo ?? 'fallback');
+
+?>
+===DONE===
```
GromNaN, line 14: Why do you need this `==DONE==`?

GromNaN (approved):
So it works exactly as we expect it to.

.evergreen/config/functions.yml
```diff
@@ -103,7 +103,7 @@ functions:
         content_type: ${content_type|application/x-gzip}
         permissions: public-read
         local_file: ${build_id}.tar.gz
-        remote_file: ${project}/${build_variant}/${revision}/${task_name}/${version}.tar.gz
+        remote_file: mongo-php-driver/${build_variant}/${revision}/${task_name}/${version_id}.tar.gz
```
jmikola, line 106: Noted that this fixes the v1.17 Evergreen failures I [mentioned in Slack earlier](https://mongodb.slack.com/archives/C026UA4GF0B/p1700074313027349).

src/BSON/Document.c
```diff
@@ -553,8 +554,14 @@ zval* php_phongo_document_read_property(phongo_compat_object_handler_type* objec
...
 	if (!php_phongo_document_get(intern, key, key_len, rv)) {
-		// Exception already thrown
-		return &EG(uninitialized_zval);
+		if (type != BP_VAR_IS) {
+			// Exception already thrown
+			return &EG(uninitialized_zval);
+		}
+
+		// php_phongo_document_get will throw. If we're invoked with a BP_VAR_IS type, clear the exception and return a null value
+		zend_clear_exception();
```
jmikola, line 563: Is it worth reworking this to avoid throwing entirely? I imagine this would necessitate a change to `php_phongo_document_get()` (e.g. adding a `null_if_missing` bool), but the method is static to this file anyway so it'd be fairly self-contained.

This would also let you honor the more serious exception if `bson_iter_init()` fails, which should likely never happen in practice:

```
phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not initialize BSON iterator");
```

src/BSON/PackedArray.c
```diff
@@ -477,13 +478,23 @@ zval* php_phongo_packedarray_read_dimension(phongo_compat_object_handler_type* o
...
 	if (Z_TYPE_P(offset) != IS_LONG) {
-		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not find index of type \"%s\" in BSON array", PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(offset));
-		return &EG(uninitialized_zval);
+		if (type != BP_VAR_IS) {
+			phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not find index of type \"%s\" in BSON array", PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(offset));
+			return &EG(uninitialized_zval);
+		}
+
+		ZVAL_NULL(rv);
+		return rv;
```
jmikola, line 487:
```suggestion
		if (type == BP_VAR_IS) {
			ZVAL_NULL(rv);
			return rv;
		}

		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not find index of type \"%s\" in BSON array", PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(offset));
		return &EG(uninitialized_zval);
```

src/BSON/Document.c
```diff
@@ -591,13 +598,25 @@ zval* php_phongo_document_read_dimension(phongo_compat_object_handler_type* obje
...
-		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not find key of type \"%s\" in BSON document", PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(offset));
-		return &EG(uninitialized_zval);
+		if (type != BP_VAR_IS) {
+			phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not find key of type \"%s\" in BSON document", PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(offset));
+			return &EG(uninitialized_zval);
+		}
+
+		ZVAL_NULL(rv);
+
+		return rv;
```
jmikola, line 608: Should improve readability.

```suggestion
		if (type == BP_VAR_IS) {
			ZVAL_NULL(rv);
			return rv;
		}

		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not find key of type \"%s\" in BSON document", PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(offset));

		return &EG(uninitialized_zval);
```

tests/bson/bson-document-array-access-003.phpt
```diff
@@ -0,0 +1,19 @@
...
+
+$document = MongoDB\BSON\Document::fromPHP([]);
+
+var_dump(isset($document['foo']));
+var_dump($document['foo'] ?? 'fallback');
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
```
jmikola, line 16:
```suggestion
--EXPECT--
```

tests/bson/bson-document-property-access-002.phpt
```diff
@@ -0,0 +1,19 @@
...
+<?php exit(0); ?>
+--EXPECTF--
```
jmikola, line 16:
```suggestion
--EXPECT--
```

tests/bson/bson-document-property-access-002.phpt
jmikola: Is this a general question or specific to this test? Generally, I can tell you that `==DONE==` has been used to separate test output from possible memory leaks info when running tests on debug version of PHP. That's also why we manually call `exit(0)`.

tests/bson/bson-packedarray-array-access-003.phpt
```diff
@@ -0,0 +1,19 @@
...
+
+$array = MongoDB\BSON\PackedArray::fromPHP([]);
+
+var_dump(isset($array[0]));
+var_dump($array[0] ?? 'fallback');
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
```
jmikola, line 16:
```suggestion
--EXPECT--
```

src/BSON/Document.c
alcaeus: Added, makes sense to still throw if the BSON structure is invalid.

src/BSON/PackedArray.c
alcaeus: Updated 👍

src/BSON/Document.c
alcaeus: Updated 👍

.evergreen/config/functions.yml
alcaeus: I extracted the commit to a separate pull request, which has since been merged.

jmikola (approved):

PR #1513, opened 2024-01-25
PHPC-1956: Remove disabled __wakeup function in non-serializable classes 

jmikola:
https://jira.mongodb.org/browse/PHPC-1956


php_phongo.c
```diff
@@ -56,27 +56,28 @@ zend_object_handlers* phongo_get_std_object_handlers(void)
 /* }}} */
 
 /* {{{ Memory allocation wrappers */
-static void* php_phongo_malloc(size_t num_bytes) /* {{{ */
```
alcaeus, line 59: I thought I had removed all of these, but I guess I haven't. Thanks for cleaning this up!

php_phongo.h
```diff
@@ -117,15 +117,4 @@ zend_object_handlers* phongo_get_std_object_handlers(void);
...
 
-#define PHONGO_DISABLED_WAKEUP(classname)                                                             \
-	static PHP_METHOD(classname, __wakeup)                                                            \
-	{                                                                                                 \
-		PHONGO_PARSE_PARAMETERS_NONE();                                                               \
-		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "MongoDB\\Driver objects cannot be serialized"); \
-	}
-
-/* Shared function entries for disabling constructors and unserialize() */
-PHP_FUNCTION(MongoDB_disabled___construct);
```
alcaeus, line 128: Noted that I added the macros in https://github.com/mongodb/mongo-php-driver/commit/6887226d34566cd772bbacb34e10fc7bd4000f57 but forgot to remove the shared functions below. Thanks for cleaning up my mess!

alcaeus (approved):

php_phongo.c
jmikola: Is "these" referring to the foldmarkers (i.e. `{{{` and `}}}`)? There are a bunch more, so I can go through and clean those up separately if we like. None of us are using vim to work on this project, so I don't have to worry about @bjori yelling at me.

PR #1514, opened 2024-02-05
PHPC-2356: Remove explicit env variables

alcaeus:
PHPC-2356
This will prevent accidental leaking of environment variables.

Note that the URI list for Atlas connectivity tests are now contained in an environment variable in the project config, which is added to the environment via the `include_expansions_in_env` option in `run tests`. This makes it available for us to use in tests without having to decrypt a file and potentially leak the encryption credentials.


.evergreen/config/test-variants.yml
```diff
@@ -19,7 +19,7 @@ buildvariants:
       - ".sharded .local !.3.6 !.4.0 !.4.2 !.4.4 !.5.0"
       - ".loadbalanced .local !.3.6 !.4.0 !.4.2 !.4.4 !.5.0"
       - "test-atlas-connectivity"
-      - ".ocsp !.4.4"
+      - ".ocsp !.4.4 !.5.0"
```
jmikola, line 22: IIUC, this drops OCSP testing on 5.0. What the impetus for this change? The [OCSP spec test README](https://github.com/mongodb/specifications/blob/master/source/ocsp-support/tests/README.rst) doesn't talk about server versions.

.evergreen/config/functions.yml
```diff
@@ -242,12 +256,6 @@ functions:
           # Don't use ${file} syntax here because evergreen treats it as an empty expansion.
           [ -f "$file" ] && sh $file || echo "$file not available, skipping"
 
-  "decrypt atlas URI list":
-    - command: shell.exec
-      params:
-        script: |
-          openssl aes-256-cbc -S "${encrypted_uris_salt}" -K "${encrypted_uris_key}" -iv "${encrypted_uris_iv}" -in ${PROJECT_DIRECTORY}/.evergreen/atlas-uris.txt.enc -out ${PROJECT_DIRECTORY}/.evergreen/atlas-uris.txt -d
```
jmikola, line 249: > This makes it available for us to use in tests without having to decrypt a file and potentially leak the encryption credentials.

jmikola, line 249: > This makes it available for us to use in tests without having to decrypt a file and potentially leak the encryption credentials.

Are other drivers doing this as well? I don't have an opinion on which is more secure, but storing the URIs directly in the Evergreen project config is certainly more convenient.

Note: you can also remove `.evergreen/atlas-uris.txt` from `.gitignore`.

.evergreen/config/functions.yml
```diff
@@ -185,28 +195,32 @@ functions:
...
-          USE_DELEGATE=${USE_DELEGATE} \
           ${PROJECT_DIRECTORY}/.evergreen/run-ocsp-responder.sh
 
   "run tests":
     - command: shell.exec
       type: test
       params:
+        include_expansions_in_env:
+          - API_VERSION
+          - ATLAS_CONNECTIVITY_URIS
```
jmikola, line 213: Should this always be specified? In the `test-atlas-connectivity` task, `run tests` is invoked with `vars: { TESTS: "tests/atlas.phpt" }`, so I would expect that's the only case where you'd actually want to pass this potentially sensitive environment variable onwards.

.evergreen/config/functions.yml
```diff
@@ -167,9 +167,19 @@ functions:
   "bootstrap mongo-orchestration":
     - command: shell.exec
       params:
+        include_expansions_in_env:
```
jmikola, line 170: Noted that `vars` (when calling a `func`) actually define expansions (per [Functions](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files#functions)). That distinction wasn't clear to me before this PR.

.evergreen/run-tests.sh
```diff
@@ -4,7 +4,7 @@ set -o errexit  # Exit the script with error if any of the commands fail
 # Supported environment variables
 API_VERSION=${API_VERSION:-} # Optional API_VERSION environment variable for run-tests.php
 CRYPT_SHARED_LIB_PATH="${CRYPT_SHARED_LIB_PATH:-}" # Optional path to crypt_shared library
-MONGODB_URI=${MONGODB_URI:-} # Connection string (including credentials and topology info)
+MONGODB_URI=${MONGODB_URI:-}${APPEND_URI:-} # Connection string (including credentials and topology info)
```
jmikola, line 7: Nit: the "Supported environment variables" section here doubles as documentation, so I think it'd be clearer to take `APPEND_URI` separately in the same form and then merge them below.

That said, it's only essential that this merging happens _before_ the "Determine if MONGODB_URI already has a query string" logic below.

alcaeus: Good point. I've moved the merging of `APPEND_URI` to before detecting the suffix.

tests/utils/skipif.php
```diff
@@ -494,3 +494,10 @@ function skip_if_no_crypt_shared()
         exit('skip crypt_shared is not available');
     }
 }
+
+function skip_if_no_atlas_connectivity_urls()
+{
+    if (getAtlasConnectivityUrls() === []) {
+        exit('skip No atlas URIs found');
```
jmikola, line 501:
```suggestion
        exit('skip No Atlas URIs found');
```

jmikola (approved):

PR #1520, opened 2024-03-07
Update submodule sources when building driver

alcaeus:
This PR adds a call to update-submodule-sources.php when compiling against a different version of libmongoc than the bundled one. This helps avoid "symbol not found" errors when compiling with libmongoc latest with new files not included in the bundled stable release.

### DIFF
--- a/.evergreen/compile-unix.sh
+++ b/.evergreen/compile-unix.sh
@@ -65,6 +65,11 @@ esac
 # Report the current PHP version
 echo "PHP: `php --version | head -n 1`"
 
+# If we're testing a specific version of libmongoc, update submodule sources
+if [ -n "$LIBMONGOC_VERSION" ]; then
+   php scripts/update-submodule-sources.php
+fi
+
 phpize
 ./configure --enable-mongodb-developer-flags
 
--- a/.evergreen/compile-unix.sh
+++ b/.evergreen/compile-unix.sh
@@ -73,9 +73,11 @@ fi
 phpize
 ./configure --enable-mongodb-developer-flags
 
-# If we're testing a specific version of libmongoc, regenerate the version file
+# configure relies on version information in libmongoc-version-current, but the target is not available until after calling configure
+# To work around this, run the make target, then run configure again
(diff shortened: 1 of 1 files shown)


.evergreen/compile-unix.sh
```diff
@@ -65,6 +65,11 @@ esac
 # Report the current PHP version
 echo "PHP: `php --version | head -n 1`"
 
+# If we're testing a specific version of libmongoc, update submodule sources
+if [ -n "$LIBMONGOC_VERSION" ]; then
+   php scripts/update-submodule-sources.php
+fi
```
jmikola, line 71: Below we have a conditional for `make libmongoc-version-current`. I suppose that still needs to remain separate because `phpize` is what builds `Makefile` from `Makefile.frag`. So, at best, we could move that _between_ the `phpize` and `./configure` lines below.

And on a related note: the `LIBMONGOC_VERSION_CURRENT` file is actually referenced _during_ `./configure` to create the substitutions, so shouldn't we be running `make libmongoc-version-current` _before_ `./configure`? As-is, I don't think the compiled driver reports the correct version info.

Do you want to handle that in this same PR?

alcaeus: Good point: `LIBMONGOC_VERSION_CURRENT` is indeed referenced during `./configure`, but we have to run `./configure` in order to have the `Makefile` generated. To avoid that conundrum, I now call `phpize` and `./configure`, then run `make libmongoc-version-current` before running `./configure` again to generate the final `Makefile` that we're building with. It's a little messy but the only way to work around the issue.

jmikola: I wouldn't complain if you want to just manually run:

```
cd src/libmongoc/ && python build/calc_release_version.py > ../LIBMONGOC_VERSION_CURRENT
```

jmikola (approved):
LGTM. I'll defer to you if you'd rather avoid the extra `./configure` call.

alcaeus:
Noted after merging that this takes care of PHPC-2006, so I'm linking that here.

PR #1522, opened 2024-03-08
Update README.md

techbelle:
added laravel framework callout

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,11 @@ languages.
  - https://php.net/manual/en/set.mongodb.php
  - https://www.mongodb.com/docs/drivers/php/
 
+## Integrations
+If you'd like to use PHP with Laravel framework, please see [laravel-mongodb](https://github.com/mongodb/laravel-mongodb)
+
+
+
 ## Installation
 
 To build and install the driver:


README.md
```diff
@@ -22,6 +22,11 @@ languages.
  - https://php.net/manual/en/set.mongodb.php
  - https://www.mongodb.com/docs/drivers/php/
 
+## Integrations
+If you'd like to use PHP with Laravel framework, please see [laravel-mongodb](https://github.com/mongodb/laravel-mongodb)
```
GromNaN, line 26: Good idea. I think we should add a link to Doctrine MongoDB ODM and the Symfony Bundle also.

We have a longer list of integrations here: https://www.mongodb.com/docs/drivers/php-libraries/#framework-integrations

jmikola: A few observations:

 * This looks like a very hasty PR. There is a period missing after the sentence, and the newlines are all over the place (three at the end and none between the heading and paragraph).
 * It looks like this was only opened on the PHPC repo and not PHPLIB. I assume we'd like equivalent content in both READMEs, so let's make a point to duplicate whatever change we add here in PHPLIB.
 * We already maintain a list of integrations in the docs, so I suggest we link to https://www.mongodb.com/docs/drivers/php-libraries/ alongside any call-out of the Laravel library.
 * Adding links to integrations we don't officially maintain may invite users to submit PRs to add their own, and I don't think any of us want to deal with that -- I still have horrible memories of the days when our docs page was a free-for-all.
 * I strongly disagree with placing "Integrations" above "Installation" in the README. Rather than introducing an extra section, consider adding a paragraph to the top section _after_ we point users to PHPLIB.

Consider this text:

```md
...[MongoDB PHP library](https://github.com/mongodb/mongo-php-library), which
implements the same higher level APIs found in MongoDB drivers for other
languages.

Please see [PHP Libraries, Frameworks, and Tools](https://www.mongodb.com/docs/drivers/php-libraries/)
for a list of MongoDB integrations for PHP, such as the official
[Laravel MongoDB](https://github.com/mongodb/laravel-mongodb) library.
```

----

> [!Warning]
> I just returning from picking up our family car to the mechanic (shout out to [r/fuckcars](https://www.reddit.com/r/fuckcars/)), so I'm trying to rein myself in over the level of kvetching in this reply. But I must reiterate how disappointed I am with the quality of this diff coming from an esteemed, published author (@techbelle... :eyes:).

PR #1534, opened 2024-04-11
Prepare 1.18.1 release

alcaeus:
Note: this PR is mainly here to try out a new release workflow with new branch protections in mind. It is not a blueprint for future releases.

This PR must not be merged using a squash merge, as it needs to retain the "Package 1.18.1" commit, which is the base for the release tag.

### DIFF
--- a/phongo_version.h
+++ b/phongo_version.h
@@ -22,9 +22,9 @@
  * publishing a release. */
 
 /* clang-format off */
-#define PHP_MONGODB_VERSION "1.18.1-dev"
-#define PHP_MONGODB_STABILITY "devel"
-#define PHP_MONGODB_VERSION_DESC 1,18,1,0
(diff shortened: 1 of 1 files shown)


phongo_version.h
```diff
@@ -22,9 +22,9 @@
  * publishing a release. */
 
 /* clang-format off */
-#define PHP_MONGODB_VERSION "1.18.1-dev"
+#define PHP_MONGODB_VERSION "1.18.2-dev"
```
jmikola, line 25: We discussed this earlier, but to repeat here: I think the process would benefit from splitting the changes up into separate PRs. That would presumably not require any special merging if automation were to create the tag after the PR is merged (in any fashion). And it would also make reviewing the diff(s) more straightforward since the most meaningful change (the "Packed 1.18.1" commit) would not longer be hidden.

jmikola (approved):

PR #1535, opened 2024-04-11
Disable fail-fast for Windows Tests

alcaeus:
Since Windows tests often suffer from long startup times of the MongoDB daemon, the failure of a single job cancels the entire workflow. Disabling `fail-fast` will prevent this and help get at least some test coverage without restarting all builds.

### DIFF
--- a/.github/workflows/windows-tests.yml
+++ b/.github/workflows/windows-tests.yml
@@ -86,6 +86,8 @@ jobs:
         shell: cmd
 
     strategy:
+      # Timeouts starting the MongoDB service are common and should not
+      # interrupt the overall test matrix.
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-tests.yml
```diff
@@ -86,7 +86,7 @@ jobs:
         shell: cmd
 
     strategy:
-      fail-fast: true
+      fail-fast: false
```
jmikola, line 89:
```suggestion
      # Timeouts starting the MongoDB service are common and should not
      # interrupt the overall test matrix.
      fail-fast: false
```

jmikola (approved):

PR #1538, opened 2024-04-18
PHPC-1674: Automate Driver Releases

alcaeus:
PHPC-1674

This PR adds release automation to reduce the number of manual steps required for the release. Manual steps are now limited to branch management (i.e. creating a new version branch before releasing a minor version) and merging up release commits. The commit process (e.g. changing to a stable version number in phongo_version.h), tagging a release, and building release artifacts including a PECL archive are now automated.

Signing release artifacts will be part of a separate PR.

Todo list:
- [x] Support filling in release notes in `make package.xml`
- [x] Build PECL package when a tag is pushed
- [x] Build Windows DLLs when a tag is pushed
- [x] TBD: Generate changelog
- [x] Create release tag automatically (including changes to phongo_version.h)
- [x] Create draft release


.github/workflows/release.yml
```diff
@@ -0,0 +1,142 @@
+name: "Release New Version"
+run-name: "Release ${{ inputs.version }}"
+
+on:
+  workflow_dispatch:
```
alcaeus, line 5: `workflow_dispatch` allows a manual trigger by anyone with write access to the repository. If we want to limit the number of people that can start a release, we'd have to add checks in this workflow, but I don't think that's necessary at this point.

.github/workflows/release.yml
```diff
@@ -0,0 +1,142 @@
...
+      # TODO: Sign tag
+      - name: "Create release tag"
+        run: git tag -a -F changelog ${{ env.PACKAGE_VERSION }}
+
+      - name: "Update version information to next patch development release"
+        run: |
+          ./bin/update-release-version.php patch
+          git commit -m "Back to -dev" phongo_version.h
+
+      # TODO: Manually merge using ours strategy. This avoids merge-up pull requests being created
```
alcaeus, line 116: I'll defer this to a separate PR as it requires some work to make this logic from the merge-up action reusable here.

phongo_version.h
```diff
@@ -22,7 +22,7 @@
  * publishing a release. */
 
 /* clang-format off */
-#define PHP_MONGODB_VERSION "1.18.2-dev"
+#define PHP_MONGODB_VERSION "1.18.2dev"
```
alcaeus, line 25: During testing, I was trying to create a package with `devel` stability and PECL complained that the `-dev` suffix is wrong. I dislike that difference, but hopefully it won't be much longer until we can finally say goodbye to pecl.

alcaeus (commented):
@jmikola left some explanatory comments in here. I still need to figure out why the changelog is not read correctly from the tag (it reads the message of the underlying commit instead of the message of the annotated tag), but other than that this is ready for a first look.

I revamped the release documentation significantly but left it in the current place. Once you've looked through and it looks good, I'd extract the release documentation to a RELEASING.md file as we've done for PHPLIB.

.github/workflows/windows/build/action.yml
```diff
@@ -0,0 +1,30 @@
...
+    required: true
+outputs:
+  build-dir:
+    description: "The build directory to be used"
+    value: ${{ steps.prepare-build-env.outputs.build-dir }}
+runs:
+  using: composite
+  steps:
+    - name: Prepare build environment
+      id: prepare-build-env
```
jmikola, line 21: Should the `windows/prepare-build` action just be consolidated into this file? Unless I overlooked something, it doesn't appear to be referenced anywhere else.

CONTRIBUTING.md
```diff
@@ -323,7 +332,64 @@ Update the version's release date and status from the
...
 
-### Update version info
+### Trigger the release workflow
+
+Releases are done automatically through a GitHub Action. Visit the corresponding
+[Release New Version](https://github.com/mongodb/mongo-php-driver/actions/workflows/release.yml)
+workflow page to trigger a new build. Select the correct branch (e.g. `v1.18`)
+and trigger a new run using the "Run workflow" button. In the following prompt,
+enter the version number and the corresponding JIRA release number for the 
+release. This will create and push the necessary commits and tag, create a draft
```
jmikola, line 342: This would be a good place to link to https://jira.mongodb.org/projects/PHPC and advise that the version ID can be copied from a link in the "Version" column (e.g. "37551" from https://jira.mongodb.org/browse/PHPC/fixforversion/37551 for 1.18.0).

CONTRIBUTING.md
```diff
@@ -323,7 +332,64 @@ Update the version's release date and status from the
...
+
+Once the packaging workflow has finished creating the PECL package, it will be
+published as a build artifact of the package workflow, as well as a release
+artifact in the draft release. This package may be published via the
+[Release Upload](https://pecl.php.net/release-upload.php) form. You will have
+one chance to confirm the package information after uploading.
+
+> **Note:** If downloading the package from the build artifacts, be aware that
+> these are always provided as zip files. You'll have to unpack the tgz archive
+> prior to uploading it to PECL.
```
jmikola, line 357: Is this a quirk of [gh release upload](https://cli.github.com/manual/gh_release_upload) that everything gets wrapped in a zip file? I don't see that mentioned in the GitHub docs.

CONTRIBUTING.md
```diff
@@ -411,24 +450,22 @@ $ git tag -a -m "Release X.Y.Z" X.Y.Z
...
 
-```diff
--#define PHP_MONGODB_VERSION "1.1.8"
--#define PHP_MONGODB_STABILITY "stable"
--#define PHP_MONGODB_VERSION_DESC 1,1,8,1
-+#define PHP_MONGODB_VERSION "1.1.9-dev"
-+#define PHP_MONGODB_STABILITY "devel"
-+#define PHP_MONGODB_VERSION_DESC 1,1,9,0
+```shell
+$ ./bin/update-release-version.php patch
```
jmikola, line 468: Should you solely demonstrating `update-release-version.php patch` here? I think there are cases where `minor` may also be appropriate, especially since you removed the "After releasing a new minor version" section.

This content was originally under [Releasing](https://github.com/mongodb/mongo-php-driver/blob/1.18.1/CONTRIBUTING.md#releasing) and presented instructions common to both minor _and_ patch releases.

If this _should_ demonstrate `update-release-version.php patch`, then I think we're missing some other section that discusses when to use `update-release-version.php minor` (after cutting a new branch).

bin/prep-release.php
```diff
@@ -190,6 +190,7 @@ function usage() {
     "%RELEASE_VERSION%" => $VERSION,
     "%RELEASE_STABILITY%" => $STABILITY,
     "%RELEASE_FILES%" => join("\n", $TREE),
+    "%RELEASE_NOTES%" => file_exists(__DIR__ . '/../changelog') ? file_get_contents(__DIR__ . '/../changelog') : '',
```
jmikola, line 193: Consider making this an optional script argument instead of hard-coding a file. That would remove coupling between this script and the GitHub Actions.

.github/workflows/package-release.yml
```diff
@@ -0,0 +1,113 @@
...
+        with:
+          version: "8.3"
+
+      - name: "Write changelog file for packaging"
+        run: git tag -l ${{ github.ref_name }} --format='%(contents)' > changelog
+
+      - name: "Build package.xml"
+        run: "make package.xml"
+
+      - name: "Build release archive"
```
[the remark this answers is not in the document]
alcaeus: The script invoked by the make target already fills in release notes. I left the steps separate to not break the current manual release process.
I can add a comment explaining that the action always places a `changelog` file and thus fills the release notes.

.github/workflows/package-release.yml
```diff
@@ -0,0 +1,113 @@
...
+        run: "make package"
+
+      # PECL always uses the version for the package name.
+      # Read it from the version file and store in env to use when uploading artifacts
+      - name: "Read current package version"
+        run: echo "PACKAGE_VERSION=$(./bin/update-release-version.php version)" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
alcaeus: I looked at that initially, but couldn't figure out the correct shell syntax to only print the filename of the generated file. Relying on the naming convention in combination with extracting the version number seemed like a sensible alternative.

.github/workflows/package-release.yml
```diff
@@ -0,0 +1,113 @@
...
+
+  build-windows:
+    name: "Create Windows package"
+    runs-on: windows-2022
+    defaults:
+      run:
+        shell: cmd
+
+    strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
alcaeus: I used `fail-fast: false` to ensure that if any build errored, other builds would be unaffected. For example, if compilation for PHP 7.4 failed, I wouldn't want to abort all previous builds but rather complete them and only have to revisit the failing one.

I'll note that this shouldn't ever happen as any build process failures would already be detected by the pull request that introduces the breakage; since the tag we're building will only contain the version update commit that didn't run through CI, I'm expecting no breakage here.

.github/workflows/release.yml
```diff
@@ -0,0 +1,142 @@
...
+          # fetch-depth 0 is required to fetch all branches, not just the branch being built
+          fetch-depth: 0
+          submodules: true
+          token: ${{ env.GH_TOKEN }}
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          tools: "phpize"
```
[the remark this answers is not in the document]
alcaeus: You are correct, it isn't used here at all and I've removed it. Revisiting the documentation, it seems like we don't need it at all:
> Tools `pear`, `pecl`, `phpize` and `php-config` are set up by default for all supported PHP versions on Linux and macOS.

I've removed this across the board.

.github/workflows/windows/build/action.yml
alcaeus: The action is used twice: once called from the build action, but also called from the `test` job in the `windows-tests` workflow. While that job reuses the build files from the `build` job in the same workflow, we still need to prepare the build environment to put the DLL file in the correct place.

CONTRIBUTING.md
alcaeus: Release assets (so what we upload using `gh release upload` are left as-is - but build artifacts are always zip files, even if we only upload a single file. Quoting the [documentation](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts):
> When an artifact is downloaded using the GitHub UI, all files that were individually uploaded as part of the artifact get zipped together into a single file.

bin/prep-release.php
alcaeus: I used a file in order to avoid dealing with escaping a multi-line string in the shell, especially as it contains potentially user-submitted data (e.g. pull request titles). The generated notes are only ever passed through files to ensure we don't open ourselves up to any script injection or data leaks due to shell expansions (e.g. because somebody placed `$GITHUB_ENV` in a pull request title which could then be expanded into the entire env value.

.github/workflows/release.yml
```diff
@@ -0,0 +1,142 @@
...
+    name: "Prepare release"
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: "Create release output"
+        run: echo '🎬 Release process for version ${{ inputs.version }} started by @${{ github.triggering_actor }}' >> $GITHUB_STEP_SUMMARY
+
+      - uses: actions/checkout@v4
+        with:
+          # fetch-depth 0 is required to fetch all branches, not just the branch being built
```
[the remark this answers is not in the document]
alcaeus: That's correct. I've removed it in the meantime and have also cleaned up other usages of `fetch-depth` where it doesn't seem necessary.

bin/prep-release.php
alcaeus: Good point. I've added an optional parameter to the script and updated the makefile and package-release workflow.

bin/update-release-version.php
```diff
@@ -0,0 +1,173 @@
...
+    global $argv;
+
+echo <<<EOT
+Usage:
+{$argv[0]} <command>
+
+Commands:
+    stable:  Mark the current version as stable
+    patch:   Update to the next patch development version
+    minor:   Update to the next minor development version
```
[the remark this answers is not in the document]
alcaeus: While the names are verbose, they are mostly invoked in automated scripts, so I don't mind it. The new naming will also make it easier to later handle pre-release versions where we won't use `to-stable` and `to-next-patch-dev` but rather just a single increment (e.g. `to-next-beta`).

.github/workflows/package-release.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: "Upload artifact"
+        uses: actions/upload-artifact@v4
+        with:
+          name: mongodb-${{ env.PACKAGE_VERSION }}.tgz
+          path: mongodb-${{ env.PACKAGE_VERSION }}.tgz
+          retention-days: 3
+
+      - name: "Upload release artifact"
+        run: gh release upload ${{ github.ref_name }} mongodb-${{ env.PACKAGE_VERSION }}.tgz
+        # Until we've changed the process to always have a release draft, ignore errors during upload
```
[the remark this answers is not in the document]
alcaeus: The comment was left over from an initial prototype where I hadn't created the release yet, so it's definitely obsolete.

However, I'm leaving `continue-on-error` in for the time being, as the process right now supports building packages for tags that don't have a release associated with them (think manually pushing a tag). Once we've confirmed that all our release types are appropriately covered, we can remove this and let the build fail if it can't find a release for the tag being built.

CONTRIBUTING.md
```diff
@@ -296,18 +296,27 @@ generation script.
...
-The following steps outline the release process for both new minor versions (e.g.
-releasing the `master` branch as X.Y.0) and patch versions (e.g. releasing the
-`vX.Y` branch as X.Y.Z).
+### Branch management
 
-The command examples below assume that the canonical "mongodb" repository has
-the remote name "mongodb". You may need to adjust these commands if you've given
-the remote another name (e.g. "upstream"). The "origin" remote name was not used
-as it likely refers to your personal fork.
+When releasing a new minor version of the driver (i.e. `vX.Y.0`), create a new
```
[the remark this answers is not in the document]
alcaeus: Fixed.

CONTRIBUTING.md
alcaeus: Re-worded and included instructions on how to find the ID.

CONTRIBUTING.md
alcaeus: I assumed that the "Branch Management" section is always done, regardless of the release process. Essentially, before releasing a new minor version we create the release branch for that minor release, then update the default branch to the next minor release. Only after that does the release process (either automated or manual) start, always working from the version branch.

.github/workflows/package-release.yml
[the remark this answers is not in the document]
jmikola: I likely left this comment before reviewing `prep-release.php`. Regardless, I think a comment above the preceding step that invokes `make package.xml` would be helpful.

.github/workflows/package-release.yml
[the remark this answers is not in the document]
jmikola: Oh, I misread this as `fail-fast: true`. I also didn't realize that was the [default behavior](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast).

_Not_ failing fast definitely makes sense. Feel free to resolve after reading.

.github/workflows/package-release.yml
```diff
@@ -0,0 +1,113 @@
...
+        run: "make package"
+
+      # PECL always uses the version for the package name.
+      # Read it from the version file and store in env to use when uploading artifacts
+      - name: "Read current package version"
+        run: echo "PACKAGE_VERSION=$(./bin/update-release-version.php version)" >> "$GITHUB_ENV"
+
+      # Currently fails with: Cannot install, php_dir for channel "pecl.php.net" is not writeable by the current user
+      # - name: "Install release archive to verify correctness"
+      #  run: pecl install mongodb-${{ env.PACKAGE_VERSION }}.tgz
```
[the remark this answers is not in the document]
alcaeus: Using `sudo` fixes this. I also changed the extensions we install when setting up PHP to avoid `pecl install` failing because the extension is already loaded.

.github/workflows/package-release.yml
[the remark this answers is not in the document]
alcaeus: I added that when I changed prep-release.php to take the filename from a CLI argument instead of hardcoding a filename.

.github/workflows/linux/build/action.yml
```diff
@@ -11,6 +11,8 @@ runs:
       uses: "shivammathur/setup-php@v2"
       with:
         php-version: "${{ inputs.version }}"
+        # Only install required extensions
+        extensions: "none,date,json,spl,standard,xml"
```
jmikola, line 15: What is "none"?

alcaeus: `none` indicates that no default extensions should be installed. The action allows disabling individual extensions using `:<extension>` or disabling all using `none` followed by specifying extensions we want. I figured that was more sensible than only excluding the mongodb extension. This will also point out if we introduce a dependency on a new extension.

.github/workflows/package-release.yml
```diff
@@ -0,0 +1,110 @@
...
+        uses: ./.github/workflows/linux/build
+        with:
+          version: "8.3"
+
+      - name: "Write changelog file for packaging"
+        run: git tag -l ${{ github.ref_name }} --format='%(contents)' > changelog
+
+      # This will fill in the release notes from the previously generated changelog
+      - name: "Build package.xml"
+        run: "make package.xml RELEASE_NOTES_FILE=$(pwd)/changelog"
```
jmikola, line 33: Why is `$(pwd)` used here, but `./` earlier? Wouldn't all of these steps run in the same directory if you're not explicitly changing paths?

.github/workflows/release.yml
jmikola: Reminder to ticket this if needed so we don't lose track of it.

I assume we already have a ticket to track the earlier TODO for release signing.

.github/workflows/windows/build/action.yml
```diff
@@ -0,0 +1,30 @@
+name: "Build DLL files for Windows"
+description: "Prepares the PHP build environment for the MongoDB driver"
+inputs:
+  version:
+    description: "PHP version to build for"
+    required: true
+  arch:
+    description: "The architecture to build for (x64 or x86)"
+    required: true
+  ts:
```
jmikola, line 10: Do you want to consider using choice and boolean fields for `arch` and `ts`, respectively? I noted that php/setup-php-sdk expects strings for both, so perhaps the choice field would make to most sense (since boolean would require some logic later to derive the correct strings).

See: https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/

CONTRIBUTING.md
```diff
@@ -323,7 +332,77 @@ Update the version's release date and status from the
...
+
+### Upload package to PECL
+
+Once the packaging workflow has finished creating the PECL package, it will be
+published as a build artifact of the package workflow, as well as a release
+asset in the draft release. This package may be published via the
+[Release Upload](https://pecl.php.net/release-upload.php) form. You will have one chance to confirm the package
+information after uploading.
+
+> **Note:** If downloading the package from the build artifacts, be aware that
```
jmikola, line 361: If you'd like to use the special syntax (learned about this while specs were being ported from RST to Markdown): https://github.com/orgs/community/discussions/16925

bin/update-release-version.php
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    $newVersions = [
+        'version' => $version,
+        'stability' => 'stable',
+        'versionComponents' => $versions['versionComponents'],
+    ];
+
+    // Increase build number
+    $newVersions['versionComponents'][3]++;
```
jmikola, line 107: This is the only place I see the build number for `PHP_MONGODB_VERSION_DESC` being modified. Is there somewhere else that resets it to zero for new releases? Or maybe I'm missing a place where it's initialized to zero.

I just want to confirm that this bit works as expected.

jmikola (commented):
A couple of questions to address but LGTM with or without changes.

.github/workflows/package-release.yml
alcaeus: The filename is used in a PHP file that is in a different directory. I could either assume that it's always a path relative to the root directory and use `__DIR__`, or ensure that we're passing an absolute path. I could be missing something, but I figured an absolute path would be the safest way to avoid issues.

.github/workflows/windows/build/action.yml
alcaeus: The reference you linked refers to inputs that are used when triggering a workflow run (as is done for release.yml). This file contains metadata for an action, where inputs are always of type string: https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputs

bin/update-release-version.php
alcaeus: It's increased here for `get_stable_version`, while `get_next_patch_version` and `get_next_minor_version` explicitly set it to 0.

CONTRIBUTING.md
alcaeus: Ah, good point :)

.github/workflows/release.yml
alcaeus: PHPC-2375

.github/workflows/package-release.yml
jmikola: Ah, I overlooked that this was being passed into a script. I concur that an absolute path is safest.

jmikola (approved):

PR #1539, opened 2024-04-30
[POC] PHPC-2326: Run static analysis

alcaeus:
PHPC-2326

This PR is a proof-of-concept to use [Semgrep](https://semgrep.dev) to run static analysis. An advantage over the Coverity setup is the ability to run scans on pull requests and prevent merging of a pull request as long as issues are available. Note that this proof of concept makes use of Semgrep OSS without an account, so there is no possibility to triage issues. I'll revisit our options (there is a free tier for up to 10 contributors, but it only talks about private repositories) before continuing with the effort, but wanted to show how this could work.


scripts/update-submodule-sources.php
```diff
@@ -32,6 +32,7 @@
 foreach ($vars as $var => $path) {
     $cutNth = 2 + substr_count($path, '/');
 
+    // nosemgrep
```
jmikola, line 35: What is the point of this? I assume `shell_exec` is the problem here.

Does Semgrep check M4 files? If not, you can exclude the entire `scripts/` directory.

Either way, the PHP files under `scripts/` shouldn't require scanning.

alcaeus: Excluded the `bin` and `scripts` directories as they only contain internal files

jmikola (approved):

PR #1540, opened 2024-05-01
PHPC-2373 and PHPC-2374: Upgrade libmongoc 1.27.0 and libmongocrypt 1.10.0

jmikola:
https://jira.mongodb.org/browse/PHPC-2373
https://jira.mongodb.org/browse/PHPC-2374


alcaeus (commented):
This PR is missing changes to the versions in .evergreen/config/templates/build/build-libmongoc.yml.

Windows builds are also throwing errors, so something is wrong on that platform.

jmikola:
I'm discussing Windows build errors in [#dbx-c-cxx](https://mongodb.slack.com/archives/GMY98JS86/p1714675846135229):

> Whilst upgrading to libmongoc 1.27.0, I'm hitting several [unexpected build errors on Windows](https://github.com/mongodb/mongo-php-driver/actions/runs/8914544355/job/24482344180?pr=1540#step:4:272) related to header includes. [mcd-nsinfo.c](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mcd-nsinfo.c) and [utlist.h](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/utlist.h) in src/libmongoc/src both use <> instead of "" for includes, which is inconsistent from other source files.
>
> I think the utlist.h problem can be resolved by declaring src/uthash as an include path (i.e. /I <path> parameter) after src/libmongoc/src. It's currently declared first, as I do for src/common (see: [config.w32](https://github.com/mongodb/mongo-php-driver/pull/1540/files#diff-cfe793e01b76cf09eab98bc88e0b4d03c4524ce3dabff25010a3a7eae4d4277eR105)).
But I don't think there's a workaround for mcd-nsinfo.c, since it's having trouble picking up both mongoc-prelude.h and mcd-nsinfo.h, which have unique names.

Will follow up once I make some headway on that.

config.w32
```diff
@@ -103,9 +103,11 @@ if (PHP_MONGODB != "no") {
     /I" + configure_module_dirname + "/src/MongoDB/Monitoring \
     /I" + configure_module_dirname + "/src/contrib \
     /I" + configure_module_dirname + "/src/libmongoc/src/common \
+    /I" + configure_module_dirname + "/src/libmongoc/src/uthash \
     /I" + configure_module_dirname + "/src/libmongoc/src/libbson/src \
     /I" + configure_module_dirname + "/src/libmongoc/src/libbson/src/jsonsl \
     /I" + configure_module_dirname + "/src/libmongoc/src/libmongoc/src \
+    /I" + configure_module_dirname + "/src/libmongoc/src/libmongoc/src/mongoc \
```
jmikola, line 110: This was the suggested workaround for libmongoc using `<>` instead of `""` for some new `#include` statements introduced in 1.27.0. CI suggests this is only necessary for Windows, but let me know if you think it'd be wise to duplicate this in `config.m4`. Alternatively, we can wait and see if any bug reports come in and address this in a patch release if needed (any failure would be quite obvious).

alcaeus (approved):

config.w32
alcaeus: I think it's fine to do this for Windows only at this time 👍

PR #1543, opened 2024-05-07
PHPC-2255: Add database name to CommandFailedEvent and CommandSucceededEvent

jmikola:
https://jira.mongodb.org/browse/PHPC-2255

Depends on https://github.com/mongodb/mongo-php-driver/pull/1540


src/MongoDB/Monitoring/CommandFailedEvent.stub.php
```diff
@@ -13,6 +13,8 @@ final private function __construct() {}
 
     final public function getCommandName(): string {}
 
+    final public function getDatabaseName(): string {}
```
alcaeus, line 16: Nothing to change here as all values are exposed through a getter, but we can consider changing all of these methods to readonly properties sometime down the line.

alcaeus (approved):
LGTM

PR #1544, opened 2024-05-07
PHPC-1584: Sign releases

alcaeus:
PHPC-1584

This PR adds release signing to the previously added automated release process. The following actions are signed using the [PHP team's GPG key](https://pgp.mongodb.com/php-driver.asc)
* The "Package X.Y.Z" commit
* The "Release X.Y.Z" tag
* The "Back to -dev" commit
* The PECL package generated as part of the release workflow
* The Windows DLLs generated as part of the release workflow

For the PECL package, the detached signature is attached as a release artifact. For Windows packages, the detached signature file is contained in the zip archive that also contains the DLL, as we only sign the DLL file. Instructions to verify the integrity of the releases have been added to the readme file.

This PR also fixes a minor issue where the message added to the GitHub release didn't properly show how to install the library using PECL.


.github/workflows/package-release.yml
```diff
@@ -38,27 +38,41 @@ jobs:
...
+        run: |
+          PACKAGE_VERSION=$(./bin/update-release-version.php get-version)
+          echo "PACKAGE_VERSION=${PACKAGE_VERSION}" >> "$GITHUB_ENV"
+          echo "PACKAGE_FILE=mongodb-${PACKAGE_VERSION}.tgz" >> "$GITHUB_ENV"
+
+      - name: "Create detached signature for PECL package"
+        uses: ./.github/actions/garasign/gpg-sign
+        with:
+          filename: ${{ env.PACKAGE_FILE }}
+          garasign_username: ${{ secrets.GRS_CONFIG_USER1_USERNAME }}
```
alcaeus, line 50: Leaving this comment as a reminder to add all necessary secrets to this and all other PHP team repositories. Will resolve when this is done.

.github/workflows/package-release.yml
```diff
@@ -38,27 +38,41 @@ jobs:
...
+      - name: "Upload release artifacts"
+        run: gh release upload ${{ github.ref_name }} ${{ env.PACKAGE_FILE }} ${{ env.PACKAGE_FILE }}.sig
         continue-on-error: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   build-windows:
     name: "Create Windows package"
-    runs-on: windows-2022
+    runs-on: windows-latest
```
alcaeus, line 75: Using windows-latest is necessary to be able to share the cached build artifacts.

.github/workflows/package-release.yml
```diff
@@ -91,20 +105,68 @@ jobs:
         env:
           BUILD_DIR: ${{ steps.build-driver.outputs.build-dir }}
 
+      - name: "Cache build artifacts for subsequent builds"
+        uses: actions/cache/save@v4
+        with:
+          key: ${{ github.sha }}-${{ matrix.php }}-${{ matrix.ts }}-${{ matrix.arch }}
+          enableCrossOsArchive: true
```
alcaeus, line 112: This is the notable difference to the same step in the windows-tests workflow. This took me longer to figure out than I care to admit.

.github/workflows/commit-and-tag.sh
```diff
@@ -0,0 +1,15 @@
...
+
+gpgloader
+
+# Create signed "Package x.y.z" commit
+echo "Create package commit"
+git commit -m "Package ${PACKAGE_VERSION}" -s --gpg-sign=${GPG_KEY_ID} phongo_version.h
+
+# Create signed "Release x.y.z" tag
+echo "Create release tag"
+git tag -F changelog -s --local-user=${GPG_KEY_ID} ${PACKAGE_VERSION}
```
jmikola, line 15: The `changelog` file is hard-coded here, which couples this script to `release.yml`. Should it be a parameter instead? Even having it defined as a script var (e.g. `TAG_MESSAGE=changelog`) would improve readability.

.github/workflows/package-release.yml
```diff
@@ -38,27 +38,41 @@ jobs:
...
+          garasign_password: ${{ secrets.GRS_CONFIG_USER1_PASSWORD }}
+          artifactory_username: ${{ secrets.ARTIFACTORY_USER }}
+          artifactory_password: ${{ secrets.ARTIFACTORY_PASSWORD }}
 
       - name: "Install release archive to verify correctness"
-        run: sudo pecl install mongodb-${{ env.PACKAGE_VERSION }}.tgz
+        run: sudo pecl install ${{ env.PACKAGE_FILE }}
 
-      - name: "Upload artifact"
+      - name: "Upload artifacts"
```
jmikola, line 58: I realize these steps predate the PR, but can you remind me why we use both [actions/upload-artifact](https://github.com/actions/upload-artifact) _and_ `gh release upload`? The former attaches the package and signature files to the _action_, while the latter attaches them to a GitHub release. Are both needed?

The only prior art that comes to mind was how we handle Windows DLLs by temporarily storing them on an action (from the build step) and having the release action transfer them to the GitHub release.

.github/workflows/package-release.yml
jmikola: Is this why `windows-latest` is required above? If so, I'd suggest adding a comment there to point out the relationship to `enableCrossOsArchive`.

Context: https://github.com/actions/cache/blob/main/tips-and-workarounds.md#cross-os-cache

.github/workflows/package-release.yml
```diff
@@ -91,20 +105,68 @@ jobs:
...
+      matrix:
+        # Note: keep this in sync with the Windows matrix in windows-tests.yml
+        php: [ "7.4", "8.0", "8.1", "8.2", "8.3" ]
+        arch: [ x64, x86 ]
+        ts: [ ts, nts ]
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
```
jmikola, line 133: Why are submodules required for signing and publish Windows DLLs that we pull from a cache?

.github/workflows/package-release.yml
```diff
@@ -91,20 +105,68 @@ jobs:
...
+
+      - name: Restore cached build artifacts
+        id: cache-build-artifacts
+        uses: actions/cache/restore@v4
+        with:
+          fail-on-cache-miss: true
+          key: ${{ github.sha }}-${{ matrix.php }}-${{ matrix.ts }}-${{ matrix.arch }}
+          enableCrossOsArchive: true
```
jmikola, line 141: Is `ubuntu-latest` required for this usage of `enableCrossOsArchive`, like `windows-latest` earlier in the file? If so, a comment around the `runs-on: ubuntu-latest` line would be prudent.

.github/workflows/package-release.yml
```diff
@@ -91,20 +105,68 @@ jobs:
...
+          garasign_username: ${{ secrets.GRS_CONFIG_USER1_USERNAME }}
+          garasign_password: ${{ secrets.GRS_CONFIG_USER1_PASSWORD }}
+          artifactory_username: ${{ secrets.ARTIFACTORY_USER }}
+          artifactory_password: ${{ secrets.ARTIFACTORY_PASSWORD }}
+
       - name: "Upload DLL and PDB files as build artifacts"
         uses: actions/upload-artifact@v4
         with:
-          name: php_mongodb-${{ github.ref_name }}-${{ matrix.php }}-${{ matrix.ts }}-${{ matrix.arch }}
+          name: php_mongodb-${{ github.sha }}-${{ matrix.php }}-${{ matrix.ts }}-${{ matrix.arch }}
```
jmikola, line 158: Why is `github.sha` used here, while `github.ref_name` remains in place below in the "Create and upload release artifact" step?

.github/workflows/release.yml
```diff
@@ -101,14 +97,34 @@ jobs:
...
-        run: |
-          ./bin/update-release-version.php to-next-patch-dev
-          git commit -m "Back to -dev" phongo_version.h
+      # This step creates the "Package x.y.z" commit that will be the base of
+      # our tag and creates the release tag. This is run inside the container in
+      # order to create signed git artifacts
+      - name: "Create package commit and release tag"
+        uses: ./.github/actions/garasign/git-sign
+        with:
+          command: "$(pwd)/.github/workflows/commit-and-tag.sh ${{ env.PACKAGE_VERSION }} ${{ vars.GPG_KEY_ID }}"
```
jmikola, line 106: If you end up parameterizing the `TAG_MESSAGE` filename, you can explicitly pass it here. I won't advocate for making "changelog" a variable in this file, as everything is self-contained.

----

I see the changelog is generated from [gh_release_create](https://cli.github.com/manual/gh_release_create). Does that become the entire body of the release draft?

This is a departure from the current git tag message we use (e.g. "Release X.Y.Z"). Considering we'll probably end up revising the body of the release notes, I have mixed feelings about preserving unedited content in the tag. Perhaps that's not a concern if this changelog is entirely derived from existing, immutable commit messages.

.github/workflows/release.yml
```diff
@@ -122,7 +138,9 @@ jobs:
 
       - name: "Prepare release message"
         run: |
-          echo "${{ format(env.default-release-message, env.PACKAGE_VERSION, inputs.jira-version-number) }}" > release-message
+          cat > release-message <<'EOL'
+          ${{ format(env.default-release-message, env.PACKAGE_VERSION, inputs.jira-version-number) }}
+          EOL
```
jmikola, line 143: Does this craft a new body for the GitHub release? If so, is there any reason to temporarily store the generated changelog in the earlier "Create draft release with generated changelog" step, or is that necessary because [gh_release_create](https://cli.github.com/manual/gh_release_create) is the only command that actually generates the changelog and can only output it to the release body?

README.md
```diff
@@ -36,7 +36,53 @@ This extension is intended to be used alongside the
...
+
+To verify the integrity of the downloaded package, run the following command:
+
+```shell
+gpg --verify mongodb-X.Y.Z.tgz.sig mongodb-X.Y.Z.tgz
+```
+
+> [!NOTE]
+> No verification is done when using `pecl` to install the package. To ensure
+> release integrity when using `pecl`, download the tarball manually from the
```
jmikola, line 68: Should you remind users that they must have imported the `php-driver.asc` public key, or is that redundant since it's discussed in the introductory paragraph?

I don't feel strongly about adding this, since it's a one-time step and the paragraph here is describing the steps that need to be done for each install.

README.md
```diff
@@ -36,7 +36,53 @@ This extension is intended to be used alongside the
...
+> GitHub release, verify the signature, then install the package from the
+> downloaded tarball using `pecl install mongodb-X.Y.Z.tgz`.
+
+### Windows
+
+Windows binaries distributed through GitHub releases contain a detached
+signature for the `php_mongodb.dll` file named `php_mongodb.dll.sig`. To verify
+the integrity of the DLL, run the following command:
```
jmikola, line 76: Just to confirm, there is no in-place signing of the DLL, correct? I remember we discussed that, but I don't recall what became of it.

.github/workflows/package-release.yml
alcaeus: Since this workflow is enabled for every tag, I wanted to upload the artifacts to the build as well in case we push a tag manually (e.g. for a pre-release) where there is no draft release yet. In that case, we'd still be able to download the artifacts manually and upload them to the release manually. I'm happy to remove this once we've covered every possible angle of the release process.

.github/workflows/package-release.yml
alcaeus: Good catch, this isn't necessary. Removed.

.github/workflows/package-release.yml
alcaeus: Ah, I missed this after fixing the cross-os cache issue where I changed this for testing.

I changed this back to use the `ref_name` and also ensured that the build artifacts contains all files that we add to the release artifact.

.github/workflows/package-release.yml
alcaeus: Added a comment in both jobs to explain this.

.github/actions/garasign/gpg-sign/action.yml
```diff
@@ -0,0 +1,47 @@
+name: "Sign artifact using garasign"
```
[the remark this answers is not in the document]
alcaeus: This action was moved to https://github.com/mongodb/drivers-github-tools. I'll fix this in that repository.

.github/actions/garasign/setup/action.yml
```diff
@@ -0,0 +1,36 @@
...
+  artifactory_registry:
+    description: "Artifactory registry to be used"
+    default: artifactory.corp.mongodb.com
+
+runs:
+  using: composite
+  steps:
+    - name: Create the envfile
+      run: |
+        cat << EOF > envfile
```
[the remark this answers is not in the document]
alcaeus: This was also moved. The env-file is used so we don't have to pass secrets in the command, as they may be exposed in process lists that way.

.github/workflows/commit-and-tag.sh
alcaeus: Ah, yes. I introduced a `TAG_MESSAGE_FILE` argument to `commit-and-tag.sh`

.github/workflows/release.yml
alcaeus: Ah, good point.

The release draft will contain the original release message as well as the full changelog generated from GitHub. The idea for the tag was to use the original title and place the full changelog in the tag body. I've now fixed this as I previously only used the changelog for the tag message.

.github/workflows/release.yml
alcaeus: Creating a draft release with the auto-generated changelog is the easiest way to get the changelog without using the GitHub API, so that's why I create the draft, extract the changelog and update the draft release in a later step.

README.md
alcaeus: I'll leave that as an exercise to the reader - I assume most people that will want to verify release integrity are already familiar with how the process works.

README.md
alcaeus: That's correct, we'd need a different key for that and DevProd confirmed that other teams also provide detached signatures for files where there is a more specific signing process available (e.g. code signing). This can be revisited at a later step.

alcaeus:
Updated the PR to address feedback. Note: successfully releasing will depend on https://github.com/mongodb-labs/drivers-github-tools/pull/3 being merged.

.github/workflows/package-release.yml
alcaeus: Secrets added.

.github/workflows/release.yml
```diff
@@ -101,14 +97,39 @@ jobs:
...
-      # TODO: Sign tag
-      - name: "Create release tag"
-        run: git tag -a -F changelog ${{ env.PACKAGE_VERSION }}
-
-      - name: "Update version information to next patch development release"
+      - name: "Prepare tag message"
         run: |
-          ./bin/update-release-version.php to-next-patch-dev
-          git commit -m "Back to -dev" phongo_version.h
+          echo "Release ${PACKAGE_VERSION}" > tag-message
```
jmikola, line 102: Can we add an additional line break after this to segregate the first line of the tag message from its body (as we do for commits)?

.github/workflows/package-release.yml
```diff
@@ -91,20 +107,73 @@ jobs:
...
           name: php_mongodb-${{ github.ref_name }}-${{ matrix.php }}-${{ matrix.ts }}-${{ matrix.arch }}
           path: |
             php_mongodb.dll
+            php_mongodb.dll.sig
             php_mongodb.pdb
+            CREDITS
+            CONTRIBUTING.md
+            LICENSE
+            README.md
+            THIRD_PARTY_NOTICES
```
jmikola, line 169: This list of filenames is duplicated in the `zip` command below. Is it worth trying to use a single variable for these, or would that be difficult given the `with.path` context here?

jmikola (approved):

.github/workflows/package-release.yml
alcaeus: In the examples for the upload-artifact action, I didn't see filenames being space separated. Since I wanted to get rid of this artifact upload once we're sure that we've covered all possible releases (including pre-releases), I'm happy to remove this upload, which solves the problem.

.github/workflows/release.yml
alcaeus: Added. In my previous attempts, this printed a literal `\n`, which I now realised can be fixed by using `-e` with echo.

Issue #1549, opened 2024-05-13
Bug & Question: How to compile mongodb statically into PHP?


crazywhalecc:
### Bug Report

I'm trying to statically compile mongodb with PHP, with following steps:

1. Download `php-src` and `mongo-php-driver`
2. Extract `mongo-php-driver` to `php-src/ext/mongodb`
3. `./buildconf --force`
4. `./configure --disable-shared --enable-static --disable-phpdbg --enable-cli --enable-mongodb --with-mongodb-system-libs=no --with-mongodb-client-side-encryption=no --with-mongodb-sasl=no --without-iconv`

And it shows error:

```
checking whether to enable MongoDB support... yes
checking for pkg-config... (cached) /opt/homebrew/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking PHP version... configure: error: php-config not found
```

From mongo source, I found it uses `php-config` directly to get PHP version, and inline build (static compilation) apparently have no any dev command. So I tried to patch `config.m4` like this:

```
// replaceFileStr is just `file_get_contents`, `str_replace`, `file_put_contents`
// getPHPVersion() assume it `8.2.18`
// getPHPVersionID assume it `80218`
FileSystem::replaceFileStr(SOURCE_PATH . '/php-src/ext/mongodb/config.m4', 'if test -z "$PHP_CONFIG"; then', 'if false; then');
FileSystem::replaceFileStr(SOURCE_PATH . '/php-src/ext/mongodb/config.m4', 'PHP_MONGODB_PHP_VERSION=`${PHP_CONFIG} --version`', 'PHP_MONGODB_PHP_VERSION=' . $this->builder->getPHPVersion());
FileSystem::replaceFileStr(SOURCE_PATH . '/php-src/ext/mongodb/config.m4', 'PHP_MONGODB_PHP_VERSION_ID=`${PHP_CONFIG} --vernum`', 'PHP_MONGODB_PHP_VERSION_ID=' . $this->builder->getPHPVersionID());
```

And in order to add and test other flags more easily, I directly use [static-php-cli](https://github.com/crazywhalecc/static-php-cli) to compile it. The main compilation methods and commands have not changed. In short, it uses commands below:

```
[06:02:15] [INFO] Extension [mongodb] patched before buildconf
[06:02:15] [INFO] Entering dir: /Users/jerry/project/git-project/static-php-cli/source/php-src
[...]

PR #1555, opened 2024-05-14
PHPC-2382: Allow static builds with php-src

jmikola:
https://jira.mongodb.org/browse/PHPC-2382

Prefer PHP_VERSION_ID for static builds, and fall back to invoking php-config otherwise.

mongocrypt.h previously used a path relative to the PHPC project root, which is not compatible with static builds. The current path is relative to src/, which is explicitly defined as an include path in config.m4.

Related: https://github.com/mongodb/mongo-php-driver/issues/1549

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -529,6 +529,7 @@ mongodb was configured with the following options:
 Build configuration:
   CC                                               : $CC
   CFLAGS                                           : $CFLAGS
+  CPPFLAGS                                         : $CPPFLAGS
   Extra flags                                      : $PHP_MONGODB_STD_CFLAGS
   Developer flags (slow)                           : $PHP_MONGODB_DEV_CFLAGS
   Code Coverage flags (extra slow                  : $PHP_MONGODB_COVERAGE_CFLAGS
--- a/config.m4
+++ b/config.m4
@@ -532,7 +532,7 @@ Build configuration:
   CPPFLAGS                                         : $CPPFLAGS
   Extra flags                                      : $PHP_MONGODB_STD_CFLAGS
   Developer flags (slow)                           : $PHP_MONGODB_DEV_CFLAGS
-  Code Coverage flags (extra slow                  : $PHP_MONGODB_COVERAGE_CFLAGS
+  Code Coverage flags (extra slow)                 : $PHP_MONGODB_COVERAGE_CFLAGS
   libmongoc                                        : $PHP_MONGODB_BSON_VERSION_STRING
   libbson                                          : $PHP_MONGODB_MONGOC_VERSION_STRING
   libmongocrypt                                    : $PHP_MONGODB_MONGOCRYPT_VERSION_STRING
--- a/config.m4
+++ b/config.m4
@@ -521,7 +521,6 @@ if test "$PHP_MONGODB" != "no"; then
   dnl This must come after PHP_NEW_EXTENSION, otherwise the srcdir won't be set
   PHP_ADD_MAKEFILE_FRAGMENT
 
-  dnl The libmongocrypt line intentionally uses the PHP_LIBBSON flag as that decides whether to build against bundled or system libraries.
   AC_CONFIG_COMMANDS_POST([
     if test "$enable_static" = "no"; then
       echo "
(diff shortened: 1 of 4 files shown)


jmikola:
I manually tested this with a build of php-8.3.7; however, I'd like to test it with an Alpine container before merging to ensure there's no regression (re: [PHPC-2195](https://jira.mongodb.org/browse/PHPC-2195)).

config.m4
```diff
@@ -46,54 +55,53 @@ if test "$PHP_MONGODB" != "no"; then
...
+    AX_CHECK_COMPILE_FLAG(-Wno-unused-parameter,            PHP_MONGODB_DEV_CFLAGS="$PHP_MONGODB_DEV_CFLAGS -Wno-unused-parameter"          ,, -Werror)
+    AX_CHECK_COMPILE_FLAG(-Wno-unused-but-set-variable,     PHP_MONGODB_DEV_CFLAGS="$PHP_MONGODB_DEV_CFLAGS -Wno-unused-but-set-variable"   ,, -Werror)
+    AX_CHECK_COMPILE_FLAG(-Wno-missing-field-initializers,  PHP_MONGODB_DEV_CFLAGS="$PHP_MONGODB_DEV_CFLAGS -Wno-missing-field-initializers",, -Werror)
 
     if test "$PHP_MONGODB_PHP_VERSION_ID" -ge "80108"; then
-      AX_CHECK_COMPILE_FLAG(-Wstrict-prototypes, _MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS -Wstrict-prototypes",, -Werror)
+      AX_CHECK_COMPILE_FLAG(-Wstrict-prototypes, PHP_MONGODB_DEV_CFLAGS="$PHP_MONGODB_DEV_CFLAGS -Wstrict-prototypes",, -Werror)
     fi
-
-    MAINTAINER_CFLAGS="$_MAINTAINER_CFLAGS"
```
jmikola, line 95: The reasons for these names are lost to time, but since I'm now initializing `$PHP_MONGODB_DEV_CFLAGS` and appending to it directly we can do without this.

config.m4
```diff
@@ -24,19 +24,28 @@ if test "$PHP_MONGODB" != "no"; then
...
 
   AC_MSG_RESULT($PHP_MONGODB_PHP_VERSION)
 
   if test "$PHP_MONGODB_PHP_VERSION_ID" -lt "70400"; then
     AC_MSG_ERROR([not supported. Need a PHP version >= 7.4.0 (found $PHP_MONGODB_PHP_VERSION)])
   fi
 
+  PHP_MONGODB_STD_CFLAGS=""
+  PHP_MONGODB_DEV_CFLAGS=""
+  PHP_MONGODB_COVERAGE_CFLAGS=""
```
jmikola, line 47: Using prefixes _and_ initializing these vars ensures we won't inherit state during a static build (admittedly unlikely).

config.m4
```diff
@@ -24,19 +24,28 @@ if test "$PHP_MONGODB" != "no"; then
...
+    fi
 
-  PHP_MONGODB_PHP_VERSION=`${PHP_CONFIG} --version`
-  PHP_MONGODB_PHP_VERSION_ID=`${PHP_CONFIG} --vernum`
+    PHP_MONGODB_PHP_VERSION=`${PHP_CONFIG} --version`
+    PHP_MONGODB_PHP_VERSION_ID=`${PHP_CONFIG} --vernum`
+  else
+    PHP_MONGODB_PHP_VERSION="${PHP_VERSION}"
+    PHP_MONGODB_PHP_VERSION_ID="${PHP_VERSION_ID}"
+  fi
```
jmikola, line 37: I successfully ran `./configure` on Alpine using the `php:8.2-alpine` [Docker image](https://hub.docker.com/_/php). This still works, although it did reveal a bug in the build config output where we print "Bundled ()" when libmongocrypt isn't configured. Opened [PHPC-2386](https://jira.mongodb.org/browse/PHPC-2386) to track that.

jmikola:
> Note that the Windows builds for PHP 7.4 are currently failing, which we'll have to investigate.

This is related to GitHub removing some older toolsets from the `windows-2022` image:

 * https://github.com/actions/runner-images/blob/win22/20240514.3/images/windows/Windows2022-Readme.md
 * https://github.com/actions/runner-images/issues/9701

I'll look into whether we can quickly work around that by using an older build image for 7.4.

PR #1559, opened 2024-05-22
Add test job to build and install PECL package

alcaeus:
This PR adds a new job to the tests workflow to build and install a PECL package for the current dev version. The idea behind this is to ensure that any changes to bundled dependencies are correctly reflected in the PECL package, as this has been a source of errors before.

Note that due to how `make package.xml` extracts the version number and the stability, a full build of the driver is necessary in order to build a PECL package. I've filed PHPC-2387 to optimise this and obtain the information from phongo_version.h as we do in the `update-release-version` script.


jmikola (approved):
Thanks for creating [PHPC-2387](https://jira.mongodb.org/browse/PHPC-2387). That seems like an entirely sensible improvement.

PR #1565, opened 2024-05-22
PHPC-2388: Install VC15 component for PHP 7.4

jmikola:
https://jira.mongodb.org/browse/PHPC-2388

### DIFF
--- a/.github/actions/windows/prepare-build/action.yml
+++ b/.github/actions/windows/prepare-build/action.yml
@@ -17,6 +17,38 @@ outputs:
 runs:
   using: composite
   steps:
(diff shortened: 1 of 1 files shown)


alcaeus (approved):
LGTM, Thanks!

I'd say this should target v1.19 in case we need to release a 1.19 patch version.

jmikola:
> I'd say this should target v1.19 in case we need to release a 1.19 patch version.

Yup. Thanks for catching that. I did intend to target v1.19 since we'll need this for releasing #1555 in 1.19.1. I missed doing the same for #1557 but that's thankfully wasn't essential.

PR #1545, opened 2024-05-07
PHPC-2378: Update libmongoc to 1.27.1

jmikola:
https://jira.mongodb.org/browse/PHPC-2378

PR #1575, opened 2024-06-05
PHPC-2394: Update libmongoc to 1.27.2

jmikola:
https://jira.mongodb.org/browse/PHPC-2394


src/LIBMONGOC_VERSION_CURRENT
```diff
@@ -1 +1 @@
-1.27.0
```
jmikola, line 1: This should have been 1.27.1, but was missed in https://github.com/mongodb/mongo-php-driver/pull/1545. That does mean our 1.19.0 and 1.19.1 releases incorrectly report a bundled version of libmongoc 1.27.0 (both releases actually include 1.27.1).

alcaeus: Ah, good catch. I'll take a look into how we can ensure this information is correct, as I'm planning to use this information for our SBOM file necessary for SSDLC.

alcaeus (approved):

PR #1586, opened 2024-06-14
Manually merge up version bump commits

alcaeus:
This PR adds logic to automatically merge up changes to the next release branch. This is necessary as the pull request for the merge-up action will always have a conflict due to the version already having changed in the next branch.

If the branch detection logic finds a newer branch to merge to (which it always should as we're releasing from version branches), it runs a `git merge --strategy=ours` in this next branch, then pushes those changes before pushing the release branch. By pushing the next branch before the release branch, the merge-up action discovers that there are no commits to be merged from the release branch into the next branch and will not create a merge-up pull request.

Note that if the release branch contains changes that have not been merged up yet, those changes will be marked as merged without applying them. I can add logic to detect this and prevent a release, but this case shouldn't come up very often as we're quite diligent about merging the merge-up pull requests so I've decided against it for the time being.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,6 +62,7 @@ jobs:
           app_id: ${{ vars.APP_ID }}
           private_key: ${{ secrets.APP_PRIVATE_KEY }}
           submodules: true
+          fetch-depth: 0
 
       - name: "Set up drivers-github-tools"
         uses: mongodb-labs/drivers-github-tools/setup@v2
@@ -109,6 +110,8 @@ jobs:
         with:
           version: ${{ inputs.version }}
           tag_message_template: 'Release ${VERSION}'
+          # Don't push tag, we'll do that after merging up
+          push_tag: false
 
       - name: "Bump to next development release and commit"
         uses: mongodb-labs/drivers-github-tools/bump-version@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -109,20 +110,43 @@ jobs:
...
+          git push origin ${NEXT_BRANCH}
+          git checkout ${RELEASE_BRANCH}
+        env:
+          NEXT_BRANCH: ${{ steps.get-next-branch.outputs.branchName }}
+          RELEASE_BRANCH: ${{ github.ref_name }}
+
+      - name: "Push tag and release branch"
+        run: |
+          git push origin ${RELEASE_BRANCH}
+          git push origin ${{ inputs.version }}
```
jmikola, line 147: Alternatively, would `git push origin tag ${{ inputs.version }}` be less ambiguous? I picked that up while reading the [git-push docs](https://git-scm.com/docs/git-push) and https://stackoverflow.com/q/5195859

I also noted the advice in the SO thread to avoid using `--tags`, which we historically did (see: [RELEASING.md](https://github.com/mongodb/mongo-php-driver/blob/master/RELEASING.md#tag-the-release)). It appears that [tag-version/action.yml](https://github.com/mongodb-labs/drivers-github-tools/blob/7edc545fcf00ff1da4910079b7488c81d7c583b0/tag-version/action.yml#L54) is still doing that, so perhaps you'd like to revisit that at some point so it only explicitly pushes one tag.

jmikola (approved):
OK with this as-is. You can decide to implement any feedback.

.github/workflows/release.yml
```diff
@@ -109,20 +110,43 @@ jobs:
...
         with:
           version: ${{ inputs.version }}
           version_bump_script: "./bin/update-release-version.php to-next-patch-dev"
           commit_template: 'Back to -dev'
+          # Don't push commit, we still need to merge up
+          push_commit: false
+
+      - name: "Determine branch to merge up to"
+        id: get-next-branch
+        uses: alcaeus/automatic-merge-up-action/get-next-branch@main
```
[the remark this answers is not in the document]
alcaeus: I'll keep the action separate for the time being. Once I've confirmed this works I'll tag the current state at v1 and update all of our repositories accordingly.

As for the 20 KLOC, all of the dependencies are compiled into a single file so that running the action from a workflow doesn't require installing them manually. The main dependencies are the actions toolkit that makes dealing with inputs, outputs, and the GitHub API easier.

.github/workflows/release.yml
alcaeus: TIL `tag <name>` is a thing. Updated to ensure what we're pushing is a tag ref.

PR #1598, opened 2024-06-27
PHPC-2376: Test against MongoDB 8.0

alcaeus:
PHPC-2376

This PR also updates libmongoc to its latest version (as the libmongoc-next-stable builds fail due to the updated zlib dependency) and fixes version detection when building with a newer libmongoc version (a problem introduced by python not being in the PATH on Debian 12).


alcaeus:
> Why do you update libmongoc in the same PR?

I discovered it was broken when working on these changes and decided to fix it here instead. I can extract it to a separate pull request if you'd like, especially since the tests won't succeed until https://github.com/mongodb-labs/drivers-evergreen-tools/pull/452 is merged.

PR #1604, opened 2024-07-05
PHPC-1489: Deprecate integer readPreference constants

alcaeus:
PHPC-1489

This deprecates the `RP_*` constants in `MongoDB\Driver\ReadPreference` constants, along with passing an int in the constructor and using `getMode`.

Note that PHP 8.3 and newer automagically triggers warnings when using deprecated warnings. Calling deprecated methods results in a warning starting in PHP 7.4 (albeit with a wrong message in PHP 7.4).

### DIFF
--- a/src/MongoDB/ReadPreference.stub.php
+++ b/src/MongoDB/ReadPreference.stub.php
@@ -12,30 +12,35 @@ final class ReadPreference implements \MongoDB\BSON\Serializable, \Serializable
     /**
      * @var int
      * @cvalue MONGOC_READ_PRIMARY
+     * @deprecated
      */
     public const RP_PRIMARY = UNKNOWN;
 
     /**
      * @var int
      * @cvalue MONGOC_READ_PRIMARY_PREFERRED
+     * @deprecated
      */
     public const RP_PRIMARY_PREFERRED = UNKNOWN;
 
     /**
      * @var int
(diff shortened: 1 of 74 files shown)


src/MongoDB/ReadPreference.stub.php
```diff
@@ -12,30 +12,35 @@ final class ReadPreference implements \MongoDB\BSON\Serializable, \Serializable
     /**
      * @var int
      * @cvalue MONGOC_READ_PRIMARY
+     * @deprecated
```
GromNaN, line 15: This will need to be sync with 3rd party stubs.

GromNaN (commented):
Do you keep any test that use the legacy constants to ensure they still work even if they are deprecated?

alcaeus:
> Do you keep any test that use the legacy constants to ensure they still work even if they are deprecated?

Added now.

src/MongoDB/ReadPreference.stub.php
alcaeus: Yes, I'll take care of those alongside the documentation

alcaeus:
Note: I'll have to revisit the failing Windows Build tasks for PHP 7.4 and 8.0 separately - deferring the merge until those have been taken care of.

GromNaN (approved):

PR #1607, opened 2024-07-10
PHPC-2347: Deprecate BSON functions

alcaeus:
PHPC-2347

This deprecates all functions in the `MongoDB\BSON` namespace in favour of the `MongoDB\BSON\Document` class. I've changed test usages of these functions to the Document class, except where we specifically want to test the functions. Note that tests often use functions like `toPHP` without a namespace - these were proxy functions and were changed to use the Document class internally, even though this is not very efficient. I plan to refactor those utility functions in a separate PR. In order to avoid accounting for deprecations in BSON corpus tests, I've removed the corresponding calls in those. Let me know if that's a concern.


tests/bson/bson-binary-001.phpt
```diff
@@ -61,48 +61,48 @@ bool(true)
 bool(true)
 bool(true)
 bool(true)
-Test#0 { "binary" : { "$binary" : "cmFuZG9tQmluYXJ5RGF0YQ==", "$type" : "00" } }
-string(73) "{ "binary" : { "$binary" : "cmFuZG9tQmluYXJ5RGF0YQ==", "$type" : "00" } }"
-string(73) "{ "binary" : { "$binary" : "cmFuZG9tQmluYXJ5RGF0YQ==", "$type" : "00" } }"
+Test#0 { "binary" : { "$binary" : { "base64" : "cmFuZG9tQmluYXJ5RGF0YQ==", "subType" : "00" } } }
```
alcaeus, line 64: The tests all used to print `toJSON`, which we did not implement in the Document class. Using relaxed extended JSON is responsible for the additional properties.

src/functions.stub.php
```diff
@@ -3,26 +3,38 @@
 /** @generate-function-entries */
 
 namespace MongoDB\BSON {
+    /** @deprecated */
```
GromNaN, line 6: You can add a hint for fixing the deprecation:
```suggestion
    /** @deprecated use MongoDB\BSON\Document::fromPHP() instead  */
```

alcaeus: Good point, added these for all functions except `toJSON`, which doesn't have a replacement.

GromNaN (approved):
> In order to avoid accounting for deprecations in BSON corpus tests, I've removed the corresponding calls in those.

LGTM, since they are aliases, no need for extensive tests.

PR #1608, opened 2024-07-10
PHPC-2349, PHPC-2411: Deprecate unused exception classes

alcaeus:
PHPC-2349, PHPC-2411

### DIFF
--- a/src/MongoDB/Exception/SSLConnectionException.stub.php
+++ b/src/MongoDB/Exception/SSLConnectionException.stub.php
@@ -7,6 +7,7 @@
 
 namespace MongoDB\Driver\Exception;
 
+/** @deprecated use ConnectionException instead */
 final class SSLConnectionException extends ConnectionException
 {
 }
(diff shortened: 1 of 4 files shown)


GromNaN (commented):
`BulkWriteException` is extending `WriteException`, does it trigger a warning when a `BulkWriteException` is thrown?
Can you make `BulkWriteException` extend `RuntimeException` instead?

alcaeus:
> Can you make `BulkWriteException` extend `RuntimeException` instead?

That's not possible, as anybody currently catching a `WriteException` would no longer be catching a `BulkWriteException`. 

As for the deprecation notices, there currently aren't any deprecation notices thrown in tests, despite us using `BulkWriteException` in a number of places. I can do some more testing with this though, especially since there's another exception class to be deprecated (`SSLConnectionException`) that I can put into this PR as well.

alcaeus:
> I can do some more testing with this though

After checking, PHP doesn't seem to emit any deprecation warnings for deprecated classes. People will have to rely on their IDE or on third-party tools to let them know about the deprecation.

GromNaN (approved):

PR #1612, opened 2024-07-17
Reduce amount of evergreen config includes

alcaeus:
As per a recent change, we can no longer include more than 35 files in an evergreen config. To work around this, the config generation script now outputs a single file for each template, containing all of the versions instead of separate files for each version.


.evergreen/config/templates/build/build-libmongoc.yml
```diff
@@ -4,7 +4,6 @@
 # These tasks are used in the build task group "build-php-libmongoc" defined in .evergreen/config/build-task-groups.yml,
 # where the "max_hosts" setting must be set to the number of enabled tasks.
 # Last but not least, make sure the corresponding test variants are enabled in .evergreen/config/test-variants.yml
-tasks:
```
alcaeus, line 7: I decided to remove the top-level key in all templates to avoid having to use a yaml library to combine individual config files. It's not optimal, but it works for the time being.

GromNaN (approved):
LGTM. I hear we're going to save on GitHub API calls.

.evergreen/config/templates/build/build-libmongoc.yml
alcaeus: I think this works fine for the time being - the less I mess with the files the better it is.

PR #1615, opened 2024-07-24
Use drivers-evergreen-tools for Windows testing

alcaeus:
This PR makes several changes, all related to how we start MongoDB on Windows. Since the previous method of starting MongoDB would often time out, the Windows workflow now uses drivers-evergreen-tools to start clusters. This should be more reliable and as a side effect uses the same configs we use for our Linux testing. As part of this change, I have also re-organised the windows workflows so that tests for a given variant (i.e. the combination of PHP version, architecture, and thread safety) start immediately after the corresponding build has completed, not only after all builds have completed. This yields faster test results.

During testing, I discovered an issue in the tests for `ServerHeartbeatFailedEvent`, where an expected integer return was actually returned as an `Int64` instance on 32-bit platforms. After some investigation I found that this was due to the type being wrong - libmongoc returns an `int64_t`, while we store a `uint64_t`. This caused issues when comparing the value to `INT32_MAX` and `INT32_MIN` on 32-bit platforms, erroneously returning an `Int64` instance which should only happen when the value exceeds the bounds of a 32-bit integer. During that investigation I also found that the same issue affects `ServerHeartbeatSucceededEvent`, but it didn't get exposed because the duration was not added to the `var_dump` output. All these issues have been fixed.

### DIFF
--- a/tests/apm/serverHeartbeatSucceededEvent-001.phpt
+++ b/tests/apm/serverHeartbeatSucceededEvent-001.phpt
@@ -68,6 +68,8 @@ object(MongoDB\Driver\Monitoring\ServerHeartbeatSucceededEvent)#%d (%d) {
   int(%d)
   ["awaited"]=>
   bool(%s)
+  ["durationMicros"]=>
+  int(%d)
   ["reply"]=>
   object(stdClass)#%d (%d) {%A
   }
--- a/.github/workflows/package-release.yml
+++ b/.github/workflows/package-release.yml
@@ -97,7 +97,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Note: keep this in sync with the Windows matrix in windows-tests.yml
+        # Note: keep this in sync with the Windows matrix in tests.yml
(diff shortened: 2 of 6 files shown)


tests/apm/serverHeartbeatSucceededEvent-001.phpt
```diff
@@ -68,6 +68,8 @@ object(MongoDB\Driver\Monitoring\ServerHeartbeatSucceededEvent)#%d (%d) {
   int(%d)
   ["awaited"]=>
   bool(%s)
+  ["durationMicros"]=>
+  int(%d)
```
GromNaN, line 72: I understand that this value was already exposed with an [`getDurationMicros`](https://www.php.net/manual/en/class.mongodb-driver-monitoring-serverheartbeatsucceededevent.php) method, but now it becomes accessible with var_dump.

GromNaN (approved):
Smart.

tests/apm/serverHeartbeatSucceededEvent-001.phpt
alcaeus: That's correct :)

PR #1616, opened 2024-07-25
PHPC-2412: Deprecate CursorId class

alcaeus:
PHPC-2412

The CursorId class was only introduced to hold the 64-bit placeholder. Since The `MongoDB\BSON\Int64` class is now more viable, it makes sense to always return a 64-bit value instead of a wrapper class. Note that we could directly return a long on 64-bit systems, but I opted against this in favour of a consistent return type regardless of platform.

This PR will be followed up by a PR in PHPLIB that updates classes implementing `CursorInterface` or wrapping a cursor. However, as the `getId` method is not used in the PHPLIB test suite, I don't expect this PR to break any tests there with the new deprecation notice.

### DIFF
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -253,8 +253,8 @@ PHP_MINIT_FUNCTION(mongodb) /* {{{ */
 	php_phongo_bulkwrite_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_clientencryption_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_command_init_ce(INIT_FUNC_ARGS_PASSTHRU);
-	php_phongo_cursorid_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_cursor_init_ce(INIT_FUNC_ARGS_PASSTHRU);
+	php_phongo_cursorid_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_manager_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_query_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_readconcern_init_ce(INIT_FUNC_ARGS_PASSTHRU);
--- a/php_phongo.c
+++ b/php_phongo.c
@@ -253,8 +253,8 @@ PHP_MINIT_FUNCTION(mongodb) /* {{{ */
 	php_phongo_bulkwrite_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_clientencryption_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_command_init_ce(INIT_FUNC_ARGS_PASSTHRU);
-	php_phongo_cursor_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_cursorid_init_ce(INIT_FUNC_ARGS_PASSTHRU);
+	php_phongo_cursor_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_manager_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_query_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_readconcern_init_ce(INIT_FUNC_ARGS_PASSTHRU);
--- a/src/MongoDB/CursorId.stub.php
+++ b/src/MongoDB/CursorId.stub.php
@@ -7,7 +7,6 @@
 
 namespace MongoDB\Driver;
 
-/** @deprecated deprecated without replacement */
 final class CursorId implements \Serializable
 {
     final private function __construct() {}
--- a/src/MongoDB/CursorId.stub.php
+++ b/src/MongoDB/CursorId.stub.php
@@ -7,6 +7,7 @@
 
 namespace MongoDB\Driver;
 
+/** @deprecated deprecated without replacement */
 final class CursorId implements \Serializable
 {
     final private function __construct() {}
--- a/src/MongoDB/CursorInterface.stub.php
+++ b/src/MongoDB/CursorInterface.stub.php
@@ -9,8 +9,13 @@ namespace MongoDB\Driver;
 
 interface CursorInterface extends \Traversable
 {
+#if PHP_VERSION_ID >= 80000
     /** @tentative-return-type */
-    public function getId(): CursorId;
+    public function getId(): CursorId|\MongoDB\BSON\Int64;
+#else
+    /** @return CursorId|\MongoDB\BSON\Int64 */
+    public function getId();
+#endif
 
     /** @tentative-return-type */
     public function getServer(): Server;
--- a/tests/cursor/cursor-session-001.phpt
+++ b/tests/cursor/cursor-session-001.phpt
@@ -26,7 +26,7 @@ $iterator = new IteratorIterator($cursor);
 $iterator->rewind();
 $iterator->next();
 
-printf("Cursor ID is zero: %s\n", (int) $cursor->getId(true) === 0 ? 'yes' : 'no');
+printf("Cursor ID is zero: %s\n", $cursor->getId(true) == 0 ? 'yes' : 'no');
 var_dump($cursor);
 
 $iterator->next();
@@ -35,7 +35,7 @@ $iterator->next();
  * is exhausted. While this is primarily done to ensure implicit sessions for
  * command cursors are returned to the pool ASAP, it also applies to explicit
  * sessions. */
-printf("\nCursor ID is zero: %s\n", (int) $cursor->getId(true) === 0 ? 'yes' : 'no');
+printf("\nCursor ID is zero: %s\n", $cursor->getId(true) == 0 ? 'yes' : 'no');
 var_dump($cursor);
 
 ?>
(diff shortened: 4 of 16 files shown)


GromNaN (approved):

php_phongo.c
```diff
@@ -253,8 +253,8 @@ PHP_MINIT_FUNCTION(mongodb) /* {{{ */
 	php_phongo_bulkwrite_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_clientencryption_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_command_init_ce(INIT_FUNC_ARGS_PASSTHRU);
-	php_phongo_cursor_init_ce(INIT_FUNC_ARGS_PASSTHRU);
 	php_phongo_cursorid_init_ce(INIT_FUNC_ARGS_PASSTHRU);
+	php_phongo_cursor_init_ce(INIT_FUNC_ARGS_PASSTHRU);
```
GromNaN, line 257: Why is this new position better?

tests/cursor/cursor-session-001.phpt
```diff
@@ -26,7 +26,7 @@ $iterator = new IteratorIterator($cursor);
 $iterator->rewind();
 $iterator->next();
 
-printf("Cursor ID is zero: %s\n", (string) $cursor->getId() === '0' ? 'yes' : 'no');
+printf("Cursor ID is zero: %s\n", $cursor->getId(true) == 0 ? 'yes' : 'no');
```
GromNaN, line 29: Note that if `$cursor->getId(true)` return `null`, this condition `null == 0` would be `true`. Should you test the type of the returned value?

alcaeus: Good point. I'll double check this, as it may have been related to the casting issues fixed in #1617.

php_phongo.c
alcaeus: Without moving the class, PHP failed to initialise the extension as the cursor class was needed elsewhere.

tests/cursorid/cursorid-001.phpt
```diff
@@ -26,6 +26,7 @@ hex_dump(fromPHP(['cid' => $cursorId]));
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
+Deprecated: MongoDB\Driver\Cursor::getId(): The method "MongoDB\Driver\Cursor::getId" will no longer return a "MongoDB\Driver\CursorId" instance in the future. Pass "true" as argument to change to the new behavior and receive a "MongoDB\BSON\Int64" instance instead. in %s
```
jmikola, line 29: Why not do `$cursor->getId(true)` above? This test is only asserting the BSON encoding, so an Int64 or CursorId makes no difference. Likewise for `cursorid-002.phpt` and `fucntional/cursorid-001.phpt`.

tests/functional/cursorid-001.phpt
```diff
@@ -32,6 +32,7 @@ var_dump($s1 > 0);
 ===DONE===
 <?php exit(0); ?>
 --EXPECTF--
+Deprecated: MongoDB\Driver\Cursor::getId(): The method "MongoDB\Driver\Cursor::getId" will no longer return a "MongoDB\Driver\CursorId" instance in the future. Pass "true" as argument to change to the new behavior and receive a "MongoDB\BSON\Int64" instance instead. in %s
```
jmikola, line 35: If you change this test to use `Cursor::getId(true)`, you can definitely remove the string cast and extra `var_dump()`. That was likely only needed because CursorId had no comparison handler.

Dumping an Int64 and checking that it's greater than zero should be sufficient.

tests/cursorid/cursorid-002.phpt
```diff
@@ -37,7 +37,8 @@ printf("Killed expected cursor: %s\n", (string) $cursorId === (string) $result->
 ?>
 ===DONE===
 <?php exit(0); ?>
---EXPECT--
+--EXPECTF--
+Deprecated: MongoDB\Driver\Cursor::getId(): The method "MongoDB\Driver\Cursor::getId" will no longer return a "MongoDB\Driver\CursorId" instance in the future. Pass "true" as argument to change to the new behavior and receive a "MongoDB\BSON\Int64" instance instead. in %s
 Killed 1 cursor(s)
 Killed expected cursor: yes
```
jmikola, line 43: I imagine `$result->cursorsKilled[0]` will always come back as a PHP integer unless we're on a 32-bit platform. The existing comparison using string casts and `===` can probably stay, since it'd still be necessary if `$cursorId` was an Int64 instead of a CursorId object.

tests/cursorid/cursorid-001.phpt
alcaeus: This test is asserting the BSON encoding of a `CursorId` instance, and while it's technically the same as an Int64, changing this would drop test coverage of CursorId encoding.

tests/functional/cursorid-001.phpt
alcaeus: Good point, updated accordingly

tests/cursor/cursor-session-001.phpt
alcaeus: TL;DR: this is not an issue, because `getId()` would never return `null`, and even if it did the comparison logic we rely on for `Int64` instances handles this correctly.

The dirty details: before the change to `getId(true)`, we'd always get a `CursorId` instance, which does not implement cast or comparison handlers, so the only sensible way to assert on it was by casting it to a string and doing a typesafe comparison.

Now that we return an `Int64` instance, we can rely on PHP's internal logic for checking equality (not identity) in objects. Note that when using `===`, there's an [early exit in `zend_is_identical`](https://github.com/php/php-src/blob/fad899e5662d0a929d8462dd8239b0489dd9b53f/Zend/zend_operators.c#L2389-L2391) if the types do not match. So `$int64 === 0` will always evaluate to `false` based on the different types. `==` on the other hand calls `zend_is_equal`, which in turn calls `zend_compare`, which relies on an object's `compare` handler, and thus defers to the logic we implemented for comparing `Int64` instances with other values (and importantly does not include a cast to avoid issues on 32-bit platforms).

I did not add a specific test for the case you mentioned, as `zend_compare` does not call into the `compare` handler when comparing an object to `null`, instead [handling it specifically to ensure correct behaviour](https://github.com/php/php-src/blob/fad899e5662d0a929d8462dd8239b0489dd9b53f/Zend/zend_operators.c#L2317-L2321).

src/MongoDB/CursorId.stub.php
```diff
@@ -7,6 +7,7 @@
 
 namespace MongoDB\Driver;
 
+/** @deprecated deprecated without replacement */
```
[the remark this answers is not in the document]
alcaeus: Good point. Looking into `zend_compile.h`, `ZEND_ACC_DEPRECATED` shares a value with `ZEND_ACC_PROTECTED_SET` (used as property flag only for asymmetric visibility), and `ZEND_ACC_USE_GUARDS` (set on the class when any `__get`, `__set`, `__isset`, or `__unset` are implemented by a class). Since `ZEND_ACC_DEPRECATED` is defined as a function flag and its value conflicts with a class flag with an entirely different meaning, I've removed the `@deprecated` comment. This seemed easier than teaching `gen_stub.php` that `@deprecated` on a class should not result in `ZEND_ACC_DEPRECATED` being added to `ce_flags`. I'll still follow up on whether gen_stub.php should be changed to account for this.

src/MongoDB/CursorInterface.stub.php
```diff
@@ -9,8 +9,13 @@
 
 interface CursorInterface extends \Traversable
 {
+#if PHP_VERSION_ID >= 80000
     /** @tentative-return-type */
-    public function getId(): CursorId;
+    public function getId(): CursorId|\MongoDB\BSON\Int64;
```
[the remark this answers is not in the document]
alcaeus: Unfortunately, `gen_stub.php` does not support this (at least in version 8.2 which I currently use to regenerate arginfo files):
```
❯❯❯ ./build/gen_stub.php
In ./src/MongoDB/Cursor.stub.php:
Unexpected node Stmt_Use
```

tests/cursorid/cursorid-001.phpt
jmikola: Oh, my mistake. I completely overlooked "cursorid" in those test names being significant.

php_phongo.c
jmikola: I meant to follow up on this. When testing this branch locally, I experienced no build error with the original order. I also don't see anything in `Cursor.c` that would depend on initialization of CursorId. There are references to `php_phongo_cursorid_ce`, but those are satisfied by the declarations in `phongo_classes.h`. The actual class entry doesn't need to be initialized until `getId()` is called.

Do you want to double-check this?

tests/cursorid/cursorid-002.phpt
jmikola: @alcaeus: I think you may want to revert this if the purpose of this test was to assert BSON serialization of _CursorId_ instances. Related to my comment in https://github.com/mongodb/mongo-php-driver/pull/1616#discussion_r1743951768.

php_phongo.c
alcaeus: Confirmed that the problem is indeed gone. I've reverted this change.

tests/cursorid/cursorid-002.phpt
alcaeus: Agree. I've reverted the change.

jmikola (approved):

PR #1617, opened 2024-07-26
PHPC-2420: Compare Int64 instances without casting

alcaeus:
PHPC-2420

Previously, we deferred to `zend_std_compare_objects` when comparing a single `Int64` instance with a long or double. By default, this will attempt to cast values to long or double, then comparing it. On 32-bit platforms, this cast truncates any value outside of the 32-bit range and yields a warning about this truncation. This means that we were not able to properly compare such values, even when we technically would easily be able to (as only the value of `ZEND_LONG` is limited to 32 bits, but `int64_t` is still available).

This PR changes this logic to specifically handle comparisons between to `Int64` instances as well as a single `Int64` instance with a long or double. Only if the instance is being compared with anything else will we attempt to cast the value. This removes warnings and fixes comparisons of 64-bit values on 32-bit platforms.

### DIFF
--- a/src/BSON/Int64.c
+++ b/src/BSON/Int64.c
@@ -345,6 +345,15 @@ static int php_phongo_int64_compare_objects(zval* o1, zval* o2)
 	return 0;
 }
 
+#if PHP_VERSION_ID < 80000
+static int php_phongo_int64_compare_zvals(zval* result, zval* op1, zval* op2)
+{
+	ZVAL_LONG(result, php_phongo_int64_compare_objects(op1, op2));
+
+	return SUCCESS;
+}
+#endif
+
 static zend_result php_phongo_int64_cast_object(phongo_compat_object_handler_type* readobj, zval* retval, int type)
 {
 	php_phongo_int64_t* intern;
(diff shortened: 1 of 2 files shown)


src/BSON/Int64.c
```diff
@@ -265,11 +265,23 @@ static zend_object* php_phongo_int64_clone_object(phongo_compat_object_handler_t
...
-	php_phongo_int64_t *intern1, *intern2;
+	if (Z_TYPE_P(object) != IS_OBJECT) {
+		return false;
+	}
 
-	ZEND_COMPARE_OBJECTS_FALLBACK(o1, o2);
+	return Z_OBJ_P(object)->ce == php_phongo_int64_ce;
+}
+
+static bool php_phongo_int64_is_long_or_double(zval* value)
```
GromNaN, line 277: How is this function called?

GromNaN (approved):
LGTM as tests shows this is correct.

src/BSON/Int64.c
alcaeus: This function is used in an `assert` call in `php_phongo_int64_compare_with_long_or_float` a few lines below.

alcaeus:
After some more investigation, I remembered that PHP 7.4 still has the old comparison logic, where the `compare_objects` handler was only used when comparing two objects, while the `compare` handler handled comparing an object to any other zval ([see code](https://github.com/php/php-src/blob/004cb827501d1ddaf98daacb185a53e0816f78a7/Zend/zend_operators.c#L2110-L2134)). In PHP 8, comparisons were simplified, with the original `compare` handler being removed and `compare_objects` renamed to `compare`.

GromNaN (approved):

PR #1624, opened 2024-09-02
PHPC-2420: Initial composer.json for pie support

alcaeus:
PHPC-2420

Note: pie support is experimental. No support will be given for installation issues with pie. The integration with pie is mainly done to be able test pie itself.

This is an initial version of a composer.json file to be used with [pie](https://github.com/php/pie). Once this PR has been merged to master, we can add the extension to packagist; after tagging a release we can install the extension using pie.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,72 @@
+{
+  "name": "mongodb/mongodb-extension",
+  "type": "php-ext",
+  "keywords": ["database", "driver", "mongodb", "persistence"],
+  "homepage": "https://jira.mongodb.org/browse/PHPC",
+  "license": "Apache-2.0",
+  "authors": [
+    { "name": "Andreas Braun", "email": "andreas.braun@mongodb.com" },
+    { "name": "Jeremy Mikola", "email": "jmikola@gmail.com" },
+    { "name": "Jérôme Tamarelle", "email": "jerome.tamarelle@mongodb.com" }
+  ],
+  "require": {
+    "php": ">=7.4,<9",
+    "ext-date": "*",
+    "ext-json": "*"
+  },
+  "php-ext": {
+    "extension-name": "ext-mongodb",
+    "configure-options": [
+      {
+        "name": "enable-mongodb-developer-flags",
+        "description": "Enable developer flags",
+        "needs-value": true
+      },
+      {
+        "name": "enable-mongodb-coverage",
+        "description": "Enable code coverage",
+        "needs-value": true
+      },
+      {
+        "name": "with-mongodb-system-libs",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -0,0 +1,72 @@
+{
```
jmikola, line 1: I spent a while getting [pie](https://github.com/php/pie) running locally just to find out that it doesn't support a `validate` command to check this JSON file. Well, at least I'll have it ready for future use :shrug:

jmikola (approved):
No objections to merging.

composer.json
```diff
@@ -0,0 +1,72 @@
...
+      },
+      {
+        "name": "with-mongodb-system-libs",
+        "description": "Use system libraries for libbson, libmongoc, and libmongocrypt",
+        "needs-value": true
+      },
+      {
+        "name": "with-mongodb-client-side-encryption",
+        "description": "Enable client-side encryption (auto/yes/no)",
+        "needs-value": true
```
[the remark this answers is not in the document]
alcaeus: @asgrim confirmed that configure options are only passed to `configure` if they are explicitly given during the installation process. Thus, the default is implicitly applied by our configuration scripts. As I understand it, this is different from pecl which always passes the configured flags and relies on the defined default value when the user does not input their own value.

composer.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "mongodb/mongodb-extension",
```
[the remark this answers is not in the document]
alcaeus: Indeed, extensions and packages all share the same namespace, so we can't choose `mongodb/mongodb` as a name.

composer.json
alcaeus: I'll use the opportunity to summon @asgrim, as they'd be in a better place to give you an idea of if and when to expect a `pie validate` command.

composer.json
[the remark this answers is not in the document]
jmikola: Got it. I was worried `needs-value: true` would require _all_ of these options to be specified at the time of package installation. Rather, it just requires a value iff the user specifies the option. That sounds like an improvement over PECL.

composer.json
asgrim: You can actually just use `composer validate`; the schema for the `php-ext` bit is actually built in there.

composer.json
```diff
@@ -0,0 +1,72 @@
...
+        "needs-value": true
+      },
+      {
+        "name": "enable-mongodb-crypto-system-profile",
+        "description": "Use system crypto profile (OpenSSL only) (yes/no)",
+        "needs-value": true
+      }
+    ]
+  }
+}
```
GromNaN, line 72: Missing newline.

PR #1632, opened 2024-09-05
Update GitHub Actions job names

alcaeus:
This PR removes the matrix strategy, as using it will include the PHP version in the name of the job, making it more difficult to have required check across branches. By setting it as an env var, we can avoid duplication as well have a consistent name for the checks after changing the PHP requirement in #1631.

### DIFF
--- a/.github/workflows/arginfo-files.yml
+++ b/.github/workflows/arginfo-files.yml
@@ -12,15 +12,13 @@ on:
       - "master"
       - "feature/*"
 
+env:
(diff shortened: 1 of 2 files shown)


GromNaN (approved):
LGTM. The [other ways of retrieving the PHP version](https://github.com/shivammathur/setup-php/blob/15b611248c3d140461e97b2eb10de461ca9ec5a3/src/utils.ts#L431-L470) are not suitable for this project.

PR #1636, opened 2024-09-09
Validate composer.json as part of CI workflows

alcaeus:
Based on https://github.com/mongodb/mongo-php-driver/pull/1624#discussion_r1748004943, this PR introduces a GitHub Actions Workflow to validate composer.json and ensure it will work. Based on that validation, I've added the missing `description` field, making it as minimalist as that of PHPLIB.


.github/workflows/composer-validate.yml
```diff
@@ -0,0 +1,35 @@
...
+      - "master"
+      - "feature/*"
+  push:
+    branches:
+      - "v*.*"
+      - "master"
+      - "feature/*"
+
+env:
+  PHP_VERSION: "8.2"
```
GromNaN, line 16: Is there any particular reason to put 8.2 and not 8.3? (which would postpone the need to modify the version for the next required version upgrades).

GromNaN (approved):
LGTM.

You can fix the [missing newline at EOF](https://github.com/mongodb/mongo-php-driver/pull/1624#discussion_r1749886051).

.github/workflows/composer-validate.yml
alcaeus: I essentially copied this from the "check arginfo" workflow. Depending on how the PHP version is determined in case it's absent, I'm also fine with not specifying a particular PHP version in this workflow.

GromNaN: Any version works. So ok to delete.

.github/workflows/composer-validate.yml
```diff
@@ -0,0 +1,30 @@
+name: "Check validity of composer.json"
```
jmikola, line 1: @alcaeus: I assume this was done for the PIE package, but did you plan to copy this workflow to [PHPLIB](https://github.com/mongodb/mongo-php-library/tree/master/.github/workflows) as well?

alcaeus: Yes, this was done to make sure we're alerted to changes to the pie schema, especially as it may change during the development process. I can copy the workflow over to PHPLIB 👍

PR #1637, opened 2024-09-09
Fix exception message expectation on MongoDB latest

alcaeus:
On MongoDB latest, the error message is different as the validation is handled by the execution planner. The additional `%S` ensures that tests pass on upcoming versions of MongoDB as well.


GromNaN (approved):
What is the message in the new version?

alcaeus:
The full error is:
> Plan executor error during update :: caused by :: Document failed validation

Using `%S` we can still assert that the `BulkWriteException` was caused by a document validation failure, instead of having to remove the exception message entirely and only checking for a write exception that could be caused by multiple other things.

GromNaN (approved):
LGTM

PR #1643, opened 2024-09-11
PHPC-2443: Deprecate string argument and accept Int64 instances in UTCDateTime constructor

alcaeus:
PHPC-2443

This PR deprecates creating a `MongoDB\BSON\UTCDateTime` instance with a string, which was previously used to pass values exceeding the 32-bit range on 32-bit platforms. With `MongoDB\BSON\Int64` available to the user, this has been replaced with the constructor accepting those values. I've updated all existing string usages in tests to `Int64` instances, except where explicitly testing string inputs.


src/BSON/UTCDateTime.c
```diff
@@ -179,36 +179,40 @@ static PHP_METHOD(MongoDB_BSON_UTCDateTime, __construct)
...
-	}
 
-	if (Z_TYPE_P(milliseconds) == IS_DOUBLE) {
-		char tmp[24];
-		int  tmp_len;
+			return;
 
-		tmp_len = snprintf(tmp, sizeof(tmp), "%.0f", Z_DVAL_P(milliseconds) > 0 ? floor(Z_DVAL_P(milliseconds)) : ceil(Z_DVAL_P(milliseconds)));
+		case IS_STRING:
+			php_error_docref(NULL, E_DEPRECATED, "Creating a %s instance with a string is deprecated and be removed in ext-mongodb 2.0", ZSTR_VAL(php_phongo_utcdatetime_ce->name));
```
jmikola:
```suggestion
			php_error_docref(NULL, E_DEPRECATED, "Creating a %s instance with a string is deprecated and will be removed in ext-mongodb 2.0", ZSTR_VAL(php_phongo_utcdatetime_ce->name));
```

tests/bson/bson-utcdatetime-001.phpt
```diff
@@ -11,7 +11,7 @@ require_once __DIR__ . "/../utils/basic.inc";
 
 $manager = create_test_manager();
 
-$utcdatetime = new MongoDB\BSON\UTCDateTime("1416445411987");
+$utcdatetime = new MongoDB\BSON\UTCDateTime(2147483647);
```
jmikola: Why did you opt to use `INT32_MAX` here instead of construct an Int64? If you wrap this value in an Int64, that should allow the expected output below to remain as-is.

tests/bson/bson-utcdatetime-002.phpt
```diff
@@ -13,6 +13,6 @@ var_dump($utcdatetime);
 --EXPECTF--
 object(MongoDB\BSON\UTCDateTime)#%d (%d) {
   ["milliseconds"]=>
-  %rint\(|string\(13\) "|%r1416445411987%r"|\)%r
+  %rint\(|string\(10\) "|%r2147483647%r"|\)%r
```
jmikola: The `int|string` expectation dates back to https://github.com/mongodb/mongo-php-driver/commit/e497c0155cd16547f3f8a314c4a2ce8500b9484d. At the time, UTCDateTime's debug handler conditionally reported an int or string depending on the build architecture. That behavior was later changed in https://github.com/mongodb/mongo-php-driver/commit/d25e565f647d824ced02146bd41d8ef17ff93cb9 to always use strings.

Therefore, I think this can just expect:

```suggestion
  string(10) "2147483647"
```

Note: related to my earlier comment, I'm not sure if you decided to use `INT32_MAX` here instead of wrapping the original value with an Int64 object; however, since you'll probably change the expect pattern anyway I don't see the problem with changing the input value.

tests/bson/bson-utcdatetime-008.phpt
```diff
@@ -0,0 +1,21 @@
...
+
+$utcdatetime = new MongoDB\BSON\UTCDateTime('1416445411987');
+
+var_dump($utcdatetime);
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+Deprecated: MongoDB\BSON\UTCDateTime::__construct(): Creating a MongoDB\BSON\UTCDateTime instance with a string is deprecated and be removed in ext-mongodb 2.0 %S
```
jmikola:
```suggestion
Deprecated: MongoDB\BSON\UTCDateTime::__construct(): Creating a MongoDB\BSON\UTCDateTime instance with a string is deprecated and will be removed in ext-mongodb 2.0 in %s
```

Added ` in %s` for consistency with the expectations in other tests.

tests/bson/bson-utcdatetime_error-003.phpt
```diff
@@ -26,13 +26,20 @@ echo throws(function() {
 ?>
 ===DONE===
 <?php exit(0); ?>
---EXPECT--
+--EXPECTF--
+Deprecated: MongoDB\BSON\UTCDateTime::__construct(): Creating a MongoDB\BSON\UTCDateTime instance with a string is deprecated and be removed in ext-mongodb 2.0 in %S
```
jmikola:
```suggestion
Deprecated: MongoDB\BSON\UTCDateTime::__construct(): Creating a MongoDB\BSON\UTCDateTime instance with a string is deprecated and be removed in ext-mongodb 2.0 in %s
```

`%S` is "zero or more", so I think you can use `%s` for all of these to actually expect a non-empty string.

src/BSON/UTCDateTime.c
```diff
@@ -179,36 +179,40 @@ static PHP_METHOD(MongoDB_BSON_UTCDateTime, __construct)
...
-
-	if (Z_TYPE_P(milliseconds) == IS_LONG) {
-		php_phongo_utcdatetime_init(intern, Z_LVAL_P(milliseconds));
-		return;
-	}
 
-	if (Z_TYPE_P(milliseconds) == IS_DOUBLE) {
-		char tmp[24];
-		int  tmp_len;
+			return;
```
jmikola, line 222: Why is this `return` outside of the brace block for `IS_DOUBLE`?

src/BSON/UTCDateTime.c
```diff
@@ -179,36 +179,40 @@ static PHP_METHOD(MongoDB_BSON_UTCDateTime, __construct)
 		return;
 	}
 
-	if (Z_TYPE_P(milliseconds) == IS_OBJECT) {
-		if (instanceof_function(Z_OBJCE_P(milliseconds), php_date_get_interface_ce())) {
-			php_phongo_utcdatetime_init_from_date(intern, Z_PHPDATE_P(milliseconds));
-		} else {
-			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected instance of DateTimeInterface, %s given", ZSTR_VAL(Z_OBJCE_P(milliseconds)->name));
+	switch (Z_TYPE_P(milliseconds)) {
```
jmikola, line 211: I don't feel strongly, but I'd rather have extra `Z_TYPE_P()` calls than all of the extra indentation. Feel free to leave as-is, but I think the block scope in `IS_DOUBLE` needs a change.

src/BSON/UTCDateTime.c
alcaeus: Whoops, fixed

tests/bson/bson-utcdatetime-001.phpt
alcaeus: I initially changed all tests to `INT32_MAX` before realising that using an `Int64` instance was better, but forgot to change these two tests. These are now consistent with the others.

tests/bson/bson-utcdatetime-002.phpt
alcaeus: Ah, yes. I've changed the expectf format to always expect a string.

tests/bson/bson-utcdatetime-008.phpt
alcaeus: Fixed

tests/bson/bson-utcdatetime_error-003.phpt
alcaeus: Fixed

src/BSON/UTCDateTime.c
alcaeus: Just the way I refactored this. Note I extracted this logic into `php_phongo_utcdatetime_init_from_double` for consistency.

src/BSON/UTCDateTime.c
alcaeus: I extracted the logic for objects and doubles so the `switch` only defers to specific methods. This makes `construct` easier to understand at a glance and helps isolating the logic for the various input formats.

jmikola: Much more readable. Thanks!

jmikola (approved):

PR #1645, opened 2024-09-12
PHPC-2350: Implement JSON methods for PackedArray

alcaeus:
PHPC-2350

We can leverage `bson_new_from_json` to create a `bson_t` for an array, and `bson_as_json_with_opts` lets us export the `bson_t` to JSON with an array being the root object.

Note that since `bson_new_from_json` was intended to read documents, the resulting `PackedArray` instance will retain the original keys in the object, but iterating it or converting it to JSON will reindex the array. However, when using the `PackedArray` instance in certain ways will cause errors (see `bson-packedarray-fromJSON-003.phpt` which fails the last operation). The only option I see is to iterate the resulting `bson_t` and either re-index it (essentially re-writing a `bson_t` array) or to throw if one of the keys isn't what we expect. Since we have the mantra of "be lenient in what you accept, but strict in what you produce" I'm inclined to go with the first option, despite this potentially causing some overhead. WDYT @jmikola?

### DIFF
--- a/tests/bson/bson-packedarray-fromJSON_error-001.phpt
+++ b/tests/bson/bson-packedarray-fromJSON_error-001.phpt
@@ -13,6 +13,10 @@ echo throws(function() {
     MongoDB\BSON\PackedArray::fromJSON('{ "foo": "bar" }');
 }, 'MongoDB\Driver\Exception\UnexpectedValueException'), PHP_EOL;
 
+echo throws(function() {
+    MongoDB\BSON\PackedArray::fromJSON('{ "00": "bar", "1": "bar" }');
+}, 'MongoDB\Driver\Exception\UnexpectedValueException'), PHP_EOL;
+
 ?>
 ===DONE===
 <?php exit(0); ?>
@@ -21,4 +25,6 @@ OK: Got MongoDB\Driver\Exception\UnexpectedValueException
 Got parse error at "o", position 1: "SPECIAL_EXPECTED"
 OK: Got MongoDB\Driver\Exception\UnexpectedValueException
 Received invalid JSON array: expected key 0, but found "foo"
+OK: Got MongoDB\Driver\Exception\UnexpectedValueException
+Received invalid JSON array: expected key 0, but found "00"
 ===DONE===
--- a/tests/bson/bson-packedarray-fromJSON_error-001.phpt
+++ b/tests/bson/bson-packedarray-fromJSON_error-001.phpt
@@ -17,6 +17,10 @@ echo throws(function() {
     MongoDB\BSON\PackedArray::fromJSON('{ "00": "bar", "1": "bar" }');
 }, 'MongoDB\Driver\Exception\UnexpectedValueException'), PHP_EOL;
 
+echo throws(function() {
+    MongoDB\BSON\PackedArray::fromJSON('{ "0": "bar", "foo": "bar" }');
+}, 'MongoDB\Driver\Exception\UnexpectedValueException'), PHP_EOL;
+
 ?>
 ===DONE===
 <?php exit(0); ?>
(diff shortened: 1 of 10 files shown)


tests/bson/bson-packedarray-fromJSON-003.phpt
```diff
@@ -0,0 +1,31 @@
...
+--FILE--
+<?php
+
+require_once __DIR__ . '/../utils/basic.inc';
+
+$bson = MongoDB\BSON\PackedArray::fromJSON('{ "foo": "bar" }');
+var_dump($bson);
+hex_dump((string) $bson);
+echo $bson->toCanonicalExtendedJSON(), PHP_EOL;
+echo MongoDB\BSON\Document::fromPHP(['array' => $bson])->toCanonicalExtendedJSON(), PHP_EOL;
```
jmikola: > Note that since `bson_new_from_json` was intended to read documents, the resulting PackedArray instance will retain the original keys in the object, but iterating it or converting it to JSON will reindex the array. However, when using the PackedArray instance in certain ways will cause errors.

I'm having trouble understanding why _this_ fails but the preceding `$bson->toCanonicalExtendedJSON()` does not.

Does `bson_json_opts_set_outermost_array()` apply some logic that disregards invalid keys, while the code path for converting a _nested_ array to extended JSON does not?

> The only option I see is to iterate the resulting bson_t and either re-index it (essentially re-writing a bson_t array) or to throw if one of the keys isn't what we expect. Since we have the mantra of "be lenient in what you accept, but strict in what you produce" I'm inclined to go with the first option, despite this potentially causing some overhead.

If [PackedArray::fromPHP()](https://www.php.net/manual/en/mongodb-bson-packedarray.fromphp.php) requires a well-formed list (see: [bson-packedarray-fromPHP_error-001.phpt](https://github.com/mongodb/mongo-php-driver/blob/v1.x/tests/bson/bson-packedarray-fromPHP_error-001.phpt)), I don't see why we can't require that in `fromJSON()` as well.

Perhaps we can't prevent libbson from accepting a JSON object with sequential numeric string keys starting at zero, but if it's easy to detect something like `{ "foo": "bar" }` I'm all for throwing an error there.

alcaeus: I added logic to iterate the array and compare the keys. I've used `strtoll` to parse the string, as I didn't want to allocate memory to convert the expected key to a string. I'm not sure if that has any performance implications, but I don't expect this method to be somewhere in the hot path given its limited usefulness.

jmikola: Recapping our Slack discussion, we discussed a caveat with `strtoll()` accepting non-canonical numeric strings. I'll review the second approach.

tests/bson/bson-packedarray-fromJSON_error-001.phpt
```diff
@@ -0,0 +1,30 @@
...
+echo throws(function() {
+    MongoDB\BSON\PackedArray::fromJSON('foo');
+}, 'MongoDB\Driver\Exception\UnexpectedValueException'), PHP_EOL;
+
+echo throws(function() {
+    MongoDB\BSON\PackedArray::fromJSON('{ "foo": "bar" }');
+}, 'MongoDB\Driver\Exception\UnexpectedValueException'), PHP_EOL;
+
+echo throws(function() {
+    MongoDB\BSON\PackedArray::fromJSON('{ "00": "bar", "1": "bar" }');
```
jmikola, line 17: Is there any need for a test where the invalid key is _not_ the first? Perhaps not, as the extra validation in PHPC means we're no longer relying on libbson's lazy check.

jmikola (approved):

tests/bson/bson-packedarray-fromJSON_error-001.phpt
alcaeus: Not necessarily, but I'll add one for the sake of completeness.

PR #1652, opened 2024-09-17
PHP-1741: Remove getServer from APM event classes

jmikola:
https://jira.mongodb.org/browse/PHPC-1741

This also removes the internal Manager reference.


src/MongoDB/Monitoring/CommandFailedEvent.c
```diff
@@ -195,10 +184,6 @@ static void php_phongo_commandfailedevent_free_object(zend_object* object)
 		zval_ptr_dtor(&intern->z_error);
 	}
 
-	if (!Z_ISUNDEF(intern->manager)) {
-		zval_ptr_dtor(&intern->manager);
```
alcaeus, line 199: You can also remove the `manager` field from the corresponding struct as it's no longer used. Same applies to the other command event structs.

alcaeus (commented):
Changes look good, but there's a leftover field in the command event structs that's no longer needed.

src/MongoDB/Monitoring/CommandFailedEvent.c
jmikola: Ah, thanks for catching that :v:

alcaeus (approved):

PR #1653, opened 2024-09-17
PHPC-2346: Remove deprecated BSON functions

alcaeus:
PHPC-2346

This removes the BSON functions that were deprecated in favour of the `MongoDB\BSON\Document` class.


alcaeus:
Waiting on #1652 to be merged to add the upgrade note.

tests/bson/bug0544.phpt
```diff
@@ -19,121 +19,105 @@ $tests = [
 ];
 
 foreach ($tests as $test) {
-    $bson = MongoDB\BSON\fromPHP($test);
+    $bson = MongoDB\BSON\Document::fromPHP($test);
     /* Note: Although libbson can parse the extended JSON representation for
      * 64-bit integers (i.e. "$numberLong"), it currently prints them as
      * doubles (see: https://jira.mongodb.org/browse/CDRIVER-375). */
-    printf("Test %s\n", MongoDB\BSON\toJSON($bson));
-    hex_dump($bson);
```
jmikola, line 27: Was asserting the parsed BSON not relevant to this test?

tests/bson/bug0592.phpt
```diff
@@ -23,8 +23,8 @@ $tests = [
 foreach ($tests as $json) {
     printf("Test %s\n", $json);
     try {
-        $encoded = MongoDB\BSON\toPHP(MongoDB\BSON\fromJSON($json));
-        var_dump( $encoded );
+        $encoded = MongoDB\BSON\Document::fromJSON($json)->toPHP();
+        var_dump($encoded);
```
jmikola:
```suggestion
        var_dump(MongoDB\BSON\Document::fromJSON($json)->toPHP());
```

tests/bson/bug0592.phpt
```diff
@@ -36,30 +36,24 @@ foreach ($tests as $json) {
...
-
-Deprecated: Function MongoDB\BSON\fromJSON() is deprecated in %s
-
-Deprecated: Function MongoDB\BSON\toPHP() is deprecated in %s
 object(stdClass)#%d (%d) {
   ["x"]=>
-  int(-2147483648)
+  object(MongoDB\BSON\Int64)#%d (%d) {
+    ["integer"]=>
+    string(11) "-2147483648"
```
jmikola, line 38: I assume this is related to changing `toPHP(fromJSON(...))` to `Document::fromJSON(...)->toPHP()`, but don't the `toPHP()` functions use the same logic internally? Just curious why this might have changed.

The current output seems more correct, as the JSON input is clearly a 64-bit integer type regardless of its value.

alcaeus: That is pretty much the only notable difference between the two implementations. Interacting with a `MongoDB\BSON\Document` instance will always yield an `Int64` instance for 64-bit integers, even when they fall within the 32-bit range. This will not matter in most cases as an `Int64` can be coerced to an int and also used in operations as if it were an integer.

tests/bson/bug0544.phpt
alcaeus: The parsed BSON is printed as part of the debug data of a `MongoDB\BSON\Document` instance. It prints the BSON bytes (base64 encoded for easier handling) as well as the PHP representation, so there's no need to use `hex_dump` as well.

jmikola: Oh, I overlooked the "data" field in the debug output. :+1:

jmikola (approved):

PR #1654, opened 2024-09-17
PHPC-2447: Extend Iterator interface in CustorInterface

alcaeus:
PHPC-2447

Making this change now will make typing changes in PHPLIB-954 much easier.


tests/cursor/cursorinterface-003.phpt
```diff
@@ -1,11 +0,0 @@
---TEST--
-MongoDB\Driver\CursorInterface does not extend Iterator
---FILE--
-<?php
-var_dump(in_array('Iterator', class_implements('MongoDB\Driver\CursorInterface')));
-?>
-===DONE===
-<?php exit(0); ?>
---EXPECT--
-bool(false)
```
GromNaN, line 10: Noted that we already have a test that checks that `CursorInterface` implements `Iterator`.

GromNaN (approved):

GromNaN:
We need to specify that `CursorInterface::key()` return `?int` instead of `mixed`.

https://github.com/mongodb/mongo-php-driver/blob/525dadb9ce0b9a1e22c5efc7d8f02cc25743272b/src/MongoDB/Cursor.stub.php#L24

This is necessary if we want [`CodecCursor`](https://github.com/mongodb/mongo-php-library/blob/89fb97ad230985d82d903306f63cae81122acf4f/src/Model/CodecCursor.php#L106-L109) to accept `CursorInterface` instead of `Cursor`.

alcaeus:
Ah, good point. I'll update the types in #1658.

PR #1658, opened 2024-09-17
PHPC-2140: Make tentative return types definitive

alcaeus:
PHPC-2140

This PR removes the `@tentative-return-type` annotation in all stub files, making the return types definitive and causing errors if an extending or implementing class does not have a covariant return type.

It also changes the return types for various methods in `MongoDB\Driver\Manager` and `MongoDB\Driver\Server` from `Cursor` to `CursorInterface`.

Note: currently contains changes from #1653 to avoid fixing tests that will be removed.

### DIFF
--- a/src/MongoDB/CursorInterface.stub.php
+++ b/src/MongoDB/CursorInterface.stub.php
@@ -9,18 +9,13 @@ namespace MongoDB\Driver;
 
 interface CursorInterface extends \Iterator
 {
-    /** @tentative-return-type */
     public function getId(): CursorId|\MongoDB\BSON\Int64;
 
-    /** @tentative-return-type */
     public function getServer(): Server;
 
-    /** @tentative-return-type */
     public function isDead(): bool;
 
-    /** @tentative-return-type */
     public function setTypeMap(array $typemap): void;
 
-    /** @tentative-return-type */
     public function toArray(): array;
 }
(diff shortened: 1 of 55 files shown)


src/MongoDB/CursorInterface.stub.php
GromNaN, line 1: The method `key(): int` needs to be added. https://github.com/mongodb/mongo-php-driver/pull/1654#issuecomment-2356219017

src/MongoDB/CursorInterface.stub.php
```diff
@@ -9,18 +9,17 @@
 
 interface CursorInterface extends \Iterator
 {
-    /** @tentative-return-type */
+    public function current(): array|object|null {}
```
jmikola, line 12: Do I assume correctly that only `current()` and `key()`  are included here because those are the only methods where we'd narrowing the return type from the parent interface?

jmikola (approved):

src/MongoDB/CursorInterface.stub.php
GromNaN: This methods are mentioned in the upgrade file:

> The return types for the `key` and `current` methods have been narrowed to the types returned by cursor instances.

GromNaN (approved):

PR #1663, opened 2024-09-18
PHPC-2248: Remove Serializable implementations

alcaeus:
PHPC-2248

This will conflict with other pull request, but I'll rebase before merging and resolve conflicts.


UPGRADE-2.0.md
```diff
@@ -4,3 +4,6 @@ UPGRADE FROM 1.x to 2.0
  * The `getServer()` method has been removed from the CommandFailedEvent,
    CommandStartedEvent, and CommandSucceededEvent event classes. The `getHost()`
    and `getPort()` methods have been added in its place.
+ * All classes that previously implemented the `Serializable` interface no
+   longer implement this. This changes the serialization format, but the 
+   ability to serialize and unserialize these classes remains as-is.
```
jmikola: I don't think we need to say "this changes the serialization format" as the [Serializable](https://www.php.net/manual/en/class.serializable.php) interface hasn't been used in some time.

In PHP 7.4+, the `__serialize()` and `__unserialize()` magic methods are preferred when available (see: [docs](https://www.php.net/manual/en/language.oop5.magic.php#language.oop5.magic.serialize)). We also noted this in [PHPC-2230](https://jira.mongodb.org/browse/PHPC-2230?focusedCommentId=5505609&focusedId=5505609&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-5505609) when dropping support for PHP 7.2 and 7.3 in 1.17.0.

----

Presumably, the only loss in functionality here is that users would no longer be able to unserialize a string in the legacy "C format" since that would depend on `Serializable::unserialize()`. All of the tests for handling that format were removed in https://github.com/mongodb/mongo-php-driver/commit/2ab1e4ba2e6b3126e7acbd7025f3a2727aaa48d2.

src/BSON/Binary.c
```diff
@@ -184,60 +184,6 @@ static PHP_METHOD(MongoDB_BSON_Binary, jsonSerialize)
 	ADD_ASSOC_STRINGL(return_value, "$type", type, type_len);
 }
 
-static PHP_METHOD(MongoDB_BSON_Binary, serialize)
-{
-	php_phongo_binary_t* intern;
-	zval                 retval;
-	php_serialize_data_t var_hash;
-	smart_str            buf = { 0 };
```
jmikola, line 192: You should be able to remove includes for `zend_smart_str.h` and `ext/standard/php_var.h` for all of these class files. Those were originally added for the Serializable implementations (e.g. https://github.com/mongodb/mongo-php-driver/commit/8ee86a15eb0c4d22ac41a7e8a0739d66594f4ce2).

UPGRADE-2.0.md
alcaeus: Good point. I removed the second sentence and only highlight that the interface is no longer implemented.

src/BSON/Binary.c
alcaeus: Ah, indeed 👍

tests/serverApi/serverApi-serialization_error-002.phpt
```diff
@@ -6,15 +6,15 @@ MongoDB\Driver\ServerApi unserialization errors (__serialize and __unserialize)
 require_once __DIR__ . '/../utils/basic.inc';
 
 echo throws(function() {
-    unserialize('C:24:"MongoDB\Driver\ServerApi":24:{a:1:{s:7:"version";i:0;}}');
```
jmikola, line 9: Ah, I was wondering if we had any examples that still relied on `Serializable::unserialize()`. Missed this earlier, but it's what I had in mind when writing about the possible loss in functionality in https://github.com/mongodb/mongo-php-driver/pull/1663#discussion_r1765303021.

tests/utils/tools.php
```diff
@@ -1,5 +1,8 @@
 <?php
 
+use MongoDB\BSON\DBPointer;
+use MongoDB\BSON\Symbol;
+use MongoDB\BSON\Undefined;
```
jmikola: The Undefined class isn't referenced anywhere.

```suggestion
```

I noted that we typically created instances of Undefined using Extended JSON. See: [`bson-undefined-serialization-002.phpt`](https://github.com/mongodb/mongo-php-driver/blob/v1.x/tests/bson/bson-undefined-serialization-002.phpt).

You already have working implementations for DBPointer and Symbol here, so probably not reason to touch those further. Thanks for parameterizing them.

Why not the same approach for DBPointer and Symbol?

jmikola (approved):
One suggestion but LGTM after that. I've disabled auto-merge so we can address that first.

PR #1666, opened 2024-09-18
PHPC-1021: Remove support for ReadPreference integer modes

jmikola:
https://jira.mongodb.org/browse/PHPC-1021

This removes the deprecated integer constants, constructor support, and getMode().

Error-checking in php_phongo_readpreference_get_mode_string() has also been removed. ReadPreference only uses mongoc_read_prefs_new() internally, and we should not need to worry about libmongoc feeding us mongoc_read_prefs_t with an invalid mode.


src/MongoDB/ReadPreference.c
```diff
@@ -526,16 +477,12 @@ static PHP_METHOD(MongoDB_Driver_ReadPreference, serialize)
 	}
 
 	tags                = mongoc_read_prefs_get_tags(intern->read_preference);
-	mode                = mongoc_read_prefs_get_mode(intern->read_preference);
```
jmikola, line 529: Note: this will conflict with https://github.com/mongodb/mongo-php-driver/pull/1663, depending on which gets merged first.

src/MongoDB/ReadPreference.c
```diff
@@ -188,44 +184,24 @@ static PHP_METHOD(MongoDB_Driver_ReadPreference, __construct)
...
-			intern->read_preference = mongoc_read_prefs_new(MONGOC_READ_SECONDARY);
-		} else if (strcasecmp(Z_STRVAL_P(mode), PHONGO_READ_SECONDARY_PREFERRED) == 0) {
-			intern->read_preference = mongoc_read_prefs_new(MONGOC_READ_SECONDARY_PREFERRED);
-		} else if (strcasecmp(Z_STRVAL_P(mode), PHONGO_READ_NEAREST) == 0) {
-			intern->read_preference = mongoc_read_prefs_new(MONGOC_READ_NEAREST);
-		} else {
-			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Invalid mode: '%s'", Z_STRVAL_P(mode));
-			return;
-		}
+	if (zend_string_equals_literal_ci(mode, PHONGO_READ_PRIMARY)) {
```
jmikola, line 193: I was delighted to find this in `zend_string.h`.

src/MongoDB/ReadPreference.c
```diff
@@ -153,9 +153,9 @@ static bool php_phongo_readpreference_init_from_hash(php_phongo_readpreference_t
 	return false;
 }
 
-static const char* php_phongo_readpreference_get_mode_string(mongoc_read_mode_t mode)
+static const char* php_phongo_readpreference_get_mode_string(const mongoc_read_prefs_t* read_prefs)
```
jmikola, line 156: This lets us eliminate some `mongoc_read_prefs_get_mode()` calls in parent contexts.

Additionally, I changed this function to return an empty string instead of throwing in the very unlikely event that the RP mode is invalid. [`_mongoc_read_mode_as_str()`](https://github.com/mongodb/mongo-c-driver/blob/1.28.0/src/libmongoc/src/mongoc/mongoc-read-prefs.c#L198) does the same, and libmongoc uses that for appending read preferences to command payloads. In our case, this is used for serialization and debugging, and the change allowed us to clean up the parent contexts.

tests/readPreference/readpreference-ctor_error-005.phpt
```diff
@@ -1,18 +0,0 @@
---TEST--
-MongoDB\Driver\ReadPreference construction (invalid string mode)
```
jmikola, line 2: This test was redundant as `readpreference-ctor_error-001.phpt` uses an invalid string value.

tests/readPreference/readpreference-ctor_error-006.phpt
```diff
@@ -1,18 +0,0 @@
---TEST--
-MongoDB\Driver\ReadPreference construction (invalid type for mode)
---FILE--
-<?php
-
-require_once __DIR__ . '/../utils/basic.inc';
-
-echo throws(function() {
-    new MongoDB\Driver\ReadPreference(3.14);
```
jmikola, line 9: Since we replaced `string|int` with `string` in the constructor signature, type coercion is now applied outside of strict mode and this would be implicitly cast to `'3.14'`. I didn't see any reason to use a different type (or declare strict types) just to assert a TypeError.

alcaeus (approved):

PR #1673, opened 2024-09-20
PHPC-2448: Update to libmongoc 1.28.0

alcaeus:
PHPC-2448


sbom.json
```diff
@@ -1,52 +1,52 @@
 {
   "components": [
     {
-      "bom-ref": "pkg:github/mongodb/libmongocrypt@1.10.1",
+      "bom-ref": "pkg:github/mongodb/libmongocrypt@1.11.0",
```
jmikola, line 4: I assume this was missed in https://github.com/mongodb/mongo-php-driver/commit/da39d53b3a7d2f5eec884c5f0ce5ebc8ab4599da.

I noted that we have [Update SBOM file](https://github.com/mongodb/mongo-php-driver/blob/v1.x/CONTRIBUTING.md#update-sbom-file) in the contributing docs, but I likely would have forgotten to do this myself.

Is it feasible to try and catch this in CI? Having it checked before allowing a release would be sufficient, as there's likely no harm in it being temporarily out of sync for the development branch.

jmikola (approved):

sbom.json
alcaeus: Good point. I filed PHPC-2450 to track this.

PR #1676, opened 2024-09-23
PHPC-2309: Remove --with-openssl-dir configure option

jmikola:
https://jira.mongodb.org/browse/PHPC-2309

For the time being, we will preserve fall back logic to detect OpenSSL with PHP_CHECK_LIBRARY; however, that will be eventually be removed (PHPC-2451).

It should be noted that PHP 7.4+ relies exclusively on pkg-config to detect OpenSSL: php/php-src@fe8fdfa3bd588d80ce60f6b3848058239e0a760f

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -30,3 +30,5 @@ UPGRADE FROM 1.x to 2.0
  * The constructor of `MongoDB\BSON\UTCDateTime` no longer accepts a `string`
    argument. To pass 64-bit integers on 32-bit platforms, use the
    `MongoDB\BSON\Int64` class instead.
+ * The `--with-openssl-dir` configure option has been removed. If using OpenSSL,
+   ensure that it is detected by `pkg-config`.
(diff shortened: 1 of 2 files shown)


UPGRADE-2.0.md
```diff
@@ -30,3 +30,5 @@ UPGRADE FROM 1.x to 2.0
  * The constructor of `MongoDB\BSON\UTCDateTime` no longer accepts a `string`
    argument. To pass 64-bit integers on 32-bit platforms, use the
    `MongoDB\BSON\Int64` class instead.
+ * The `--with-openssl-dir` configure option has been removed. If using OpenSSL,
+   ensure that it is detected by `pkg-config`.
```
jmikola, line 34: I just realized that we never added a warning when using the option in 1.x. Opened [PHPC-2452](https://jira.mongodb.org/browse/PHPC-2452) and https://github.com/mongodb/mongo-php-driver/pull/1677 to address that.

alcaeus (approved):

PR #1683, opened 2024-09-24
PHPC-2456: Correctly dereference arrays in type maps

alcaeus:
PHPC-2456

This PR adds dereferencing logic to the PHP array API polyfill. This logic has already been added for longs, strings, and doubles, but not for arrays and objects. This can cause issues when passing a reference to an array in a type map, as we extract an array but receive `NULL`. To the user, this makes no sense as from their point of view, a reference to an array is an array like any other.

Note that I'm targeting v1.19 here as that's our current stable branch, but we may opt to not release another 1.19 patch release and only fix this in 1.20.0.

### DIFF
--- a/src/contrib/php_array_api.h
+++ b/src/contrib/php_array_api.h
@@ -420,7 +420,9 @@ try_again:
  * zval *php_array_fetchz_array(zval *zarr, zval *key)
  */
 static inline zval *php_array_zval_to_array(zval *zarr) {
-	return (zarr && (Z_TYPE_P(zarr) == IS_ARRAY)) ? zarr : NULL;
+	if (!zarr) { return NULL; }
+	ZVAL_DEREF(zarr);
+	return Z_TYPE_P(zarr) == IS_ARRAY ? zarr : NULL;
 }
 PHP_ARRAY_FETCH_TYPE_MAP(zval*, array)
 #define php_array_fetchc_array(zarr, litstr) \
@@ -493,8 +495,10 @@ void *php_array_zval_to_resource(zval *z, int le) {
  */
 static inline
 zval *php_array_zval_to_object(zval *z, zend_class_entry *ce) {
-	return (z && (Z_TYPE_P(z) == IS_OBJECT) &&
-	        ((!ce) || instanceof_function(Z_OBJCE_P(z), ce))) ? z : NULL;
+	if (!z) { return NULL; }
+	ZVAL_DEREF(z);
+	if (Z_TYPE_P(z) != IS_OBJECT) { return NULL; }
+	return (!ce) || instanceof_function(Z_OBJCE_P(z), ce) ? z : NULL;
 }
 #define php_array_fetch_object(zarr, key, ce) \
 	php_array_zval_to_object(php_array_fetch(zarr, key), ce)
--- a/tests/bson/bug2456-001.phpt
+++ b/tests/bson/bug2456-001.phpt
@@ -1,0 +1,22 @@
+--TEST--
+PHPC-2456: References passed in a typeMap
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$fieldPaths = [];
+
+$typeMap = ['fieldPaths' => &$fieldPaths];
+
+var_dump(MongoDB\BSON\Document::fromPHP([])->toPHP($typeMap));
+var_dump(MongoDB\BSON\PackedArray::fromPHP([])->toPHP($typeMap));
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+object(stdClass)#%d (0) {
+}
+array(0) {
+}
+===DONE===


GromNaN (approved):

src/contrib/php_array_api.h
```diff
@@ -420,7 +420,17 @@ char *php_array_zval_to_string(zval *z, int *plen, zend_bool *pfree) {
...
+	if (!zarr) { return NULL; }
+	switch (Z_TYPE_P(zarr)) {
+		case IS_REFERENCE:
+			ZVAL_DEREF(zarr);
+			goto try_again;
+		case IS_ARRAY:
+			return zarr;
+		default:
+			return NULL;
+	}
```
jmikola: This looks related to a previous issue we had with dereferencing strings. See https://github.com/mongodb/mongo-php-driver/commit/c68cdcda7654a630b6da52d99f7ce5e2b3fdabfc from [PHPC-1839](https://jira.mongodb.org/browse/PHPC-1839).

Also, you should contribute all of these `ZVAL_DEREF()` changes upstream to [sgolemon/php-array-api](https://github.com/sgolemon/php-array-api/) when you get a chance. It's your ticket to the very exclusive [contributors club](https://github.com/sgolemon/php-array-api/graphs/contributors).

Maybe @sgolemon can make us t-shirts.

src/contrib/php_array_api.h
```diff
@@ -420,7 +420,17 @@ char *php_array_zval_to_string(zval *z, int *plen, zend_bool *pfree) {
  * zval *php_array_fetchz_array(zval *zarr, zval *key)
  */
 static inline zval *php_array_zval_to_array(zval *zarr) {
-	return (zarr && (Z_TYPE_P(zarr) == IS_ARRAY)) ? zarr : NULL;
+try_again:
+	if (!zarr) { return NULL; }
+	switch (Z_TYPE_P(zarr)) {
+		case IS_REFERENCE:
+			ZVAL_DEREF(zarr);
+			goto try_again;
```
jmikola: I realize this is the same logic from https://github.com/mongodb/mongo-php-driver/commit/c68cdcda7654a630b6da52d99f7ce5e2b3fdabfc, but I'm curious this could be replaced with a fall-through.

[Dereferencing and Unwrapping](https://www.phpinternalsbook.com/php7/zvals/references.html#dereferencing-and-unwrapping) includes the following code example:

```c
zval *val;
ZEND_HASH_FOREACH_VAL(ht, val) {
    ZVAL_DEREF(val);

    /* Do something with val, now a guaranteed non-reference. */
} ZEND_HASH_FOREACH_END();
```

That implies that dereferencing would never yield another `IS_REFERENCE` type. That is also supported by this earlier text in [Initializing references](https://www.phpinternalsbook.com/php7/zvals/references.html#initializing-references):

> ```c
> zval *zv = /* ... */;
> ZVAL_MAKE_REF(zv);
> ```
> 
> If zv was already a reference, this does nothing. It if wasn’t a reference yet, this will change zv into a reference and set its initial value to the old value of zv.

It's a bit too early for me to dive through the macros in [zend_types.h](https://heap.space/xref/PHP-8.1/Zend/zend_types.h) to confirm this.

jmikola: Looking through [php-src](https://github.com/search?q=repo%3Aphp%2Fphp-src%20ZVAL_DEREF&type=code), I don't see any examples of `ZVAL_DEREF()` being called iteratively. It tends to precede zval access, and code relies on it being a no-op for non-references. In that case, perhaps you could unconditionally call `ZVAL_DEREF(zarr)` after the initial `NULL` check and eliminate the `goto`.

Likewise for the original patch in https://github.com/mongodb/mongo-php-driver/commit/c68cdcda7654a630b6da52d99f7ce5e2b3fdabfc.

If you'd rather not make this change now, I'd consider it for the upstream contribution to [sgolemon/php-array-api](https://github.com/sgolemon/php-array-api/).

jmikola (approved):
I have a suggestion to eliminate the `goto`, but I'm sure this works as-is given the regression test (and previous, related issue).

src/contrib/php_array_api.h
alcaeus: Looking at this, I don't think that will work. After calling `ZVAL_DEREF` we still have to check the type of the dereferenced zval before accessing the ce, which wouldn't work with a fallthrough. As much as I hate the `goto` way of doing things, the only other option I see is a `while (Z_TYPE_P(z) == IS_REFERENCE)`, which is equally bad.

src/contrib/php_array_api.h
alcaeus: Will contribute this upstream 👍

src/contrib/php_array_api.h
alcaeus: Removed the `goto` after all by calling `ZVAL_DEREF` unconditionally after a `NULL` check.

PR #1687, opened 2024-09-24
PHPC-2144 Throw a LogicException when getting info from unacknowledged write result

GromNaN:
Fix PHPC-2144

Transform deprecation to exception. The exception message comes from PHPLIB: https://github.com/mongodb/mongo-php-library/blob/v1.x/src/Exception/BadMethodCallException.php#L45

And remove nullable return types from `WriteResult::get*Count` accessors

### DIFF
--- a/src/MongoDB/WriteResult.c
+++ b/src/MongoDB/WriteResult.c
@@ -31,10 +31,9 @@
 #include "MongoDB/WriteError.h"
 #include "WriteResult_arginfo.h"
 
-#define PHONGO_WRITERESULT_CHECK_ACKNOWLEDGED(method)                                                                                                                                        \
-	if (!mongoc_write_concern_is_acknowledged(intern->write_concern)) {                                                                                                                      \
-		php_error_docref(NULL, E_DEPRECATED, "Calling MongoDB\\Driver\\WriteResult::" method "() for an unacknowledged write is deprecated and will throw an exception in ext-mongodb 2.0"); \
-		RETURN_NULL();                                                                                                                                                                       \
+#define PHONGO_WRITERESULT_CHECK_ACKNOWLEDGED(method)                                                                                                     \
+	if (!mongoc_write_concern_is_acknowledged(intern->write_concern)) {                                                                                   \
+		phongo_throw_exception(PHONGO_ERROR_LOGIC, "MongoDB\\Driver\\WriteResult::" method "() should not be called for an unacknowledged write result"); \
 	}
 
 #define PHONGO_WRITERESULT_RETURN_LONG_FROM_BSON_INT32(iter, bson, key)                \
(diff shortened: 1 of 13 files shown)


src/MongoDB/WriteResult_arginfo.h
```diff
@@ -1,10 +1,10 @@
 /* This is a generated file, edit the .stub.php file instead.
- * Stub hash: 84f1690181585d88c4e7233a4c959d5f50dc1b40 */
+ * Stub hash: 217742ba620620cfaee2758070ff13e651620f87 */
 
 ZEND_BEGIN_ARG_INFO_EX(arginfo_class_MongoDB_Driver_WriteResult___construct, 0, 0, 0)
 ZEND_END_ARG_INFO()
 
-ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(arginfo_class_MongoDB_Driver_WriteResult_getInsertedCount, 0, 0, IS_LONG, 1)
+ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(arginfo_class_MongoDB_Driver_WriteResult_getInsertedCount, 0, 0, IS_LONG, 0)
```
GromNaN, line 7: Using the first method definition for other methods looks strange, but since that's generated code it might be an optimization.

tests/writeResult/writeresult-getdeletedcount-002.phpt
```diff
@@ -19,12 +19,13 @@ $bulk->delete(['x' => 1]);
 
 $result = $manager->executeBulkWrite(NS, $bulk, new MongoDB\Driver\WriteConcern(0));
 
-var_dump($result->getDeletedCount());
+throws(function() use ($result) {
+    $result->getDeletedCount();
+}, MongoDB\Driver\Exception\LogicException::class, 'getDeletedCount');
```
GromNaN: Even if it's optional, adding an assertion to the throwing method costs nothing.

src/MongoDB/WriteResult.c
```diff
@@ -31,10 +31,9 @@
...
-	if (!mongoc_write_concern_is_acknowledged(intern->write_concern)) {                                                                                                                      \
-		php_error_docref(NULL, E_DEPRECATED, "Calling MongoDB\\Driver\\WriteResult::" method "() for an unacknowledged write is deprecated and will throw an exception in ext-mongodb 2.0"); \
-		RETURN_NULL();                                                                                                                                                                       \
+#define PHONGO_WRITERESULT_CHECK_ACKNOWLEDGED(method)                                                                                                     \
+	if (!mongoc_write_concern_is_acknowledged(intern->write_concern)) {                                                                                   \
+		phongo_throw_exception(PHONGO_ERROR_LOGIC, "MongoDB\\Driver\\WriteResult::%s() should not be called for an unacknowledged write result", method); \
```
jmikola:
```suggestion
		phongo_throw_exception(PHONGO_ERROR_LOGIC, "MongoDB\\Driver\\WriteResult::" method "() should not be called for an unacknowledged write result"); \
```

`method` is a string literal so we can just concatenate it directly and avoid printf-style interpolation.

jmikola (approved):

PR #1690, opened 2024-09-24
Document how to run part of the test suite

GromNaN:


CONTRIBUTING.md
```diff
@@ -31,7 +31,7 @@ $ php --ri mongodb
 ## Generating arginfo from stub files
 
 Arginfo structures are generated from stub files using the `gen_stub.php`
-file. Note that this requires `phpize` to be run for PHP 8.2 to make use
+file. Note that this requires `phpize` to be run for **PHP 8.2** to make use
```
jmikola, line 34: Should this read PHP 8.2+? Locally, I've been running this with PHP 8.3 for some time.

jmikola (approved):
Optional suggestion for `phpize` instructions.

CONTRIBUTING.md
GromNaN: With PHP 8.3, generated files are different: https://github.com/mongodb/mongo-php-driver/pull/1687#discussion_r1773485175

jmikola: Is it only PHP 8.3's `gen_stub.php` that has the optimization to re-use arginfo for methods? It's not clear to me what you used in that PR.

I'll note that [workflows/arginfo-files.yml](https://github.com/mongodb/mongo-php-driver/blob/v1.x/.github/workflows/arginfo-files.yml) uses PHP 8.2, so no objections to requiring that here. If PHP 8.3 actually generates different output, I suppose CI would fail.

PR #1694, opened 2024-09-24
PHPC-2457 Fix modifiers and other Query options by reference

GromNaN:
Fix PHPC-2457

Similar to #1683.

### DIFF
--- a/src/MongoDB/Query.c
+++ b/src/MongoDB/Query.c
@@ -58,7 +58,6 @@ static bool php_phongo_query_opts_append_document(bson_t* opts, const char* opts
 	zval*  value = php_array_fetch(zarr, zarr_key);
 	bson_t b     = BSON_INITIALIZER;
 
-	ZVAL_DEREF(value);
 	if (Z_TYPE_P(value) != IS_OBJECT && Z_TYPE_P(value) != IS_ARRAY) {
 		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected \"%s\" %s to be array or object, %s given", zarr_key, zarr_key[0] == '$' ? "modifier" : "option", PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(value));
 		return false;
--- a/src/MongoDB/Query.c
+++ b/src/MongoDB/Query.c
@@ -58,6 +58,7 @@ static bool php_phongo_query_opts_append_document(bson_t* opts, const char* opts
 	zval*  value = php_array_fetch(zarr, zarr_key);
 	bson_t b     = BSON_INITIALIZER;
 
+	ZVAL_DEREF(value);
 	if (Z_TYPE_P(value) != IS_OBJECT && Z_TYPE_P(value) != IS_ARRAY) {
 		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected \"%s\" %s to be array or object, %s given", zarr_key, zarr_key[0] == '$' ? "modifier" : "option", PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(value));
 		return false;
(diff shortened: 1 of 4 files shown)


src/MongoDB/Query.c
```diff
@@ -58,6 +58,7 @@ static bool php_phongo_query_opts_append_document(bson_t* opts, const char* opts
 	zval*  value = php_array_fetch(zarr, zarr_key);
 	bson_t b     = BSON_INITIALIZER;
 
+	ZVAL_DEREF(value);
```
GromNaN: Should be done by `php_array_fetch`?

src/MongoDB/Query.c
```diff
@@ -333,6 +334,7 @@ bool phongo_query_init(zval* return_value, zval* filter, zval* options)
 	if (php_array_existsc(options, "modifiers")) {
 		modifiers = php_array_fetchc(options, "modifiers");
 
+		ZVAL_DEREF(modifiers);
```
GromNaN, line 336: Should be done by `php_array_fetchc`?

src/MongoDB/Query.c
GromNaN: Done.

src/contrib/php_array_api.h
```diff
@@ -188,7 +188,12 @@ zval *php_array_fetchl(zval *zarr, const char *key, int key_len) {
 }
 static inline
 zval *php_array_fetch(zval *zarr, const char *key) {
-	return php_array_fetchl(zarr, key, strlen(key));
+	zval *ret = php_array_fetchl(zarr, key, strlen(key));
+	if (ret) {
+		ZVAL_DEREF(ret);
+	}
+
+	return ret;
```
GromNaN, line 196: This is used to read options in `Query`, `BulkWrite` and `phongo_lient`.

tests/query/bug2457-001.phpt
```diff
@@ -0,0 +1,34 @@
+--TEST--
+PHPC-2457: Query modifiers can be passed reference
+--FILE--
+<?php
+
+$modifiers = ['$orderby' => ['x' => 1]];
```
jmikola, line 6: You noted the following error in [PHPC-2457](https://jira.mongodb.org/browse/PHPC-2457):

> PHP Fatal error: Uncaught MongoDB\Driver\Exception\InvalidArgumentException: Expected "$orderby" modifier to be array or object, bool given

The exception originates from [`php_phongo_query_opts_append_document()`](https://github.com/mongodb/mongo-php-driver/blob/1.20.0/src/MongoDB/Query.c#L62), but I'm not sure how that's possible as the code calls `PHONGO_ZVAL_CLASS_OR_TYPE_NAME_P(value)`. That macro delegates to [`zend_get_type_by_const()`](https://github.com/php/php-src/blob/php-8.3.12/Zend/zend_API.c#L110). `IS_REFERENCE` isn't even included in that function, and I can't imagine why the type would match for the function returning "bool".

src/contrib/php_array_api.h
jmikola: Note that there are a whole family of `php_array_fetch` functions, so even this change might be incomplete (in terms of fixing the general issue and not just the regression tests).

In [this Slack thread](https://mongodb.slack.com/archives/C026UA4GF0B/p1727248369999559), @alcaeus suggested a new method that conditionally applies `ZVAL_DEREF()`. That sounded reasonable, along with leaving this as-is and just adding our own `ZVAL_DEREF()` calls as needed.

As for the `php_array_zval_to_<type>` family of functions, I think those are easy candidates for automatic dereferencing, as @alcaeus did in previous PRs.

alcaeus: Created #1697 to update php_array_api and fix the deref issue in remaining places.

GromNaN:
Replaced by https://github.com/mongodb/mongo-php-driver/pull/1697

PR #1695, opened 2024-09-25
PHPC-2458: Deprecate float arg for UTCDateTime constructor

jmikola:
https://jira.mongodb.org/browse/PHPC-2458

I realized this was overlooked while going through [documentation-needed](https://jira.mongodb.org/issues/?jql=project%20%3D%20PHPC%20AND%20labels%20%3D%20documentation-needed%20ORDER%20BY%20fixVersion%20ASC) tickets for 1.20. Floats, like strings, were only supported as an alternative for 32-bit platforms.

We can discuss whether this should slip into a 1.20.1 patch release (in which case I'll re-target v1.20) or be deferred to 1.21.0. Either way, I'm going to preemptively document this as deprecated along with strings.

### DIFF
--- a/src/BSON/UTCDateTime.c
+++ b/src/BSON/UTCDateTime.c
@@ -218,6 +218,8 @@ static PHP_METHOD(MongoDB_BSON_UTCDateTime, __construct)
 			return;
 
 		case IS_DOUBLE:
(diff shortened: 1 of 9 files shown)


alcaeus (approved):
LGTM.

SemVer suggests that deprecations should be introduced in minor versions, so considering this and that we've already planned for a 1.21 release, I'm fine moving this to 1.21.

PR #1697, opened 2024-09-26
PHPC-2457: Fix using array items by reference

alcaeus:
PHPC-2457

Previous fixes for this were made in #1683 and #1223, but both attempts resulted in missed cases. #1694 fixes this for query modifiers and options, but a number of other occurrences were affected, as evidenced by the changes in this PR.

To fix this on a wider scale, I made a few adjustments to `php_array_api.h`, some of them unrelated:
* (unrelated) remove support for zend_engine 2 (PHP < 7)
* (unrelated) simplify some switch statements and fix casts
* add `php_array_fetch*_deref`
* change `php_array_fetch*_<type>` to always use `*_deref` version

After making this change, I audited all usages of `php_array_fetch*` and updated them accordingly. If references were correctly handled (such as in `phongo_zval_to_bson_value` which always calls `ZVAL_DEREF`), the call was not changed. Most cases immediately checked types after calling `php_array_fetch`, which was changed to `php_array_fetch_deref`.

In addition, I changed a number of usages of `php_array_fetch` to `php_array_fetchc` when a constant string was used.

I also included the tests added in #1694, but did not add tests for each and every usage that was changed. Let me know if you think those tests are necessary.

### DIFF
--- a/src/contrib/php_array_api.h
+++ b/src/contrib/php_array_api.h
@@ -187,17 +187,53 @@ zval *php_array_fetchz(zval *zarr, zval *key) {
 	}
 }
 
+static inline zval* zval_deref_safe(zval* z) {
+	if (z) { ZVAL_DEREF(z); }
+	return z;
+}
+
+/* Fetch zval*, resolving references
+ *
+ * zval *php_array_fetch_deref(zval *zarr, const char *key)
+ * zval *php_array_fetchl_deref(zval *zarr, const char *key, int key_len)
+ * zval *php_array_fetchl_safe_deref(zval *zarr, const char *key, int key_len)
+ * zval *php_array_fetchn_deref(zval *zarr, unsigned long idx)
+ * zval *php_array_fetchc_deref(zval *zarr, const char *litstr)
+ * zval *php_array_fetchz_deref(zval *zarr, zval *key)
+ */
+static inline
+	zval *php_array_fetchl_deref(zval *zarr, const char *key, int key_len) {
+	return zval_deref_safe(php_array_fetchl(zarr, key, key_len));
+}
+static inline
+	zval *php_array_fetch_deref(zval *zarr, const char *key) {
+	return zval_deref_safe(php_array_fetch(zarr, key));
+}
+#define php_array_fetchc_deref(zarr, litstr) zval_deref_safe(php_array_fetchl(zarr, litstr, sizeof(litstr)-1))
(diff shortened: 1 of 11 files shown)


src/contrib/php_array_api.h
```diff
@@ -193,65 +159,81 @@ zval *php_array_fetch(zval *zarr, const char *key) {
...
+ * zval *php_array_fetchl_deref(zval *zarr, const char *key, int key_len)
+ * zval *php_array_fetchl_safe_deref(zval *zarr, const char *key, int key_len)
+ * zval *php_array_fetchn_deref(zval *zarr, unsigned long idx)
+ * zval *php_array_fetchc_deref(zval *zarr, const char *litstr)
+ * zval *php_array_fetchz_deref(zval *zarr, zval *key)
+ */
+static inline
+	zval *php_array_fetchl_deref(zval *zarr, const char *key, int key_len) {
+	return zval_deref_safe(php_array_fetchl(zarr, key, key_len));
+}
```
jmikola, line 207: I assume this is `clang-format`, but I find this formatting very bizarre.

tests/query/bug2457-002.phpt
```diff
@@ -0,0 +1,49 @@
+--TEST--
+PHPC-2457: Query options can be passed reference
```
jmikola: I'd suggest swapping the names on these two tests, so once `modifiers` is removed in 2.0 we can delete its test and just be left with a `bug2457-001.phpt` file.

jmikola (approved):
Suggested test file rename and a code formatting complaint, but content LGTM.

src/contrib/php_array_api.h
alcaeus: It's copy-paste from a previous occurrence, but I'll fix it

tests/query/bug2457-002.phpt
```diff
@@ -1,45 +1,30 @@
...
-      string(5) "fr_FR"
-      ["strength"]=>
-      int(2)
-    }
-    ["let"]=>
-    object(stdClass)#4 (1) {
-      ["x"]=>
-      int(1)
-    }
+  object(stdClass)#4 (1) {
```
jmikola, line 24: In the future, note that you can use `%d` for the object identifiers here, since those don't need to be asserted (and could change if we did internal refactoring).

PR #1703, opened 2024-09-27
PHPC-2350: Test that PackedArray::fromJSON() accepts objects with valid keys

jmikola:
https://jira.mongodb.org/browse/PHPC-2350

This came to me while writing PHP docs and I realized it was never tested in the original PR (https://github.com/mongodb/mongo-php-driver/pull/1645). I don't think this warrants a separate ticket.


tests/bson/bson-packedarray-fromJSON-001.phpt
```diff
@@ -10,6 +10,8 @@ $tests = [
     '[ 1, 2, 3 ]',
     '[[ 1, 2, 3 ]]',
     '[{ "bar": 1 }]',
+    // JSON objects are accepted iff the keys are valid
```
alcaeus, line 13: I was gonna nitpick the typo, but TIL that [iff is an actual word](https://en.wiktionary.org/wiki/iff).

alcaeus (approved):

tests/bson/bson-packedarray-fromJSON-001.phpt
jmikola: Yes, I don't like how easily this can be mixed up for a typo but it'd be a complete waste of space to write "if and only if"

PR #1717, opened 2024-10-10
Include the compiler variant in Windows release file names for PHP 8.3

mickverm:
Fixes https://github.com/mongodb/mongo-php-driver/issues/1716

PR #1718, opened 2024-10-11
Include the compiler variant in Windows release file names for PHP 8.3

jmikola:
Rebasing https://github.com/mongodb/mongo-php-driver/pull/1717 on v1.20.


GromNaN (approved):

alcaeus (approved):
LGTM. I created PHPC-2469 to switch our release actions to use the output variable from setup-php-sdk to avoid having to update the list for every new PHP version.

mickverm:
@alcaeus please take the following into account when updating the release actions: https://github.com/php/pie/blob/main/docs/extension-maintainers.md#windows-support

I figured out that the package should be named `x86_64` instead of `x64` and that the name of the `.dll` also doesn't follow their spec.

alcaeus:
Thanks for pointing this out. I've created #1732 to use the setup-php-sdk output and change the filename of the `.dll` and `.zip` files.

PR #1732, opened 2024-10-17
PHPC-2649: Use output from setup-php-sdk in packaging step

alcaeus:
PHPC-2649

This PR adds the `vs` output variable from the `setup-php-sdk` action to all actions involved in building Windows packages so that it can be used when determining the filename for the archive. This removes the need for maintaining a list of PHP versions with their compilers.

In addition to this, the PR makes two more adjustments for compatibility with PIE:
* The archive name should contain `x86_64` for 64-bit platforms, not `x64`
* The built DLL should be named like the archive instead of the default `php_mongodb.dll`. PIE will rename the file in its install step.

cc @mickverm - this should get rid of the issues you encountered, but we won't know until we release 1.20.1. Please bear with us while we sort out the initial difficulties with PIE. We do appreciate your testing efforts for PIE compatibility 💪

### DIFF
--- a/.github/workflows/build-windows-package.yml
+++ b/.github/workflows/build-windows-package.yml
@@ -38,6 +38,8 @@ jobs:
     defaults:
       run:
         shell: cmd
+    outputs:
+      vs: ${{ steps.build-driver.outputs.vs }}
 
     steps:
       - uses: actions/checkout@v4
@@ -114,27 +116,9 @@ jobs:
       - name: "Copy signature file"
         run: cp ${RELEASE_ASSETS}/php_mongodb.dll.sig .
 
-      - name: "Set compiler environment variable"
-        run: |
-          case "$PHP_VERSION" in
-            "7.4")
-              COMPILER="vc15"
-              ;;
-            "8.0" | "8.1" | "8.2" | "8.3")
-              COMPILER="vs16"
-              ;;
-            "8.4")
-              COMPILER="vs17"
-              ;;
-          esac
-          echo "COMPILER=${COMPILER}" >> "$GITHUB_ENV"
-        shell: bash
-        env:
-          PHP_VERSION: ${{ inputs.php }}
-
       - name: "Create and upload release asset"
         if: ${{ inputs.upload_release_asset }}
         run: |
-          ARCHIVE=php_mongodb-${{ inputs.version }}-${{ inputs.php }}-${{ inputs.ts }}-${{ env.COMPILER }}-${{ inputs.arch }}.zip
+          ARCHIVE=php_mongodb-${{ inputs.version }}-${{ inputs.php }}-${{ inputs.ts }}-${{ needs.build.outputs.vs }}-${{ inputs.arch }}.zip
           zip ${ARCHIVE} php_mongodb.dll php_mongodb.dll.sig php_mongodb.pdb CREDITS CONTRIBUTING.md LICENSE README.md THIRD_PARTY_NOTICES
           gh release upload ${{ inputs.version }} ${ARCHIVE}
--- a/.github/workflows/build-windows-package.yml
+++ b/.github/workflows/build-windows-package.yml
@@ -119,6 +119,6 @@ jobs:
       - name: "Create and upload release asset"
         if: ${{ inputs.upload_release_asset }}
         run: |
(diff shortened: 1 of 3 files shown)


mickverm:
I'll test it out as soon as a new version is released, thank you for keeping me up to date. 😄

.github/workflows/build-windows-package.yml
```diff
@@ -110,31 +112,22 @@ jobs:
...
+          mv php_mongodb.dll ${{ env.FILENAME }}.dll
+          cp ${RELEASE_ASSETS}/php_mongodb.dll.sig ${{ env.FILENAME }}.dll.sig
 
       - name: "Create and upload release asset"
         if: ${{ inputs.upload_release_asset }}
         run: |
-          ARCHIVE=php_mongodb-${{ inputs.version }}-${{ inputs.php }}-${{ inputs.ts }}-${{ env.COMPILER }}-${{ inputs.arch }}.zip
-          zip ${ARCHIVE} php_mongodb.dll php_mongodb.dll.sig php_mongodb.pdb CREDITS CONTRIBUTING.md LICENSE README.md THIRD_PARTY_NOTICES
+          ARCHIVE=${{ env.FILENAME }}.zip
+          zip ${ARCHIVE} ${{ env.FILENAME }}.dll ${{ env.FILENAME }}.dll.sig php_mongodb.pdb CREDITS CONTRIBUTING.md LICENSE README.md THIRD_PARTY_NOTICES
```
mickverm, line 132: If I'm not mistaken, `php_mongodb.pdb` needs to become `${{ env.FILENAME }}.pdb`.

`php_mongodb-1.20.1-8.3-ts-vs16-x86_64.dll` would be converted to `php_mongodb-1.20.1-8.3-ts-vs16-x86_64.pdb` in https://github.com/php/pie/blob/eaf77ea181f44c7711887474c4f33501a17cb39f/src/Installing/WindowsInstall.php#L39

alcaeus: Hmm, according to the [maintainer docs](https://github.com/php/pie/blob/main/docs/extension-maintainers.md#contents-of-the-windows-zip), it doesn't:

> `php_{extension-name}.pdb` - this will be moved alongside the `C:\path\to\php\ext\php_{extension-name}.dll`

Looking at the code you linked, I do agree that it should be named the same way as the DLL file.

Let me try summoning @asgrim to clarify 😉 

---

Another concern that comes to mind is that we're now requiring people that install their extensions manually. Right now, they only need to copy `php_mongodb.dll` to their extensions directory and they're done. This is valid regardless of whether they download our packages or the ones offered on pecl.php.net.

I'm sure there was a particular reason for naming the DLL file this way, but I wonder if that's truly necessary. Clearly, we can extract the relevant information (compiler, thread safety, PHP version, architecture) from the archive filename before we extract the DLL file, so we shouldn't need to re-verify this when we have the DLL file. If a user packs a DLL file into an archive for a different platform, that's their problem, not PIE's.

asgrim: Regarding the first part, yes the implementation is incorrect, it was always meant to be `php_{extension-name}.pdb` - I've reported issue php/pie#71 and I will look into getting this fixed.

The second part; the naming was agreed since it aligns with what [php/php-windows-builder](https://github.com/php/php-windows-builder) action produces

.github/workflows/build-windows-package.yml
```diff
@@ -110,31 +112,22 @@ jobs:
         with:
           filenames: php_mongodb.dll
 
-      # Copy the signature file from the release asset directory to avoid directory issues in the ZIP file
-      - name: "Copy signature file"
-        run: cp ${RELEASE_ASSETS}/php_mongodb.dll.sig .
+      - name: "Generate file name for DLL and archive"
+        run:
+          echo FILENAME="php_mongodb-${{ inputs.version }}-${{ inputs.php }}-${{ inputs.ts }}-${{ needs.build.outputs.vs }}-${{ inputs.arch == 'x64' && 'x64_86' || inputs.arch }}" >> "$GITHUB_ENV"
```
jmikola, line 117: Noted that this uses the [`needs` context](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context).

jmikola (approved):
LGTM assuming `php_mongodb.pdb` is using the correct name.

PR #1745, opened 2024-11-14
Use correct name for PDB file for pie installs.

mickverm:
Follow up of https://github.com/mongodb/mongo-php-driver/pull/1732#discussion_r1804635245 and https://github.com/php/pie/pull/85


alcaeus (approved):

alcaeus:
Thank you for providing the patch!

PR #1748, opened 2024-11-15
PHPC-2473: Bump to libmongoc 1.29.0

jmikola:
https://jira.mongodb.org/browse/PHPC-2473
https://jira.mongodb.org/browse/PHPC-2475
https://jira.mongodb.org/browse/PHPC-2470


src/phongo_atomic.c
```diff
@@ -0,0 +1,119 @@
+/*
```
alcaeus, line 1: I suppose this and the corresponding header file were created from bson_atomic, with symbols renamed to have a `phongo_` prefix. As such, I trust our tests to ensure that this works properly and am not reviewing the logic contained. Let me know if my assumption is wrong.

alcaeus (approved):
LGTM!

src/phongo_atomic.c
jmikola: Correct. I reduced the file as much as possible to just keep the APIs necessary from our use case. So we've only got `int32_t` and the `int8_t` logic that requires. I did leave in the extra functions (e.g. "exchange") since they were all defined in the same block.

Issue #1750, opened 2024-11-19
Wrong PIE extension name


macintoshplus:
In the file `composer.json` the extension name starts with `ext-`.

https://github.com/mongodb/mongo-php-driver/blob/v1.x/composer.json#L19

According to the php/pie [documentation](https://github.com/php/pie/blob/main/docs/extension-maintainers.md#extension-name), the extension name should not be prefixed with `ext-`.

alcaeus:
We have released 1.20.1 which brings the correct extension name as well as some other fixes for PIE support.

PR #1751, opened 2024-11-19
Remove prefix ext- from extension name.

macintoshplus:
See #1750.

Remove the wrong extension name prefix into `composer.json` configuration used by php/pie.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "ext-json": "*"
   },
   "php-ext": {
-    "extension-name": "ext-mongodb",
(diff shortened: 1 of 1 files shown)


alcaeus:
Ah, I forgot that the guidance was updated to exclude the `ext-` prefix. Do you mind re-targeting v1.20 for this change? You can do so by resetting your branch to v1.20, then cherry-picking the above commit onto it and force-pushing to this PR.

Also, if you'd like to apply this contribution to more extensions, I saw that xdebug also includes the `ext-` prefix in its name.

PR #1763, opened 2024-12-10
PHPC-2478: Implement API for bulkWrite command 

jmikola:
https://jira.mongodb.org/browse/PHPC-2478


src/MongoDB/BulkWriteCommand.stub.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace MongoDB\Driver;
+
+/** @not-serializable */
+final class BulkWriteCommand implements \Countable
+{
+    final public function __construct(?array $options = null) {}
+
+    // TODO: decide if Countable is still desirable
+    public function count(): int {}
```
GromNaN: I don't see the use-case for counting. Is it the number of operations it contains? Maybe only to know if it's empty.

src/MongoDB/BulkWriteCommand.stub.php
```diff
@@ -0,0 +1,31 @@
...
+
+    final public function insertOne(string $namespace, array|object $document): mixed {}
+
+    final public function replaceOne(string $namespace, array|object $filter, array|object $replacement, ?array $options = null): void {}
+
+    final public function updateOne(string $namespace, array|object $filter, array|object $update, ?array $options = null): void {}
+
+    final public function updateMany(string $namespace, array|object $filter, array|object $update, ?array $options = null): void {}
+
+    final public function execute(Manager|Server $managerOrServer): mixed {}
```
GromNaN: By placing the execute method on this side, it becomes essential to be able to mock this class. But this class is `final`.

How do you test a function that creates a BulkWriteCommand and executes it?

```php
function saveDocuments(Manager $manager, array $documents) {
    $command = new BulkWriteCommand();
    foreach ($documents as $document) {
        $command->replaceOne(/** do something very domain specific */);
    }
    // I have no idea how to track the call to $manager
    $command->execute($manager);

    // This implementation would be easier to mock
    $manager->execute($command);
}
```

src/MongoDB/BulkWriteCommand.stub.php
jmikola: It would be the number of appended operations. I'm not convinced this is necessary, and it would be trivial to add at a later point if some user requested it (presumably because they relied on it on the original BulkWrite class).

src/MongoDB/BulkWriteCommand.stub.php
jmikola: Thanks for the feedback here. I clearly don't deal with mocking objects anymore, so I overlooked how this would hinder that. I suppose that takes us back to `Manager::executeBulkWriteCommand()`, and a corresponding Server method.

jmikola: Revisiting this since we discussed it earlier in the week with @alcaeus. In terms of mocking, having `execute()` here isn't much different than the Manager and Server classes, which are also both `final`.

We also discussed having `execute()` methods oh other objects, and I believe only Command would be problematic since that currently relies on several variations (read, read/write, write) and to replace that with a singular `execute()` method would require PHPC to inspect command documents.

Anyway, I'm going to proceed with `BulkWriteCommand::execute()` for now and we can revisit later if needed.

jmikola: Resolving this, as I fell back to `Manager::execute()` and `Server::execute()`.

src/MongoDB/BulkWriteCommand.c
```diff
@@ -0,0 +1,853 @@
...
+	if (bson_empty(array)) {
+		return true;
+	}
+
+	if (bson_iter_init(&iter, array)) {
+		char key[12];
+		int  count = 0;
+
+		while (bson_iter_next(&iter)) {
+			bson_snprintf(key, sizeof(key), "%d", count);
```
alcaeus, line 108: I've probably thought about this before, but I'm wondering if there's a performance difference between using `snprintf` and `atoll`, i.e. converting the expected int to string and doing string comparison, vs. parsing the string key and comparing it vs. the expected int.

Not sure if we're able to use PHP's logic for dealing with packed arrays for this, but I agree with unifying this across the code base.

src/MongoDB/BulkWriteCommand.c
```diff
@@ -0,0 +1,853 @@
...
+/* Extracts the "_id" field of a BSON document into a return value. */
+static void phongo_bwc_extract_id(bson_t* doc, zval** return_value)
+{
+	zval*                 id = NULL;
+	php_phongo_bson_state state;
+
+	PHONGO_BSON_INIT_STATE(state);
+	state.map.root.type = PHONGO_TYPEMAP_NATIVE_ARRAY;
+
+	/* TODO: Instead of converting the entire document, iterate BSON to obtain
```
alcaeus, line 75: Yep, using `bson_iter_init_find` and `bson_iter_value` is probably significantly faster than walking the entire BSON structure 👍

src/MongoDB/BulkWriteCommand.c
```diff
@@ -0,0 +1,853 @@
...
+	intern->num_ops = 0;
+	intern->ordered = true;
+	intern->verbose = false;
+	intern->write_concern = NULL;
+
+	if (!zoptions) {
+		return;
+	}
+
+	if (php_array_existsc(zoptions, "bypassDocumentValidation")) {
```
alcaeus, line 148: We do have some macros that we use for queries - would it make sense to make these macros universal? I'm thinking of something like this:

```
PHONGO_EXTRACT_OPTION_BOOL(zoptions, "bypassDocumentValidation", intern->bypass)
```

If so, I can create a ticket for this.

src/MongoDB/BulkWriteCommand.c
```diff
@@ -0,0 +1,853 @@
...
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected \"let\" option to be array or object, %s given", zend_get_type_by_const(Z_TYPE_P(value)));
+			return;
+		}
+
+		intern->let = bson_new();
+		php_phongo_zval_to_bson(value, PHONGO_BSON_NONE, intern->let, NULL);
+
+		if (EG(exception)) {
+			return;
+		}
```
alcaeus, line 176: I only now realised that we can't easily check the return value of `php_phongo_zval_to_bson` to figure out if something went wrong and instead have to check `EG(exception)`. This is something we may want to refactor in the future.

src/MongoDB/BulkWriteCommand.c
```diff
@@ -0,0 +1,853 @@
...
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	RETURN_LONG(intern->num_ops);
+}
+
+static bool phongo_bwc_parse_hint(zval* zhint, bson_value_t* bhint)
+{
+	if (Z_TYPE_P(zhint) != IS_STRING && Z_TYPE_P(zhint) != IS_OBJECT && Z_TYPE_P(zhint) != IS_ARRAY) {
+		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected \"hint\" option to be string, array, or object, %s given", zend_get_type_by_const(Z_TYPE_P(zhint)));
```
alcaeus, line 213: Nothing to change in this PR, but we may want to revisit these exceptions to communicate the "document" type as an "object or hash" to avoid the confusing case that it needs to be an array but may not be a packed array.

src/MongoDB/BulkWriteCommand.c
```diff
@@ -0,0 +1,853 @@
...
+
+			if (!phongo_bwc_parse_document(php_array_fetchc_deref(zoptions, "collation"), &bcollation, "collation")) {
+				bson_destroy(&bcollation);
+				goto cleanup;
+			}
+
+			mongoc_bulkwrite_deletemanyopts_set_collation(opts, &bcollation);
+			bson_destroy(&bcollation);
+		}
+	}
```
alcaeus, line 327: Consider extracting this to a static function here and in other methods that populate options for bulkwrite operations. Some of the options are also shared between the structs, so we may be able to further unify them, e.g. by having a function that extracts a `bson_value_t*` for a valid hint or `NULL` if no hint was given while returning `false` if the hint could not be parsed.

src/MongoDB/BulkWriteCommand.c
```diff
@@ -0,0 +1,853 @@
...
+		phongo_throw_exception_from_bson_error_t(&error);
+		goto cleanup;
+	}
+
+	intern->num_ops++;
+
+	if (!bson_out) {
+		phongo_throw_exception(PHONGO_ERROR_LOGIC, "php_phongo_zval_to_bson() did not return document identifier. Please file a bug report.");
+		goto cleanup;
+	}
```
alcaeus: Should we throw this directly after calling `php_phongo_zval_to_bson` instead of deferring after adding the operation to the bulkwrite? As is, this can lead to strange behaviour where the exception can be caught and the document will be inserted (ultimately with an id assigned by the server) despite the exception.

src/MongoDB/BulkWriteCommand.c
```diff
@@ -0,0 +1,853 @@
...
+	if (intern->comment) {
+		zval zv;
+
+		if (!phongo_bson_value_to_zval_legacy(intern->comment, &zv)) {
+			zval_ptr_dtor(&zv);
+			goto done;
+		}
+
+		ADD_ASSOC_ZVAL_EX(&retval, "comment", &zv);
+	}
```
alcaeus, line 808: Do we want to add `NULL` for comment in this case to always return the same fields?

src/MongoDB/BulkWriteCommandResult.c
```diff
@@ -0,0 +1,482 @@
...
+{
+	bson_iter_t iter;
+
+	array_init(return_value);
+
+	if (!intern->write_concern_errors) {
+		return true;
+	}
+
+	for (bson_iter_init(&iter, intern->write_concern_errors); bson_iter_next(&iter);) {
```
alcaeus: Not sure if it matters here, but I'll note that we don't throw an exception for an invalid iterator here. That may be irrelevant as the bson may always be valid, but I wanted to make sure this was a conscious choice.

src/MongoDB/BulkWriteCommandResult.c
```diff
@@ -0,0 +1,482 @@
...
+	php_phongo_bulkwritecommandresult_t* intern;
+
+	intern = Z_BULKWRITECOMMANDRESULT_OBJ_P(getThis());
+
+	PHONGO_PARSE_PARAMETERS_NONE();
+
+	PHONGO_BULKWRITECOMMANDRESULT_CHECK_ACKNOWLEDGED("getServer");
+
+	// TODO: null handling
+	phongo_server_init(return_value, &intern->manager, intern->server_id);
```
alcaeus: I remember doing away with Server instances in APM events - is that something we want to do for exceptions as well?

src/MongoDB/BulkWriteCommandResult.c
```diff
@@ -0,0 +1,482 @@
...
+static zend_object_handlers php_phongo_handler_bulkwritecommandresult;
+
+static void php_phongo_bulkwritecommandresult_free_object(zend_object* object)
+{
+	php_phongo_bulkwritecommandresult_t* intern = Z_OBJ_BULKWRITECOMMANDRESULT(object);
+
+	zend_object_std_dtor(&intern->std);
+
+	if (intern->insert_results) {
+		bson_destroy(intern->insert_results);
```
alcaeus: Since `bson_destroy` is a no-op if the `bson_t*` is `NULL`, you don't need the conditionals here.

src/MongoDB/BulkWriteCommandResult.c
```diff
@@ -0,0 +1,482 @@
...
+	ADD_ASSOC_LONG_EX(&retval, "upsertedCount", intern->upserted_count);
+	ADD_ASSOC_LONG_EX(&retval, "deletedCount", intern->deleted_count);
+
+	if (intern->insert_results) {
+		zval insert_results;
+
+		phongo_document_new(&insert_results, intern->insert_results, true);
+		ADD_ASSOC_ZVAL_EX(&retval, "insertResults", &insert_results);
+	} else {
+		ADD_ASSOC_NULL_EX(&retval, "indexResults");
```
alcaeus: Typo: should be insertResults

src/MongoDB/Exception/BulkWriteCommandException.stub.php
```diff
@@ -0,0 +1,15 @@
...
+ * @generate-function-entries static
+ */
+
+namespace MongoDB\Driver\Exception;
+
+class BulkWriteCommandException extends ServerException
+{
+    protected \MongoDB\Driver\BulkWriteCommandResult $bulkWriteCommandResult;
+
+    final public function getBulkWriteCommandResult(): \MongoDB\Driver\BulkWriteCommandResult {}
```
alcaeus, line 14: You could use a public readonly property to do away with the public getter.

src/MongoDB/WriteConcernError.c
```diff
@@ -147,14 +147,18 @@ zend_bool phongo_writeconcernerror_init(zval* return_value, bson_t* bson)
 		intern->code = bson_iter_int32(&iter);
 	}
 
-	if (bson_iter_init_find(&iter, bson, "errmsg") && BSON_ITER_HOLDS_UTF8(&iter)) {
+	// Additionally check for field name used by mongoc_bulkwriteexception_t
```
alcaeus, line 156: Oof. Yay, I guess?

alcaeus (commented):
Initial feedback on the code - I haven't yet gone through all the tests.

src/MongoDB/BulkWriteCommandResult.c
jmikola: This pertains to the BulkWriteCommandResult, which applies to both the successful return _and_ the exception. I originally added this field for consistency with [WriteResult](https://www.php.net/manual/en/class.mongodb-driver-writeresult.php) (and will note that we also report the executing server for cursors); however, this isn't actually required by the [Bulk Write spec](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md) so we can remove it if you prefer.

Doing so would mean we could ignore the issue with libmongoc not always populating this (see: [CDRIVER-5843](https://jira.mongodb.org/browse/CDRIVER-5843)).

src/MongoDB/BulkWriteCommand.c
jmikola: In the case of `comment`, reporting null here might be ambiguous with a BSON null value. I opted to also omit `let` when it's unset, although that wouldn't be ambiguous since it can only be a document. The treatment of both fields is consistent with what we did in [BulkWrite.c](https://github.com/mongodb/mongo-php-driver/blob/1.20.1/src/MongoDB/BulkWrite.c#L621).

src/MongoDB/BulkWriteCommandResult.c
jmikola: This was intentional as I don't expect invalid BSON from libmongoc, but it's technically problematic since the `for` loop could proceed with calling `bson_iter_next()` after initialization failed.

Replacing this with a conditional on `bson_iter_init()` will be consistent with WriteResult.c. I'd prefer to just NOP on a false return value instead of dealing with exceptions.

On the flip side, there are some subtle changes I can make to WriteResult.c as well. If `phongo_writeerror_init()` or `phongo_writeconcernerror_init` ever fail, we should return `false` immediately instead of continuing with iteration (and risk throwing multiple exceptions).

src/MongoDB/BulkWriteCommand.c
jmikola: This is the same approach used in libbson, so I'm inclined to keep it as-is. Also note that we're re-using a stack-allocated string, so there isn't much memory allocation happening.

> Not sure if we're able to use PHP's logic for dealing with packed arrays for this, but I agree with unifying this across the code base.

We could use [`zend_array_is_list()`](https://heap.space/s?refs=zend_array_is_list&project=PHP-8.3) to avoid _some_ calls to `phongo_bwc_bson_array_has_valid_keys()`, but we'd still need our own check after running a non-array value through `php_phongo_zval_to_bson()`. Assuming that's faster than checking the BSON, it still seems like a micro-optimization given how seldom we'd hit this code path ("arrayFilters" option parsing).

alcaeus: Sounds good - feel free to resolve after reading.

src/MongoDB/BulkWriteCommand.c
alcaeus: Ah, I see. While it can be confusing, I'm fine leaving this as-is for the sake of consistency.

src/MongoDB/BulkWriteCommandResult.c
alcaeus: Replacing this with a NOP in case of failing to initialise the iterator sounds fine to me, as well as improving handling in WriteResult.c.

src/MongoDB/BulkWriteCommandResult.c
alcaeus: I'm fine not having this, especially if it works around a libmongoc issue.

src/MongoDB/BulkWriteCommand.c
jmikola: Created [PHPC-2511](https://jira.mongodb.org/browse/PHPC-2511) to track this.

src/MongoDB/BulkWriteCommand.c
jmikola: Created [PHPC-2512](https://jira.mongodb.org/browse/PHPC-2512) to track this.

src/MongoDB/BulkWriteCommandResult.c
jmikola: Added a commit to remove this. We can always add it later once libmongoc addresses its inconsistency.

src/MongoDB/Exception/BulkWriteCommandException.stub.php
jmikola: Going to disregard this since we don't have prior art for doing so in PHPC.

src/MongoDB/BulkWriteCommand.c
jmikola: I concur. Also opened [PHPC-2513](https://jira.mongodb.org/browse/PHPC-2513) to address the original code in `BulkWrite::insert()`.

src/MongoDB/BulkWriteCommand.c
jmikola: I don't think there is sufficient overlap with the macros in `Query.c` to make that worthwhile. In Query, the parsed options go directly to a `bson_t`. Here, they're stashed on the object struct. For BulkWriteCommand's operation methods, they're going to be assigned via distinct setter methods (similar to most of the option handling in `ClientEncryption.c`).

src/MongoDB/BulkWriteCommand.c
jmikola: `phongo_bwc_parse_document()` was the attempt to extract common logic here. Unfortunately, each operation has its own assortment of setter methods that are going to require distinct code.

`phongo_bwc_parse_hint()` was the attempt to extract common handling for the `hint` option.

src/MongoDB/BulkWriteCommand.c
jmikola: Created [PHPC-2516](https://jira.mongodb.org/browse/PHPC-2516) to track this.

jmikola:
Superseded by https://github.com/mongodb/mongo-php-driver/pull/1790, which is rebased on v2.x.

PR #1781, opened 2025-02-04
PHPC:2498: Accept integer types for Document array access

jmikola:
https://jira.mongodb.org/browse/PHPC-2498

### DIFF
--- a/src/BSON/Document.c
+++ b/src/BSON/Document.c
@@ -192,6 +192,36 @@ static bool php_phongo_document_get(php_phongo_document_t* intern, char* key, si
 	return true;
 }
 
+static bool php_phongo_document_get_by_zval(php_phongo_document_t* intern, zval* key, zval* return_value, bool null_if_missing)
+{
+	if (Z_TYPE_P(key) != IS_STRING && Z_TYPE_P(key) != IS_LONG) {
+		if (null_if_missing) {
+			ZVAL_NULL(return_value);
+			return true;
+		}
+
+		phongo_throw_exception(PHONGO_ERROR_RUNTIME, "Could not find key of type \"%s\" in BSON document", zend_zval_type_name(key));
+		return false;
+	}
+
+	zend_string* tmp_str;
+	zend_string* str = zval_try_get_tmp_string(key, &tmp_str);
+
+	if (!str) {
+		// Exception already thrown
+		return false;
+	}
+
+	if (!php_phongo_document_get(intern, ZSTR_VAL(str), ZSTR_LEN(str), return_value, null_if_missing)) {
+		// Exception already thrown
+		zend_tmp_string_release(tmp_str);
+		return false;
+	}
+
+	zend_tmp_string_release(tmp_str);
+	return true;
+}
+
 static PHP_METHOD(MongoDB_BSON_Document, get)
 {
 	php_phongo_document_t* intern;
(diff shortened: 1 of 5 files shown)


src/BSON/Document.c
```diff
@@ -204,10 +234,8 @@ static PHP_METHOD(MongoDB_BSON_Document, get)
 
 	intern = Z_DOCUMENT_OBJ_P(getThis());
 
-	if (!php_phongo_document_get(intern, key, key_len, return_value, false)) {
-		// Exception already thrown
-		RETURN_NULL();
```
jmikola, line 209: This was redundant, as `null` is already the default value for the `return_value` zval.

src/BSON/Document.c
```diff
@@ -229,6 +257,30 @@ static bool php_phongo_document_has(php_phongo_document_t* intern, char* key, si
...
 }
 
+static bool php_phongo_document_has_by_zval(php_phongo_document_t* intern, zval* key)
+{
+	if (Z_TYPE_P(key) != IS_STRING && Z_TYPE_P(key) != IS_LONG) {
+		return false;
+	}
+
+	zend_string* tmp_str;
+	zend_string* str = zval_try_get_tmp_string(key, &tmp_str);
```
jmikola, line 267: C99 allows declarations anywhere, but if there is a concern I can position these at the top of the function for consistency (granted, I don't think we have anything to enforce that moving forward since it's no longer a build error).

src/BSON/Document.c
```diff
@@ -192,6 +192,36 @@ static bool php_phongo_document_get(php_phongo_document_t* intern, char* key, si
 	return true;
 }
 
+static bool php_phongo_document_get_by_zval(php_phongo_document_t* intern, zval* key, zval* return_value, bool null_if_missing)
```
jmikola, line 195: I created two `by_zval` variants for `get` and `has`, which are used by the ArrayAccess methods and dimension handlers. These both abstraction the `zval_try_get_tmp_string` logic, which was necessary to avoid possibly modifying the zval argument with a basic cast (see: [Casts and Operations](https://www.phpinternalsbook.com/php7/zvals/casts_and_operations.html)).

I also added a test for this using an assigned variable instead of a literal to assert that we don't modify anything inadvertently.

alcaeus (approved):
LGTM. Thanks for the refactoring, definitely looks cleaner this way!

tests/bson/bson-document-array-access-005.phpt
```diff
@@ -0,0 +1,35 @@
...
+
+require_once __DIR__ . '/../utils/basic.inc';
+
+$document = MongoDB\BSON\Document::fromPHP([
+    '0' => 'foo',
+    '1' => 'bar',
+]);
+
+// Use a variable to assert that conversion doesn't affect the original zval
+$key = 1;
```
GromNaN, line 14: Interesting, how the variable could have been updated?

alcaeus: Since we are operating directly on a `zval`, calling `ZVAL_NULL(key)` would update the variable, even when it wasn't passed by reference. There's no good reason to do this on purpose (despite it being used in the legacy driver), but it can happen by accident. I believe the last occurrence of us wrongly messing with the key when we were inadvertently freeing interned strings that occurred in the type map. We've gotten more careful since.

jmikola: [`Manager::__construct()`](https://github.com/mongodb/mongo-php-driver/blob/1.20.1/src/MongoDB/Manager.c#L256) is another good example of this, where we need to separate the zvals to ensure any conversions don't inadvertently change the input arguments.

This is also discussed in the [Casts and Operations](https://www.phpinternalsbook.com/php7/zvals/casts_and_operations.html) article I linked in https://github.com/mongodb/mongo-php-driver/pull/1781#discussion_r1941954106

PR #1782, opened 2025-02-04
Use pre-packaged-source for pie downloads

asgrim:
We worked on php/pie#39 to allow the `mongodb` extension to be built from pre-packaged source URLs instead of relying on the git-generated ZIP archive that GitHub etc. uses (and thus Composer/PIE).

Note: `build-path` is also needed, and I added a template for it, since the TGZ produced has a subdirectory in it based on the version; so in the case of `mongodb` at least, we'll need both `download-url-method` and `build-path` to be specified.

Note 2: needs php/pie#187 to be merged first

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,8 @@
   },
   "php-ext": {
     "extension-name": "mongodb",
+    "download-url-method": "pre-packaged-source",
+    "build-path": "mongodb-{version}",
     "configure-options": [
       {
         "name": "enable-mongodb-developer-flags",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -17,6 +17,8 @@
   },
   "php-ext": {
     "extension-name": "mongodb",
+    "override-download-url-method": "pre-packaged-source",
```
asgrim: Needs php/pie#187 to be merged before this will work

alcaeus (approved):

composer.json
asgrim: Above is merged, release will be very soon, just finishing off the uninstall feature in PIE and will make a next release :)

PR #1791, opened 2025-02-19
Fix openssl detection on Silicon Macs

alcaeus:
Targeting 1.21 for this change, as it could lead to a different SSL library being picked, which I wouldn't expect in a patch release. This is consistent with #1463 which was the original attempt to prefer openssl over Secure Transport and also targeted a minor version to avoid unexpected issues on a patch update.

On Silicon Macs, the paths for homebrew have changed, so if the `PKG_CHECK_MODULES` macro does not find openssl on MacOS and no `OPENSSL_DIR` option has been provided, it usually won't be found. As a result, Secure Transport will be picked in those circumstances.

On Silicon platforms, homebrew installs openssl into `/opt/homebrew/opt/openssl`, so adding that path to `PHP_OPENSSL_DIR` fixes the issue.

To make debugging easier, I also added the selected SSL library to the configure output.

NB: I checked whether pkg-config is available on my system, and it isn't by default. A quick search through homebrew also didn't reveal it, so my assumption is that pkg-config won't be available on the typical developer machine, hence this workaround.

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -536,6 +536,7 @@ Build configuration:
   libmongoc                                        : $PHP_MONGODB_BSON_VERSION_STRING
   libbson                                          : $PHP_MONGODB_MONGOC_VERSION_STRING
   libmongocrypt                                    : $PHP_MONGODB_MONGOCRYPT_VERSION_STRING
+  SSL                                              : $PHP_MONGODB_SSL
   LDFLAGS                                          : $LDFLAGS
   EXTRA_LDFLAGS                                    : $EXTRA_LDFLAGS
   MONGODB_SHARED_LIBADD                            : $MONGODB_SHARED_LIBADD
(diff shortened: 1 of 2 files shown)


jmikola:
[pkgconf](https://formulae.brew.sh/formula/pkgconf) is the Homebrew package for `pkg-config`. We already note that `pkg-config` is a requirement in the [PECL install docs](https://www.php.net/manual/en/mongodb.installation.php), but we could certainly repeat this for Homebrew and link to the specific package.

I'm not sure how the PHP formulae detect OpenSSL without `pkg-config`, since PHP itself has used it exclusively since 7.4 (see: [php/php-src@fe8fdfa3](https://github.com/php/php-src/commit/fe8fdfa3bd588d80ce60f6b3848058239e0a760f)). That was the point where PHP removed all of the legacy path detection that we still have as a fallback. But if Homebrew is all binaries, I suppose `pkg-config` might only be a requirement for the package maintainer(s) -- in contrast to PECL, where users are compiling the extension themselves.

alcaeus:
Agree, after install the pkgconf package the detection now works as expected. I'm closing this PR and I've filed PHPC-2520 to document how to install pkg-config on MacOS. We can separately consider removing the legacy path-based detection in favour of the same mechanism used in PHP itself.

PR #1794, opened 2025-02-24
PHPC-2518: Support sort option for updateOne and replaceOne

jmikola:
https://jira.mongodb.org/browse/PHPC-2518

updateMany does not support sort. That validation is handled by mongoc_bulk_operation_update_many_with_opts().

### DIFF
--- a/tests/bulk/bulkwrite-update-008.phpt
+++ b/tests/bulk/bulkwrite-update-008.phpt
@@ -1,0 +1,81 @@
+--TEST--
+MongoDB\Driver\BulkWrite::update() with sort option
+--SKIPIF--
+<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
+<?php skip_if_not_live(); ?>
+<?php skip_if_server_version('<', '8.0'); ?>
+<?php skip_if_not_clean(); ?>
+--FILE--
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+class CommandLogger implements MongoDB\Driver\Monitoring\CommandSubscriber
+{
+    public function commandStarted(MongoDB\Driver\Monitoring\CommandStartedEvent $event): void
+    {
+        if ($event->getCommandName() !== 'update') {
+            return;
+        }
+
+        printf("update included sort: %s\n", json_encode($event->getCommand()->updates[0]->sort));
+    }
+
+    public function commandSucceeded(MongoDB\Driver\Monitoring\CommandSucceededEvent $event): void
+    {
+    }
+
+    public function commandFailed(MongoDB\Driver\Monitoring\CommandFailedEvent $event): void
+    {
+    }
+}
+
+$manager = create_test_manager();
+
+$bulk = new MongoDB\Driver\BulkWrite();
+$bulk->insert(['_id' => 1]);
+$bulk->insert(['_id' => 2]);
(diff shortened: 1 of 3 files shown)


tests/bulk/bulkwrite-update-008.phpt
```diff
@@ -0,0 +1,81 @@
...
+$manager->executeBulkWrite(NS, $bulk);
+
+MongoDB\Driver\Monitoring\addSubscriber(new CommandLogger);
+
+$bulk = new MongoDB\Driver\BulkWrite;
+$bulk->update(['_id' => ['$gt' => 1]], ['$set' => ['x' => 11]], ['sort' => ['_id' => 1]]);
+$manager->executeBulkWrite(NS, $bulk);
+
+$bulk = new MongoDB\Driver\BulkWrite;
+$bulk->update(['_id' => ['$gt' => 1]], ['x' => 22], ['sort' => ['_id' => -1]]);
```
GromNaN, line 47: Did you want to test `replaceOne` here instead? This other method is in the [spec commit](https://github.com/mongodb/specifications/commit/da8e120514fd5e7a5934513f4fb7a63c7cadbb43). Otherwise I don't get the benefit of this 2nd update.

tests/bulk/bulkwrite-update_error-009.phpt
```diff
@@ -0,0 +1,22 @@
...
+echo throws(function() use ($manager, $bulk) {
+    $bulk->update(['x' => ['$gt' => 1]], ['$set' => ['y' => 11]], ['multi' => true, 'sort' => ['x' => 1]]);
+}, MongoDB\Driver\Exception\InvalidArgumentException::class), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Invalid option 'sort'
```
GromNaN, line 21: Is this exception coming from the C driver?

jmikola:
FYI: Evergreen failures are all due to failed downloads when bootstrapping mongo-orchestration. https://github.com/mongodb-labs/drivers-evergreen-tools/pull/612 was intended to address that but I don't think the retry logic was sufficient.

tests/bulk/bulkwrite-update-008.phpt
jmikola: This is actually testing the code path for `mongoc_bulk_operation_replace_one_with_opts`. PHPC's original bulk write API only provides a single `BulkWrite::update()` method, and we branch according to the structure of the second argument and `multi` option.

See: https://github.com/mongodb/mongo-php-driver/blob/1.20.1/src/MongoDB/BulkWrite.c#L452

This isn't relevant to the upcoming API for the bulkWrite _command_, which will have specific methods in PHPC.

tests/bulk/bulkwrite-update_error-009.phpt
jmikola: Yes. I opted to use a more friendly exception message https://github.com/mongodb/mongo-php-library/pull/1605, but did not do so here. Let me know if you'd like a change.

GromNaN: LGTM

GromNaN (approved):

alcaeus (approved):

PR #1790, opened 2025-02-18
PHPC-2478: Implement API for bulkWrite command

jmikola:
https://jira.mongodb.org/browse/PHPC-2478

This supersedes https://github.com/mongodb/mongo-php-driver/pull/1763 and rebases the work on the v2.x branch.

PR #1818, opened 2025-04-16
Feature: Client Bulk Write

GromNaN:
- [x] #1790
- [x] #1819
- [x] #1820


alcaeus (approved):

tests/bulkwritecommand/bulkwritecommand-debug-001.phpt
```diff
@@ -0,0 +1,39 @@
...
+    ["foo"]=>
+    int(1)
+  }
+  ["ordered"]=>
+  bool(true)
+  ["verboseResults"]=>
+  bool(true)
+  ["session"]=>
+  NULL
+}
```
GromNaN, line 38: The `BulkWriteCommand` class doesn't expose the list of operations in `var_dump` or xdebug. The only way to see its state is by executing it and using a subscriber to get the raw command.
Can we add a (virtual) property to show them or is there a limitation in libmongoc?

GromNaN, line 38: The `BulkWriteCommand` doesn't expose the list of operations when using `var_dump` or xdebug. There is no way to inspect its contents without executing it and using a subscriber to get the raw command.

I think the `executed` property should also be exposed, as it cannot be executed multiple times.

alcaeus: Adding `executed` would make sense if we have the information already. As the operations list could potentially be big, I'd rather not expose it in the debug output.

GromNaN: Tracked in [PHPC-2582](https://jira.mongodb.org/browse/PHPC-2582)

PR #1819, opened 2025-04-16
PHPC-2571 Import BulkWrite tests for Client BulkWriteCommand

GromNaN:
Fix PHPC-2571


tests/bulkwritecommand/bulkwritecommand-deleteOne_error-004.phpt
```diff
@@ -0,0 +1,33 @@
...
+
+echo throws(function() use ($bulk) {
+    $bulk->deleteOne(NS, [], ['collation' => MongoDB\BSON\PackedArray::fromPHP([])]);
+}, MongoDB\Driver\Exception\UnexpectedValueException::class), "\n";
+
+// @TODO: ALMOST: Got MongoDB\Driver\Exception\InvalidArgumentException - expected MongoDB\Driver\Exception\UnexpectedValueException
+// Expected "hint" option to yield string or document but got "array"
+echo throws(function() use ($bulk) {
+    $bulk->deleteOne(NS, [], ['hint' => MongoDB\BSON\PackedArray::fromPHP([])]);
+}, MongoDB\Driver\Exception\UnexpectedValueException::class), "\n";
```
GromNaN: The exception type is different:
- Collection BulkWrite: `UnexpectedValueException`
- Client BulkWriteCommand: `InvalidArgumentException`

Should we use the same.

tests/bulkwritecommand/bulkwritecommand-insertOne_error-001.phpt
```diff
@@ -0,0 +1,29 @@
...
+
+// @TODO: FAILED: Expected MongoDB\Driver\Exception\InvalidArgumentException, but no exception thrown!
+echo throws(function() use ($bulk) {
+    $bulk->insertOne(NS, ['' => 1]);
+}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n\n";
+
+// @TODO: FAILED: Expected MongoDB\Driver\Exception\InvalidArgumentException, but no exception thrown!
+echo throws(function() use ($bulk) {
+    $bulk->insertOne(NS, ["\xc3\x28" => 1]);
+}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n";
```
GromNaN: This is not permitted with collection bulk write: https://github.com/mongodb/mongo-php-driver/blob/d3a079fb4d6f9173a0c81c0d0250d1304e0292a3/tests/bulk/bulkwrite-insert_error-001.phpt#L10-L27

tests/bulkwritecommand/bulkwritecommand-updateOne-001.phpt
```diff
@@ -0,0 +1,110 @@
...
+
+$document = new MyClass('foo', new MyClass('bar', new MyClass('baz')));
+
+$bulk = new MongoDB\Driver\BulkWriteCommand();
+// @TODO: Fatal error: Uncaught MongoDB\Driver\Exception\InvalidArgumentException: Invalid key '__pclass': update only works with $ operators and pipelines
+$bulk->updateOne(NS,
+    ['_id' => 'foo'],
+    $document,
+    ['upsert' => true]
+);
```
GromNaN: We should wrap of documents in a `$set`, the simplest would be to do it in PHPLIB, but for the best compatibility we can do it in PHPC.

tests/bulkwritecommand/bulkwritecommand-updateOne_error-003.phpt
```diff
@@ -0,0 +1,57 @@
...
+}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n\n";
+
+echo throws(function() use ($bulk) {
+    $bulk->updateOne(NS, ['x' => 1], ['$set' => ['y' => 1]], ['arrayFilters' => 1]);
+}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n\n";
+
+// @TODO: Expected "arrayFilters" option to yield array but got non-sequential keys
+echo throws(function() use ($bulk) {
+    $bulk->updateOne(NS, ['x' => 1], ['$set' => ['y' => 1]], ['arrayFilters' => ['foo' => 'bar']]);
+}, 'MongoDB\Driver\Exception\InvalidArgumentException'), "\n\n";
```
GromNaN: The error isn't the same as with collection bulk write: 
- Collection BulkWrite: `"arrayFilters" option has invalid keys for a BSON array`
- Client BulkWriteCommand: `Expected "arrayFilters" option to yield array but got non-sequential keys`

https://github.com/mongodb/mongo-php-driver/blob/d3a079fb4d6f9173a0c81c0d0250d1304e0292a3/tests/bulk/bulkwrite-update_error-003.phpt#L26-L28

tests/bulkwritecommand/bulkwritecommand-updateOne_error-006.phpt
```diff
@@ -0,0 +1,33 @@
...
+
+echo throws(function() use ($bulk) {
+    $bulk->updateOne(NS, [], [], ['collation' => MongoDB\BSON\PackedArray::fromPHP([])]);
+}, MongoDB\Driver\Exception\UnexpectedValueException::class), "\n";
+
+// @TODO: ALMOST: Got MongoDB\Driver\Exception\InvalidArgumentException - expected MongoDB\Driver\Exception\UnexpectedValueException
+// Expected "hint" option to yield string or document but got "array"
+echo throws(function() use ($bulk) {
+    $bulk->updateOne(NS, [], [], ['hint' => MongoDB\BSON\PackedArray::fromPHP([])]);
+}, MongoDB\Driver\Exception\UnexpectedValueException::class), "\n";
```
GromNaN: The error message is different:
- Collection BulkWrite: `MongoDB\BSON\PackedArray cannot be serialized as a root document`
- Client BulkWriteCommand: `Expected "hint" option to yield string or document but got "array"`

tests/bulkwritecommand/bulkwritecommand-updateOne_error-009.phpt
```diff
@@ -0,0 +1,23 @@
...
+<?php
+require_once __DIR__ . "/../utils/basic.inc";
+
+$manager = create_test_manager();
+$bulk = new MongoDB\Driver\BulkWriteCommand;
+
+// @TODO: FAILED: Expected MongoDB\Driver\Exception\InvalidArgumentException, but no exception thrown!
+echo throws(function() use ($manager, $bulk) {
+    $bulk->updateOne(NS, ['x' => ['$gt' => 1]], ['$set' => ['y' => 11]], ['multi' => true, 'sort' => ['x' => 1]]);
+}, MongoDB\Driver\Exception\InvalidArgumentException::class), "\n";
```
GromNaN: I don't know why we don't validate the `sort` option for `updateOne`.

GromNaN (commented):
@alcaeus I spotted differences between existing collection BulkWrite and the new client BulkWiteCommand. I need your opinion wether we converge to the same behavior or we keep things different for good reasons.

tests/bulkwritecommand/bulkwritecommand-deleteOne_error-004.phpt
alcaeus: Technically, the `InvalidArgumentException` is the correct one, but we shouldn't change the legacy bulk write classes to throw different exceptions. I think we can adapt the test and leave the exception classes as-is.

tests/bulkwritecommand/bulkwritecommand-insertOne_error-001.phpt
alcaeus: Interesting, I'll have to take another look.

tests/bulkwritecommand/bulkwritecommand-updateOne-001.phpt
alcaeus: > We should wrap of documents in a `$set`

We shouldn't - technically submitting a document is a replace operation. It would be an update once we are able to compute what has changed on the doc and only submit that. The error is expected here.

tests/bulkwritecommand/bulkwritecommand-updateOne_error-003.phpt
alcaeus: This is probably due to newer (read, better) logic in client bulk write. We could change the collection bulk write to use the same exception message, but I'm not sure it's worth it.

tests/bulkwritecommand/bulkwritecommand-updateOne_error-009.phpt
alcaeus: I'm not sure what you mean. The given sort order looks correct?

tests/bulkwritecommand/bulkwritecommand-updateOne-001.phpt
GromNaN: Ok, I'll update the test to put the document in the `$set` operator.

tests/bulkwritecommand/bulkwritecommand-updateOne_error-009.phpt
GromNaN: Collection bulk write is throwing an exception with this options: https://github.com/mongodb/mongo-php-driver/blob/d3a079fb4d6f9173a0c81c0d0250d1304e0292a3/tests/bulk/bulkwrite-update_error-009.phpt#L12-L21

alcaeus: Ah. I'm not sure if legacy bulk write even supports `sort`, I'll have to double-check.

GromNaN: We can check that option `sort` is not used with `updateMany`.
PHPLIB has an [exception for that](https://github.com/mongodb/mongo-php-library/blob/a4ccccd9011bad978e6f0d52efd1a512ef64d35a/src/Operation/Update.php#L156C29-L158), we can add it to `ClientBulkWrite::updateMany` and `ClientBulkWrite::deleteMany` in https://github.com/mongodb/mongo-php-library/pull/1630

For now, I'm removing this test.

tests/bulkwritecommand/bulkwritecommand-insertOne_error-001.phpt
GromNaN: This is something that absolutely must be corrected. Currently, you can insert an invalid BSON.

```php
$bulk = new MongoDB\Driver\BulkWriteCommand;
$bulk->insertOne(NS, ["\xc3\x28" => 1]);
$manager->executeBulkWriteCommand($bulk);
$manager->executeQuery(NS, new MongoDB\Driver\Query([]))->toArray();
```

```
Fatal error: Uncaught MongoDB\Driver\Exception\UnexpectedValueException: Detected corrupt BSON data for field path '_id'
```

GromNaN: I'll remove this tests as the validation is not required by the spec, as mentioned in https://github.com/mongodb/mongo-c-driver/pull/1590

> Since UTF-8 validation is already broken, there are requests to disable, validation negatively impacts performance, and it is not a spec requirement, this PR excludes default validation in MongoClient.bulkWrite.

tests/bulkwritecommand/bulkwritecommand-updateOne_error-009.phpt
GromNaN: Tracked in [PHPLIB-1670](https://jira.mongodb.org/browse/PHPLIB-1670)

GromNaN:
Changes:
- Remove tests for BSON validation https://github.com/mongodb/mongo-c-driver/pull/1590
> Since UTF-8 validation is already broken, there are requests to disable, validation negatively impacts performance, and it is not a spec requirement, this PR excludes default validation in MongoClient.bulkWrite.
- Use the `$set` operator for every `updateOne` tests what were passing a document directly.
- Update the tests for the new exception class and message

tests/bulkwritecommand/bulkwritecommand-insertone-001.phpt
GromNaN, line 1: This test is replaced by [bulkwritecommand-insertOne-004.phpt](https://github.com/mongodb/mongo-php-driver/pull/1819/files#diff-2c7fe26de6f85aec2d1578581083e57bb0486b80b2d9348ead0e37e13e81a787)

alcaeus (approved):

PR #1821, opened 2025-05-07
Add details to the generic `BulkWriteCommandException` message

GromNaN:
I'm trying to add more context to the exception. When a `BulkWriteCommandException` is thrown for individual operation error, the message is not helpful. By mentioning "write concern errors" and "write errors", we give a hint for debugging.

### DIFF
--- a/src/phongo_execute.c
+++ b/src/phongo_execute.c
@@ -440,6 +440,8 @@ bool phongo_execute_bulkwritecommand(zval* manager, php_phongo_bulkwritecommand_
 			(void) spprintf(&message, 0, "Bulk write failed due to previous %s: %s", PHONGO_ZVAL_EXCEPTION_NAME(EG(exception)), error.message);
 			zend_throw_exception(php_phongo_bulkwritecommandexception_ce, message, 0);
 			efree(message);
+		} else if (!has_top_level_error && bw_ret.res) {
+			zend_throw_exception(php_phongo_bulkwritecommandexception_ce, "Bulk write failed with %d write concern errors and %d write errors", error.code, bw_ret.res->n_write_concern_errors, bw_ret.res->n_write_errors);
 		} else {
 			zend_throw_exception(php_phongo_bulkwritecommandexception_ce, has_top_level_error ? error.message : "Bulk write failed", error.code);
 		}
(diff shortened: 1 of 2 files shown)


src/phongo_execute.c
```diff
@@ -440,6 +440,8 @@ bool phongo_execute_bulkwritecommand(zval* manager, php_phongo_bulkwritecommand_
 			(void) spprintf(&message, 0, "Bulk write failed due to previous %s: %s", PHONGO_ZVAL_EXCEPTION_NAME(EG(exception)), error.message);
 			zend_throw_exception(php_phongo_bulkwritecommandexception_ce, message, 0);
 			efree(message);
+		} else if (!has_top_level_error && bw_ret.res) {
+			zend_throw_exception(php_phongo_bulkwritecommandexception_ce, "Bulk write failed with %d write concern errors and %d write errors", error.code, bw_ret.res->n_write_concern_errors, bw_ret.res->n_write_errors);
```
GromNaN, line 444: @alcaeus I don't know how to debug this error. 
```
src/phongo_execute.c:444:158: error: incomplete definition of type 'mongoc_bulkwriteresult_t' (aka 'struct _mongoc_bulkwriteresult_t')
  444 |                         zend_throw_exception(php_phongo_bulkwritecommandexception_ce, "Bulk write failed with %d write concern errors and %d write errors", error.code, bw_ret.res->n_write_concern_errors, bw_ret.res->n_write_errors);
      |                                                                                                                                                                         ~~~~~~~~~~^
```

alcaeus: This is because of how the `mongoc_bulkwriteresult_t` is defined. While `mongoc_bulkwriteresult_t` is defined in the .h file (which we include), it points to a struct that is defined in the .c file. This means that you can access members of a variable in the C file, but not outside of it, for example here as the internal `_mongoc_bulkwriteresult_t` struct doesn't exist. This makes up for the fact that there are no visibilities for structural members in C, and this way libmongoc limits the scope of its external API.

To access struct fields, libmongoc exposes [some functions](http://mongoc.org/libmongoc/current/mongoc_bulkwriteresult_t.html) to extract data from a `mongoc_bulkwriteresult_t`, but notably it doesn't expose the `errorscount` field. The `n_write_concern_errors` and `n_write_errors` field don't exist in the struct at all.

IMO, we should tell people to check the exception object for error details for the time being; we can then discuss whether the C team can add an API to expose the total number of errors.

GromNaN: The `BulkWriteCommandException` class exposes `getWriteErrors` and `getWriteConcernErrors`. It should be possible to count this values to inject this counts in the exception message. Or maybe wrap this exception in PHPLIB to improve the message.

PR #1828, opened 2025-05-21
Bump libmongoc, libmongocrypt and get tests running

alcaeus:
This PR is the first part of PHPC-2435 to update the libmongoc dependency. In this PR:
- PHPC-2581 bumps libmongoc to 2.0.1
- PHPC-2578 bumps libmongocrypt to 1.14.0 (1.13+ is required for libmongoc 2)
- PHPC-2548 and PHPC-2540 handle deprecations

One part that currently isn't tested is building with system dependencies for libmongoc and libmongocrypt. I've created PHPC-2584 to add tests with system libraries.

I've also updated the version file handling for libmongoc as we no longer need to compute the version but instead use the file from the bundled repo directly. The downside to this is that we no longer have commit granularity when compiling with a development version of libmongoc. Since this is only relevant for internal testing, I think it's an acceptable change as we can easily check the submodule to find the version.

I've also updated the build config to test latest and dev versions of libmongoc with PHP 8.4, but can extract this to a separate PR if we want this change to land sooner.

Note: this PR targets a feature branch, as I anticipate more changes and want to test with PHPLIB before merging this to a stable branch. I'll create a separate draft PR into v2.x after this PR is merged.


GromNaN (approved):
I don't fully understand this code, but nothing seems wrong.

PR #1829, opened 2025-05-22
PHPC-2435: Support libmongoc 2.0

alcaeus:
PHPC-2435


bin/prep-release.php
```diff
@@ -49,6 +48,8 @@ function get_files() {
         "src/contrib/*.{c,h}",
 
         "src/libmongoc/src/common/src/*.{c,h,h.in}",
+        // Note: src/libmongoc/src/common/src/mlib/ does not contain source files (as of libmongoc 2.0.1)
+        "src/libmongoc/src/common/src/mlib/*.{c,h}",
```
kevinAlbs:
```suggestion
        "src/libmongoc/src/common/src/mlib/*.h",
```

Remove unnecessary `*.c` glob? I expect the purpose of the comment is to indicate there are no `*.c` source files needed.

bin/prep-release.php
```diff
@@ -64,10 +65,12 @@ function get_files() {
         "src/libmongocrypt-compat/mongocrypt/*.{c,h}",
         "src/libmongocrypt/src/*.{c,h,h.in}",
         "src/libmongocrypt/src/crypto/*.{c,h}",
-        // Note: src/libmongocrypt/src/mlib/ does not contain source files (as of libmongocrypt 1.3.1)
+        "src/libmongocrypt/src/mlib/*.{c,h}",
```
kevinAlbs:
```suggestion
```

I expect this is redundant with the glob below for `src/libmongocrypt/src/mlib/*.h`. The only `*.c` source files are tests.

tests/manager/manager-ctor-write_concern-error-001.phpt
```diff
@@ -12,12 +12,12 @@ echo throws(function() {
     create_test_manager(null, ['w' => 1.0]);
 }, "MongoDB\Driver\Exception\InvalidArgumentException"), "\n";
 
-/* Note: Values of w < -1 are invalid, but libmongoc's URI string parsing only
+/* Note: Values of w < 0 are invalid, but libmongoc's URI string parsing only
  * logs a warning instead of raising an error (see: CDRIVER-2234), so we cannot
  * test for this. */
```
kevinAlbs: Comment appears incorrect. libmongoc [returns an error](https://github.com/kevinAlbs/mongo-c-driver/blob/56a89c5359c165420cdca5df4652737cb8c80886/src/libmongoc/src/mongoc/mongoc-uri.c#L1733) for unsupported values, e.g. `w=-1`:

```php
echo throws(function() {
    create_test_manager("mongodb://localhost:27017/?w=-1", []);
}, "MongoDB\Driver\Exception\InvalidArgumentException"), "\n";
```
Results in output:
```
Failed to parse MongoDB URI: 'mongodb://localhost:27017/?w=-1'. Error while parsing the 'w' URI option: Unsupported w value [w=-1].
```

.github/actions/linux/build-libmongocrypt/action.yml
```diff
@@ -0,0 +1,29 @@
...
+
+  steps:
+    - name: Add repository key
+      shell: bash
+      run: sudo sh -c 'curl -s --location https://pgp.mongodb.com/libmongocrypt.asc | gpg --dearmor >/etc/apt/trusted.gpg.d/libmongocrypt.gpg'
+
+    - name: Add repository
+      shell: bash
+      working-directory: /tmp
+      # Note: no packages for Ubuntu 24.04 noble exist, so we use those for 22.04
```
kevinAlbs, line 18:
```suggestion
      # Note: no packages for Ubuntu 24.04 noble exist, so we use those for 22.04. Update after MONGOCRYPT-813 is released.
```

Aside: MONGOCRYPT-813 is planned for the next 1.14.1 patch release.

kevinAlbs (approved):
LGTM with minor comments.

bin/prep-release.php
alcaeus: Updated - these are indeed not necessary (here and below).

tests/manager/manager-ctor-write_concern-error-001.phpt
alcaeus: Thank you for pointing this out - I guess that comment was never updated after CDRIVER-2234 was fixed.

.github/actions/linux/build-libmongocrypt/action.yml
alcaeus: Thank you, I'll update this alongside 1.14.1 then :)

paulinevos (approved):

bin/prep-release.php
```diff
@@ -65,7 +65,7 @@ function get_files() {
         "src/libmongocrypt-compat/mongocrypt/*.{c,h}",
         "src/libmongocrypt/src/*.{c,h,h.in}",
         "src/libmongocrypt/src/crypto/*.{c,h}",
-        "src/libmongocrypt/src/mlib/*.{c,h}",
+        "src/libmongocrypt/src/mlib/*.h",
```
jmikola, line 68: @alcaeus: This line appears to be redundant with the pattern appearing _after_ the subsequent comment. Perhaps this should be removed entirely?

.github/actions/linux/build-libmongoc/action.yml
```diff
@@ -0,0 +1,55 @@
+name: "Build libmongoc"
+description: "Builds and install libmongoc"
+inputs:
+  prefix:
+    description: "Prefix to install libmongoc to"
+    default: "/usr/local"
+  version:
+    description: "Libmongoc version to build"
```
jmikola, line 8: Noted that the description here is missing "(major.minor)", which you had in the build-libmongocrypt action.

.github/actions/linux/build-libmongocrypt/action.yml
```diff
@@ -0,0 +1,29 @@
...
+    required: true
+runs:
+  using: composite
+
+  steps:
+    - name: Add repository key
+      shell: bash
+      run: sudo sh -c 'curl -s --location https://pgp.mongodb.com/libmongocrypt.asc | gpg --dearmor >/etc/apt/trusted.gpg.d/libmongocrypt.gpg'
+
+    - name: Add repository
```
jmikola, line 15: Is there a particular reason that libmongocrypt is installed as a package, while libmongoc is compiled for source?

Wouldn't the approach here assume that the build is running on an Ubuntu or Debian OS? I suppose that's OK since `runs-on: "ubuntu-latest"` is specified for the `test-system-libs` workflow.

Makefile.frag
```diff
@@ -91,6 +88,28 @@ generate-function-map: all
 		echo "ERROR: Cannot generate function maps without CLI sapi."; \
 	fi
 
+show-config: all
```
jmikola, line 91: Did you write this yourself, or was it pulled from another extension/skeleton?

bin/prep-release.php
```diff
@@ -64,10 +65,12 @@ function get_files() {
         "src/libmongocrypt-compat/mongocrypt/*.{c,h}",
         "src/libmongocrypt/src/*.{c,h,h.in}",
         "src/libmongocrypt/src/crypto/*.{c,h}",
-        // Note: src/libmongocrypt/src/mlib/ does not contain source files (as of libmongocrypt 1.3.1)
+        "src/libmongocrypt/src/mlib/*.h",
+        // Note: src/libmongocrypt/src/mlib/ does not contain source files (as of libmongocrypt 1.14.0)
         "src/libmongocrypt/src/mlib/*.h",
         "src/libmongocrypt/src/os_posix/*.{c,h}",
         "src/libmongocrypt/src/os_win/*.{c,h}",
+        "src/libmongocrypt/src/unicode/*.{c,h}",
```
jmikola, line 73: Noted that this new directory was also included in `config.m4` 👍

scripts/autotools/libmongoc/CheckSSL.m4
```diff
@@ -1,24 +1,32 @@
...
              [auto],
              [no])
 
+dnl PHP_ARG_WITH without a value assigns "yes". Treat it like "auto" but
+dnl require a crypto library.
+if test "$PHP_MONGODB_SSL" = "yes"; then
+  PHP_MONGODB_SSL="auto"
+fi
+
+dnl TODO 3.0: Remove libressl from valid options
```
jmikola, line 14: Is this already tracked in a PHPC ticket?

.github/actions/linux/build-libmongocrypt/action.yml
alcaeus: IIRC, this was because I didn't find a 2.0 package for libmongoc. It's also very likely that we'll need to support building from a branch once we start supporting features only included in an upcoming libmongoc version. Thinking about this, the same may apply to libmongocrypt, but I assume it's less likely as libmongocrypt follows a more aggressive release strategy.

.github/actions/linux/build-libmongoc/action.yml
alcaeus: Ah, yes. Added in #1848

bin/prep-release.php
alcaeus: Indeed. Removed in #1848

Makefile.frag
alcaeus: This was shamelessly copied over from the `test` target. If there's a way to extract all the setup fluff to a target or similar that then runs whatever command is given, I'd definitely go for that, but my Makefile knowledge doesn't quite extend that far.

scripts/autotools/libmongoc/CheckSSL.m4
alcaeus: It is now: PHPC-2605.

.github/actions/linux/build-libmongocrypt/action.yml
jmikola: OK to cross that bridge when we come to it.

PR #1831, opened 2025-05-23
PHPC-2584: Run driver test with system libraries

alcaeus:
PHPC-2584

This PR adds tests with system libraries to ensure that the driver can function when built not with the bundled versions of libmongoc and libmongocrypt, but instead is able to locate and build with system libraries. I'm using GitHub Actions for this build as it's easier to isolate this build from others on this platform. The process is relatively simple:
- Install libmongocrypt
- Build an install libmongoc/libbson
- Run a driver build with the `--with-mongodb-system-libs=yes` configure option
- Make sure the driver builds correctly (hence the `make show-config` output so this can be verified)
- Run tests against a standalone 8.0 server (no need to test multiple topologies or server versions)

Once we merge the feature branch to v2.x, I would also mark this particular build as required so any breakage when building with system libs prevents a merge.


.github/actions/linux/build-libmongocrypt/action.yml
```diff
@@ -0,0 +1,29 @@
...
+
+  steps:
+    - name: Add repository key
+      shell: bash
+      run: sudo sh -c 'curl -s --location https://pgp.mongodb.com/libmongocrypt.asc | gpg --dearmor >/etc/apt/trusted.gpg.d/libmongocrypt.gpg'
+
+    - name: Add repository
+      shell: bash
+      working-directory: /tmp
+      # Note: no packages for Ubuntu 24.04 noble exist, so we use those for 22.04
```
GromNaN, line 18: Are there any plans to have one later? If not, it will get complicated when the upgrade to Ubuntu 24.04 is enforced by GitHub.

.github/actions/linux/build-libmongocrypt/action.yml
```diff
@@ -0,0 +1,29 @@
...
+      env:
+        LIBMONGOCRYPT_VERSION: ${{ inputs.version }}
+
+    - name: Update apt sources
+      shell: bash
+      run: sudo apt-get update
+
+    - name: Install libmongocrypt
+      shell: bash
+      run: sudo apt-get install -y libmongocrypt-dev
```
GromNaN, line 29: It might be possible to cache the apt dependency: https://github.com/marketplace/actions/cache-apt-packages

GromNaN (commented):
SGTM, with 2 questions.

.github/actions/linux/build-libmongocrypt/action.yml
alcaeus: I created [MONGOCRYPT-813](https://jira.mongodb.org/browse/MONGOCRYPT-813) for this. Note that I'm already using 24.04 and the PPA works for the time being. Once the ticket is resolved, I'd update this accordingly.

.github/actions/linux/build-libmongocrypt/action.yml
alcaeus: Noted for a separate PR.

GromNaN (approved):

PR #1837, opened 2025-05-26
PHPC-2367: Add SSPI SASL, drop Cyrus on Windows

alcaeus:
PHPC-2367

This build adds support for building with SSPI SASL on Windows, in turn dropping Cyrus as it's no longer supported by libmongoc. Since people would be using the default value (`yes`) to enable Cyrus support, we add a warning that we're falling back to SSPI. The setting also supports an explicit sspi setting (`--with-mongodb-sasl=sspi`) that does not emit a warning.

Note that our GitHub Windows builds test with SASL support enabled, so this change is properly tested.

### DIFF
--- a/config.w32
+++ b/config.w32
@@ -232,27 +232,33 @@ if (PHP_MONGODB != "no") {
     WARNING("mongodb libopenssl support not enabled, libs not found");
   }
 
-  if (PHP_MONGODB_SASL != "no" &&
-      CHECK_LIB("libsasl.lib", "mongodb", PHP_MONGODB) &&
-      CHECK_HEADER_ADD_INCLUDE("sasl/sasl.h", "CFLAGS_MONGODB")) {
+  has_sasl_libs = CHECK_LIB("libsasl.lib", "mongodb", PHP_MONGODB) &&
+    CHECK_HEADER_ADD_INCLUDE("sasl/sasl.h", "CFLAGS_MONGODB");
+  if (PHP_MONGODB_SASL != "no") {
+    if (has_sasl_libs) {
+      // TODO 3.0: Remove warning on "yes" as it implies "sspi"
+      if (PHP_MONGODB_SASL == "yes") {
+        WARNING("Cyrus SASL support for Windows was removed. Falling back to SSPI; use '--with-mongodb-sasl=sspi' to avoid this warning.");
+        PHP_MONGODB_SASL = "sspi";
+      }
 
-    // TODO 3.0: Remove warning on "yes" as it implies "sspi"
-    if (PHP_MONGODB_SASL == "yes") {
-      WARNING("Cyrus SASL support for Windows was removed. Falling back to SSPI; use '--with-mongodb-sasl=sspi' to avoid this warning.");
-    }
+      if (PHP_MONGODB_SASL == "sspi") {
+        mongoc_opts.MONGOC_ENABLE_SASL = 1;
+        mongoc_opts.MONGOC_ENABLE_SASL_SSPI = 1;
+      } else {
+        ERROR("MongoDB SASL support not enabled, unknown value for --with-mongodb-sasl: " + PHP_MONGODB_SASL);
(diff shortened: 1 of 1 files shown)


config.w32
```diff
@@ -235,17 +235,17 @@ if (PHP_MONGODB != "no") {
   if (PHP_MONGODB_SASL != "no" &&
       CHECK_LIB("libsasl.lib", "mongodb", PHP_MONGODB) &&
       CHECK_HEADER_ADD_INCLUDE("sasl/sasl.h", "CFLAGS_MONGODB")) {
```
kevinAlbs: Suggest removing checks for the Cyrus `libsasl.lib` and `sasl/sasl.h`:

```sh
if (PHP_MONGODB_SASL != "no") {
```

The C driver does not appear to have header/library checks when configuring with `ENABLE_SASL=SSPI`. I expect SSPI is assumed to be present on Windows.

config.w32
```diff
@@ -235,17 +235,17 @@ if (PHP_MONGODB != "no") {
...
 
     if (PHP_MONGODB_SASL == "yes" || PHP_MONGODB_SASL == "sspi") {
+      mongoc_opts.MONGOC_ENABLE_SASL = 1;
       mongoc_opts.MONGOC_ENABLE_SASL_SSPI = 1;
     } else {
-      WARNING("mongodb sasl support not enabled, unknown value for PHP_MONGODB_SASL: " + PHP_MONGODB_SASL);
+      WARNING("MongoDB SASL support not enabled, unknown value for PHP_MONGODB_SASL: " + PHP_MONGODB_SASL);
     }
 
     if (CHECK_FUNC_IN_HEADER("sasl/sasl.h", "sasl_client_done")) {
```
kevinAlbs: If the check above is updated, suggest also updating the warning message below (when `PHP_MONGODB_SASL != "no"`):

```sh
WARNING("MongoDB SASL support not enabled");
```

Consider changing to a non-WARNING message since I expect this would only be printed if a user chose `--with-mongodb-sasl=no`.

config.w32
```diff
@@ -232,15 +232,21 @@ if (PHP_MONGODB != "no") {
...
+    // TODO 3.0: Remove warning on "yes" as it implies "sspi"
+    if (PHP_MONGODB_SASL == "yes") {
+      WARNING("Cyrus SASL support for Windows was removed. Falling back to SSPI; use '--with-mongodb-sasl=sspi' to avoid this warning.");
+    }
+
+    if (PHP_MONGODB_SASL == "yes" || PHP_MONGODB_SASL == "sspi") {
+      mongoc_opts.MONGOC_ENABLE_SASL = 1;
+      mongoc_opts.MONGOC_ENABLE_SASL_SSPI = 1;
+    } else {
+      WARNING("MongoDB SASL support not enabled, unknown value for PHP_MONGODB_SASL: " + PHP_MONGODB_SASL);
```
kevinAlbs:
```suggestion
      WARNING("MongoDB SASL support not enabled, unknown value for --with-mongodb-sasl: " + PHP_MONGODB_SASL);
```

Suggest using `--with-mongodb-sasl` to match how users set the option (IIUC).

config.w32
alcaeus: Thank you. I wasn't sure whether those were still needed, so I erred on the side of caution, but removed the checks now.

config.w32
alcaeus: Agree. With the removed checks for system libraries, this condition will no longer apply. I did change the warning on an unknown value for `--with-mongodb-sasl` to an error as we shouldn't continue in that case.

config.w32
alcaeus: Update: I added the checks back in to support the default case of users not specifying anything. When specifying `--with-mongodb-sasl=sspi` we print an error that the necessary libraries were not found. If `yes` was specified, we print a warning and leave SASL support disabled.

PR #1838, opened 2025-05-27
Ignore branches that are no longer maintained when merging up

alcaeus:
The new `IGNORED_MERGE_UP_BRANCHES` action variable is already configured to exclude the `v1.x` and `v2.0` branches which we no longer support.


GromNaN (approved):
Related to https://github.com/alcaeus/automatic-merge-up-action/pull/70

PR #1846, opened 2025-06-19
Update installation instruction to use `pie`

GromNaN:
pie 1.0.0 has been released, it's the new recommended way of installing PHP extension as it works with Windows, Linux and Mac. It provides more security as the extension metadata are extracted from GitHub releases and published on Packagist [`mongodb/mongodb-extension`](https://packagist.org/packages/mongodb/mongodb-extension), and the source are downloaded from GitHub releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,9 @@ languages.
  - https://www.php.net/mongodb
  - https://www.mongodb.com/docs/drivers/php-drivers/
 
-## Installation with pie
+## Installation
+
+### With pie
 
 To install this extension, you need [pie](https://github.com/php/pie) installed on your system. `pie` is a modern tool for managing PHP extensions.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -22,12 +22,30 @@ languages.
...
 
-## Installation
+## Installation with pie
+
+To install this extension, you need [pie](https://github.com/php/pie) installed on your system. `pie` is a modern tool for managing PHP extensions.
+
+Install the [`mongodb/mongodb-extension`](https://packagist.org/packages/mongodb/mongodb-extension) package from Packagist using the following command:
+
+```shell
+pie install mongodb/mongodb-extension
```
alcaeus, line 34: We should add an example on how to use version constraints - we could show how to install version 2.1 or newer:
```shell
pie install "mongodb/mongodb-extension:^2.1"
```

README.md
```diff
@@ -40,7 +58,7 @@ distributed as the
 [`mongodb/mongodb`](https://packagist.org/packages/mongodb/mongodb) package for
 [Composer](https://getcomposer.org).
 
-## Release Integrity
+### Release Integrity
```
alcaeus: Release integrity is relevant even when not installing with PECL, and moving it to a sub-header doesn't make sense IMO. We can add a general "Installation" header above and make the sections for pie and pecl sub-sections of that installation section.

Since pie doesn't support signature verification and also doesn't support installing from a downloaded binary, we should add a note that integrity validation is not yet supported with pie.

GromNaN: Title updated.
Note added.

alcaeus (approved):
Great work!

PR #1850, opened 2025-07-24
Fix foreach after the garbage collector is invoked

### LINKED ISSUE #1776: foreach over ObjectId and Garbage Collection
### Bug Report

foreach over ObjectId changes behavior after circular reference collector.

Before GC foreach over ObjectId executes one loop iteration with $key="oid".
After GC foreach over ObjectId executes zero loop iterations.
 
### Environment

PHP inside Docker on Windows 10.
Docker Image: php:8.2.27-apache
PHP: 8.2.27
MongoDB PHP Driver: 1.20.1
MongoDB PHP Library: 1.20.0
MongoDB: 8.0.0
mongodb/laravel-mongodb: 4.8.1

MongoDB: I run DB in separate Docker container locally and use in PHP only HOST and PORT of DB.

```
www-data@5c2174446320:~/html$ php -i | grep -E 'mongodb|libmongoc|libbson'
Cannot load Xdebug - it was already loaded
/usr/local/etc/php/conf.d/mongodb.ini,
mongodb
libbson bundled version => 1.28.1
libmongoc bundled version => 1.28.1
libmongoc SSL => enabled
libmongoc SSL library => OpenSSL
libmongoc crypto => enabled
libmongoc crypto library => libcrypto
libmongoc SASL => enabled
libmongoc SRV => enabled
libmongoc compression snappy => enabled
libmongocrypt bundled version => 1.11.0
libmongocrypt crypto library => libcrypto
```

### Test Script

I've used mongodb/laravel-mongodb to reproduce the bug but there is not so much Laravel specific and main issue in \MongoDB\BSON\ObjectId.

ObjectIdTestModel.php
```php
<?php

namespace App\Models;

class ObjectIdTestModel extends \MongoDB\Laravel\Eloquent\Model
{
}
```

Test.php
```php
\dump();

// Uncommenting following line fixes the test
//gc_disable();

$objectId1 = new ObjectId();
$objectId2 = new ObjectId();

// Here we see public property "oid"
echo 'objectId1: ';
xdebug_debug_zval('objectId1');
echo 'objectId2: ';
xdebug_debug_zval('objectId2');

// Check possible assumption about iterating
$is_traversable = ($objectId1 instanceof \Traversable);
echo 'is_traversable: ' . var_export($is_traversable, true) . "\n";

// Check there are no properties for iterating in foreach hence there is not property "oid"
$reflect    = new \ReflectionClass($objectId1);
$properties = $reflect->getProperties();
echo 'properties: ' . print_r($properties, true);

// Allocate a lot of objects
for ($i = 0; $i < 200; $i++) {
    $model = new ObjectIdTestModel();
    $model->save();
}
$a = ObjectIdTestModel::all();

// First foreach over \MongoDB\BSON\ObjectId
$number1 = 0;
foreach ($objectId1 as $key => $value) {
    $number1 += 1;
}

// Allocate a lot of objects again
$a = ObjectIdTestModel::all();

// Second foreach over \MongoDB\BSON\ObjectId
$number2 = 0;
foreach ($objectId2 as $key => $value) {
    $number2 += 1;
}

echo 'objectId1: ';
xdebug_debug_zval('objectId1');
echo 'objectId2: ';
xdebug_debug_zval('objectId2');

$this->assertEquals(1, $number1, 'number1');
$this->assertEquals(1, $number2, 'number2');
```

Test output:
```
objectId1: objectId1: (refcount=1, is_ref=0)=class MongoDB\BSON\ObjectId { public $oid = (refcount=1, is_ref=0)='679a97b2d1cc47db6503f004' }
objectId2: objectId2: (refcount=1, is_ref=0)=class MongoDB\BSON\ObjectId { public $oid = (refcount=1, is_ref=0)='679a97b2d1cc47db6503f005' }
is_traversable: false
properties: Array
(
)
objectId1: objectId1: (refcount=1, is_ref=0)=class MongoDB\BSON\ObjectId { public $oid = (refcount=1, is_ref=0)='679a97b2d1cc47db6503f004' }
objectId2: objectId2: (refcount=1, is_ref=0)=class MongoDB\BSON\ObjectId { public $oid = (refcount=1, is_ref=0)='679a97b2d1cc47db6503f005' }

...

  number2
  Failed asserting that 0 matches expected 1.
```

If the bug is not repoduced you can increase number of created objects in line `for ($i = 0; $i < 200; $i++) {`

#### Expected
$number1 === 1
$number2 === 1
Circular reference collector do not change script behavior.

##### Actual
$number1 === 1
$number2 === 0
Calling gc_disable() "fixes" the bug.

### Preffered solution

Prohibit to iterate over \MongoDB\BSON\ObjectId especially by **hidden** public property "oid".
I have iterated it by mistake and prefer empty loop over object without public properties instead of strange behavior.

### Debug Log

[object_id_gc.zip](https://github.com/user-attachments/files/18594277/object_id_gc.zip)

danog:
Fixes https://github.com/mongodb/mongo-php-driver/issues/1776, https://jira.mongodb.org/browse/PHPC-2505, and the equivalent bug triggered when assigning dynamic properties and then unsetting them again.

The underlying issue in both cases is the invocation of `zend_std_get_properties`, which populates the `properties` hashtable in the zend object (not in the interned mongo one).

This, in turn, causes the iteration logic to completely skip iteration due to the `if (zend_hash_num_elements(properties) == 0) {` check in the `ZEND_FE_RESET_R_SPEC_CV_HANDLER` 

The fix adds custom handlers for `get_property`, `write_property`, `unset_property`, `has_property`, `get_property_ptr_ptr` which use a new interned php_properties hashtable, instead of the zend hashtable property.

The `zend_gc` handler is also modified to invoke `get_properties` instead of `zend_std_get_properties`.

The overall implementation mostly works as expected, the only issue is some strange refcounting behavior within the garbage collector on the hashtable returned by get_gc, which sometimes increments the refcount without decrementing it again, leading to a leak of the interned `properties` hashtable (detectable using ASAN); I have asked clarifications regarding that, marking the PR as draft in the meantime.

There's also a leftover issue with foreach loops not exiting for whatever reason, still looking into that.

### DIFF
--- a/php_phongo.h
+++ b/php_phongo.h
@@ -63,8 +63,8 @@ zend_object_handlers* phongo_get_std_object_handlers(void);
 #define PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_temp, intern, props, size) \
 	do {                                                                  \
 		if (!(intern)->php_properties) {                                  \
-			ALLOC_HASHTABLE(props);                                       \
-			zend_hash_init((props), (size), NULL, ZVAL_PTR_DTOR, 0);      \
+			ALLOC_HASHTABLE((intern)->php_properties);                    \
+			zend_hash_init((intern)->php_properties, 0, NULL, ZVAL_PTR_DTOR, 0); \
 		}                                                                 \
 		if (is_temp) {                                                    \
 			(props) = zend_array_dup((intern)->php_properties);           \
--- a/php_phongo.h
+++ b/php_phongo.h
@@ -111,7 +111,11 @@ zend_object_handlers* phongo_get_std_object_handlers(void);
 			zend_hash_init(props, 0, NULL, ZVAL_PTR_DTOR, 0);                                                                      \
 			_intern_extractor(zobj)->php_properties = props;                                                                       \
 		}                                                                                                                          \
-		return zend_hash_find(props, member);                                                                                      \
+		zval* ret = zend_hash_find(props, member);                                                                                 \
+		if (ret) {                                                                                                                 \
+			return ret;                                                                                                            \
+		}                                                                                                                          \
+		return &EG(uninitialized_zval);                                                                                            \
 	}                                                                                                                              \
                                                                                                                                    \
 	static zval* php_phongo_##_name##_write_property(zend_object* zobj, zend_string* name, zval* value, void** cache_slot)         \
--- a/php_phongo.h
+++ b/php_phongo.h
@@ -60,6 +60,12 @@ ZEND_TSRMLS_CACHE_EXTERN()
 
 zend_object_handlers* phongo_get_std_object_handlers(void);
 
+#define PHONGO_RETURN_PROPS(is_temp, props) \
+	if (!(is_temp)) { \
+		GC_ADDREF(props); \
+	} \
+	return props;
+
 #define PHONGO_GET_PROPERTY_HASH_INIT_PROPS(is_temp, intern, props, size) \
 	do {                                                                  \
 		if (!(intern)->php_properties) {                                  \
--- a/php_phongo.h
+++ b/php_phongo.h
@@ -84,6 +84,79 @@ zend_object_handlers* phongo_get_std_object_handlers(void);
 		}                                                   \
 	} while (0)
 
+#define PHONGO_GET_PROPERTY_HANDLERS(_name, _intern_extractor) \
+	static zval* php_phongo_##_name##_read_property(zend_object *zobj, zend_string *member, int type, void **cache_slot, zval *rv) \
+	{ \
+		HashTable *props = _intern_extractor(zobj)->php_properties; \
+		if (!props) { \
+			ALLOC_HASHTABLE(props); \
+			zend_hash_init(props, 0, NULL, ZVAL_PTR_DTOR, 0); \
+			_intern_extractor(zobj)->php_properties = props; \
+		} \
+		return zend_hash_find(props, member); \
+	} \
+	\
+	static zval *php_phongo_##_name##_write_property(zend_object *zobj, zend_string *name, zval *value, void **cache_slot) \
+	{ \
+		Z_TRY_ADDREF_P(value); \
+		HashTable *props = _intern_extractor(zobj)->php_properties; \
+		if (!props) { \
+			ALLOC_HASHTABLE(props); \
+			zend_hash_init(props, 0, NULL, ZVAL_PTR_DTOR, 0); \
+			_intern_extractor(zobj)->php_properties = props; \
+		} \
+		return zend_hash_add_new(props, name, value); \
+	} \
+	static int php_phongo_##_name##_has_property(zend_object *zobj, zend_string *name, int has_set_exists, void **cache_slot) \
+	{ \
+		HashTable *props = _intern_extractor(zobj)->php_properties; \
+		if (!props) { \
+			ALLOC_HASHTABLE(props); \
+			zend_hash_init(props, 0, NULL, ZVAL_PTR_DTOR, 0); \
+			_intern_extractor(zobj)->php_properties = props; \
+		} \
+		zval *value = zend_hash_find(props, name); \
+		if (value) { \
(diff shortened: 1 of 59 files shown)


php_phongo.h
```diff
@@ -60,28 +60,112 @@ ZEND_TSRMLS_CACHE_EXTERN()
 
 zend_object_handlers* phongo_get_std_object_handlers(void);
 
+#define PHONGO_RETURN_PROPS(is_temp, props) \
+	if (!(is_temp)) { \
+		GC_ADDREF(props); \
```
danog: This is done to mark that the hashtable is also stored in (intern)->properties; can't be done in `PHONGO_GET_PROPERTY_HASH_INIT_PROPS` because `zend_hash_str_update` to add the native properties requires the refcount of the hashtable to be 1.

danog:
Fixed up everything!

danog:
The codebase was successfully formatted using `scripts/clang-format.sh` locally with clang-format 20, not sure why is it complaining here; can't access the evergreen tests so I can't see what's wrong there

alcaeus:
I can take a look at the clang-format issue.

It looks like there is a memory leak in some tests:
```
[2025/07/28 09:44:20.753] FAILURE:  (FAILED)
[2025/07/28 09:44:20.753] --
[2025/07/28 09:44:20.753]      same serverConnectionId as last commandStarted: yes
[2025/07/28 09:44:20.753]      int(%d)
[2025/07/28 09:44:20.753]      OK: Got MongoDB\Driver\Exception\CommandException
[2025/07/28 09:44:20.753] 007+ [Mon Jul 28 07:43:20 2025]  Script:  '/data/mci/c0332320a42282d79356368a6fe838a7/src/tests/apm/commandFailedEvent-getServerConnectionId-001.php'
[2025/07/28 09:44:20.753] 008+ /data/mci/5683cf19da3859207a5bbd4900e33449/src/src/MongoDB/Manager.c(253) :  Freeing 0x00007f3fcc07c180 (56 bytes), script=/data/mci/c0332320a42282d79356368a6fe838a7/src/tests/apm/commandFailedEvent-getServerConnectionId-001.php
[2025/07/28 09:44:20.753] 009+ === Total 1 memory leaks detected ===
```

Our tests on GHA don't use debug versions of PHP, so these leaks are not reported there. The offending line in Manager.c allocates the `intern->subscribers` hash table: https://github.com/mongodb/mongo-php-driver/blob/ae08bbc06e96109839f21787cbd63e34609db0d1/src/MongoDB/Manager.c#L253
I can see that in a number of classes, you have replaced the calls to `zend_hash_destroy` followed by `FREE_HASHTABLE` with a single call to `zend_hash_release`, but left just a single call to `zend_hash_destroy` in Manager.c.

If you are testing with a debug version of PHP locally, I recommend calling `./configure` with the `--enable-mongodb-developer-flags` switch to enable memory leak detection and other development checks.

danog:
I simply compiled both php and the extension with ASAN/LSAN, which is even better than PHP's builtin leak detection, but the issue is that there are many unfixed leaks in PHP itself when dealing with fatal errors (such as the invalid class extension tests, etc), so I just focused on finding leaks in the GC/cycle tests, that's why I missed this one :)

Re: the usage of zend_hash_release, I'd actually also replace the remaining places where zend_hash_destroy/HASH_RELEASE is used as it considers the refcount/immutability (not present in this case, but who knows in the future) and automatically frees the hashtable struct as well, though probably it would be better to do this in a separate PR...

alcaeus:
@danog the windows failure is unrelated, and the first impression from Evergreen is that the memory leak is no more. As for the remaining `zend_hash_destroy` usage, I have no objection to changing those as well, but a separate PR might indeed be the way to go for this. Please be aware that @jmikola is at Laracon these days, so it might be a few days until we can review and merge this PR.

jmikola:
> As for the remaining zend_hash_destroy usage, I have no objection to changing those as well, but a separate PR might indeed be the way to go for this.

Created [PHPC-2682](https://jira.mongodb.org/browse/PHPC-2682) to track that, as I noticed that several calls in `phongo_apm.c` were left untouched.

PR #1853, opened 2025-07-24
PHPC-2474: Support Binary vector subtype

jmikola:
https://jira.mongodb.org/browse/PHPC-2474


src/BSON/BinaryVector.stub.php
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public const VECTOR_TYPE_FLOAT32 = UNKNOWN;
+
+    /**
+     * @var int
+     * @cvalue PHONGO_BSON_VECTOR_TYPE_PACKED_BIT
+     */
+    public const VECTOR_TYPE_PACKED_BIT = UNKNOWN;
+
+    final public function __construct(array $vector, int $type) {}
```
GromNaN: Is it possible to specify an advanced type in this stub? It should be `@param list<int|float> $vector`.

GromNaN: I find the `$type` param confusing as the parent constructor has a param with the same name but a different meaning. It should be `$vectorType`.

src/BSON/BinaryVector.stub.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+/**
+ * @generate-class-entries static
+ * @generate-function-entries static
+ */
+
+namespace MongoDB\BSON;
+
+class BinaryVector extends Binary
```
GromNaN: By extending the `Binary` class you inherit from the `Binary::TYPE_*` constants that are not accurate for this class.

src/BSON/Binary.h
```diff
@@ -19,6 +19,13 @@
 
 #define PHONGO_BINARY_UUID_SIZE 16
 
+typedef enum {
+	PHONGO_BSON_VECTOR_TYPE_INVALID    = 0,
+	PHONGO_BSON_VECTOR_TYPE_INT8       = 0x03,
+	PHONGO_BSON_VECTOR_TYPE_FLOAT32    = 0x27,
+	PHONGO_BSON_VECTOR_TYPE_PACKED_BIT = 0x10,
+} phongo_bson_vector_type_t;
```
GromNaN: If you can use an enum in the C code, you should be able to use a PHP userland enum for the vector type. The constructor signature would be more strict.

src/BSON/Binary.c
```diff
@@ -308,3 +316,250 @@ bool phongo_binary_new(zval* object, const char* data, size_t data_len, bson_sub
 
 	return true;
 }
+
+/* MongoDB\BSON\BinaryVector implementation */
```
GromNaN: Why isn't it in a new file `src/BSON/BinaryVector.c`?

src/BSON/Binary.stub.php
```diff
@@ -7,7 +7,7 @@
 
 namespace MongoDB\BSON;
 
-final class Binary implements BinaryInterface, \JsonSerializable, Type, \Stringable
+class Binary implements BinaryInterface, \JsonSerializable, Type, \Stringable
```
GromNaN: This classes does not define any method that is not in one of the interfaces. This means you don't need to make it extensible: the new `BinaryVector` class could implement all this interfaces (re-using the underlying C functions).

In the library, [I used `MongoDB\BSON\Binary` type in the builder](https://github.com/mongodb/mongo-php-library/blob/25dd588cdd3360294fa8539c29f1cec66dc33c56/generator/config/expressions.php#L27), but I would change it to use `BinaryInterface`. Why this interface doesn't extend `MongoDB\BSON\Type`?

GromNaN (commented):
In the python implementation, the `Binary` class has a [`from_vector` static method](https://github.com/mongodb/mongo-python-driver/commit/ae6cfd6d102d885ac6b0873d31f0dac139b1ddae#diff-a2613ff5c658ceeb40e3c60a6ff0cd4ebaacf6b8f43867af5ce46a816c40982eR401). You have certainly considered this solution, of having only a factory method instead of a new class. But the new methods `toArray` and `getVectorType` are specific to this binary type, so having a dedicated class makes sense. But if you create a `new Binary($raw, Binary::TYPE_VECTOR)` instance with the vector type from the raw binary data, you don't have this methods. The current state doesn't allow creating a `BinaryVector` from raw data. I think you would have to wrap in into a BSON Document and encode/decode it to get a `BinaryVector` instance. I don't find any other binary type that would require a specific class.

I see 3 possible implementations:
1. Add `Binary::fromVector(array $vector, VectorType $vectorType)` and `toArray()` to the existing `Binary` class
2. ~Create `BinaryVector` class that extends `Binary` (current state of this PR).~
3. ~Create `BinaryVector` class that implements `BinaryInterface, \JsonSerializable, Type, \Stringable`.~

We need to compare pro/con of each solution.

src/BSON/BinaryVector.stub.php
jmikola: We can do that if it'd help Psalm or other static analysis, but it'd have no effect for generated stubs.

jmikola: Can change to `$vectorType` when moving to `Binary::fromVector()`.

src/BSON/BinaryVector.stub.php
jmikola: That is a good argument for not extending.

src/BSON/Binary.h
jmikola: Second suggestion you had was to place the `Binary::fromVector()` factory method on the enum class.

src/BSON/Binary.c
jmikola: We depended on so many Binary.c functions (many of which are static) that it was simpler to share the same file rather than export those symbols.

src/BSON/Binary.stub.php
jmikola: The interfaces were never very useful for PHPC. One of the original use cases was have an embedded object masquerade as a UTCDateTime and incorporate a timezone, but the benefits are mainly for userland applications.

I think sharing a single Binary class is ideal for anything that would expect the BSON type, and avoiding a separate BinaryVector child class seems like it'd make things easier for users (and simplify what comes back from BSON-to-PHP conversion).

src/BSON/Binary.h
GromNaN: The enum method would look like this:

```php
enum VectorType
{
    case Int;
    case Float;
    case PackedBit;

    public function createBinaryVector(array $vector): Binary
    {
        return Binary::fromVector($vector, $this);
    }
}
```

Usage:
```php
VectorType::Int->createBinaryVector([1, 2, 3]);
```

GromNaN: I think using enum method to create the `Binary` instance is confusing, I don't think it's a good idea.

src/BSON/BinaryVector.stub.php
```diff
@@ -0,0 +1,46 @@
...
+
+    final static public function fromInt8Array(array $vector): BinaryVector {}
+
+    //final static public function fromInt8Array2(array $vector): BinaryVector {}
+
+    final static public function fromFloat32Array(array $vector): BinaryVector {}
+
+    final static public function fromPackedBitArray(array $vector): BinaryVector {}
+
+    final public function getVectorType(): int {}
```
paulinevos: Do we want to return `VectorType` here?

GromNaN: Yes, once this method is moved to the `MongoDB\BSON\Binary` class.

src/BSON/BinaryVector.stub.php
```diff
@@ -0,0 +1,46 @@
...
+     * @cvalue PHONGO_BSON_VECTOR_TYPE_PACKED_BIT
+     */
+    public const VECTOR_TYPE_PACKED_BIT = UNKNOWN;
+
+    // TODO: Implement and branch
+    public function __construct(array $vector, int $vectorType) {}
+
+    final static public function fromInt8Array(array $vector): BinaryVector {}
+
+    //final static public function fromInt8Array2(array $vector): BinaryVector {}
```
jmikola: Reminder to remove this since we benchmarked both implementations and this was worse.

src/BSON/BinaryVector.stub.php
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * @var int
+     * @cvalue PHONGO_BSON_VECTOR_TYPE_PACKED_BIT
+     */
+    public const VECTOR_TYPE_PACKED_BIT = UNKNOWN;
+
+    // TODO: Implement and branch
+    public function __construct(array $vector, int $vectorType) {}
+
+    final static public function fromInt8Array(array $vector): BinaryVector {}
```
GromNaN: I think distinct factory names `fromInt8Vector`, `fromFloat32Vector` and `fromPackedBitVector` would make sense if we can provide array types `list<int>`, `list<float>`, `list<bool|0|1>`

src/BSON/Binary.h
GromNaN: Closing this comment as it's too unconventional.

src/BSON/BinaryVector.stub.php
jmikola: AFAIK there's no prior art for using Psalm type hints in stub files. I'm not familiar with how PHPStorm parses the stubs, but would these have a significant benefit for developers? For instance, would they be able to influence Psalm's analysis of an app?

Also, does this mean we prefer factory methods on Binary in lieu of factories on the vector type enum? Either way I expect we'll get a `Binary::fromVector()` method.

alcaeus: There's no point adding these -- if anything the parser might stumble over them. We do however provide a [stub file for psalm](https://github.com/vimeo/psalm/blob/6.x/stubs/extensions/mongodb.phpstub) that includes more advanced types, including generics.

src/BSON/Binary.h
```diff
@@ -19,6 +19,13 @@
 
 #define PHONGO_BINARY_UUID_SIZE 16
 
+typedef enum {
+	PHONGO_BSON_VECTOR_TYPE_INVALID    = 0,
```
GromNaN: Why do you have this enum case here, that is not documented in the stub? Is it returned by `VectorType::cases()`?
This can be covered by a test.

src/BSON/Binary.c
```diff
@@ -297,14 +328,339 @@ void php_phongo_binary_init_ce(INIT_FUNC_ARGS)
...
+
+	if (bson_vector_float32_const_view_init(NULL, data, data_len)) {
+		return PHONGO_BSON_VECTOR_TYPE_FLOAT32;
+	}
+
+	if (bson_vector_packed_bit_const_view_init(NULL, data, data_len)) {
+		return PHONGO_BSON_VECTOR_TYPE_PACKED_BIT;
+	}
+
+	return PHONGO_BSON_VECTOR_TYPE_INVALID;
```
GromNaN: Can't we throw the exception here, or return `null`? I don't like having this invalid enum case.

tests/bson/bson-binary-fromVector-001.phpt
```diff
@@ -0,0 +1,71 @@
...
+    float(1)
+    [1]=>
+    float(-1)
+    [2]=>
+    float(0.5)
+    [3]=>
+    float(-0.5)
+  }
+  ["vectorType"]=>
+  int(39)
```
GromNaN: This is the internal value, do we really want to expose it?

tests/bson/bson-binary-fromVector_error-001.phpt
```diff
@@ -0,0 +1,24 @@
...
+}, MongoDB\Driver\Exception\InvalidArgumentException::class), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected vector to be a list
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected vector[1] to be a float, int given
```
GromNaN, line 23: This is stricter than PHP that would cast the int into float. That makes sense for a vector.

tests/bson/bson-binary-fromVector_error-002.phpt
```diff
@@ -0,0 +1,30 @@
...
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected vector to be a list
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected vector[1] to be an integer, float given
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected vector[1] to be a signed 8-bit integer, 256 given
```
GromNaN: Should you also test the lower boundary?

tests/bson/bson-binary-fromVector-001.phpt
```diff
@@ -0,0 +1,71 @@
+--TEST--
+MongoDB\BSON\Binary::fromVector() construction of various vector types
+--FILE--
+<?php
+
+var_dump(MongoDB\BSON\Binary::fromVector([1.0, -1.0, 0.5, -0.5], MongoDB\BSON\VectorType::Float32));
+var_dump(MongoDB\BSON\Binary::fromVector([1, 2, 3, 4], MongoDB\BSON\VectorType::Int8));
```
GromNaN: You can insert a negative int, as we expect "signed 8-bit integer".

tests/bson/bson-binary-fromVector_error-003.phpt
```diff
@@ -0,0 +1,30 @@
...
+}, MongoDB\Driver\Exception\InvalidArgumentException::class), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECT--
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected vector to be a list
+OK: Got MongoDB\Driver\Exception\InvalidArgumentException
+Expected vector[1] to be an integer or boolean, float given
```
GromNaN: The error message could be more specific.
```suggestion
Expected vector[1] to be 0, 1 or boolean, float given
```

src/BSON/Binary.c
```diff
@@ -297,14 +328,339 @@ void php_phongo_binary_init_ce(INIT_FUNC_ARGS)
...
+	size_t i = 0;
+
+	ZEND_HASH_FOREACH_VAL_IND(vector, val)
+	{
+		if (Z_TYPE_P(val) != IS_LONG && Z_TYPE_P(val) != IS_TRUE && Z_TYPE_P(val) != IS_FALSE) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected vector[%zu] to be an integer or boolean, %s given", i, zend_zval_type_name(val));
+			return;
+		}
+
+		if (Z_TYPE_P(val) == IS_LONG && (Z_LVAL_P(val) < 0 || Z_LVAL_P(val) > 1)) {
```
GromNaN: Why do you use `<` and `>` instead of `!=`?
```suggestion
		if (Z_TYPE_P(val) == IS_LONG && (Z_LVAL_P(val) != 0 || Z_LVAL_P(val) != 1)) {
```

src/BSON/Binary.h
jmikola: I wanted a failure case for `phongo_binary_get_vector_type_from_data()`, which is called by `phongo_binary_get_vector_type()`. This invalid return value is already tested by `bson-binary-set_state_error-004.phpt`, but I can add additional tests for the constructor and unserialization.

It's not applicable to the enum class, since we don't want to expose this to users and `Binary::getVectorType()` would also throw before attempting to return an enum. That said, there's no way to test that code path since we should never end up with a Binary instance containing invalid vector data.

src/BSON/Binary.c
jmikola: Even if I moved `phongo_throw_exception()` here, we'd still have to return a value. While C would technically allow any integer to be returned, it's still cleaner to use an actual enum value for hinting. The same is done for some enums in libmongoc, although they tend to name the case `UNKNOWN` instead of `INVALID`, which I actually think might be preferable -- then the "invalid" terminology would only get used in our exception messages.

The alternative to all of this would be changing the API to instead return a success/failure boolean and instead populate a `phongo_bson_vector_type_t*` output variable, but that API feels much more cumbersome and wouldn't allow me to concisely call the getter as part of an `if` condition, for example.

tests/bson/bson-binary-fromVector-001.phpt
jmikola: Good point. This predated the introduction of a VectorType enum, but I think it makes more sense to return an enum case instead.

tests/bson/bson-binary-fromVector_error-001.phpt
jmikola: > That makes sense for a vector

Just so I understand correctly, are you suggesting we _do_ accept integer values for float vectors or are you OK with being strict here?

Presently, we only cast `int(0)` and `int(1)` to booleans for packed bit vectors.

I'm amenable to accepting integers here for the user's convenience but want to be sure that's what you're proposing.

tests/bson/bson-binary-fromVector_error-002.phpt
jmikola: This technically isn't even testing the upper boundary. I'll change this to use `128` (`INT8_MAX + 1`) and `-129` (`INT8_MIN - 1`).

tests/bson/bson-binary-fromVector-001.phpt
jmikola: I'll change this to include the boundary conditions `-128` and `127`.

src/BSON/Binary.c
jmikola: I've typically used `<` and `>` for range validation, as the `||` can short-circuit. This is a special case where we need only check for two values, so I can switch to:

```c
Z_TYPE_P(val) == IS_LONG && Z_LVAL_P(val) != 0 && Z_LVAL_P(val) != 1
```

src/BSON/Binary.h
GromNaN: I think I confused th C enum with the PHP enum. Having this extra case in the C enum is fine as soon as we don't have it in the PHP enum.

tests/bson/bson-binary-fromVector_error-001.phpt
GromNaN: Being stricter that PHP here is a good feature. Keep it like this.

src/BSON/Binary.h
jmikola: Noted. I'll still rename the C enum to `UNKNOWN` as that's more consistent with other names in libmongoc (we don't have many enums ourselves).

tests/bson/bson-binary-fromVector_error-001.phpt
jmikola: Fantastic. I think it'd also be helpful if folks end up using typed arrays for their vectors, as I expect Psalm would be strict as well.

tests/bson/bson-vectortype-001.phpt
```diff
@@ -0,0 +1,17 @@
+--TEST--
+MongoDB\BSON\VectorType
+--FILE--
+<?php
+
+var_dump(MongoDB\BSON\VectorType::Float32);
+var_dump(MongoDB\BSON\VectorType::Int8);
+var_dump(MongoDB\BSON\VectorType::PackedBit);
```
GromNaN: To get all the cases:
```suggestion
var_dump(MongoDB\BSON\VectorType::cases());
```

jmikola:
One failing test appears to be from Atlas connectivity tests failing to connect to a serverless cluster. Per [PHPLIB-1667](https://jira.mongodb.org/browse/PHPLIB-1667), I expect that should have been removed; however, I don't see a related PHPC ticket. Please review ignoring that particular test failure and I'll address that with a separate PR and ticket.

alcaeus (approved):
LGTM. No objection to adding corpus tests in a follow-up PR.

GromNaN (approved):

tests/bson/bson-vectortype-001.phpt
jmikola: I assume `cases()` returns everything in the declaration order (from the stub/header), but I'm not sure if that's an implementation detail. A more robust solution might be to construct an ordered list of test cases and then do a separate assertion to ensure it's consistent with what `cases()` returns.

GromNaN: From the [documentation](https://www.php.net/manual/en/unitenum.cases.php):
> This method will return a packed array of all cases in an enumeration, **in order of declaration**.

So you don't have to sort them.

GromNaN:
```php
var_dump(VectorType::cases());
```
```
array(3) {
  [0]=>
  enum(VectorType::Float32)
  [1]=>
  enum(VectorType::Int8)
  [2]=>
  enum(VectorType::PackedBit)
}
```

tests/bson/bson-vectortype-001.phpt
```diff
@@ -0,0 +1,17 @@
+--TEST--
+MongoDB\BSON\VectorType cases
+--FILE--
+<?php
+
+foreach (VectorType::cases() as $case) {
```
jmikola: Note: added usage of `cases()`.

tests/bson/bson-binary-serialization_error-007.phpt
```diff
@@ -0,0 +1,18 @@
+--TEST--
+MongoDB\BSON\Binary unserialization requires valid vector data (__serialize and __unserialize)
```
jmikola, line 2: Note: this complements `bson-binary-set_state_error-004.phpt`

GromNaN (approved):

PR #1858, opened 2025-08-07
PHPC-2613: Test x509 authentication on Atlas

alcaeus:
PHPC-2613

This adds two new Atlas environments to the connectivity tests. These test X509 authentication and rely on extracting a certificate string to a file and modifying the connection string accordingly.

### DIFF
--- a/tests/atlas.phpt
+++ b/tests/atlas.phpt
@@ -12,7 +12,7 @@ function extractUri(string $env): ?string
 	return getenv($env) ?: null;
 }
 
-function extractUriWithCertificate(string $env): ?string
+function extractUriWithCertificate(string $env): ?array
 {
 	$uri = getenv($env);
 	if (! is_string($uri)) {
@@ -24,10 +24,19 @@ function extractUriWithCertificate(string $env): ?string
 		return null;
 	}
 
-	$certPath = '/tmp/cert.pem';
-	file_put_contents($certPath, base64_decode($cert));
+	$certPath = tempnam(sys_get_temp_dir(), 'cert_');
+	$certContents = base64_decode($cert);
+	if (! $certPath || ! $certContents) {
+		return null;
+	}
 
-	return $uri . '&tlsCertificateKeyFile=' . $certPath;
+	file_put_contents($certPath, $certContents);
+	chmod($certPath, 0600);
+
+	return [
+		'uri' => $uri . '&tlsCertificateKeyFile=' . $certPath,
+		'certPath' => $certPath,
+	];
 }
 
 function testConnection(string $uri): void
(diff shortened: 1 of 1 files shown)


tests/atlas.phpt
```diff
@@ -21,28 +71,41 @@ $envs = [
...
-$query = new \MongoDB\Driver\Query([]);
+$x509Envs = [
+	'ATLAS_X509',
+	'ATLAS_X509_DEV',
+];
 
 foreach ($envs as $env) {
 	echo $env, ': ';
-	$uri = getenv($env);
+	$uri = extractUri($env);
```
jmikola: It looks like the only purpose of `extractUri()` is to convert `false` return value from `getenv()` into `null`. That seems unnecessary given the conditional below that fails a non-string value.

tests/atlas.phpt
```diff
@@ -7,6 +7,56 @@ Atlas Connectivity Tests
...
+	$certContents = base64_decode($cert);
+	if (! $certPath || ! $certContents) {
+		return null;
+	}
+
+	file_put_contents($certPath, $certContents);
+	chmod($certPath, 0600);
+
+	return [
+		'uri' => $uri . '&tlsCertificateKeyFile=' . $certPath,
```
jmikola, line 31: Should `$certPath` get URL-encoded with [`rawurlencode()`](https://www.php.net/manual/en/function.rawurlencode.php)? I expect `tempnam()` will only produce alphanumeric strings, so that may be superfluous.

tests/atlas.phpt
```diff
@@ -7,6 +7,56 @@ Atlas Connectivity Tests
...
+	}
+
+	$certPath = tempnam(sys_get_temp_dir(), 'cert_');
+	$certContents = base64_decode($cert);
+	if (! $certPath || ! $certContents) {
+		return null;
+	}
+
+	file_put_contents($certPath, $certContents);
+	chmod($certPath, 0600);
```
jmikola: Is this necessary? [`tempnam()`](https://www.php.net/tempnam) should already create the file with these perms.

jmikola (approved):
Some questions but you can decide what is worth addressing.

tests/atlas.phpt
alcaeus: Good point, I was a little eager with my refactoring. Removed it in favour of `getenv`.

tests/atlas.phpt
alcaeus: Yeah, I think it should be safe (famous last words)

tests/atlas.phpt
alcaeus: TIL:
> Creates a file with a unique filename, with access permission set to 0600

Removed the `chmod` call.

PR #1868, opened 2025-09-12
PHPC-2638 and PHPC-2630: Bump to libmongoc 2.1.2 and libmongocrypt 1.16.0

jmikola:
https://jira.mongodb.org/browse/PHPC-2638
https://jira.mongodb.org/browse/PHPC-2630

The libmongocrypt upgrade is necessary due to a collision between "mlib" headers in libmongoc and libmongocrypt (addressed in [MONGOCRYPT-845](https://jira.mongodb.org/browse/MONGOCRYPT-845)).


.evergreen/config/templates/build/build-libmongoc.yml
```diff
@@ -10,7 +10,7 @@
       - func: "compile driver"
         vars:
           PHP_VERSION: "%phpVersion%"
-          LIBMONGOC_VERSION: "2.0.2"
```
alcaeus, line 13: Reminder to regenerate config as you've only updated the template

src/LIBMONGOCRYPT_VERSION_CURRENT
```diff
@@ -1 +1 @@
-1.14.1
+1.15.2-20250912+git03cba710f7
```
alcaeus: Do we depend on the development version of libmongocrypt 1.15.2? Would it make sense to wait for a release?

jmikola: Originally opened this before the release. It has since been bumped to the 1.15.2 tag, though.

alcaeus (approved):

jmikola:
Note: we've got two regularly failing builds on Windows 8.4 x86 ([nts](https://github.com/mongodb/mongo-php-driver/actions/runs/17775530868/job/50522138488?pr=1868#step:3:623) and [ts](https://github.com/mongodb/mongo-php-driver/actions/runs/17775530868/job/50522138628?pr=1868#step:3:623)) due to:

> C1060: compiler is out of heap space

The most recent failures are at [mongoc-handshake.c:702](https://github.com/mongodb/mongo-c-driver/blob/2.1.0/src/libmongoc/src/mongoc/mongoc-handshake.c#L702) but I'm not sure how pertinent that is. Will come back to this and investigate further.

tests/manager/manager-ctor-006.phpt
```diff
@@ -12,7 +12,7 @@ echo throws(function() {
 ?>
 ===DONE===
 <?php exit(0); ?>
---EXPECT--
+--EXPECTF--
 OK: Got MongoDB\Driver\Exception\InvalidArgumentException
-Failed to parse MongoDB URI: 'invalid'. Invalid URI Schema, expecting 'mongodb://' or 'mongodb+srv://'.
+Failed to parse MongoDB URI: 'invalid'. Invalid URI%s
```
GromNaN: Noted the new exception message is
> Failed to parse MongoDB URI: 'invalid'. Invalid URI, no scheme part specified.

Why not update the expected output instead of adding this place-holder?

jmikola: Good question. I must have done this during an earlier iteration of this PR when I was testing against both libmongoc and 2.0.x and 2.1.x, but there is no reason not to use the newer pattern since we require 2.1.2+.

.gitmodules
```diff
@@ -1,7 +1,8 @@
 [submodule "src/libmongoc"]
 	path = src/libmongoc
-	url = https://github.com/mongodb/mongo-c-driver.git
+	url = https://github.com/jmikola/mongo-c-driver.git
 	ignore = untracked
+	branch = 2.1.2-handshake
```
GromNaN: Is the fix you added on [your branch](https://github.com/jmikola/mongo-c-driver/commits/2.1.2-handshake/) necessary for this PR?

jmikola: I'm still attempting to isolate the OOM compiler error in order to provide meaningful feedback to the C team. It's likely this is the result of ludicrous macro expansion in their BSON DSL (see `mongoc-handshake.c` in https://github.com/mongodb/mongo-c-driver/commit/f142c6138321d186f19fa2735662337e7c0aafe9).

jmikola: Quoting [How to: Enable a 64-Bit, x64 hosted MSVC toolset on the command line](https://learn.microsoft.com/en-us/cpp/build/how-to-enable-a-64-bit-visual-cpp-toolset-on-the-command-line?view=msvc-170):

> If you run into memory limits when you use the 32-bit tools, try the 64-bit tools.

Hat tip to @kevinAlbs.

GromNaN (approved):

PR #1849, opened 2025-07-17
use ZSTR_IS_INTERNED()

remicollet:
* `IS_INTERNED` removed in 8.5
* `ZSTR_IS_INTERNED` available since 7.0

PR #1871, opened 2025-09-16
PHPC-2617: Support PHP 8.5

alcaeus:
PHPC-2617

Adds testing and support for PHP 8.5, backporting the fix for `IS_INTERNED` from #1849.

### DIFF
--- a/.github/actions/windows/prepare-build/action.yml
+++ b/.github/actions/windows/prepare-build/action.yml
@@ -22,7 +22,7 @@ runs:
   steps:
     - name: Setup PHP SDK
       id: setup-php
-      uses: php/setup-php-sdk@v0.10
+      uses: php/setup-php-sdk@v0.11
       with:
         version: ${{ inputs.version }}
         arch: ${{ inputs.arch }}
--- a/.github/workflows/package-release.yml
+++ b/.github/workflows/package-release.yml
@@ -98,6 +98,6 @@ jobs:
       fail-fast: false
       matrix:
         # Note: keep this in sync with the Windows matrix in tests.yml
-        php: [ "8.1", "8.2", "8.3", "8.4" ]
+        php: [ "8.1", "8.2", "8.3", "8.4", "8.5" ]
         arch: [ x64, x86 ]
         ts: [ ts, nts ]
--- a/tests/bson/bson-decimal128-002.phpt
+++ b/tests/bson/bson-decimal128-002.phpt
@@ -4,8 +4,6 @@ MongoDB\BSON\Decimal128 NaN values
 <?php
 
 $tests = [
-    acos(8),
-    NAN,
     'nan',
     'Nan',
     'NaN',
@@ -20,8 +18,6 @@ foreach ($tests as $test) {
 ===DONE===
 <?php exit(0); ?>
 --EXPECT--
-NaN
-NaN
 NaN
 NaN
 NaN
--- a/tests/exception/bulkwriteexception-haserrorlabel-001.phpt
+++ b/tests/exception/bulkwriteexception-haserrorlabel-001.phpt
@@ -9,7 +9,6 @@ $labels = ['test', 'foo'];
 $reflection = new ReflectionClass($exception);
 
 $resultDocumentProperty = $reflection->getProperty('errorLabels');
-$resultDocumentProperty->setAccessible(true);
 $resultDocumentProperty->setValue($exception, $labels);
 
 var_dump($exception->hasErrorLabel('foo'));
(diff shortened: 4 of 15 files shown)


tests/exception/bulkwriteexception-haserrorlabel-001.phpt
```diff
@@ -9,7 +9,6 @@ $labels = ['test', 'foo'];
 $reflection = new ReflectionClass($exception);
 
 $resultDocumentProperty = $reflection->getProperty('errorLabels');
-$resultDocumentProperty->setAccessible(true);
```
jmikola, line 12: Noted that this is a no-op since 8.1 (per [docs](https://www.php.net/manual/en/reflectionproperty.setaccessible.php)), and require 8.1+ per [package.xml](https://github.com/mongodb/mongo-php-driver/blob/v2.x/bin/package.xml.in#L61); however, why does [composer.json](https://github.com/mongodb/mongo-php-driver/blob/v2.x/composer.json#L14) still say 7.4+?

alcaeus: That was an oversight in #1631, where we didn't update composer.json. We can update this (in 1.21.x and 2.1.x as the current stable branches), but it won't fix old versions. We'll just have to rely on not enough people trying to install the extension using pie on PHP 7.4 or 8.0.

.github/workflows/package-release.yml
```diff
@@ -98,6 +98,6 @@ jobs:
       fail-fast: false
       matrix:
         # Note: keep this in sync with the Windows matrix in tests.yml
-        php: [ "8.1", "8.2", "8.3", "8.4" ]
+        php: [ "8.1", "8.2", "8.3", "8.4", "8.5" ]
```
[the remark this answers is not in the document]
alcaeus: Nice catch as I hadn't updated the description yet. With RC1 out, the Zend API version has been bumped, so we can now publish Windows builds for PHP 8.5.

.github/actions/windows/prepare-build/action.yml
```diff
@@ -22,7 +22,7 @@ runs:
   steps:
     - name: Setup PHP SDK
       id: setup-php
-      uses: php/setup-php-sdk@v0.10
+      uses: php/setup-php-sdk@v0.11
```
jmikola, line 25: Noted that [v0.11](https://github.com/php/setup-php-sdk/releases/tag/v0.11) added support for PHP 8.5.

.github/workflows/package-release.yml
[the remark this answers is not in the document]
jmikola: Noted. Feel free to resolve.

tests/bson/bson-decimal128-002.phpt
```diff
@@ -4,8 +4,6 @@ MongoDB\BSON\Decimal128 NaN values
 <?php
 
 $tests = [
-    acos(8),
-    NAN,
```
jmikola, line 8: Per [our earlier Slack conversation](https://mongodb.slack.com/archives/C026UA4GF0B/p1758718647713839), you opted to remove any Decimal128 test cases that depended on type-coercion. I have no objections here, as the [documentation](https://www.php.net/manual/en/mongodb-bson-decimal128.construct.php) has always indicated that the constructor expects a string value.

I wasn't able to find any mention of the type-coercion behavioral change in PHP 8.5's [UPGRADING](https://github.com/php/php-src/blob/84ec65c5992045a52643dcb674f0d694be58e3ab/UPGRADING) file. If you happen to find the change responsible for this I'd be interested in seeing it.

tests/server/server-construct-001.phpt
```diff
@@ -20,7 +20,7 @@ $bulk->insert(array('foo' => 'bar'));
 $server = $manager->executeBulkWrite(NS, $bulk)->getServer();
 
 $expectedHost = $parsed['host'];
-$expectedPort = (integer) (isset($parsed['port']) ? $parsed['port'] : 27017);
+$expectedPort = (int) (isset($parsed['port']) ? $parsed['port'] : 27017);
```
jmikola, line 23: Noted: https://github.com/php/php-src/blob/84ec65c5992045a52643dcb674f0d694be58e3ab/UPGRADING#L364

jmikola (approved):

tests/bson/bson-decimal128-002.phpt
alcaeus: The offending commit is https://github.com/php/php-src/commit/320fe2975bfd084c0e6528c3f617043a19bb1d24, which is the result of the large PHP 8.5 deprecation vote.

PR #1900, opened 2025-12-10
PHPC-2654: Update to libmongoc 2.2.1

alcaeus:
PHPC-2654. This update also fixes PHPC-2653.


.evergreen/config/templates/build/build-libmongoc.yml
```diff
@@ -19,7 +19,7 @@
       - func: "compile driver"
         vars:
           PHP_VERSION: "%phpVersion%"
-          LIBMONGOC_VERSION: "r2.1"
+          LIBMONGOC_VERSION: "r2.2"
```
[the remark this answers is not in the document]
alcaeus: This is wrong -- libmongoc uses `master` as a development branch for the next minor version.

scripts/autotools/libmongoc/Versions.m4
```diff
@@ -12,3 +12,5 @@ AC_SUBST(libmongoc_VERSION_MAJOR)
 AC_SUBST(libmongoc_VERSION_MINOR)
 AC_SUBST(libmongoc_VERSION_PATCH)
 AC_SUBST(libmongoc_VERSION_PRERELEASE)
+AC_SUBST(MONGOC_CXX_COMPILER_ID, "Unknown")
+AC_SUBST(MONGOC_CXX_COMPILER_VERSION, "")
```
jmikola, line 16: Noted these are the defaults in libmongoc: https://github.com/mongodb/mongo-c-driver/blob/2.2.1/src/libmongoc/CMakeLists.txt#L754-L755

It doesn't seem ideal that libmongoc pollutes the handshake string with `MONGOC_CXX_COMPILER_ID` when it's "Unknown", though: https://github.com/mongodb/mongo-c-driver/blob/2.2.1/src/libmongoc/src/mongoc/mongoc-handshake.c#L494-L498

cc: @kevinAlbs

jmikola (approved):

scripts/autotools/libmongoc/Versions.m4
kevinAlbs: Suggest using an empty string instead of "Unknown" to signal "did not check". Filed CDRIVER-6175 to omit "CXX=" entirely if `MONGOC_CXX_COMPILER_ID` is empty.

Quoting https://github.com/mongodb/mongo-c-driver/pull/2114:

> If a C++ compiler is not found, `CXX=Unknown` is included. Alternatively: `CXX` could be omitted if not detected. But I expect it may help with analysis to easily distinguish between: "driver too old to check" and "checked and did not find".

alcaeus: Good point, I had copied this over from the defaults. I've changed it to an empty string.

PR #1907, opened 2026-01-09
PHPC-2661: Use php-windows-builder for Windows extension builds

alcaeus:
PHPC-2661

This PR changes the build and release process to use php-windows-builder for Windows extension builds. Since php-windows-builder provides fully built zip packages for Windows extensions, this also changes the release process slightly: instead of packaging a detached signature file for the DLL in the package, we're now uploading a detached signature file for the entire package to the release, as we already do for the pecl package.


GromNaN (dismissed):

.github/actions/windows/build/action.yml
```diff
@@ -10,24 +13,22 @@ inputs:
...
-  vs:
-    description: "The Visual Studio version"
-    value: ${{ steps.prepare-build-env.outputs.vs }}
-  build-dir:
-    description: "The build directory to be used"
-    value: ${{ steps.prepare-build-env.outputs.build-dir }}
+  run-tests:
+    description: "Whether to run tests after building"
+    required: false
+    default: "false"
```
jmikola, line 19: Why a string value instead of a bool?

.github/actions/windows/build/action.yml
```diff
@@ -10,24 +13,22 @@ inputs:
...
-        version: ${{ inputs.version }}
+        extension-ref: ${{ inputs.ref }}
+        php-version: ${{ inputs.php-version }}
         arch: ${{ inputs.arch }}
         ts: ${{ inputs.ts }}
-
-    - name: Build driver
-      shell: cmd
-      run: nmake /nologo
+        run-tests: ${{ inputs.run-tests == 'true' && true || false }}
```
jmikola, line 30: Noted that you just look for a string value here. What is `&& true || false` accomplishing?

.github/workflows/build-windows-packages.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: "Build Driver"
+        id: build-driver
+        uses: ./.github/actions/windows/build
+        with:
+          ref: ${{ inputs.ref }}
+          php-version: ${{ matrix.php-version }}
+          arch: ${{ matrix.arch }}
+          ts: ${{ matrix.ts }}
+          run-tests: false
```
jmikola, line 55: Here you are passing a bool and overriding the default string value (`"false"`).

.github/workflows/build-windows-packages.yml
```diff
@@ -0,0 +1,99 @@
...
+          filenames: php_mongodb*.zip
+
+      - name: "Move signatures from release assets folder"
+        run: |
+          mv ${RELEASE_ASSETS}/php_mongodb*.sig artifacts/
+
+      - name: "Upload assets to release"
+        if: ${{ inputs.upload_release_assets }}
+        run: |
+          gh release upload ${{ inputs.version }} artifacts/php_mongodb* --clobber
```
jmikola, line 99: Probably not necessary but you could restrict the extension to "zip" and "sig". I assume the directory starts empty, though.

jmikola (approved):
Some questions on the bool option being passed around, but this looks fine otherwise.

.github/actions/windows/build/action.yml
alcaeus: Inputs for composite actions are always strings, I wanted to make this explicit here.

.github/actions/windows/build/action.yml
alcaeus: This is a ternary since I don't trust type coercion in YAML. Could potentially be simplified but I didn't want to run the risk of it misbehaving

.github/workflows/build-windows-packages.yml
alcaeus: It does :)

PR #1911, opened 2026-01-21
PHPC-2664: Drop Windows x86 builds

alcaeus:
PHPC-2664

Before php-windows-builder, we could use the x64 toolset for x86 builds, working around the heap space error we're receiving. With php-windows-builder, this no longer works. Since x86 is pretty much dead at this point, we will stop providing prebuilt x86 binaries for Windows.


jmikola (approved):
Good night, sweet prince.

PR #1940, opened 2026-02-10
PHPC-2505: Fix foreach after GC invocation

### LINKED ISSUE #1776: foreach over ObjectId and Garbage Collection

jmikola:
https://jira.mongodb.org/browse/PHPC-2505

Supersedes #1850 and fixes #1776

@danog: I ended up rebasing your original PR and making a number of revisions, but preserved your authorship. Thanks for your initial work on this, and all of the research on the php-src repo.

---

As previously reported, iterating objects with custom get_properties handlers (i.e. serializable classes) was inconsistent depending on whether or not garbage collection had occurred.

Remove unconditional invocation of zend_std_get_properties from the extension's standard get_gc handler. This prevents population of properties on the internal zend_object. Serializable classes that define a custom get_properties handler (also used for var_export) will now define a separate php_properties HashTable. Additional property handlers (including get_gc) will operate exclusively on php_properties.

Long-term, get_properties may still be replaced with get_properties_for (see: PHPC-1622). Doing so may allow us to eliminate get_properties in most cases.

This commit also utilizes zend_hash_release() from PHP 8.0 to free HashTables. Additional instances of zend_hash_destroy() can be revised in PHPC-2682.

### DIFF
--- a/src/BSON/Binary.c
+++ b/src/BSON/Binary.c
@@ -225,8 +225,11 @@ static void php_phongo_binary_free_object(zend_object* object)
 	}
 
 	if (intern->properties) {
-		zend_hash_destroy(intern->properties);
-		FREE_HASHTABLE(intern->properties);
+		zend_hash_release(intern->properties);
+	}
+
+	if (intern->php_properties) {
+		zend_hash_release(intern->php_properties);
 	}
 }
 
@@ -320,6 +323,8 @@ static HashTable* php_phongo_binary_get_properties(zend_object* object)
 	return php_phongo_binary_get_properties_hash(object, false);
 }
 
+PHONGO_DEFINE_PROPERTY_HANDLERS(binary, Z_OBJ_BINARY);
+
 void php_phongo_binary_init_ce(INIT_FUNC_ARGS)
 {
 	php_phongo_binary_ce                = register_class_MongoDB_BSON_Binary(php_phongo_binary_interface_ce, php_phongo_json_serializable_ce, php_phongo_type_ce, zend_ce_stringable);
@@ -332,6 +337,8 @@ void php_phongo_binary_init_ce(INIT_FUNC_ARGS)
 	php_phongo_handler_binary.get_properties = php_phongo_binary_get_properties;
 	php_phongo_handler_binary.free_obj       = php_phongo_binary_free_object;
 	php_phongo_handler_binary.offset         = XtOffsetOf(php_phongo_binary_t, std);
+
+	PHONGO_ASSIGN_PROPERTY_HANDLERS(binary);
 }
 
 bool phongo_binary_new(zval* object, const char* data, size_t data_len, bson_subtype_t type)
--- a/src/phongo_compat.h
+++ b/src/phongo_compat.h
@@ -99,9 +99,7 @@
 #define PHONGO_RETVAL_SMART_STR(val) RETVAL_STRINGL(ZSTR_VAL((val).s), ZSTR_LEN((val).s));
 #define ZVAL_STATIC_INIT \
 	{                    \
-		{                \
-			0            \
-		}                \
+		{ 0 }            \
 	}
 
 #define ADD_NEXT_INDEX_INT64_OBJ(_zv, _value) \
--- a/src/phongo_structs.h
+++ b/src/phongo_structs.h
@@ -122,12 +122,14 @@ typedef struct {
 typedef struct {
 	mongoc_read_concern_t* read_concern;
 	HashTable*             properties;
+	HashTable*             php_properties;
 	zend_object            std;
 } php_phongo_readconcern_t;
 
 typedef struct {
 	mongoc_read_prefs_t* read_preference;
 	HashTable*           properties;
+	HashTable*           php_properties;
 	zend_object          std;
 } php_phongo_readpreference_t;
 
@@ -141,12 +143,14 @@ typedef struct {
 typedef struct {
 	mongoc_server_api_t* server_api;
 	HashTable*           properties;
+	HashTable*           php_properties;
 	zend_object          std;
 } php_phongo_serverapi_t;
 
 typedef struct {
 	mongoc_server_description_t* server_description;
 	HashTable*                   properties;
+	HashTable*                   php_properties;
 	zend_object                  std;
 } php_phongo_serverdescription_t;
 
@@ -160,12 +164,14 @@ typedef struct {
 typedef struct {
 	mongoc_topology_description_t* topology_description;
 	HashTable*                     properties;
+	HashTable*                     php_properties;
 	zend_object                    std;
 } php_phongo_topologydescription_t;
 
 typedef struct {
-	HashTable*              properties;
 	mongoc_write_concern_t* write_concern;
+	HashTable*              properties;
+	HashTable*              php_properties;
 	zend_object             std;
 } php_phongo_writeconcern_t;
 
@@ -197,12 +203,14 @@ typedef struct {
 	size_t      data_len;
 	uint8_t     type;
 	HashTable*  properties;
+	HashTable*  php_properties;
 	zend_object std;
 } php_phongo_binary_t;
 
 typedef struct {
 	bson_t*     bson;
 	HashTable*  properties;
+	HashTable*  php_properties;
 	zend_object std;
 } php_phongo_packedarray_t;
 
@@ -220,6 +228,7 @@ typedef struct {
 	size_t      key;
 	zval        current;
 	HashTable*  properties;
+	HashTable*  php_properties;
 	zend_object std;
 } php_phongo_iterator_t;
 
@@ -228,6 +237,7 @@ typedef struct {
 	size_t      ref_len;
 	char        id[25];
 	HashTable*  properties;
+	HashTable*  php_properties;
 	zend_object std;
 } php_phongo_dbpointer_t;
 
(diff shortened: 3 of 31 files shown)


src/BSON/Binary.c
```diff
@@ -225,8 +225,11 @@ static void php_phongo_binary_free_object(zend_object* object)
...
 
 	if (intern->properties) {
-		zend_hash_destroy(intern->properties);
-		FREE_HASHTABLE(intern->properties);
+		zend_hash_release(intern->properties);
+	}
+
+	if (intern->php_properties) {
+		zend_hash_release(intern->php_properties);
 	}
```
alcaeus, line 233: Refactoring suggestion: we could extract this to a `PHONGO_RELEASE_INTERN_PROPERTIES` macro as this code is the same in all classes. Feel free to defer this to a follow-up ticket.

src/phongo_compat.h
```diff
@@ -99,9 +99,7 @@
 #define PHONGO_RETVAL_SMART_STR(val) RETVAL_STRINGL(ZSTR_VAL((val).s), ZSTR_LEN((val).s));
 #define ZVAL_STATIC_INIT \
 	{                    \
-		{                \
-			0            \
-		}                \
+		{ 0 }            \
```
alcaeus, line 102: Note that this diff is causing the coding standards check to break. I'm not sure what the cause of this is, but it would seem that the version of `clang-format` installed on MacOS produces different output of the one on GitHub Actions.

src/phongo_structs.h
```diff
@@ -160,12 +164,14 @@ typedef struct {
 typedef struct {
 	mongoc_topology_description_t* topology_description;
 	HashTable*                     properties;
+	HashTable*                     php_properties;
 	zend_object                    std;
 } php_phongo_topologydescription_t;
 
 typedef struct {
-	HashTable*              properties;
```
alcaeus, line 167: I'm assuming the order of struct members doesn't matter as long as `std` is the last one? I vaguely remember there being a difference between PHP 5 and PHP 7.

tests/bson/bug1598-002.phpt
```diff
@@ -1,8 +1,5 @@
 --TEST--
 PHPC-1598: BSON type get_gc should delegate to zend_std_get_properties
---SKIPIF--
-<?php require __DIR__ . "/../utils/basic-skipif.inc"; ?>
-<?php skip_if_php_version('>=', '8.1.99'); ?>
```
alcaeus, line 5: I don't remember the reason why this test only applied on PHP 8.2+, but what about these changes makes it applicable to all versions?

alcaeus (changes requested):
I left some comments in the code. Besides that, I'm wondering whether this is in fact the right approach here. Part of the problem seems to be that we're treating our "debug" properties as actual properties, which requires us to duplicate this internal handling here. However, it sounds like implementing `get_properties_for` would (at least partially) resolve this. My worry is that with this implementation, we're adding custom dynamic property access for all of our classes, which will block us from effectively adding public properties for things (e.g. for `Server` instances) in the future. The implementation of `write_property` also bypasses the default behaviour of newer PHP versions, which emit a deprecation notice about defining dynamic properties on objects that aren't explicitly marked as supporting them:
```
php > $rp = new MongoDB\Driver\ReadPreference('primary');
php > $rp->foo = 'bar';

Deprecated: Creation of dynamic property MongoDB\Driver\ReadPreference::$foo is deprecated in php shell code on line 1

Call Stack:
   13.9274     426584   1. {main}() php shell code:0
```

In my opinion, we should not introduce this custom property behaviour, but only focus on ensuring that we don't go out of scope if someone holds a reference to our properties, e.g. for iteration.

src/BSON/Binary.c
jmikola: I did originally have a `PHONGO_FREE_PROPS` macro but removed it before finalizing the commit because I felt it made the `free_object` handler a bit less clear. And I didn't like it was selectively used only in classes that defined both `properties` _and_ `php_properties` on their internal struct. Happy to add that back, though, as it would remove duplication.

src/phongo_compat.h
jmikola: Very likely. I can revert this, as the file itself wasn't actually relevant to the PR and I debated including this.

src/phongo_structs.h
jmikola: `std` must be last for the `XtOffsetOf` calculations to work. The order of other properties should not matter, though.

alcaeus:
Moving this back to draft state as we'll want to implement the `get_properties_for` handler in PHPC-1622 first. Either way, dynamic properties aren't something we're necessarily concerned with, but iterating the default properties of our objects should work without issues.

tests/bson/bug2505-002.phpt
```diff
@@ -0,0 +1,69 @@
+-TEST--
```
GromNaN, line 1:
```suggestion
--TEST--
```

GromNaN:
I'm taking over this PR. I'll try running the tests from https://github.com/mongodb/mongo-php-driver/pull/1850

PR #1952, opened 2026-03-13
PHPC-1971: Use debug builds of PHP for testing

alcaeus:
PHPC-1971

Using debug builds ensures we catch errors due to wrong return types, as PHP only validates return types for built-in methods in debug builds, but not other builds.

### DIFF
--- a/.github/actions/linux/build/action.yml
+++ b/.github/actions/linux/build/action.yml
@@ -14,7 +14,7 @@ runs:
     - name: "Install PHP"
       uses: "shivammathur/setup-php@v2"
       with:
-        php-version: "${{ inputs.version }}"
+        php-version: "${{ inputs.version }}-debug"
         # Only install required extensions
         extensions: "none,date,json,spl,standard,xml"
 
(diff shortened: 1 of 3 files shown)


alcaeus:
Note: the PHP 8.6 failure is unrelated and caused by PHP calling the `interface_gets_implemented` handlers in a different order. I'll create a separate pull request to make the test more resilient.

GromNaN:
> Note: the PHP 8.6 failure is unrelated and caused by PHP calling the `interface_gets_implemented` handlers in a different order. I'll create a separate pull request to make the test more resilient.

Please give us the opportunity to learn by working on the extension.

GromNaN:
Is it possible for the build to succeed in php-debug mode but fail in php-no-debug mode? For example, if we’re using features specific to php-debug.

alcaeus:
> Is it possible for the build to succeed in php-debug mode but fail in non-debug mode? For example, if we’re using features specific to php-debug.

Not that I'm aware of -- there's a `config_get_hash` function that's only available in debug PHP, but I don't think we're gonna ever use that as it deals with returning ini values. The other differences all relate to detecting memory leaks and performing extra checks.

GromNaN (approved):

PR #1954, opened 2026-03-13
Stabilize enum interface error handling and tests

GromNaN:
Follow-up https://github.com/mongodb/mongo-php-driver/pull/1953#issuecomment-4056107155

### DIFF
--- a/src/BSON/Unserializable.c
+++ b/src/BSON/Unserializable.c
@@ -24,7 +24,15 @@ zend_class_entry* php_phongo_unserializable_ce;
 static int php_phongo_implement_unserializable(zend_class_entry* interface, zend_class_entry* class_type)
 {
 	if (class_type->ce_flags & ZEND_ACC_ENUM) {
-		zend_error_noreturn(E_ERROR, "Enum class %s cannot implement interface %s", ZSTR_VAL(class_type->name), ZSTR_VAL(interface->name));
+		zend_class_entry* error_interface = interface;
+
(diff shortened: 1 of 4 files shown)


src/BSON/Unserializable.c
```diff
@@ -24,7 +24,15 @@ zend_class_entry* php_phongo_unserializable_ce;
...
 {
 	if (class_type->ce_flags & ZEND_ACC_ENUM) {
-		zend_error_noreturn(E_ERROR, "Enum class %s cannot implement interface %s", ZSTR_VAL(class_type->name), ZSTR_VAL(interface->name));
+		zend_class_entry* error_interface = interface;
+
+		/* Persistable extends Unserializable, so this handler may be called for
+		 * Persistable enums; prefer Persistable in the fatal message when applicable. */
+		if (interface == php_phongo_unserializable_ce && php_phongo_persistable_ce && instanceof_function(class_type, php_phongo_persistable_ce)) {
+			error_interface = php_phongo_persistable_ce;
+		}
```
GromNaN, line 33: Still not convinced to keep this part. That's the `Unserializable` behavior that is not supported.

GromNaN: If `Unserializable::bsonUnserialize()` was static, then it could return an enum case and this would not be an issue.

alcaeus (approved):
LGTM. We can revisit the discussion around `Unserializable`, although I think that codecs are the better variant to deal with BSON data than the `Persistable` solution.

PR #1956, opened 2026-03-19
Add configuration for Claude Code

alcaeus:
This PR adds configuration for Claude Code to provide better assistance for developers. It also includes a code review agent to help with code reviews before creating pull requests, especially around memory management and common security issues.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -35,7 +35,12 @@ make test
 
 # Run a single test file
 make test TESTS=tests/path/to/test.phpt
+
+# TESTS accepts a glob pattern:
+make test TESTS="tests/path/to/pattern*.phpt"
 ```
+
+For failed tests, all intermediate files (e.g. expected output, actual output, diff) are left alongside the test file. This is helpful for debugging purposes. 
 
 Key environment variables for testing:
 - `MONGODB_URI` – connection string (default: `mongodb://127.0.0.1/`)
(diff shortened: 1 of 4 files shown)


.claude/CLAUDE.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+## Testing
+
+```sh
+# Run all tests (requires a running MongoDB server on localhost:27017)
+make test
+
+# Run a single test file
+make test TESTS=tests/path/to/test.phpt
```
GromNaN, line 37: should you add that in case of error the expected, out, diff files are left near to the test file, so the agent can read it.

Also, `TESTS` accepts glob and list of files. This should be in the example.

.claude/CLAUDE.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Stub Files and Arginfo
+
+PHP method signatures are defined in `*.stub.php` files alongside each class's `.c` file. After changing a stub file, regenerate the arginfo:
+
+```sh
+./build/gen_stub.php
+```
+
+This requires `phpize` to have been run with **PHP 8.2** to use all features. Commit both the stub and the generated arginfo file.
```
GromNaN, line 60: PHP 8.2 is required, or the minimum version?
I would not indicate the version here, to not have to maintain it.

alcaeus: PHP 8.2 is the version we have to use. I remember some issues with `gen_stub.php` on PHP 8.3 producing arginfo that no longer works on PHP 8.2. There is a legacy arginfo mode (see https://php.github.io/php-src/miscellaneous/stubs.html#maintaining-backward-compatibility), but we don't use it (yet).

.claude/CLAUDE.md
alcaeus: Added

GromNaN (approved):

PR #1957, opened 2026-03-20
Harden code after AI-assisted review

alcaeus:
This PR is the result of an AI-assisted coding session where I tested Claude Code to help review the code and look for memory leaks and potential security issues. None of the issues found here is critical or exploitable, but the memory leaks and other potential memory issues could become a problem in future.

### DIFF
--- a/src/BSON/Binary.c
+++ b/src/BSON/Binary.c
@@ -371,7 +371,7 @@ static void phongo_binary_init_vector_from_float32_array(phongo_binary_t* intern
 
 	if (!BSON_APPEND_VECTOR_FLOAT32_UNINIT(&doc, "vector", vector_len, &view)) {
 		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "BSON_APPEND_VECTOR_FLOAT32_UNINIT failed for vector of size %zu", vector_len);
-		return;
+		goto cleanup;
 	}
 
 	zval*  val;
@@ -381,14 +381,14 @@ static void phongo_binary_init_vector_from_float32_array(phongo_binary_t* intern
 	{
 		if (Z_TYPE_P(val) != IS_DOUBLE) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected vector[%zu] to be a float, %s given", i, zend_zval_type_name(val));
-			return;
+			goto cleanup;
 		}
 
 		float v = (float) Z_DVAL_P(val);
 
 		if (!bson_vector_float32_view_write(view, &v, 1, i)) {
 			phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "bson_vector_float32_view_write failed to write vector[%zu]", i);
-			return;
+			goto cleanup;
 		}
 
 		i += 1;
@@ -396,6 +396,9 @@ static void phongo_binary_init_vector_from_float32_array(phongo_binary_t* intern
 	ZEND_HASH_FOREACH_END();
 
 	phongo_binary_init_vector_from_bson_key(intern, &doc, "vector");
+
+cleanup:
+	bson_destroy(&doc);
 }
 
 static void phongo_binary_init_vector_from_int8_array(phongo_binary_t* intern, HashTable* vector)
@@ -412,7 +415,7 @@ static void phongo_binary_init_vector_from_int8_array(phongo_binary_t* intern, H
 
 	if (!BSON_APPEND_VECTOR_INT8_UNINIT(&doc, "vector", vector_len, &view)) {
 		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "BSON_APPEND_VECTOR_INT8_UNINIT failed for vector of size %zu", vector_len);
-		return;
+		goto cleanup;
 	}
 
 	zval*  val;
@@ -422,19 +425,19 @@ static void phongo_binary_init_vector_from_int8_array(phongo_binary_t* intern, H
 	{
 		if (Z_TYPE_P(val) != IS_LONG) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected vector[%zu] to be an integer, %s given", i, zend_zval_type_name(val));
-			return;
+			goto cleanup;
 		}
 
 		if (Z_LVAL_P(val) < INT8_MIN || Z_LVAL_P(val) > INT8_MAX) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected vector[%zu] to be a signed 8-bit integer, %" PHONGO_LONG_FORMAT " given", i, Z_LVAL_P(val));
-			return;
+			goto cleanup;
 		}
 
 		int8_t v = (int8_t) Z_LVAL_P(val);
 
 		if (!bson_vector_int8_view_write(view, &v, 1, i)) {
 			phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "bson_vector_int8_view_write failed to write vector[%zu]", i);
-			return;
+			goto cleanup;
 		}
 
 		i += 1;
@@ -442,6 +445,9 @@ static void phongo_binary_init_vector_from_int8_array(phongo_binary_t* intern, H
 	ZEND_HASH_FOREACH_END();
 
 	phongo_binary_init_vector_from_bson_key(intern, &doc, "vector");
+
+cleanup:
+	bson_destroy(&doc);
 }
 
 static void phongo_binary_init_vector_from_packed_bit_array(phongo_binary_t* intern, HashTable* vector)
@@ -458,7 +464,7 @@ static void phongo_binary_init_vector_from_packed_bit_array(phongo_binary_t* int
 
 	if (!BSON_APPEND_VECTOR_PACKED_BIT_UNINIT(&doc, "vector", vector_len, &view)) {
 		phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "BSON_APPEND_VECTOR_PACKED_BIT_UNINIT failed for vector of size %zu", vector_len);
-		return;
+		goto cleanup;
 	}
 
 	zval*  val;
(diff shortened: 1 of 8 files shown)


src/BSON/Binary.c
```diff
@@ -381,21 +381,24 @@ static void phongo_binary_init_vector_from_float32_array(phongo_binary_t* intern
...
 		}
 
 		i += 1;
 	}
 	ZEND_HASH_FOREACH_END();
 
 	phongo_binary_init_vector_from_bson_key(intern, &doc, "vector");
+
+cleanup:
+	bson_destroy(&doc);
```
alcaeus, line 401: While the `bson_t` is stack allocated and thus freed automatically after leaving this function, this doesn't cover every use case. `bson_t` is aligned to 128 bytes, so once the BSON document gets sufficiently large it will need to allocate memory on the heap. Without the explicit `bson_destroy` here, we'd leak this allocated memory if an error occurs after we allocate that memory. This applies to all init functions in this file.

src/BSON/Binary.c
```diff
@@ -381,21 +381,24 @@ static void phongo_binary_init_vector_from_float32_array(phongo_binary_t* intern
...
-			return;
+			goto cleanup;
 		}
 
 		float v = (float) Z_DVAL_P(val);
 
 		if (!bson_vector_float32_view_write(view, &v, 1, i)) {
 			phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "bson_vector_float32_view_write failed to write vector[%zu]", i);
-			return;
+			goto cleanup;
```
GromNaN, line 391: I see that the `goto cleanup` is a common pattern in other functions. We could have called `bson_destroy(&doc); return;`  directly, but that becomes more verbose in the event that we add more instructions.

src/BSON/Javascript.c
```diff
@@ -46,6 +46,14 @@ static bool phongo_javascript_init(phongo_javascript_t* intern, const char* code
 	if (scope && (Z_TYPE_P(scope) == IS_OBJECT || Z_TYPE_P(scope) == IS_ARRAY)) {
 		intern->scope = bson_new();
 		phongo_zval_to_bson(scope, PHONGO_BSON_NONE, intern->scope, NULL);
+
+		/* If an exception was thrown during BSON conversion, clean up and return false */
+		if (EG(exception)) {
+			efree(intern->code);
+			intern->code = NULL;
+			bson_clear(&intern->scope);
+			return false;
```
GromNaN: Returning `false` seems to be a bug fix. Should it be covered by a test?

src/MongoDB/ClientEncryption.c
```diff
@@ -469,6 +469,7 @@ static PHP_METHOD(MongoDB_Driver_ClientEncryption, rewrapManyDataKey)
 	}
 
 	bson_destroy(&filter);
+	bson_destroy(&reply);
```
alcaeus, line 472: `RETVAL_BSON_T` converts the `bson_t` to a `zval` and calls `RETVAL_ZVAL`. After that, we need to call `bson_destroy` to free any potential heap-allocated memory.

src/MongoDB/ClientEncryption.c
```diff
@@ -625,6 +626,7 @@ static mongoc_client_encryption_opts_t* phongo_clientencryption_opts_from_zval(z
 
 		phongo_zval_to_bson(tls_options, PHONGO_BSON_NONE, &bson_options, NULL);
 		if (EG(exception)) {
+			bson_destroy(&bson_options);
```
alcaeus, line 629: Need to free potentially heap-allocated memory from `bson_t`.

src/phongo_bson.c
```diff
@@ -1287,7 +1299,7 @@ static void map_add_field_path_element(phongo_bson_typemap* map, phongo_field_pa
 	/* Make sure we have allocated enough */
 	if (map->field_paths.allocated_size < map->field_paths.size + 1) {
 		map->field_paths.allocated_size += PHONGO_FIELD_PATH_EXPANSION;
-		map->field_paths.map = erealloc(map->field_paths.map, sizeof(phongo_field_path_map_element) * map->field_paths.allocated_size);
+		map->field_paths.map = erealloc(map->field_paths.map, sizeof(phongo_field_path_map_element*) * map->field_paths.allocated_size);
```
alcaeus, line 1301: This was allocating too much memory, namely the size of an entire element structure instead of just a pointer to it. While the memory was correctly freed later on, it's still unnecessary and can cause issues.

src/phongo_bson.c
```diff
@@ -1339,10 +1351,10 @@ static bool phongo_bson_state_add_field_path(phongo_bson_typemap* map, char* fie
 			return false;
 		}
 
-		tmp = calloc(1, segment_end - ptr + 1);
+		tmp = ecalloc(1, segment_end - ptr + 1);
```
alcaeus, line 1353: We always want to use PHP's memory management functions.

src/phongo_bson.c
```diff
@@ -98,14 +98,23 @@ char* phongo_field_path_as_string(phongo_field_path* field_path)
...
+		size_t element_len;
+
 		if (!field_path->elements[i]) {
 			continue;
 		}
+
+		element_len = strlen(field_path->elements[i]);
+
+		/* Assert that we have enough space in the buffer for this element plus separator */
+		BSON_ASSERT((size_t) (ptr - path) + element_len + 1 < length);
```
alcaeus, line 110: Defensive hardening -- we want to make sure we're not running into a buffer overflow here.

src/phongo_bson.c
```diff
@@ -122,11 +131,13 @@ phongo_field_path* phongo_field_path_alloc(bool owns_elements)
 
 void phongo_field_path_free(phongo_field_path* field_path)
 {
-	if (field_path->owns_elements) {
+	if (field_path->owns_elements && field_path->elements) {
```
alcaeus, line 134: Defensive programming: if for some reason `elements` isn't allocated, we don't want to access invalid memory.

src/phongo_bson_encode.c
```diff
@@ -602,12 +602,21 @@ static void phongo_zval_to_bson_internal(zval* data, phongo_field_path* field_pa
...
+		if (!bson_iter_init_find(&iter, bson, "_id")) {
+			phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE, "Error copying \"_id\" field from encoded document");
+
+			goto cleanup;
+		}
+
 		*bson_out = bson_new();
 
-		if (bson_iter_init_find(&iter, bson, "_id") && !bson_append_iter(*bson_out, NULL, 0, &iter)) {
+		if (!bson_append_iter(*bson_out, NULL, 0, &iter)) {
```
alcaeus, line 615: Previously, we'd allocate a `bson_t` and throw an exception if we found an `_id` field but failed to copy it; however, we didn't free the allocated memory causing a potential leak.

The new logic splits this, and throws before allocating memory if we don't find an identifier, but also freeing memory if we fail to copy an identifier. This is a small behavioural change, as we now throw if `PHONGO_BSON_RETURN_ID` is set and we don't find an identifier, where we previously just did nothing. A quick review reveals that we always combine `PHONGO_BSON_RETURN_ID` with `PHONGO_BSON_ADD_ID`, so there shouldn't be instances where we now throw even though we previously didn't. However, I believe that this behaviour is actually more correct and we should not use `PHONGO_BSON_RETURN_ID` when we're not sure if an ID was added. I'll see if it would make sense to have `PHONGO_BSON_RETURN_ID` imply `PHONGO_BSON_ADD_ID`.

src/phongo_bson_encode.h
```diff
@@ -30,6 +30,5 @@ typedef enum {
 
 void phongo_zval_to_bson(zval* data, phongo_bson_flags_t flags, bson_t* bson, bson_t** bson_out);
 bool phongo_zval_to_bson_value(zval* data, bson_value_t* value);
-void phongo_bson_append_zval(bson_t* bson, const char* key, long key_len, zval* value);
```
alcaeus, line 33: This declaration was added in #1387 without an implementation. This was likely the result of me not cleaning up properly after refactoring.

tests/bulk/bulkwrite-ctor-comment_error-002.phpt
```diff
@@ -0,0 +1,18 @@
...
+echo throws(function() {
+    new MongoDB\Driver\BulkWrite(['comment' => ['key' => "\xc3\x28"]]);
+}, MongoDB\Driver\Exception\UnexpectedValueException::class), "\n";
+
+?>
+===DONE===
+<?php exit(0); ?>
+--EXPECTF--
+OK: Got MongoDB\Driver\Exception\UnexpectedValueException
+Detected invalid UTF-8 for field path "data.key": %s
```
alcaeus, line 17: Note: the output is `data.key` as this data goes through `phongo_zval_to_bson_value`. For objects and arrays, it creates an array with a single `data` key containing the `zval` and calls `phongo_zval_to_bson` on that object, producing a BSON document with a single `data` field. It then copies a `bson_value_t` from the resulting `bson_t`. This avoids re-implementing a lot of encoding logic, but creates the additional `data` prefix on any error paths.

alcaeus (commented):
Left some comments to explain the changes.

src/BSON/Binary.c
alcaeus: Yes, the `cleanup` label is a common strategy in the extension. When we use a label, we no longer use `return` to ensure we're cleaning up everything without having to repeat a bunch of cleanup logic as we allocate memory throughout the function.

src/BSON/Javascript.c
alcaeus: There was a previous case where the function would return `false`, even though the return value isn't used anywhere. `phongo_javascript_init` is called from the constructor and from the `__unserialize` method (through `phongo_javascript_init_from_hash`), which both correctly propagate the exception thrown by `phongo_zval_to_bson` without needing to check the return value.

However, I have removed this code in between: since the `free_obj` handler correctly cleans up memory and is invoked by PHP once an object goes out of scope (which it does if the constructor or `__unserialize` throw), this memory is properly cleaned up in the exception case and this logic is unnecessary.

GromNaN (approved):

PR #1958, opened 2026-03-23
PHPC-2160: Extract calls to bsonSerialize and remove unnecessary check function

alcaeus:
PHPC-2160

The ticket calls for removing the duplicate invocation in `phongo_zval_to_bson_internal` and `phongo_bson_append_object`. Implementing this would require a more significant refactoring of the entire encoding logic, which may not be warranted. I've extracted the calls to `bsonSerialize` and removed the now unnecessary check function (which is now enforced by the return type on `bsonSerialize` directly), but refrained from further modifying the functions to remove all of the duplication.

### DIFF
--- a/src/phongo_bson_encode.c
+++ b/src/phongo_bson_encode.c
@@ -88,51 +88,18 @@ static int phongo_is_array_or_document(zval* val)
 	return IS_ARRAY;
 }
 
-/* Checks the return type of a bsonSerialize() method. Returns true on
- * success; otherwise, throws an exception and returns false.
- *
- * TODO: obsolete once the tentative return type in Serializable::bsonSerialize is enforced.
- */
-static inline bool phongo_check_bson_serialize_return_type(zval* retval, zend_class_entry* ce)
+static bool phongo_bson_encode_serializable(zval* object, zval* out_data)
 {
-	if (instanceof_function(ce, phongo_persistable_ce)) {
-		// Instances of Persistable must return an array, stdClass, or MongoDB\BSON\Document
-		if (
-			Z_TYPE_P(retval) != IS_ARRAY && !(Z_TYPE_P(retval) == IS_OBJECT && (instanceof_function(Z_OBJCE_P(retval), zend_standard_class_def) || instanceof_function(Z_OBJCE_P(retval), phongo_document_ce)))) {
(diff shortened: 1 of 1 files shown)


src/phongo_bson_encode.c
```diff
@@ -88,51 +88,18 @@ static int phongo_is_array_or_document(zval* val)
 	return IS_ARRAY;
 }
 
-/* Checks the return type of a bsonSerialize() method. Returns true on
- * success; otherwise, throws an exception and returns false.
- *
- * TODO: obsolete once the tentative return type in Serializable::bsonSerialize is enforced.
```
paulinevos, line 94: Is this TODO obsolete now? I might not have enough reading comprehension of the code here but it feels like it might still apply?

src/phongo_bson_encode.c
```diff
@@ -88,51 +88,18 @@ static int phongo_is_array_or_document(zval* val)
...
-			phongo_throw_exception(PHONGO_ERROR_UNEXPECTED_VALUE,
-								   "Expected %s::%s() to return an array, stdClass, or %s, %s given",
-								   ZSTR_VAL(ce->name),
-								   BSON_SERIALIZE_FUNC_NAME,
-								   ZSTR_VAL(phongo_document_ce->name),
-								   zend_zval_type_name(retval));
-			return false;
-		}
+	ZVAL_UNDEF(out_data);
+	zend_call_method_with_0_params(Z_OBJ_P(object), NULL, NULL, BSON_SERIALIZE_FUNC_NAME, out_data);
```
paulinevos, line 94: Or is this what that is?

e: I think I understand now that this is what that is

paulinevos (approved):
Assuming we have test cases that cover this

src/phongo_bson_encode.c
alcaeus: This was a leftover from the 1.x series. The return type for `Serializable::bsonSerialize` was marked as tentative and not enforced, which means that somebody could've returned an invalid type from that method. To prevent this, we enforced the return type in the extension.
In 2.0, the return type became definitive (seen in the [stub file](https://github.com/mongodb/mongo-php-driver/blob/bff1a60b906a96c7adc4fe166bd7bc41381a4be4/src/BSON/Serializable.stub.php#L12)), which means we no longer need to enforce this ourselves. We just hadn't removed this until now.

PR #1959, opened 2026-03-23
PHPC-2024: Introduce macros for simpler class declarations

alcaeus:
PHPC-2024

Class declarations in the extension need the following parts:
* an external declaration for a `zend_class_entry*` variable
* an external declaration for the function to init the class entry
* (optional) a function to get the underlying struct for an object
* (optional) helper macros to get the underlying struct for a `zval*` or `zend_object*`

Until now, these have been declared using copy/paste/replace, which is unnecessarily complex. This PR introduces helper macros:
* `PHONGO_DECLARE_CLASS` declares the `zend_class_entry*` variable and `init_ce` functions
* `PHONGO_DECLARE_CLASS_WITH_HELPERS` declares the above as well as functions to get the underlying structure for an object.

As a result of this, the `Z_<class>_OBJ_P` and `Z_OBJ_<class>` macros change to inline functions, as we can't declare macros from inside a macro body. This doesn't change anything in the final product, but is worth noting.


src/phongo_classes.h
```diff
@@ -19,465 +19,178 @@
...
-extern void phongo_writeresult_init_ce(INIT_FUNC_ARGS);
+#define PHONGO_DECLARE_CLASS_WITH_HELPERS(ucname, name) \
+	PHONGO_DECLARE_CLASS(name)                          \
+	CLASS_FETCH_OBJ_DECL(name);                         \
+	Z_OBJ_P_FETCH_DECL(ucname, name);                   \
+	Z_OBJ_DECL(ucname, name);
 
-extern void phongo_cursor_interface_init_ce(INIT_FUNC_ARGS);
+/* Export zend_class_entry dependencies, which are initialized in MINIT */
+extern zend_class_entry* phongo_json_serializable_ce;
```
alcaeus, line 69: This is declared and populated in phongo.c and is mostly used by BSON classes when registering class information.

src/phongo_classes.h
```diff
@@ -19,465 +19,178 @@
...
-extern void phongo_encryptionexception_init_ce(INIT_FUNC_ARGS);
-extern void phongo_exception_init_ce(INIT_FUNC_ARGS);
-extern void phongo_executiontimeoutexception_init_ce(INIT_FUNC_ARGS);
-extern void phongo_invalidargumentexception_init_ce(INIT_FUNC_ARGS);
-extern void phongo_logicexception_init_ce(INIT_FUNC_ARGS);
-extern void phongo_runtimeexception_init_ce(INIT_FUNC_ARGS);
-extern void phongo_serverexception_init_ce(INIT_FUNC_ARGS);
-extern void phongo_sslconnectionexception_init_ce(INIT_FUNC_ARGS);
-extern void phongo_unexpectedvalueexception_init_ce(INIT_FUNC_ARGS);
+PHONGO_DECLARE_CLASS_WITH_HELPERS(CLIENTENCRYPTION, clientencryption);
```
alcaeus: For all classes that are used throughout the extension, we declare class definitions with the `Z_<name>_OBJ_P` and `Z_OBJ_<name>` helpers. This allows us to quickly get the internal struct (e.g. `phongo_binary_t` from a `zval*` or `zend_object*`.

src/phongo_classes.h
```diff
@@ -19,465 +19,178 @@
...
+PHONGO_DECLARE_CLASS(regex);
+CLASS_FETCH_OBJ_DECL(regex);
+PHONGO_STATIC_INLINE phongo_regex_t* Z_REGEX_OBJ_P(const zval* zv)
+{
+	return php_regex_fetch_object(Z_OBJ_P(zv));
+}
+PHONGO_STATIC_INLINE phongo_regex_t* Z_OBJ_REGEX(const zend_object* zo)
+{
+	return php_regex_fetch_object(zo);
+}
```
alcaeus: This took way longer than I care to admit. On Windows platforms with PHP < 8.4, using `PHONGO_DECLARE_CLASS_WITH_HELPERS` for the `MongoDB\BSON\Regex` class, the `##ucname##` replacement inside the macro caused issues, leading to the symbols for `Z_REGEX_OBJ_P` and `Z_OBJ_REGEX` to not be emitted. To work around this, we use `PHONGO_DECLARE_CLASS`, use the `CLASS_FETCH_OBJ_DECL` macro to generate the `php_regex_fetch_object` function (which wouldn't attempt to use `REGEX`), then declare the remaining functions manually.

src/phongo_classes.h
```diff
@@ -19,465 +19,178 @@
...
+
+PHONGO_DECLARE_CLASS(cursor_interface);
```
alcaeus: Remaining classes without a dedicated object struct are declared using `PHONGO_DECLARE_CLASS` as there's no additional data to be read from them. This includes all interfaces.

src/phongo_classes.h
```diff
@@ -19,465 +19,163 @@
...
-extern zend_class_entry* phongo_invalidargumentexception_ce;
-extern zend_class_entry* phongo_connectionexception_ce;
-extern zend_class_entry* phongo_authenticationexception_ce;
-extern zend_class_entry* phongo_encryptionexception_ce;
-extern zend_class_entry* phongo_executiontimeoutexception_ce;
-extern zend_class_entry* phongo_connectiontimeoutexception_ce;
-extern zend_class_entry* phongo_bulkwriteexception_ce;
-extern zend_class_entry* phongo_bulkwritecommandexception_ce;
+#define CLASS_FETCH_OBJ_DECL(name)                                                            \
+	static inline phongo_##name##_t* php_##name##_fetch_object(const zend_object* obj) \
```
alcaeus: The `Z_<name>_OBJ_P` and `Z_OBJ_<name>` macros are now `static inline` functions. This shouldn't make a difference in practice, but is necessary as we can't use `#define` in a macro body.

src/phongo_classes.h
```diff
@@ -19,465 +19,163 @@
...
+PHONGO_DECLARE_CLASS(logsubscriber);
+PHONGO_DECLARE_CLASS(sdamsubscriber);
+
+#undef CLASS_FETCH_OBJ_DECL
+#undef CLASS_ENTRY_DECL
+#undef CE_INIT_FUNC_DECL
+#undef Z_OBJ_P_FETCH_DECL
+#undef Z_OBJ_DECL
+#undef PHONGO_DECLARE_CLASS
+#undef PHONGO_DECLARE_CLASS_WITH_HELPERS
```
alcaeus, line 179: I'm undefining the macros here on purpose so that they don't become part of the API of the extension and can only be used inside this file. There shouldn't be a reason to declare classes outside of here. If we want to further split this file (e.g. to have BSON class declarations in a header file inside the BSON directory), we should include that header file from inside the block where the macro definitions exist.

alcaeus (commented):
Left some explanatory comments

paulinevos (approved):

PR #1960, opened 2026-03-24
PHPC-1988: base64 encode binary data when debugging

paulinevos:
As printing the literal bytes is not useful for a human debugging

### DIFF
--- a/src/BSON/Binary.c
+++ b/src/BSON/Binary.c
@@ -74,7 +74,7 @@ static bool phongo_binary_init_from_hash(phongo_binary_t* intern, HashTable* pro
 	return false;
 }
 
-static HashTable* phongo_binary_get_properties_hash(zend_object* object, bool is_temp)
+static HashTable* phongo_binary_get_properties_hash(zend_object* object, bool is_temp, bool is_debug)
 {
 	phongo_binary_t* intern;
 	HashTable*       props;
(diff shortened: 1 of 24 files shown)


paulinevos:
The one thing not covered here is the concern @jmikola mentioned:

> One concern there is that there is no indication to the user that the debug output itself has changed, so we might want to consider renaming the field to something like "data_b64" instead. 

Lmk if that's something we still want to do

alcaeus (commented):
Changes LGTM. Failures are from test that assert against `var_dump` output for binary objects. Where relevant for the test, we should update the tests; if the contents of the binary is not relevant we can use `%s` (and `%d` for the string length) to avoid havinng to base64-encode data for tests.

alcaeus (approved):

src/BSON/Binary.c
```diff
@@ -90,7 +90,12 @@ static HashTable* phongo_binary_get_properties_hash(zend_object* object, bool is
 	{
 		zval data, type;
 
-		ZVAL_STRINGL(&data, intern->data, intern->data_len);
+		if (is_debug) {
+			ZVAL_NEW_STR(&data, php_base64_encode((unsigned char*) intern->data, intern->data_len));
```
jmikola, line 94: Late to the game here and I noted [PHPC-1988](https://jira.mongodb.org/browse/PHPC-1988) was already released in 2.3.0, but was there any consideration to rename `data` to `data_b64` or something to indicate a change?

Edit: Just noticed @paulinevos reiterated my original concern in https://github.com/mongodb/mongo-php-driver/pull/1960#issuecomment-4118945736 so I assume not ✌️

PR #1961, opened 2026-03-25
Use `zend_array_is_list` in `phongo_is_array_or_document`

paulinevos:
As it has some performance improvements, and is more readable than iterating through array keys to determine if the given value is a list or not

### DIFF
--- a/src/phongo_bson_encode.c
+++ b/src/phongo_bson_encode.c
@@ -53,7 +53,6 @@ static void phongo_zval_to_bson_internal(zval* data, phongo_field_path* field_pa
 static int phongo_is_array_or_document(zval* val)
 {
 	HashTable* ht_data = HASH_OF(val);
-	int        count;
 
 	if (Z_TYPE_P(val) != IS_ARRAY) {
 		if (Z_TYPE_P(val) == IS_OBJECT && instanceof_function(Z_OBJCE_P(val), phongo_packedarray_ce)) {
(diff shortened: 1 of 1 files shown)


src/phongo_bson_encode.c
```diff
@@ -63,26 +61,8 @@ static int phongo_is_array_or_document(zval* val)
...
-				if (index != idx) {
-					return IS_OBJECT;
-				}
-			}
-			idx++;
-		}
-		ZEND_HASH_FOREACH_END();
-	} else {
-		return Z_TYPE_P(val);
+	if (! zend_array_is_list(ht_data)) {
```
GromNaN: Just to check. I'm not sure if there's a test where an empty array returns `IS_ARRAY` and an empty object returns `IS_OBJECT`. Does `zend_array_is_list` return `true` for an empty array?

src/phongo_bson_encode.c
```diff
@@ -16,7 +16,6 @@
 
 #include "bson/bson.h"
```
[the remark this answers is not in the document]
alcaeus: This comment is correct. Header files use an `#ifndef` definition to emulate an `include_once` behaviour: https://github.com/php/php-src/blob/f6f6fa9d6f4db14f80a08612a5afce50f9b492d0/main/php.h#L18

That way, you can include all header files that contain symbols used in the source file without having to worry whether it has been included before.

src/phongo_bson_encode.c
```diff
@@ -53,7 +52,6 @@ static void phongo_zval_to_bson_internal(zval* data, phongo_field_path* field_pa
 static int phongo_is_array_or_document(zval* val)
```
alcaeus, line 53: Minor: this refactoring may be an opportunity to change the signature of the method.

```suggestion
static bool phongo_is_bson_array(zval* val)
```

We return either `IS_ARRAY` or `IS_OBJECT`, although `true` and `false` would be more meaningful and would make other checks simpler. It's only used twice in this file, so this would be a small change to make while we're at it.

Feel free to create a separate ticket for this change and defer it to a second pull request.

src/phongo_bson_encode.c
paulinevos: It does: https://github.com/php/php-src/pull/6070/changes#diff-5b12cd0a570bddf0dfdc4db8c9156fbf49ee8d284de218222eb6acf850b65ccdR72

alcaeus (approved):
Change LGTM, with a suggestion on how to improve the function signature for more clarity

PR #1962, opened 2026-03-26
PHPC-2530: Deprecate hedged reads

alcaeus:
PHPC-2530

### DIFF
--- a/src/MongoDB/ReadPreference.c
+++ b/src/MongoDB/ReadPreference.c
@@ -117,6 +117,8 @@ static bool phongo_readpreference_init_from_hash(phongo_readpreference_t* intern
 		if (Z_TYPE_P(hedge) == IS_ARRAY || Z_TYPE_P(hedge) == IS_OBJECT) {
 			bson_t* hedge_doc = bson_new();
 
+			php_error_docref(NULL, E_DEPRECATED, "The \"hedge\" option is deprecated as of MongoDB 8.0 and will be removed in a future release");
+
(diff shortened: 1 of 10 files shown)


src/MongoDB/ReadPreference.stub.php
```diff
@@ -53,6 +53,7 @@ final class ReadPreference implements \MongoDB\BSON\Serializable
 
     final public function __construct(string $mode, ?array $tagSets = null, ?array $options = null) {}
 
+    /** @deprecated Hedged reads are deprecated in MongoDB 8.0 and will be removed in a future release */
```
alcaeus, line 56: This stub change adds the `ZEND_ACC_DEPRECATED` to the method arginfo, which then automatically triggers a deprecation warning whenever the method is invoked.

src/MongoDB/ReadPreference.c
```diff
@@ -117,6 +117,8 @@ static bool phongo_readpreference_init_from_hash(phongo_readpreference_t* intern
 		if (Z_TYPE_P(hedge) == IS_ARRAY || Z_TYPE_P(hedge) == IS_OBJECT) {
 			bson_t* hedge_doc = bson_new();
 
+			php_error_docref(NULL, E_DEPRECATED, "The \"hedge\" option is deprecated as of MongoDB 8.0 and will be removed in a future release");
```
alcaeus, line 120: This triggers the deprecation notice when the hedge option is used from `__set_state`; the change below does the same for constructor usage.

paulinevos (approved):

PR #1963, opened 2026-03-26
PHPC-2486: Require non-null namespace and database arguments in Manager

alcaeus:
PHPC-2486

I also audited Server.c to confirm that it uses `Z_PARAM_STRING` in all the right places.


paulinevos (approved):
I know you audited usages, but these are also covered by tests? Do we check coverage in CI actually?

alcaeus:
> I know you audited usages, but these are also covered by tests? Do we check coverage in CI actually?

Nope, we don't have any tests for `null` parameters. At the moment we don't have any coverage information, I'd like to add that eventually.

PR #1965, opened 2026-03-27
Add macros to simplify handling of internal structs

alcaeus:
Following the changes made in #1959, this PR introduces additional macros to handle declarations of the `intern` variables used throughout and working with `$this` and object initialisation.

@paulinevos and @GromNaN, I'll let you review and decide whether you find the code more legible with the macros or if you prefer the current style.


alcaeus:
@GromNaN @paulinevos do you feel like these new macros increase readability, or should we keep the original code?

src/phongo_classes.h
```diff
@@ -33,6 +33,25 @@
...
+#define PHONGO_INTERN_FROM_Z_OBJ(name, obj) phongo_##name##_t* intern = php_##name##_fetch_object((obj))
+#define PHONGO_INTERN_FROM_THIS(name) PHONGO_INTERN_FROM_ZVAL(name, getThis())
+
+#define PHONGO_INTERN_OBJECT_ALLOC(name, class_type)                       \
+	phongo_##name##_t* intern;                                             \
+	do {                                                                   \
+		intern = zend_object_alloc(sizeof(phongo_##name##_t), class_type); \
+		zend_object_std_init(&intern->std, class_type);                    \
+		object_properties_init(&intern->std, class_type);                  \
+	} while (0)
```
GromNaN, line 46: Why do you need this `while` loop?

alcaeus: This pattern is a standard C idiom for writing multi-statement macros safely. It lets us wrap multiple statements in a macro that behaves like a single statement. This is important in certain contexts, most importantly when used in an if/else without braces (which we fortunately don't use):

```
    if (condition)
        MY_MACRO(x);  // expands safely; no dangling-else or missing-brace bugs
    else
        other();
```

Without the do/while, an if-less `{ ... }` block would break the else, and a bare statement list would only execute the first line under the `if`. The `while (0)` is optimized away by the compiler, so there's no runtime cost — it's purely a syntactic wrapper.

GromNaN: And the variable declaration is moved up to the begining of the function by the compiler, so it's not an issue.

GromNaN (approved):
It seems to me that this makes things clearer and more consistent.

src/BSON/Binary.c
```diff
@@ -145,20 +143,16 @@ static PHP_METHOD(MongoDB_BSON_Binary, __set_state)
 /* Return the Binary's data string. */
 static PHP_METHOD(MongoDB_BSON_Binary, __toString)
 {
-	phongo_binary_t* intern;
+	PHONGO_INTERN_FROM_THIS(binary);
```
paulinevos, line 146: If we're gonna do this I'd suggest using `INTERNAL` and do away with the old abbreviations (which I personally found confusing seeing the code base for the first time)

paulinevos (approved):
I like, just one small naming suggestion

src/BSON/Binary.c
alcaeus: Do you mean the macro name or the variable name as well? I can rename the macros to `PHONGO_INTERNAL_STRUCT_*` to clarify that this extracts the internal struct from the `zend_object` or `zval`. Changing the variable name across the code base is something I'd like to avoid.

paulinevos: Naw just the macro name. I figured if we're introducing a new thing it may as well clarify. I think the variable name is clear enough if it can be inferred from the macro name.

GromNaN: This macro name is consistent with the variable name. I would not touch it.

alcaeus: After discussion, we decided to leave the name as-is for the time being.

PR #1970, opened 2026-04-01
Improve testing notes in CLAUDE.md

GromNaN:
Clarify two things in the Testing section:

- `make test` compiles and loads the extension automatically — `make all` and `make install` are not needed before running tests.
- The `.log` file is the recommended starting point when debugging a failed test, as it contains all intermediate output (expected, actual, diff) in one place.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -40,7 +40,10 @@ make test TESTS=tests/path/to/test.phpt
 make test TESTS="tests/path/to/pattern*.phpt"
 ```
 
-For failed tests, all intermediate files (e.g. expected output, actual output, diff) are left alongside the test file. This is helpful for debugging purposes. 
+`make test` compiles the extension and loads it automatically — no need to run `make all` or `make install` first.
+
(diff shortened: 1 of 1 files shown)


paulinevos (approved):

paulinevos:
Is this the norm for all PHP extension and/or core work? Asking because of the MCP server I'm building.

GromNaN:
> Is this the norm for all PHP extension and/or core work? Asking because of the MCP server I'm building.

Yes, the `make test` task and `run-tests.php` script comes from php-src when runninng `configure` (I guess).

alcaeus:
> Yes, the `make test` task and `run-tests.php` script comes from php-src when runninng `configure` (I guess).

You guessed correctly. The Makefile mostly comes from PHP itself and is generated through `phpize` and `configure`. There's a `Makefile.frag` file that we can use to add our own targets, such as the `test-no-build` target which doesn't have `all` as a dependency (used on Evergreen where we compile the extension in one task but run tests in another).

PR #1971, opened 2026-04-01
PHPC-2234: Implement object comparison handler for read preference

paulinevos:
Same concept as #1968.

Docs PR lives here: https://github.com/php/doc-en/pull/5454

### DIFF
--- a/src/MongoDB/ReadPreference.c
+++ b/src/MongoDB/ReadPreference.c
@@ -520,12 +520,60 @@ static HashTable* phongo_readpreference_get_properties(zend_object* object)
 	return phongo_readpreference_get_properties_hash(object, false);
 }
 
+static int phongo_readpreference_compare_objects(zval* o1, zval* o2)
+{
+	ZEND_COMPARE_OBJECTS_FALLBACK(o1, o2);
+
+	phongo_readpreference_t *intern1, *intern2;
+
+	const mongoc_read_prefs_t* rp1 = NULL;
+	const mongoc_read_prefs_t* rp2 = NULL;
+
+	intern1 = Z_READPREFERENCE_OBJ_P(o1);
+	intern2 = Z_READPREFERENCE_OBJ_P(o2);
+
+	if (intern1 && intern1->read_preference) {
+		rp1 = intern1->read_preference;
+	}
+
+	if (intern2 && intern2->read_preference) {
+		rp2 = intern2->read_preference;
+	}
+
+	if (rp1 == rp2) {
+		return 0;
+	}
+
+	if (!rp1 || !rp2) {
+		return -1;
+	}
+
+	if (mongoc_read_prefs_get_mode(rp1) != mongoc_read_prefs_get_mode(rp2)) {
+		return -1;
+	}
+
+	if (mongoc_read_prefs_get_max_staleness_seconds(rp1) != mongoc_read_prefs_get_max_staleness_seconds(rp2)) {
+		return -1;
+	}
+
+	if (bson_compare(mongoc_read_prefs_get_tags(rp1), mongoc_read_prefs_get_tags(rp2)) != 0) {
+		return -1;
+	}
+
+	if (bson_compare(mongoc_read_prefs_get_hedge(rp1), mongoc_read_prefs_get_hedge(rp2)) != 0) {
+		return -1;
+	}
+
+	return 0;
+}
+
 void phongo_readpreference_init_ce(INIT_FUNC_ARGS)
 {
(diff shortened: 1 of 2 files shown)


src/MongoDB/ReadPreference.c
```diff
@@ -520,12 +520,64 @@ static HashTable* phongo_readpreference_get_properties(zend_object* object)
...
+		rp1 = intern1->read_preference;
+	}
+
+	if (intern2 && intern2->read_preference) {
+		rp2 = intern2->read_preference;
+	}
+
+	if (rp1 == rp2) {
+		return 0;
+	}
```
GromNaN, line 545: Is it the strict equal case? Doing `$rp === $rp`.

src/MongoDB/ReadPreference.c
```diff
@@ -520,12 +520,64 @@ static HashTable* phongo_readpreference_get_properties(zend_object* object)
...
+
+	if (mongoc_read_prefs_get_max_staleness_seconds(rp1) != mongoc_read_prefs_get_max_staleness_seconds(rp2)) {
+		return -1;
+	}
+
+	const bson_t* tags1 = mongoc_read_prefs_get_tags(rp1);
+	const bson_t* tags2 = mongoc_read_prefs_get_tags(rp2);
+
+	if (tags1 == NULL && tags2 == NULL) {
+	} else if (bson_compare(tags1, tags2) != 0) {
```
GromNaN: The empty `if` block is unusual, the conditions can be merged.
```suggestion
	if ((tags1 != NULL || tags2 != NULL) && bson_compare(tags1, tags2) != 0) {
```

What's the result of doing `bson_compare(tags1, tags2) != 0` when both `tag1` and `tag2` are `NULL`?

src/MongoDB/ReadPreference.c
alcaeus: `===` will hit a completely different code path. The `compare` handler is only invoked when checking equality (including `<` and other comparison operators), not when checking identity.

tests/readPreference/readpreference-compare.phpt
```diff
@@ -0,0 +1,40 @@
+--TEST--
+MongoDB\Driver\ReadPreference equality comparison
+--FILE--
+<?php
+
+// True: same mode
+var_dump(new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::PRIMARY) == new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::PRIMARY));
```
alcaeus: I know we usually write tests like this, but I just thought of a way to make them more readable. Consider a `$tests` array that stores a test name and the result, then prints the name along with the result:

```php
$tests = [
    'True: same mode' => new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::PRIMARY) == new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::PRIMARY),
    'True: same mode and tag sets' => new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::SECONDARY, [['dc' => 'east']]) == new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::SECONDARY, [['dc' => 'east']]),
    // ...
];

foreach ($tests as $test => $result) {
    echo $test . "\n";
    var_dump($result);
}
```

This would make debugging failed tests easier as you wouldn't have to count through the results to figure out which line failed.

src/MongoDB/ReadPreference.c
```diff
@@ -520,12 +520,64 @@ static HashTable* phongo_readpreference_get_properties(zend_object* object)
...
+		return -1;
```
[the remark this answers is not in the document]
paulinevos: We're not supporting anything but equality on this.

src/MongoDB/ReadPreference.c
paulinevos: You can't pass null `bson_compare` iirc but I'll doublecheck

paulinevos: I doublechecked and was wrong, lol.

tests/readPreference/readpreference-compare.phpt
```diff
@@ -0,0 +1,48 @@
...
+--EXPECTF--
+bool(true)
+bool(true)
+bool(true)
+bool(false)
+bool(false)
+bool(false)
+bool(false)
+bool(false)
+%A
```
paulinevos, line 43: These need to be here because of deprecation notices that may or may not be present depending on versions

paulinevos:
Closing this as we opted for the solution in PHPC-2379 instead

PR #1973, opened 2026-04-07
[PoC] PHPC-2379: Add properties to ReadPreference class

alcaeus:
PHPC-2379

I revisited the topic of proper properties for value objects based on recent work on object comparisons and the `get_properties_for` handler. For the time being, I've only done the work for the `MongoDB\Driver\ReadPreference` class, but the concept applies to all other value objects (see list below).

I've left comments in the code itself to elaborate on some of the changes, but the TL;DR is that we can do away with the static `get_properties_hash` function (which serves no other purpose than to build a list of properties for the various methods that need them) and some of the other handlers. In fact, we no longer need to implement the `__serialize` method and the `get_debug_info` object handler as these dump public properties by default. There are multiple test changes associated with this as our own handlers often would not serialise properties that were "unset" or contained default values.

Here is the list of classes this concept would apply to:
- `MongoDB\Driver\ReadConcern`
- `MongoDB\Driver\WriteConcern`
- `MongoDB\Driver\WriteError`
- `MongoDB\Driver\WriteResult`
- `MongoDB\Driver\Exception\CommandException`
- `MongoDB\Driver\Exception\WriteException`
- `MongoDB\Driver\Monitoring\CommandFailedEvent`
- `MongoDB\Driver\Monitoring\CommandStartedEvent`
- `MongoDB\Driver\Monitoring\CommandSucceededEvent`
- `MongoDB\Driver\Monitoring\ServerChangedEvent`
- `MongoDB\Driver\Monitoring\ServerClosedEvent`
- `MongoDB\Driver\Monitoring\ServerHeartbeatFailedEvent`
- `MongoDB\Driver\Monitoring\ServerHeartbeatStartedEvent`
- `MongoDB\Driver\Monitoring\ServerHeartbeatSucceededEvent`
- `MongoDB\Driver\Monitoring\ServerOpeningEvent`
- `MongoDB\Driver\Monitoring\TopologyChangedEvent`
- `MongoDB\Driver\Monitoring\TopologyClosedEvent`
- `MongoDB\Driver\Monitoring\TopologyOpeningEvent`

These changes would also apply to all BSON classes, although they would not be as useful as the corresponding methods (which should eventually be deprecated) are mandated by the interfaces, which can't contain properties until PHP 8.4. For some of the other classes (e.g. `MongoDB\Driver\Server`), the value of a property may change over time (e.g. the server description or latency) so I'm not sure whether we would want to use hooked properties for those. I would suggest discussing that separately and splitting PHPC-2379 if we decide to go this route.

### DIFF
--- a/src/MongoDB/ReadPreference.c
+++ b/src/MongoDB/ReadPreference.c
@@ -31,6 +31,71 @@
 
 zend_class_entry* phongo_readpreference_ce;
 
+static const char* phongo_readpreference_get_mode_string(const mongoc_read_prefs_t* read_prefs)
+{
+	switch (mongoc_read_prefs_get_mode(read_prefs)) {
+		case MONGOC_READ_PRIMARY:
+			return PHONGO_READ_PRIMARY;
+		case MONGOC_READ_PRIMARY_PREFERRED:
+			return PHONGO_READ_PRIMARY_PREFERRED;
+		case MONGOC_READ_SECONDARY:
+			return PHONGO_READ_SECONDARY;
+		case MONGOC_READ_SECONDARY_PREFERRED:
+			return PHONGO_READ_SECONDARY_PREFERRED;
+		case MONGOC_READ_NEAREST:
+			return PHONGO_READ_NEAREST;
+		default:
+			return "";
+	}
+}
+
+static void phongo_readpreference_update_properties(phongo_readpreference_t* intern)
+{
+	const bson_t* bson;
+
+	zend_update_property_string(phongo_readpreference_ce, &intern->std, ZEND_STRL("mode"), phongo_readpreference_get_mode_string(intern->read_preference));
+	zend_update_property_long(phongo_readpreference_ce, &intern->std, ZEND_STRL("maxStalenessSeconds"), mongoc_read_prefs_get_max_staleness_seconds(intern->read_preference));
+
+	bson = mongoc_read_prefs_get_tags(intern->read_preference);
+
+	if (!bson_empty0(bson)) {
+		phongo_bson_state state;
+
+		PHONGO_BSON_INIT_STATE(state);
+		state.map.root.type = PHONGO_TYPEMAP_NATIVE_ARRAY;
+
+		if (!phongo_bson_to_zval_ex(bson, &state)) {
+			// Exception already thrown
+			zval_ptr_dtor(&state.zchild);
+			return;
+		}
+
+		zend_update_property(phongo_readpreference_ce, &intern->std, ZEND_STRL("tags"), &state.zchild);
+		zval_ptr_dtor(&state.zchild);
+	} else {
+		zend_update_property_null(phongo_readpreference_ce, &intern->std, ZEND_STRL("tags"));
+	}
+
+	bson = mongoc_read_prefs_get_hedge(intern->read_preference);
+
+	if (!bson_empty0(bson)) {
+		phongo_bson_state state;
+
+		PHONGO_BSON_INIT_STATE(state);
+
+		if (!phongo_bson_to_zval_ex(bson, &state)) {
+			// Exception already thrown
+			zval_ptr_dtor(&state.zchild);
+			return;
+		}
+
+		zend_update_property(phongo_readpreference_ce, &intern->std, ZEND_STRL("hedge"), &state.zchild);
+		zval_ptr_dtor(&state.zchild);
+	} else {
+		zend_update_property_null(phongo_readpreference_ce, &intern->std, ZEND_STRL("hedge"));
+	}
+}
+
 /* Initialize the object from a HashTable and return whether it was successful.
  * An exception will be thrown on error. */
 static bool phongo_readpreference_init_from_hash(phongo_readpreference_t* intern, HashTable* props)
@@ -83,7 +148,7 @@ static bool phongo_readpreference_init_from_hash(phongo_readpreference_t* intern
 
 			mongoc_read_prefs_set_tags(intern->read_preference, tags);
 			bson_destroy(tags);
-		} else {
+		} else if (Z_TYPE_P(tagSets) != IS_NULL) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "%s initialization requires \"tags\" field to be array", ZSTR_VAL(phongo_readpreference_ce->name));
 			goto failure;
 		}
@@ -134,7 +199,7 @@ static bool phongo_readpreference_init_from_hash(phongo_readpreference_t* intern
 
 			mongoc_read_prefs_set_hedge(intern->read_preference, hedge_doc);
 			bson_destroy(hedge_doc);
-		} else {
+		} else if (Z_TYPE_P(hedge) != IS_NULL) {
 			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "%s initialization requires \"hedge\" field to be an array or object", ZSTR_VAL(phongo_readpreference_ce->name));
 			goto failure;
 		}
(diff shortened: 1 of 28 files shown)


src/MongoDB/ReadPreference.c
```diff
@@ -31,6 +31,71 @@
 
 zend_class_entry* phongo_readpreference_ce;
 
+static const char* phongo_readpreference_get_mode_string(const mongoc_read_prefs_t* read_prefs)
```
alcaeus, line 34: This function isn't new, I merely had to pull it up as it's used in `update_properties`.

src/MongoDB/ReadPreference.c
```diff
@@ -83,7 +148,7 @@ static bool phongo_readpreference_init_from_hash(phongo_readpreference_t* intern
 
 			mongoc_read_prefs_set_tags(intern->read_preference, tags);
 			bson_destroy(tags);
-		} else {
+		} else if (Z_TYPE_P(tagSets) != IS_NULL) {
```
alcaeus, line 151: This is necessary as we now serialise `null` values for empty tag sets, while previously we would omit the `tags` element entirely.

src/MongoDB/ReadPreference.c
```diff
@@ -384,89 +435,34 @@ static PHP_METHOD(MongoDB_Driver_ReadPreference, getTagSets)
 	}
 }
 
-static HashTable* phongo_readpreference_get_properties_hash(zend_object* object, bool is_temp)
```
alcaeus, line 387: `get_properties_hash` becomes entirely useless.

src/MongoDB/ReadPreference.c
```diff
@@ -384,89 +435,34 @@ static PHP_METHOD(MongoDB_Driver_ReadPreference, getTagSets)
...
-		PHONGO_BSON_INIT_STATE(state);
+		zend_string* string_key;
+		zval*        val;
+
+		ZEND_HASH_FOREACH_STR_KEY_VAL_IND(HASH_OF(getThis()), string_key, val)
+		{
+			if (
+				(Z_TYPE_P(val) == IS_NULL) || (!strcmp(ZSTR_VAL(string_key), "maxStalenessSeconds") && Z_TYPE_P(val) == IS_LONG && Z_LVAL_P(val) == MONGOC_NO_MAX_STALENESS)) {
+				continue;
+			}
```
alcaeus, line 453: `bsonSerialize` gets a bit of special logic to avoid adding default values to the output.
`HASH_OF(getThis())` gets us the property table, which we can then iterate over to skip `null` values, as well as the default value for `maxStalenessSeconds`.

Note the `IND` suffix on the foreach macro above. Without this, the type of each zval is `IS_INDIRECT`, while the `_IND` set of foreach macros gives us access to the actual zval so we can do type checking.

src/MongoDB/ReadPreference.c
```diff
@@ -384,89 +435,34 @@ static PHP_METHOD(MongoDB_Driver_ReadPreference, getTagSets)
...
 
-		if (!phongo_bson_to_zval_ex(hedge, &state)) {
-			zval_ptr_dtor(&state.zchild);
-			goto done;
+			// Increase the refcount of our zval, as add_assoc_zval takes ownership, leading to the property value being
+			// freed once the return value goes out of scope
+			Z_TRY_ADDREF_P(val);
+			add_assoc_zval(return_value, ZSTR_VAL(string_key), val);
```
alcaeus, line 458: `add_assoc_zval` assumes ownership of the zval. Since we're passing the zval from the internal properties table, we have to use `Z_TRY_ADDREF_P` to add to the refcount of each refcounted zval to avoid the internal property value being freed once the return value goes out of scope. Finding this took me longer than I care to admit.

src/MongoDB/ReadPreference.c
```diff
@@ -384,89 +435,34 @@ static PHP_METHOD(MongoDB_Driver_ReadPreference, getTagSets)
...
-
-static PHP_METHOD(MongoDB_Driver_ReadPreference, bsonSerialize)
-{
-	PHONGO_PARSE_PARAMETERS_NONE();
-
-	ZVAL_ARR(return_value, phongo_readpreference_get_properties_hash(Z_OBJ_P(getThis()), true));
 	convert_to_object(return_value);
 }
 
-static PHP_METHOD(MongoDB_Driver_ReadPreference, __serialize)
```
alcaeus, line 463: No longer need to implement `__serialize` as this now corresponds to the default behaviour.

src/MongoDB/ReadPreference.c
```diff
@@ -509,15 +500,13 @@ static zend_object* phongo_readpreference_create_object(zend_class_entry* class_
 	return &intern->std;
 }
 
-static HashTable* phongo_readpreference_get_debug_info(zend_object* object, int* is_temp)
```
alcaeus, line 512: No longer need to implement a `get_debug_info` handler as this now corresponds to the default behaviour.

src/MongoDB/ReadPreference.c
```diff
@@ -509,15 +500,13 @@ static zend_object* phongo_readpreference_create_object(zend_class_entry* class_
...
+static zval* phongo_readpreference_read_property(zend_object* zobj, zend_string* name, int type, void** cache_slot, zval* rv)
```
alcaeus, line 503: Implementing the `read_property` handler is necessary to emit a deprecation notice for the `hedge` property (which is deprecated like the `getHedge` method). Since PHP does not support marking properties as deprecated internally (the mechanism only exists for functions/methods), we have to work around this and emit a deprecation notice before deferring to `zend_std_read_property`.

src/phongo_structs.h
```diff
@@ -127,7 +127,6 @@ typedef struct {
 
 typedef struct {
 	mongoc_read_prefs_t* read_preference;
-	HashTable*           properties;
```
alcaeus, line 130: This struct member was only needed to store our own properties table assembled by `get_properties_hash`. The internal properties table itself is stored in `zend_object`.

tests/manager/bug0851-001.phpt
```diff
@@ -51,5 +51,9 @@ object(MongoDB\Driver\ReadPreference)#%d (%d) {
     object(stdClass)#%d (%d) {
     }
   }
+  ["maxStalenessSeconds"]=>
+  int(-1)
+  ["hedge"]=>
+  NULL
```
alcaeus, line 57: All test changes account for the extra output generated by `var_dump` which now always includes all properties.

tests/readPreference/readpreference-compare-001.phpt
```diff
@@ -0,0 +1,45 @@
+--TEST--
+MongoDB\Driver\ReadPreference equality comparison
```
alcaeus, line 2: This test was lifted from #1971. I changed the way we print the output for easier debugging and moved the deprecation notices to the top of the output as we run all comparisons before printing output.

Other than that, none of the results needed changing. Most importantly though, since we now defer to the standard object comparator, we can rely on a well-defined order between different `ReadPreference` instances based on the order and values of their internal properties.

GromNaN: I don't see any tests using the `>` and `<` operators, so I assume they aren't supported natively with object properties. The value-equal `==` operator seems perfectly adequate to me as it is.

src/MongoDB/ReadPreference.c
GromNaN: Is there an opportunity for a PHP RFC to enable property deprecation?

tests/readPreference/readpreference-compare-001.phpt
alcaeus: They are supported -- I merely wanted to make sure that the tests added in #1971 pass as expected with this change. I would actually change the tests to `var_dump($rp1 <=> $rp2);` so we can assert against the result.

For context, when PHP compares objects of the same type that don't have a `compare` handler implemented, the engine goes through all properties of the objects comparing their values until a difference has been found, in which case the `-1` or `1` value is returned. This follows PHP's standard semantics for value comparisons.

src/MongoDB/ReadPreference.c
alcaeus: This has been previously discussed, and at least for properties it doesn't seem very straightforward: https://externals.io/message/127305. I would definitely appreciate being able to formally deprecate all kinds of symbols in PHP.

src/MongoDB/ReadPreference.stub.php
```diff
@@ -51,6 +51,13 @@ final class ReadPreference implements \MongoDB\BSON\Serializable
      */
     public const SMALLEST_MAX_STALENESS_SECONDS = UNKNOWN;
 
+    public readonly string $mode;
+    public readonly array|null $tags;
+    public readonly int $maxStalenessSeconds;
+
+    /** @deprecated */
+    public readonly object|null $hedge;
```
GromNaN, line 59: The reason all these properties weren't declared was because we couldn't make them read-only. But that's been possible since PHP 8.1. That explains why we can do it now, and it seems like a cleaner design.
However, as discussed, PHP does not allow modifications to a read-only property. Instead, we would use `private(set)` for properties whose values may change over time.

tests/readPreference/readpreference-compare-001.phpt
GromNaN: Using spaceship operator in the test would be more specific; that's a good idea if we are confident PHP default sorting behavior will not change.
That would also prevent us from changing properties declaration order.

src/MongoDB/ReadPreference.stub.php
alcaeus: That's correct -- however, `private(set)` requires us to bump our version requirement to PHP 8.4, while for value objects `readonly` achieves the same thing on PHP 8.1 already. Since all these classes are final, I don't see a concern to switching to `private(set)` once we bump the version requirement.

alcaeus:
Continued in #1979.

PR #1974, opened 2026-04-07
PHPC-2500: Skip errorLabels assignment for non-RuntimeException exceptions

GromNaN:
Fixes PHPC-2500.

While implementing `BulkWriteCommandException`, a dynamic property deprecation notice was observed at runtime:

```
Deprecated: Creation of dynamic property Exception::$bulkWriteCommandResult is deprecated
```

This was caused by `phongo_exception_add_error_labels()` unconditionally calling `phongo_add_exception_prop()` with `"errorLabels"` after any exception thrown via `phongo_throw_exception_from_bson_error_t_and_reply()` — including `InvalidArgumentException`, which does not inherit from `MongoDB\Driver\Exception\RuntimeException` and does not declare that property.

An initial approach added a `ZEND_ASSERT` in `phongo_add_exception_prop()` to catch mismatched property names in debug builds. However, `ZEND_ASSERT` is a no-op in release builds, leaving the dynamic property creation silently in place in production.

This commit instead adds a proper runtime guard in `phongo_exception_add_error_labels()`: the assignment is skipped entirely if the current exception is not an instance of `RuntimeException`. This is the correct fix because:

- `errorLabels` is semantically meaningful only on `RuntimeException` and its subclasses (`ServerException`, `CommandException`, `BulkWriteException`, etc.), which represent server-side or network failures that can carry labels such as `TransientTransactionError`.
- Exception types outside this hierarchy (e.g. `InvalidArgumentException`) represent client-side validation errors that occur before any server interaction and can never carry error labels.
- The guard is effective in both debug and release builds, unlike `ZEND_ASSERT`.

Also fixes `prop_len` type in `phongo_add_exception_prop()` from `int` to `size_t` to match the `zend_update_property()` and `zend_hash_str_exists()` signatures.

### DIFF
--- a/src/phongo_error.c
+++ b/src/phongo_error.c
@@ -32,7 +32,6 @@ void phongo_add_exception_prop(const char* prop, size_t prop_len, zval* value)
 	if (EG(exception)) {
 		zval ex;
 		ZVAL_OBJ(&ex, EG(exception));
-		ZEND_ASSERT(zend_hash_str_exists(&Z_OBJCE(ex)->properties_info, prop, prop_len) && "Exception class does not declare property");
 		zend_update_property(Z_OBJCE(ex), Z_OBJ_P(&ex), prop, prop_len, value);
 	}
 }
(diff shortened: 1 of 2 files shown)


src/phongo_error.c
```diff
@@ -32,6 +32,7 @@ void phongo_add_exception_prop(const char* prop, int prop_len, zval* value)
 	if (EG(exception)) {
 		zval ex;
 		ZVAL_OBJ(&ex, EG(exception));
+		ZEND_ASSERT(zend_hash_str_exists(&Z_OBJCE(ex)->properties_info, prop, prop_len) && "Exception class does not declare property");
```
GromNaN: The added message is a hack to give more context when the error is output; https://stackoverflow.com/a/3692961/450789

alcaeus (approved):

paulinevos (approved):

PR #1975, opened 2026-04-07
Add tests to improve coverage of untested or under-tested APIs

GromNaN:
## Context

This PR is the result of a systematic test coverage audit across all public PHP classes in the extension. The audit searched for methods and behaviors with no dedicated tests or only tested indirectly (e.g., via debug output or exception message strings).

### `tests/session/session-getServer-001.phpt`
`Session::getServer()` had no test at all. Documents that the method returns `null` when the session is not pinned to a server. Note: mongos pinning for sharded transactions was removed in MongoDB 6.0+, so `getServer()` now consistently returns `null` in all topologies.

### `tests/session/session-getTransactionOptions-001.phpt` (modified)
The existing test covered each transaction option (`maxCommitTimeMS`, `readConcern`, `readPreference`, `writeConcern`) in isolation. A combined case was added to verify all four options coexist correctly and are all reflected by `getTransactionOptions()`.

### `tests/exception/executiontimeoutexception-haserrorlabel-001.phpt`
`hasErrorLabel()` was tested on `RuntimeException` and `BulkWriteException` but not on `ExecutionTimeoutException` despite inheriting the method. Uses the same `ReflectionClass` pattern as existing tests.

### `tests/exception/connectiontimeoutexception-haserrorlabel-001.phpt`
Same gap as above for `ConnectionTimeoutException`.

### `tests/bulkwritecommand/bulkwritecommand-ctor-ordered-002.phpt` (modified)
`BulkWriteCommandException::getWriteConcernErrors()` appeared in debug output but was never explicitly asserted. Added `var_dump($e->getWriteConcernErrors())` to the existing `ordered=false` test, which already triggers the same duplicate key scenario, verifying it returns an empty array when only write errors (not write concern errors) occur.

### `tests/query/query-ctor-maxAwaitTimeMS-001.phpt`
`maxAwaitTimeMS` in `Query::__construct()` had range validation tests but no test verifying that valid values are accepted without error. Covers `0` and `1000`.

### `tests/query/query-ctor-maxAwaitTimeMS-002.phpt`
Complements the above with the upper boundary value `UINT32_MAX` (4294967295), guarded with a 64-bit platform skip to avoid integer overflow on 32-bit builds (same pattern as `query-ctor_error-006.phpt`).

### `tests/bulkwritecommand/bulkwritecommand-ctor-options-combined-001.phpt`
`BulkWriteCommand` constructor options (`ordered`, `comment`, `verboseResults`) were each tested in isolation. This test verifies they work correctly when combined, using APM to assert the options appear in the sent command.

### `tests/command/command-ctor_error-002.phpt`
`Command::__construct()` validates `maxAwaitTimeMS >= 0` in `Command.c` (same logic as `Query`), but had no corresponding test.


alcaeus:
Test failure is related to #1953 which was merged to v2.x only. @GromNaN I think we can merge this despite the failure or alternatively target 2.x with the changes.

tests/exception/connectiontimeoutexception-haserrorlabel-001.phpt
```diff
@@ -0,0 +1,22 @@
+--TEST--
+MongoDB\Driver\Exception\ConnectionTimeoutException::hasErrorLabel()
+--FILE--
+<?php
+
+$exception = new MongoDB\Driver\Exception\ConnectionTimeoutException();
+$labels = ['test', 'foo'];
+
+$reflection = new ReflectionClass($exception);
+$errorLabelsProperty = $reflection->getProperty('errorLabels');
```
alcaeus, line 10: I'll note that this property will become public as part of the properties rework. I'll make a note to update the tests and remove reflection where we no longer need it.

alcaeus (approved):

PR #1979, opened 2026-04-08
PHPC-2699: Add typed, read-only properties for value objects and events

alcaeus:
PHPC-2699; continuation of #1973

This PR adds public, read-only properties to the stub files for the following classes:

- `MongoDB\Driver\ReadConcern`
- `MongoDB\Driver\ReadPreference`
- `MongoDB\Driver\WriteConcern`
- `MongoDB\Driver\WriteError`
- `MongoDB\Driver\WriteResult`
- `MongoDB\Driver\Exception\CommandException` (make existing property `public readonly`)
- `MongoDB\Driver\Exception\BulkWriteException` (make existing property `public readonly`)
- `MongoDB\Driver\Monitoring\CommandFailedEvent`
- `MongoDB\Driver\Monitoring\CommandStartedEvent`
- `MongoDB\Driver\Monitoring\CommandSucceededEvent`
- `MongoDB\Driver\Monitoring\ServerChangedEvent`
- `MongoDB\Driver\Monitoring\ServerClosedEvent`
- `MongoDB\Driver\Monitoring\ServerHeartbeatFailedEvent`
- `MongoDB\Driver\Monitoring\ServerHeartbeatStartedEvent`
- `MongoDB\Driver\Monitoring\ServerHeartbeatSucceededEvent`
- `MongoDB\Driver\Monitoring\ServerOpeningEvent`
- `MongoDB\Driver\Monitoring\TopologyChangedEvent`
- `MongoDB\Driver\Monitoring\TopologyClosedEvent`
- `MongoDB\Driver\Monitoring\TopologyOpeningEvent`

This allows us to rely on PHP's default handling for several things:
- The `__serialize` implementation falls away for classes that implement it, as our property table now contains serialisable properties
- The `get_debug_info` handler falls away as it exposes properties by default
- By extension, the `get_properties_hash` function in these classes falls away; `bsonSerialize` implementations instead use the standard property table
- No `get_properties_for` handler needs to be implemented (makes PHPC-1622 obsolete for these classes)
- For `ReadConcern`, `ReadPreference`, and `WriteConcern`, we no longer need to implement a `compare` handler to make them comparable as there's now a property table to compare (this makes PHPC-2234 obsolete)

I strongly recommend a review of individual commits.

This PR opens avenues for further simplification. For example, the various event classes copy data from the libmongoc event object into the internal structure, and also store them in the property table as `zval`. By making the getters return the values of the property, we could do away with the internal structure for all these objects and only store data in the `zval`.

In future, we can also deprecate the various getters for removal in 3.0. I've left this out of this PR as the changes are already quite verbose due to the occasional change in debug output.

Other classes, such as BSON objects and "functional" objects like `MongoDB\Driver\Server` or `MongoDB\Driver\Cursor`, will be handled at a separate time.

### DIFF
--- a/src/MongoDB/Exception/BulkWriteException.stub.php
+++ b/src/MongoDB/Exception/BulkWriteException.stub.php
@@ -9,8 +9,7 @@ namespace MongoDB\Driver\Exception;
 
 class BulkWriteException extends ServerException
 {
-    /** @var \MongoDB\Driver\WriteResult */
-    protected $writeResult;
+    public readonly \MongoDB\Driver\WriteResult $writeResult;
 
     final public function getWriteResult(): \MongoDB\Driver\WriteResult {}
 }
--- a/src/MongoDB/Monitoring/CommandFailedEvent.stub.php
+++ b/src/MongoDB/Monitoring/CommandFailedEvent.stub.php
@@ -10,6 +10,18 @@ namespace MongoDB\Driver\Monitoring;
 /** @not-serializable */
 final class CommandFailedEvent
 {
+    public readonly string $host;
+    public readonly int $port;
+    public readonly string $commandName;
+    public readonly string $databaseName;
+    public readonly int $durationMicros;
+    public readonly \Exception $error;
+    public readonly object $reply;
+    public readonly string $operationId;
+    public readonly string $requestId;
+    public readonly ?\MongoDB\BSON\ObjectId $serviceId;
+    public readonly ?int $serverConnectionId;
+
     final private function __construct() {}
 
     final public function getCommandName(): string {}
--- a/src/MongoDB/ReadConcern.c
+++ b/src/MongoDB/ReadConcern.c
@@ -25,6 +25,17 @@
 
 zend_class_entry* phongo_readconcern_ce;
 
+static void phongo_readconcern_update_properties(phongo_readconcern_t* intern)
+{
+	const char* level = mongoc_read_concern_get_level(intern->read_concern);
+
+	if (level) {
+		zend_update_property_string(phongo_readconcern_ce, &intern->std, ZEND_STRL("level"), level);
+	} else {
+		zend_update_property_null(phongo_readconcern_ce, &intern->std, ZEND_STRL("level"));
+	}
+}
+
 /* Initialize the object from a HashTable and return whether it was successful.
  * An exception will be thrown on error. */
 static bool phongo_readconcern_init_from_hash(phongo_readconcern_t* intern, HashTable* props)
@@ -36,12 +47,13 @@ static bool phongo_readconcern_init_from_hash(phongo_readconcern_t* intern, Hash
 	if ((level = zend_hash_str_find(props, "level", sizeof("level") - 1))) {
 		if (Z_TYPE_P(level) == IS_STRING) {
 			mongoc_read_concern_set_level(intern->read_concern, Z_STRVAL_P(level));
-			return true;
+		} else if (Z_TYPE_P(level) != IS_NULL) {
+			phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "%s initialization requires \"level\" string field", ZSTR_VAL(phongo_readconcern_ce->name));
+			goto failure;
 		}
-
-		phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "%s initialization requires \"level\" string field", ZSTR_VAL(phongo_readconcern_ce->name));
-		goto failure;
 	}
+
+	phongo_readconcern_update_properties(intern);
 
 	return true;
 
@@ -69,6 +81,8 @@ static PHP_METHOD(MongoDB_Driver_ReadConcern, __construct)
 	if (level) {
 		mongoc_read_concern_set_level(intern->read_concern, level);
 	}
+
+	phongo_readconcern_update_properties(intern);
 }
 
 static PHP_METHOD(MongoDB_Driver_ReadConcern, __set_state)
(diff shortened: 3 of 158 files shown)


src/MongoDB/Exception/BulkWriteException.stub.php
```diff
@@ -9,8 +9,7 @@
 
 class BulkWriteException extends ServerException
 {
-    /** @var \MongoDB\Driver\WriteResult */
-    protected $writeResult;
```
GromNaN, line 13: It's technically a breaking change if anyone extends the `BulkWriteException` and overrides this property, which is highly unlikely.

src/MongoDB/Monitoring/CommandFailedEvent.stub.php
```diff
@@ -10,6 +10,18 @@
 /** @not-serializable */
 final class CommandFailedEvent
 {
+    public readonly string $host;
+    public readonly int $port;
+    public readonly string $commandName;
+    public readonly string $databaseName;
+    public readonly int $durationMicros;
+    public readonly \Exception $error;
```
GromNaN: Should it be more specific? `MongoDB\Driver\Exception` or `MongoDB\Driver\RuntimeException`

src/MongoDB/Monitoring/CommandFailedEvent.stub.php
```diff
@@ -10,6 +10,18 @@
...
+    public readonly object $reply;
+    public readonly string $operationId;
+    public readonly string $requestId;
+    public readonly ?\MongoDB\BSON\ObjectId $serviceId;
```
GromNaN, line 22: No `use` statement?

src/MongoDB/ReadConcern.stub.php
```diff
@@ -50,7 +52,4 @@ final public static function __set_state(array $properties): ReadConcern {}
     final public function bsonSerialize(): \stdClass {}
 
     final public function __unserialize(array $data): void {}
-
-    final public function __serialize(): array {}
```
GromNaN, line 54: Why is it removed?

tests/readConcern/readconcern-compare-001.phpt
```diff
@@ -0,0 +1,59 @@
...
+    'Object comparison fallback if one value is not a ReadConcern' => [
+        'a' => new MongoDB\Driver\ReadConcern(MongoDB\Driver\ReadConcern::LOCAL),
+        'b' => new MongoDB\BSON\Int64('1')
+    ],
+];
+
+foreach ($tests as $name => $test) {
+    echo "\nTesting $name:\n";
+    var_dump($test['a'] <=> $test['b']);
+    var_dump($test['b'] <=> $test['a']);
```
GromNaN, line 32: Perfect use of spaceship operator you suggested in https://github.com/mongodb/mongo-php-driver/pull/1973#discussion_r3044696562

tests/readConcern/readconcern-compare-001.phpt
```diff
@@ -0,0 +1,59 @@
...
+int(-1)
+int(1)
+
+Testing one has level and the other does not:
+int(1)
+int(-1)
+
+Testing Object comparison fallback if one value is not a ReadConcern:
+int(1)
+int(1)
```
GromNaN, line 58: So unrelated objects still return inconsistent comparison. I checked this is the default PHP behavior: https://3v4l.org/mbRA3

src/MongoDB/ReadConcern.stub.php
alcaeus: `__serialize` is no longer necessary as the default behaviour is to serialise properties. We needed it because we didn't have real properties, but now that we do it works with PHP's default implementation for serialisation.

src/MongoDB/Monitoring/CommandFailedEvent.stub.php
alcaeus: We don't use `use`, and off the top of my head I don't know if gen_stub.php even supports it.

src/MongoDB/Monitoring/CommandFailedEvent.stub.php
alcaeus: I reused the return type of `getError`. I'll take a look to see if we can make it more specific.

src/MongoDB/Exception/BulkWriteException.stub.php
alcaeus: I'm willing to take the risk :)

alcaeus:
GitHub Copilot had one low-confidence comment it suppressed:
> Property type is nullable (ObjectId|null) but getTopologyId() is declared to always return ObjectId. Either make $topologyId non-nullable (and ensure it is always initialized), or update the getter signature/implementation to allow null and ensure the property is set to null when appropriate; otherwise, this creates an inconsistent API surface for the same value.

I'm looking at this to see what the discrepancy is about.

src/MongoDB/Monitoring/CommandFailedEvent.stub.php
GromNaN: Confirmed that `gen_stub.php` does not support use statements.
```
In ./src/MongoDB/Monitoring/CommandFailedEvent.stub.php:
Unexpected node Stmt_Use
```

src/MongoDB/Monitoring/CommandFailedEvent.stub.php
alcaeus: I'm now using the `MongoDB\Driver\Exception\Exception` interface as it's the only common ancestor between the various exceptions that could be added: `phongo_exception_from_mongoc_domain` might return a `RuntimeException` or `LogicException`. The only problem is that this interface only extends `\Throwable` but does not share an ancestor with `\Exception`. It should be fine, but I suspect that's why `getError` originally used `\Exception` for its return type instead of a more specific type.

GromNaN: So that's a case for an intersection type `\Exception&MongoDB\Driver\Exception\Exception`

GromNaN: MongoDB\Driver\Exception\Exception doesn't provide any method or property, that's not worth the change.
Alternatively, a union of `MongoDB\Driver\Exception\RuntimeException| MongoDB\Driver\Exception\LogicException`.

alcaeus: I like the intersection type and made the change :)

src/MongoDB/ReadConcern.c
```diff
@@ -115,46 +129,31 @@ static PHP_METHOD(MongoDB_Driver_ReadConcern, isDefault)
...
+		zend_string* string_key;
+		zval*        val;
 
-	if (level) {
-		zval z_level;
+		ZEND_HASH_FOREACH_STR_KEY_VAL_IND(HASH_OF(getThis()), string_key, val)
+		{
+			if (Z_TYPE_P(val) == IS_NULL) {
+				continue;
+			}
```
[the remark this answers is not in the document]
alcaeus: We decided to ignore this for the time being as dynamic property assignments are done. We can also switch to read-only classes when dropping support for PHP 8.1: https://3v4l.org/BaU6M

src/MongoDB/Monitoring/CommandFailedEvent.stub.php
GromNaN: The intersection type doesn't work.

GromNaN: May be too easy to fix: https://github.com/php/php-src/pull/21717

alcaeus: Removed the intersection type for the time being. We can clarify that in a later version once the arginfo generator has been fixed.

GromNaN (approved):

PR #1981, opened 2026-04-13
Generate legacy arginfo for older PHP versions

alcaeus:
The arginfo generator allows for [maintaining backward compatibility](https://php.github.io/php-src/miscellaneous/stubs.html#maintaining-backward-compatibility) through the use of the `generate-legacy-arginfo` annotation, together with the minimum PHP version being targeted. Using this, we can remove the limitation to only generate arginfo on PHP 8.2 and instead use the latest stable PHP version.

### DIFF
--- a/src/BSON/BinaryInterface.stub.php
+++ b/src/BSON/BinaryInterface.stub.php
@@ -1,8 +1,8 @@
 <?php
 
 /**
- * @generate-class-entries static
- * @generate-function-entries
+ * @generate-class-entries
+ * @generate-legacy-arginfo 80100
  */
 
 namespace MongoDB\BSON;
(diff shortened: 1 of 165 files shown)


src/BSON/BinaryInterface.stub.php
```diff
@@ -1,8 +1,8 @@
 <?php
 
 /**
- * @generate-class-entries static
- * @generate-function-entries
+ * @generate-class-entries
```
alcaeus, line 4: When using `gen_stub.php` from PHP 8.5, I noticed that using `static` with `generate-class-entries` for interfaces causes invalid files. Haven't yet had time to investigate, but since we're not generating an `PHP_METHOD` entries we can omit the `static` keyword for interfaces.

GromNaN (approved):

PR #1983, opened 2026-04-13
Remove `phongo_write_concern_to_zval`

paulinevos:
It was only used in `bulkwrite_get_debug_info`, where we could really just instantiate the `WriteConcern` instance and add that to the debug info, which would result in the `WriteConcern` object being var_dumped and it more in line with what you'd expect as a PHP user anyway.

### DIFF
--- a/tests/bulk/write-0002.phpt
+++ b/tests/bulk/write-0002.phpt
@@ -70,9 +70,11 @@ object(MongoDB\Driver\BulkWrite)#%d (%d) {
   ["session"]=>
   NULL
   ["write_concern"]=>
-  array(%d) {
+  object(MongoDB\Driver\WriteConcern)#%d (%d) {
     ["w"]=>
     int(1)
+    ["j"]=>
+    NULL
     ["wtimeout"]=>
     int(1000)
   }
--- a/src/MongoDB/BulkWrite.c
+++ b/src/MongoDB/BulkWrite.c
@@ -642,7 +642,7 @@ static HashTable* phongo_bulkwrite_get_debug_info(zend_object* object, int* is_t
 	if (mongoc_bulk_operation_get_write_concern(intern->bulk)) {
 		zval write_concern;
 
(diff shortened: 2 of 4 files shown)


tests/bulk/write-0002.phpt
```diff
@@ -70,7 +70,7 @@ object(MongoDB\Driver\BulkWrite)#%d (%d) {
   ["session"]=>
   NULL
   ["write_concern"]=>
-  array(%d) {
+  object(MongoDB\Driver\WriteConcern)#%d (%d) {
```
GromNaN, line 73: To make our intent explicit, we can add an assertion stating that this property is an instance of `WriteConcern`.

GromNaN (approved):

tests/bulk/write-0002.phpt
GromNaN, line 73: The `WriteConcern` object always has a `j` property (set to `NULL` when journal is not configured by `phongo_writeconcern_update_properties`), unlike the old array produced by `phongo_write_concern_to_zval` which only included `j` when `mongoc_write_concern_journal_is_set()` was true. The expected output is missing:
```
    ["j"]=>
    NULL
```
between `["w"]` and `["wtimeout"]`. This is why the test fails in CI.

paulinevos: How is it possible that it passes locally? e: and it fails locally when I do add the `j` prop 🤔

paulinevos: I don't understand this one. Is this not already clear from the debug output?

GromNaN: Yes, it's already clear. It’s just that I find it easier to modify an output snapshot by telling myself it’s a minor change. Whereas if it’s a specific assertion, we know that the result was intended.

alcaeus: Did you rebase after I merged the big properties PR?

paulinevos: Ahh nope, that'll do it lol

PR #1985, opened 2026-04-13
Use `json_decode` to produce a non-interned string in typeMap test

GromNaN:
Replace `chr(ord('a')).'rray'` with `json_decode('"array"')` to produce a non-interned string that opcache cannot fold, allowing the test to run under all configurations.

### DIFF
--- a/tests/bson/bug1839-005.phpt
+++ b/tests/bson/bug1839-005.phpt
@@ -6,9 +6,9 @@ require_once __DIR__ . "/../utils/basic.inc";
 
 function createTypemap()
 {
-    // Assemble the string so as to not have an interned string
-    $rootValue = chr(ord('a')) . 'rray';
-    $documentValue = chr(ord('a')) . 'rray';
+    // Use json_decode to produce a non-interned string that opcache cannot optimise
+    $rootValue = json_decode('"array"');
+    $documentValue = json_decode('"array"');
 
     // Use a reference to this non-interned string in the type map
     $typemap = ['root' => &$rootValue, 'document' => &$documentValue];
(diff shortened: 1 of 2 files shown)


tests/bson/bug1839-005.phpt
GromNaN: `chr` and `ord` are [optimized functions](https://support.tideways.com/documentation/reference/observations/compiler-optimized-functions.html).

alcaeus: I took another look, and an alternative approach would be to use `$rootValue = json_decode('"array"');`. This produces a non-interned string as well and can't be optimised by opcache.

GromNaN: Good point, change that.

GromNaN:
Updated: replaced `chr(ord('a')).'rray'` with `json_decode('"array"')` as suggested. No need to skip the test anymore.

tests/bson/bug1839-005.phpt
```diff
@@ -6,9 +6,9 @@ require_once __DIR__ . "/../utils/basic.inc";
 
 function createTypemap()
 {
-    // Assemble the string so as to not have an interned string
-    $rootValue = chr(ord('a')) . 'rray';
-    $documentValue = chr(ord('a')) . 'rray';
+    // Use json_decode to produce a non-interned string that opcache cannot optimise
+    $rootValue = json_decode('"array"');
```
alcaeus, line 10: Note there's another occurrence in bug1839-006.phpt

GromNaN: Good catch, I don't know how I missed this other one.

alcaeus (approved):
LGTM

PR #1988, opened 2026-04-14
PHPC-1519: Remove phongo_server_to_zval and include Server objects in Manager debug output

GromNaN:
Fix PHPC-1519
`phongo_server_to_zval` was used by both `phongo_server_get_debug_info` and `phongo_manager_get_debug_info` to build an array representation of a server. The Manager's debug handler now creates `Server` objects via `phongo_server_init` instead, consistent with how other value objects appear in debug output. The array-building logic is inlined into `phongo_server_get_debug_info` with proper error handling.

### DIFF
--- a/src/MongoDB/Manager.c
+++ b/src/MongoDB/Manager.c
@@ -802,6 +802,7 @@ static HashTable* phongo_manager_get_debug_info(zend_object* object, int* is_tem
 	size_t                        i, n = 0;
 	zval                          retval = ZVAL_STATIC_INIT;
 	zval                          cluster;
+	zval                          manager_zval;
 
 	*is_temp = 1;
 	intern   = Z_OBJ_MANAGER(object);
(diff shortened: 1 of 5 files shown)


src/MongoDB/Manager.c
```diff
@@ -814,16 +815,12 @@ static HashTable* phongo_manager_get_debug_info(zend_object* object, int* is_tem
 
 	array_init_size(&cluster, n);
 
+	ZVAL_OBJ(&manager_zval, object);
```
alcaeus, line 822: Minor: you could add a comment here to explain that `phongo_server_init` increases the refcount of the underlying object, which is why we can assign this using `ZVAL_OBJ` instead of using one of the `ZVAL_COPY` macros or `Z_ADDREF_P` here.

alcaeus (approved):

src/MongoDB/Manager.c
GromNaN: Done in d8cbc860. The comment explains that `ZVAL_OBJ` is used without incrementing the refcount because `phongo_server_init` stores the manager reference via `ZVAL_ZVAL` with copy=1, so it takes ownership of the refcount itself.

src/MongoDB/Manager.c
```diff
@@ -814,16 +815,16 @@ static HashTable* phongo_manager_get_debug_info(zend_object* object, int* is_tem
...
-		if (!phongo_server_to_zval(&obj, intern->client, sds[i])) {
-			/* Exception already thrown */
-			zval_ptr_dtor(&obj);
-			zval_ptr_dtor(&cluster);
-			goto done;
-		}
-
+		phongo_server_init(&obj, &manager_zval, mongoc_server_description_id(sds[i]));
 		add_next_index_zval(&cluster, &obj);
 	}
```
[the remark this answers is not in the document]
GromNaN: Good catch, fixed in c0affd1d.

PR #1989, opened 2026-04-16
PHPC-1514 Remove `php_phongo_read_concern_to_zval`

paulinevos:
Fix PHPC-1514
It's only used by the `Query` debug handler and can be replaced by adding a `ReadConcern` object to `Query`'s debug info instead

### DIFF
--- a/src/MongoDB/Query.c
+++ b/src/MongoDB/Query.c
@@ -396,7 +396,7 @@ static HashTable* phongo_query_get_debug_info(zend_object* object, int* is_temp)
 	if (intern->read_concern) {
 		zval read_concern;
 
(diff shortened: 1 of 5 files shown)


GromNaN:
I'm surprised that a test like [tests/query/query-debug-001.phpt](https://github.com/mongodb/mongo-php-driver/blob/09c56d64d742ed98cdbaa360479e1aaaefa80229/tests/query/query-debug-001.phpt#L44-L48) doesn't fail. The format has changed from an array to an instance of `ReadConcern`.

paulinevos:
@GromNaN I'm actually surprised the test isn't included in the PR, cause I did update it and I'm using JJ (which should include any unstaged changes with the revision). Weird

GromNaN (approved):

PR #1990, opened 2026-04-16
PHPC-2699: Add readonly properties to WriteConcernError class

alcaeus:
PHPC-2699

This class was missed in #1979, but deserves the same treatment.


tests/writeConcernError/writeconcernerror-getinfo-001.phpt
```diff
@@ -61,4 +62,15 @@ object(stdClass)#%d (%d) {
     string(14) "clientSupplied"
   }
 }
+object(stdClass)#%d (%d) {
+  ["writeConcern"]=>
```
GromNaN, line 66: This should be a `WriteConcern` instance.

alcaeus: This is essentially the `details` or `errInfo` field from the command response. We don't know what the server will return here, even if it looks like a write concern.

GromNaN (approved):

PR #1991, opened 2026-04-16
PHPC-2699: Rename durationMicros properties

alcaeus:
PHPC-2699

After consideration, we're using this opportunity to fix the naming of the `duration` properties on APM event classes. `durationMicros` may be clearer in terms of units, but the specification actually calls for a `duration` property, so we're a bit more consistent with the spec here. Note that libmongoc also uses `duration` as naming, which brought this to my attention.


alcaeus:
Ignoring Copilot's comments about the rename being a BC break since we haven't yet released this.

GromNaN (approved):

PR #1992, opened 2026-04-16
Remove extra structs for objects with properties

alcaeus:
With many classes storing internal properties in `zend_object`, we're now storing data twice in multiple classes: once in the `struct` fields, and once in the property. The `struct` fields are only used for the getters, which strictly speaking isn't necessary. To be clear, we were doing this before already once `get_properties_hash` was invoked and we populated our own `properties` table.

This PR does away with that duplicate data handling for everything that returns static information (i.e. where we don't need to keep a reference to an underlying libmongoc structure):
- The new `PHONGO_PROPERTY_GETTER` macro is used to declare the boilerplate for getters, consisting of parameter parsing and reading the property value into the `return_value` zval
- For classes affected by this, we remove the class helpers and the struct

Not only does this remove duplicate data handling, but it greatly simplifies those value object classes by doing away with class handlers, allocations, etc.

### DIFF
--- a/src/MongoDB/Monitoring/CommandFailedEvent.c
+++ b/src/MongoDB/Monitoring/CommandFailedEvent.c
@@ -32,82 +32,17 @@ zend_class_entry* phongo_commandfailedevent_ce;
 
 PHONGO_DISABLED_CONSTRUCTOR(MongoDB_Driver_Monitoring_CommandFailedEvent)
 
-static PHP_METHOD(MongoDB_Driver_Monitoring_CommandFailedEvent, getCommandName)
-{
-	PHONGO_PARSE_PARAMETERS_NONE();
-
-	PHONGO_RETURN_PROPERTY(commandfailedevent, "commandName");
-}
-
-static PHP_METHOD(MongoDB_Driver_Monitoring_CommandFailedEvent, getDatabaseName)
-{
-	PHONGO_PARSE_PARAMETERS_NONE();
-
-	PHONGO_RETURN_PROPERTY(commandfailedevent, "databaseName");
-}
-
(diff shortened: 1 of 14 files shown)


alcaeus:
Leaving in draft state until #1990 has been merged. Feel free to review already.

src/MongoDB/Monitoring/CommandFailedEvent.c
```diff
@@ -32,82 +32,17 @@ zend_class_entry* phongo_commandfailedevent_ce;
...
-	PHONGO_RETURN_PROPERTY(commandfailedevent, "serviceId");
-}
-
-static PHP_METHOD(MongoDB_Driver_Monitoring_CommandFailedEvent, getServerConnectionId)
-{
-	PHONGO_PARSE_PARAMETERS_NONE();
-
-	PHONGO_RETURN_PROPERTY(commandfailedevent, "serverConnectionId");
-}
+PHONGO_PROPERTY_GETTER(MongoDB_Driver_Monitoring_CommandFailedEvent, getCommandName, commandfailedevent, "commandName")
```
paulinevos, line 35: Very satisfying

paulinevos (approved):
LGTM, but is the thing Copilot mentioned worth looking into, also in terms of test coverage?

alcaeus:
> LGTM, but is the thing Copilot mentioned worth looking into, also in terms of test coverage?

I didn't want to include it, but common sense prevailed and I added the warnings back in for 32-bit platforms. We should really get rid of that at some point in time (or lobby for PHP's `zend_long` becoming an `int64_t` regardless of platform int size)

PR #1996, opened 2026-04-21
PHPC-2695 Replace deprecated bson_append_array_begin with bson_append_array_unsafe_begin

GromNaN:
Replace the deprecated `bson_append_array_begin` with `bson_append_array_unsafe_begin`, the non-deprecated equivalent introduced in libmongoc 2.3.0.

Fix PHPC-2695

### DIFF
--- a/src/phongo_bson_encode.c
+++ b/src/phongo_bson_encode.c
@@ -129,7 +129,7 @@ static void phongo_bson_append_object(bson_t* bson, phongo_field_path* field_pat
 				phongo_zval_to_bson_internal(&obj_data, field_path, flags, &child, NULL);
 				bson_append_document_end(bson, &child);
 			} else {
-				bson_append_array_begin(bson, key, key_len, &child);
+				bson_append_array_unsafe_begin(bson, key, key_len, &child);
 				phongo_zval_to_bson_internal(&obj_data, field_path, flags, &child, NULL);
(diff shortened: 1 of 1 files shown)


alcaeus (approved):
LGTM. Interesting this is now marked as "unsafe" -- I wonder what the safe alternative to this would be.

GromNaN:
Good question:

> The distinction is:
> 
> - `_unsafe_begin` uses a plain `bson_t *child` — the caller is responsible for appending keys as `"0"`, `"1"`, `"2"`, etc. (BSON array keys must be consecutive integer strings). If you append arbitrary string keys into a `bson_t` obtained this way, you get a malformed array. There's no enforcement.
> - `bson_array_builder_begin/end` uses an opaque `bson_array_builder_t **child` — this builder automatically generates the correct integer keys, so you can't accidentally produce an invalid array.
> 
> The `_unsafe` suffix means "we don't enforce correct array key sequencing for you" — it's the same underlying operation but without the guardrail.
> 
> For the PHP driver's use case, `_unsafe_begin` is perfectly fine because `phongo_zval_to_bson_internal` already generates the correct `"0"`, `"1"`, ... keys from PHP array indices. There's no benefit from the builder's auto-keying, and migrating to 
> `bson_array_builder_t` would require a more invasive refactor of that recursive encoding path.

PR #1999, opened 2026-04-27
Remove unnecessary structs for BSON MinKey and MaxKey classes

alcaeus:
Missed these empty structs when cleaning up structs and class declarations in #1992.


GromNaN (approved):
Less code is a great feature!

PR #2007, opened 2026-04-30
PHPC-2715: fix out-of-sources-tree builds

remicollet:
To reproduce

```
cd /path/to/src
phpize
mkdir mybuild
cd mybuild
../configure
make

```
Will output
```
/dev/shm/BUILD/php-pecl-mongodb2-2.3.1-build/php-pecl-mongodb2-2.3.1/mongo-php-driver-2.3.1/src/libmongocrypt/src/mongocrypt.h:36:8: error: #error No "mongocrypt-config.h" header is available. That file must be generated in order to use libmongocrypt.
   36 |       #error No "mongocrypt-config.h" header is available. That file must \
      |        ^~~~~

```

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -387,6 +387,7 @@ if test "$PHP_MONGODB" != "no"; then
     dnl of build layout.  Using $PWD here breaks PHP in-tree builds because $PWD is
     dnl the PHP source/build root, not the extension's subdirectory within it.
     php_mongodb_ext_builddir=PHP_EXT_BUILDDIR(mongodb)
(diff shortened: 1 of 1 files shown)


GromNaN:
Thank you @remicollet, for reporting this issue and providing the fix. We will try to add a test job to ensure that there are no regressions in this build environment.

alcaeus:
Thanks @remicollet! I cherry-picked your commit into #2010 to fix this in 2.3 and added a test to prevent regressions

PR #2009, opened 2026-04-30
Fix ZVAL_STATIC_INIT macro formatting for clang-format compatibility

GromNaN:
Use a compact single-line form wrapped in `clang-format off/on` to avoid formatting differences between clang-format versions (v18 on CI vs v22 locally).


alcaeus (approved):
👏 👏 👏

PR #2016, opened 2026-05-05
Implement memory handling correctness fixes

alcaeus:
This PR introduces a bunch of defensive hardening and consistency fixes across the extension. Each commit is scoped to a single change to keep review and bisecting straightforward.

- Manager cache key — store a SHA-256 digest of the URI rather than the raw connection string in phongo_manager_make_client_hash.
- Int64 shift operators — match PHP's native integer-shift handlers: throw `ArithmeticError` on negative shift counts and return `0`/`-1` at or above the operand width.
- BSON traversal — halt the document and array visitors cleanly when `bsonUnserialize` throws, instead of inserting a partially-constructed object into the parent.
- BSON value encoder — propagate failure from `phongo_zval_to_bson_value_ex` and validate UTF-8 in the scalar path so it matches `phongo_bson_append`.
- Field path helpers — guard `phongo_field_path_pop` against an empty path and fix the trailing-separator overwrite in `phongo_field_path_as_string` when all entries are NULL.
- Binary / Regex / Javascript — free pre-existing buffers before re-init, surface scope-encoding errors from `phongo_javascript_init`, route `phongo_regex_new` through `phongo_regex_init` so BSON-decoded Regex instances pick up the same flag canonicalisation as PHP-constructed ones.
- Type cleanups — use `size_t` for BSON data lengths and for `phongo_regex_t` length fields; drop a tautological `sparsity > INT64_MAX` check in ClientEncryption.

### DIFF
--- a/src/BSON/Int64.c
+++ b/src/BSON/Int64.c
@@ -443,12 +443,20 @@ static zend_result phongo_int64_do_operation_ex(zend_uchar opcode, zval* result,
 
 		case ZEND_SL:
 			PHONGO_GET_INT64(value2, op2);
-			OPERATION_RESULT_INT64(value1 << value2);
+			if (value2 < 0) {
+				zend_throw_exception(zend_ce_arithmetic_error, "Bit shift by negative number", 0);
+				return FAILURE;
+			}
+			OPERATION_RESULT_INT64(value2 >= 64 ? 0 : (int64_t) ((uint64_t) value1 << value2));
 			return SUCCESS;
 
 		case ZEND_SR:
 			PHONGO_GET_INT64(value2, op2);
-			OPERATION_RESULT_INT64(value1 >> value2);
+			if (value2 < 0) {
+				zend_throw_exception(zend_ce_arithmetic_error, "Bit shift by negative number", 0);
+				return FAILURE;
+			}
+			OPERATION_RESULT_INT64(value2 >= 64 ? (value1 < 0 ? -1 : 0) : value1 >> value2);
 			return SUCCESS;
 
 		case ZEND_POW:
--- a/src/phongo_bson.c
+++ b/src/phongo_bson.c
@@ -787,6 +787,15 @@ static bool phongo_bson_visit_document(const bson_iter_t* iter ARG_UNUSED, const
 			object_init_ex(&obj, obj_ce);
 
 			zend_call_method_with_1_params(Z_OBJ_P(&obj), NULL, NULL, BSON_UNSERIALIZE_FUNC_NAME, NULL, &state.zchild);
+
+			if (EG(exception)) {
+				zval_ptr_dtor(&obj);
+				zval_ptr_dtor(&state.zchild);
+				phongo_bson_state_dtor(&state);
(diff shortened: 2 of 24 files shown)


src/phongo_bson.c
```diff
@@ -205,6 +205,10 @@ bool phongo_field_path_push(phongo_field_path* field_path, const char* element,
 
 bool phongo_field_path_pop(phongo_field_path* field_path)
 {
+	if (field_path->size == 0) {
+		return false;
+	}
```
GromNaN, line 214: Should it be covered by a test?

src/BSON/Int64.c
```diff
@@ -443,12 +443,20 @@ static zend_result phongo_int64_do_operation_ex(zend_uchar opcode, zval* result,
 
 		case ZEND_SL:
 			PHONGO_GET_INT64(value2, op2);
-			OPERATION_RESULT_INT64(value1 << value2);
+			if (value2 < 0) {
+				zend_throw_exception(zend_ce_arithmetic_error, "Bit shift by negative number", 0);
```
GromNaN, line 447: This exception can also be tested.

src/phongo_client.c
```diff
@@ -831,12 +833,21 @@ static char* phongo_manager_make_client_hash(const char* uri_string, zval* optio
 	php_serialize_data_t var_hash;
 	zval*                serializable_driver_options = NULL;
 	bool                 free_driver_options         = false;
+	PHP_SHA256_CTX       sha_ctx;
+	unsigned char        sha_digest[32];
+	char                 sha_hex[65];
 
 	zval args;
 
+	PHP_SHA256Init(&sha_ctx);
```
GromNaN, line 841: I confirm `sh256` is the best algo for this use-case: crypto-safe for credentials and fast in the execution hot path (Hardware Acceleration).

src/phongo_client.c
```diff
@@ -831,12 +833,21 @@ static char* phongo_manager_make_client_hash(const char* uri_string, zval* optio
...
+	PHP_SHA256Update(&sha_ctx, (const unsigned char*) uri_string, strlen(uri_string));
+	PHP_SHA256Final(sha_digest, &sha_ctx);
+	php_hash_bin2hex(sha_hex, sha_digest, sizeof(sha_digest));
```
GromNaN: Why do you need to convert to hexadecimal? Is it displayed?

src/phongo_bson.c
alcaeus: There's no way to currently produce this case; it could only happen if we make a mistake in new code if we create an imbalance between `push` and `pop` operations on the field path stack.

src/BSON/Int64.c
alcaeus: Tests added.

src/phongo_client.c
alcaeus: Good point, I've removed the hex conversion as the binary string is perfectly fine for the hash.

GromNaN (approved):

PR #2017, opened 2026-05-19
Use debug env var for debug builds

alcaeus:
A recent change to setup-php broke installs when the version contains the `-debug` suffix. This way of installing debug versions of PHP was never supported, it should've been requested using the `debug` env var.

### DIFF
--- a/.github/actions/linux/build/action.yml
+++ b/.github/actions/linux/build/action.yml
@@ -14,9 +14,11 @@ runs:
     - name: "Install PHP"
       uses: "shivammathur/setup-php@v2"
       with:
-        php-version: "${{ inputs.version }}-debug"
+        php-version: "${{ inputs.version }}"
         # Only install required extensions
         extensions: "none,date,json,spl,standard,xml"
+      env:
+        debug: true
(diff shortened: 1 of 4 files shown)


GromNaN (approved):

GromNaN:
Hi @alcaeus, I noticed that in this PR you changed `LIBMONGOCRYPT_VERSION` from `1.18.0` to `1.18`, but the bundled libmongocrypt submodule is still at 1.17.3. Was this intentional — i.e. is the system library test meant to test against a newer version than the bundled one? I'm asking because I'm working on PHPC-2724 to bump the bundled submodule to 1.19.0 and I want to understand whether the system libs test should match the bundled version or stay ahead of it.

alcaeus:
I had to make the change as 1.18.0 was no longer available, so the constraint couldn't be satisfied. Changing the constraint to `1.18` allowed installing newer versions.

We typically would want to test with the same versions, but if the system in question doesn't contain an old enough version of libmongocrypt to be the same as the one we bundle, we fall back to the newer version.

Issue #2020, opened 2026-05-21
Release 2.3.2 breaks pecl install in docker (zstd.h not found)


tux-rampage:
### Bug Report

Since release 2.3.2 the pecl install/build fails in docker environments complaining about a missing `zstd.h` header while version 2.3.1 builds and installs without any issues.

### Environment

- Docker (Image `docker.io/library/php:8.4-cli`)
- Debian GNU/Linux 13
- PHP 8.4.21 (cli) (built: May 19 2026 23:08:19) (NTS)
- MongoDB 2.3.2 (pecl)

### Test Script

```bash
pecl install mongodb
```

### Expected and Actual Behavior

**Expected:** Build and install the mongodb extension
**Actual:** Build failure

### Debug Log

Truncated to the last relevant lines:

```
/tmp/pear/temp/mongodb/src/libmongoc/src/libmongoc/src/mongoc/mongoc-compression.c:34:10: fatal error: zstd.h: No such file or directory
   34 | #include <zstd.h>
      |          ^~~~~~~~
compilation terminated.
make: *** [Makefile:502: src/libmongoc/src/libmongoc/src/mongoc/mongoc-compression.lo] Error 1
ERROR: `make' failed
```

alcaeus:
I'm currently debugging the source of the problem. There seems to be an issue with recent changes regarding out-of-source builds and `pecl` not liking them. On systems where installing via `pecl` failed, I was able to successfully install it using [PIE](https://github.com/php/pie), so if that's an option this would be a good opportunity to move away from `pecl` and provide relief while we work on fixing the issue. Instructions on how to use `pie` within docker can be found in [PIE's usage docs](https://github.com/php/pie/blob/1.5.x/docs/usage.md#docker-installation).

a8ns:
Confirmed on **Alpine 3.23 / musl / `php:8.4.16-fpm-alpine3.23`** — same root cause, different symptom.

## Environment

- Base image: `php:8.4.16-fpm-alpine3.23` (`sha256:56aed89b8fd022516b400526db7d4b699fd437372c585cd5785de9634bb7cdd7`)
- GCC: 15.2.0-r2 (Alpine)
- linux-headers: 6.16.12-r0

## Symptom

```
mongoc-client.c:461:7: error: implicit declaration of function 'res_ninit'; did you mean 'res_init'?
[...]

PR #2021, opened 2026-05-21
PHPC-2715: fix PHP_MONGODB_ADD_BUILD_INCLUDE and include ordering for phpize/PECL builds on Alpine/musl

a8ns:
## Summary

Fixes a regression introduced in 2.3.2 (commit 5d19cfa) where \`pecl install mongodb\` fails on Alpine/musl with implicit-declaration errors for \`res_ninit\`, \`res_nsearch\`, and \`res_nclose\` (see #2020).

## Root cause (two bugs)

**Bug 1 — wrong \`$abs_builddir\` in \`PHP_MONGODB_ADD_BUILD_INCLUDE\`:**
\`AC_CONFIG_FILES\` writes generated headers (e.g. \`mongoc-config.h\`) relative to \`$abs_builddir\` — the directory where \`config.status\` runs. In a phpize/PECL build this is the pear build directory (e.g. \`/tmp/pear/temp/pear-build-.../mongodb-x.y.z/\`), **not** the extension source directory.

\`PHP_MONGODB_ADD_BUILD_INCLUDE\` was using \`PHP_EXPAND_PATH(PHP_EXT_BUILDDIR(mongodb), ...)\`, which in phpize builds resolves to the extension source directory (e.g. \`/tmp/pear/temp/mongodb/\`). These two directories differ, so the compiler never found the configure-generated headers.

**Bug 2 — include ordering:**
In 2.3.2, the three \`PHP_MONGODB_ADD_BUILD_INCLUDE\` calls were placed *after* the \`PHP_MONGODB_ADD_INCLUDE\` (source-dir) calls in \`config.m4\`. Even with Bug 1 fixed, if build-dir includes appear after source-dir includes in the \`-I\` list, the compiler finds the pre-bundled source-tree \`mongoc-config.h\` first.

The pre-bundled copy was generated on a glibc system with \`MONGOC_HAVE_RES_NSEARCH=1\` hardcoded. On musl/Alpine, \`configure\` correctly sets \`MONGOC_HAVE_RES_NSEARCH=0\` in the generated header — but with wrong ordering, the wrong header wins, causing compilation to include \`res_ninit\`/\`res_nsearch\` call sites that musl does not implement, which fail with GCC 15's new error-on-implicit-declaration.

## Fix

**Commit 1** ([23233d8](https://github.com/a8ns/mongo-php-driver/commit/23233d84)): Replace \`PHP_EXPAND_PATH(PHP_EXT_BUILDDIR(...), ...)\` with \`\$abs_builddir\` directly in \`PHP_MONGODB_ADD_BUILD_INCLUDE\`. This is the same path \`AC_CONFIG_FILES\` uses.

**Commit 2** ([08d8528](https://github.com/a8ns/mongo-php-driver/commit/08d8528b)): Move \`PHP_MONGODB_ADD_BUILD_INCLUDE\` calls before \`PHP_MONGODB_ADD_INCLUDE\` calls in \`config.m4\`. Remove the duplicate block that was placed after \`AC_CONFIG_FILES\`.

## Testing

Verified that \`pecl install mongodb-2.3.2\` builds successfully on \`php:8.4.16-fpm-alpine3.23\` (Alpine 3.23, musl libc, GCC 15.2.0) with both patches applied:

```
Build complete.
install ok: channel://pecl.php.net/mongodb-2.3.2
```
```
mongodb version: 2.3.2
```

Without the fix, the build fails with:
```
error: implicit declaration of function 'res_ninit'; did you mean 'res_init'?
error: implicit declaration of function 'res_nsearch'
error: implicit declaration of function 'res_nclose'
```

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -374,6 +374,13 @@ if test "$PHP_MONGODB" != "no"; then
     PHP_MONGODB_ADD_SOURCES([src/libmongoc/src/libbson/src/jsonsl/], $PHP_MONGODB_JSONSL_SOURCES, $PHP_MONGODB_BUNDLED_CFLAGS)
     PHP_MONGODB_ADD_SOURCES([src/libmongoc/src/libmongoc/src/mongoc/], $PHP_MONGODB_MONGOC_SOURCES, $PHP_MONGODB_BUNDLED_CFLAGS)
 
+    dnl Build-dir includes must come before source-dir includes so that the
+    dnl compiler finds configure-generated headers (e.g. mongoc-config.h with
+    dnl correct MONGOC_HAVE_RES_* values) rather than the pre-bundled copies.
+    PHP_MONGODB_ADD_BUILD_INCLUDE([src/libmongoc/src/common/src/])
+    PHP_MONGODB_ADD_BUILD_INCLUDE([src/libmongoc/src/libbson/src/])
+    PHP_MONGODB_ADD_BUILD_INCLUDE([src/libmongoc/src/libmongoc/src/])
+
     PHP_MONGODB_ADD_INCLUDE([src/libmongoc/src/common/src/])
     PHP_MONGODB_ADD_INCLUDE([src/libmongoc/src/uthash/])
     PHP_MONGODB_ADD_INCLUDE([src/libmongoc/src/libbson/src/])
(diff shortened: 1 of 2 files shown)


GromNaN:
Thank you for reporting and investigating this issue. It should be fixed by https://github.com/mongodb/mongo-php-driver/pull/2022

PR #2033, opened 2026-06-29
PHPC-2728: Add `stringOpts` support and bump libmongoc 2.4.0

GromNaN:
Adds support for the `stringOpts` option in `ClientEncryption::encrypt()` for QE string encryption (prefix, suffix, substring query types with case/diacritic sensitivity).

Jira: https://jira.mongodb.org/browse/PHPC-2728


src/MongoDB/ClientEncryption.c
```diff
@@ -832,6 +832,157 @@ static void phongo_clientencryption_create_datakey(phongo_clientencryption_t* cl
 	bson_value_destroy(&keyid);
 }
 
+static mongoc_client_encryption_encrypt_text_opts_t* phongo_clientencryption_encrypt_text_opts_from_zval(zval* options)
```
GromNaN: The name `text_opts` is inherited from libmongoc/libmongocrypt.

kevinAlbs: Heads up: the "text" named C driver API is deprecated for newly added "string" names in CDRIVER-6293. But this is not-yet released. It will be released in C driver 2.4.0.

src/MongoDB/ClientEncryption.c
```diff
@@ -832,6 +832,157 @@ static void phongo_clientencryption_create_datakey(phongo_clientencryption_t* cl
...
+
+		if (prefix_zval && Z_TYPE_P(prefix_zval) == IS_ARRAY) {
+			if (php_array_existsc(prefix_zval, "strMaxQueryLength")) {
+				int64_t v = php_array_fetchc_long(prefix_zval, "strMaxQueryLength");
+
+				if (v < 0 || v > INT32_MAX) {
+					phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected \"strMaxQueryLength\" to be a positive 32-bit integer, %" PRId64 " given", v);
+					mongoc_client_encryption_encrypt_text_prefix_opts_destroy(prefix_opts);
+					goto cleanup;
+				}
```
kevinAlbs:
```suggestion
				if (v < INT32_MIN || v > INT32_MAX) {
					phongo_throw_exception(PHONGO_ERROR_INVALID_ARGUMENT, "Expected \"strMaxQueryLength\" to be a 32-bit integer, %" PRId64 " given", v);
					mongoc_client_encryption_encrypt_text_prefix_opts_destroy(prefix_opts);
					goto cleanup;
				}
```

Minor: consider only checking that `v` fits within an int32 before the `(int32_t)` cast. libmongocrypt further [validates](https://github.com/mongodb/libmongocrypt/blob/96caed1bdbcd473cd019b8f4f3c2e5981a3f2da6/src/mc-textopts.c#L81-L93) the value is greater than 0. This may help avoid a needed change in the PHP driver if that validation is ever changed in libmongocrypt to include 0.

Similar suggestion applies to other int32 checks.

kevinAlbs (approved):
LGTM with a minor suggestion.

src/MongoDB/ClientEncryption.c
GromNaN: So, this PR can wait the release of C driver 2.4.0.

src/MongoDB/ClientEncryption.c
GromNaN: Done. Note that the behavior changes slightly: values that are negative but fit in an int32 (e.g. `-1`) are no longer rejected by the PHP driver and instead reach libmongocrypt, which throws an `EncryptionException` ("'strMaxQueryLength' must be greater than zero"). Values that overflow int32 (e.g. `PHP_INT_MAX`) are still caught early as `InvalidArgumentException`. The phpt test has been updated to cover both cases.

PR #2054, opened 2026-08-25
PHPC-2751: Upgrade libmongoc to 2.5.0

paulinevos:
Also bumps `utf8proc` as it was bumped for `libmongoc` 2.5.0

### DIFF
--- a/.github/workflows/check-sbom.yml
+++ b/.github/workflows/check-sbom.yml
@@ -1,0 +1,46 @@
+name: "Check SBOM"
+
+on:
+  pull_request:
+    branches:
+      - "v*.*"
+      - "feature/*"
+    paths:
+      - ".github/workflows/check-sbom.yml"
+      - "sbom.json"
+      - "src/libmongoc"
+      - "src/libmongocrypt"
+  push:
+    branches:
+      - "v*.*"
+      - "feature/*"
+    paths:
+      - ".github/workflows/check-sbom.yml"
+      - "sbom.json"
+      - "src/libmongoc"
+      - "src/libmongocrypt"
+
+jobs:
+  check-sbom:
+    name: "Check SBOM component versions"
+    runs-on: "ubuntu-24.04"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v7"
+        with:
+          # The purls are derived from version files inside the submodules
+          submodules: true
+
+      # silkbomb percent-encodes the version segment, so a git-describe version such as
+      # 1.10.1-20260825+git048fa8ef67 is stored as ...%2Bgit048fa8ef67. Decode before comparing.
+      - name: "Check SBOM purls against bundled dependency versions"
+        run: |
+          ./scripts/generate-purls.sh | sort > expected-purls.txt
+          jq -r '.components[].purl' sbom.json |
+            python3 -c 'import sys, urllib.parse; sys.stdout.writelines(urllib.parse.unquote(line) for line in sys.stdin)' |
+            sort > actual-purls.txt
+          if ! diff --unified expected-purls.txt actual-purls.txt; then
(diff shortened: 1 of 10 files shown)


.github/workflows/check-sbom.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v7"
+        with:
+          # The purls are derived from version files inside the submodules
+          submodules: true
+
+      # Runs the purl generation from scripts/update-sbom.sh without silkbomb, which lives in a
+      # private ECR registry and is therefore unavailable to pull request builds. This catches the
+      # case that matters in review: a submodule bump that forgot to update sbom.json.
```
GromNaN: The ECR is available in github actions. This assertion it not true.
However, I do understand the motivation to have a more direct validation.

.github/workflows/check-sbom.yml
GromNaN, line 1: You should filter the path that trigger this action
- .github/workflows/check-sbom.yml
- sbom.json
- src/libmongoc
- src/libmongocrypt

scripts/update-submodule-sources.php
```diff
@@ -13,7 +13,7 @@
     'PHP_MONGODB_BSON_SOURCES' => 'src/libmongoc/src/libbson/src/bson',
     'PHP_MONGODB_JSONSL_SOURCES' => 'src/libmongoc/src/libbson/src/jsonsl',
     'PHP_MONGODB_MONGOC_SOURCES' => 'src/libmongoc/src/libmongoc/src/mongoc',
-    'PHP_MONGODB_UTF8PROC_SOURCES' => 'src/libmongoc/src/utf8proc-2.8.0',
+    'PHP_MONGODB_UTF8PROC_SOURCES' => 'src/libmongoc/src/utf8proc-2.11.3',
```
GromNaN, line 16: Has this package been updated in libmongoc? Why has it been updated now, and not for the latests libmongoc updates?

.github/workflows/check-sbom.yml
paulinevos: I'm gonna remove the comment though since it's just confusing to whoever reads it next

scripts/update-submodule-sources.php
paulinevos: No I think this is leftover from something I did wrong before

.github/workflows/check-sbom.yml
paulinevos: Ah! So, actually. Turns out the issue is that secrets are not exposed to forks, so it wouldn't be able to run anyway

scripts/update-submodule-sources.php
paulinevos: (Incorrect, it actually was bumped in 2.5.0: https://jira.mongodb.org/browse/CDRIVER-6422)

GromNaN (approved):

config.m4
```diff
@@ -243,9 +243,9 @@ if test "$PHP_MONGODB" != "no"; then
   PHP_MONGODB_MONGOCRYPT_VERSION_STRING="None"
 
   if test "$PHP_MONGODB_SYSTEM_LIBS" != "no"; then
-    PHP_MONGODB_MIN_LIBBSON_VERSION="2.4.0"
-    PHP_MONGODB_MIN_LIBMONGOC_VERSION="2.4.0"
     PHP_MONGODB_MIN_LIBMONGOCRYPT_VERSION="1.20.2"
+    PHP_MONGODB_MIN_LIBBSON_VERSION="2.5.0"
+    PHP_MONGODB_MIN_LIBMONGOC_VERSION="2.5.0"
```
GromNaN, line 248: I've just noticed that you changed the order of this declaration. That does not change the behavior, but complicates merges.

PR #2056, opened 2026-08-25
PHPC-2755: Fix test suite failures with PHP 8.6

GromNaN:
Since PHP 8.6, `run-tests.php` spawns several test workers by default. The suite is not parallel safe, as all tests share a single MongoDB deployment where fail points and namespaces are global, so `-j1` is now passed explicitly.

Also replaces `is_integer()`, deprecated since PHP 8.6, with `is_int()`.


GromNaN:
Idea for a follow-up ticket, out of scope here.

Making the suite parallel safe would need two changes:

- Scope fail points to a single client. `failCommand` accepts an `appName` filter, so a fail point only applies to connections using that application name. Tests would create their manager with `appName=<test name>` in the URI and pass the same value in the fail point data. This removes the case where a fail point with `times: 1` is consumed by another test.
- Give each test its own namespace, including the CSFLE key vault, which is currently the fixed `keyvault.datakeys` for every test.

The measured gain does not justify the work on its own: the suite runs in 53 seconds sequentially and in 22.6 seconds with 4 workers, on a job that takes 3 minutes 19 seconds in total, dominated by the server setup and the build.

Splitting the suite into a parallel group (tests/bson, tests/bson-corpus, tests/bson-binary-vector, 1451 of 2250 tests) and a sequential group was also considered and rejected. `run-tests.php` has no exclusion option, so the second group would have to list every other directory by hand, and a directory left out would silently stop being tested.

.github/workflows/tests.yml
```diff
@@ -80,7 +80,10 @@ jobs:
       - name: "Run Tests"
         run: make test
         env:
-          TEST_PHP_ARGS: "-q -x --show-diff -g FAIL,XFAIL,BORK,WARN,LEAK,SKIP"
+          # Tests cannot run in parallel, as they all share a single MongoDB
+          # deployment where fail points and namespaces are global. PHP 8.6
+          # spawns several test workers by default, so -j1 is required.
+          TEST_PHP_ARGS: "-q -x -j1 --show-diff -g FAIL,XFAIL,BORK,WARN,LEAK,SKIP"
```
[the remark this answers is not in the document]
GromNaN: Good catch. `Makefile.frag` now sets `TEST_PHP_ARGS ?= -j1` and exports it, so `make test` and `make test-no-build` run serially by default on developer machines. An environment or command line value still wins, which is why CI keeps `-j1` in its own `TEST_PHP_ARGS`.

paulinevos:
@GromNaN Can you restart the evergreen builds so we can ensure it's not related to this fix?

Makefile.frag
```diff
@@ -5,6 +5,13 @@ PHP_TEST_SHARED_EXTENSIONS := "-d" "extension=$(EXTENSION_DIR)/json.so" $(PHP_TE
 PHP_TEST_SETTINGS := "-d" "extension=$(EXTENSION_DIR)/json.so" $(PHP_TEST_SETTINGS)
 endif
 
+# Tests cannot run in parallel, as they all share a single MongoDB deployment
+# where fail points and namespaces are global. Since PHP 8.6, run-tests.php
+# spawns several test workers by default, so -j1 is required. Environments that
+# set TEST_PHP_ARGS themselves must include -j1 as well.
+TEST_PHP_ARGS ?= -j1
```
paulinevos, line 9: Tbh I feel like these sorts of descriptions belong in the commit history. Claude is very eager to add them as code comments.

I'd also like to see an explanation of what `j1` actually is/does so we don't have to go and look it up.

GromNaN: I'll simplify this comment.
`-j` is the number of parallel processes. Documented in [`run-tests.php`](https://github.com/php/php-src/blob/846773a4e9bfc6df7e672737e3b7b07c4d57b61f/run-tests.php#L37-L40).

paulinevos (approved):

PR #2094, opened 2026-09-04
Warn against premature release publish

kevinAlbs:
Motivated by recent release difficulties.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,6 +23,10 @@ env:
 
     TODO: one or more paragraphs describing important changes in this release
 
+    > [!WARNING]
+    > Do _NOT_ publish release until after artifacts are attached from the "Create Release Artifacts" workflow.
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -23,6 +23,10 @@ env:
 
     TODO: one or more paragraphs describing important changes in this release
 
+    > [!WARNING]
+    > Do _NOT_ publish release until after artifacts are attached from the "Create Release Artifacts" workflow.
```
GromNaN: Maybe provide a link?
```suggestion
    > Do _NOT_ publish release until after artifacts are attached from the [Create Release Artifacts](https://github.com/mongodb/mongo-php-driver/actions/workflows/create-release-artifacts.yml) workflow.
```

kevinAlbs: Good idea. Accepting suggestion.

GromNaN (approved):
Thanks!

